Issue #831, opened 2018-10-18
Surveillance officers should be able to create and assign tasks for informants [2]


MartinWahnschaffeSymeda:
Creating and assigning tasks is a new functionality for the Android app.
It should be possible to do so from the tasks view in the app. Similar to creating a new case, a"+" sign in the top right will allow the user to create a new task. This has to be limited based on the user rights.

MartinWahnschaffeSymeda:
Preparation - sormas-api

- [ ] add SUVEILLANCE_OFFICER to TASK_CREATE and TASK_ASSIGN UserRight

Implementation details - sormas-app

Important: There are four ways to create a task (one for each TaskContext):

1. Create a case task from within the tasks list of a case (CaseEditTaskListFragment)
2. Create a contact task from within the tasks list of a contact (ContactEditTaskListFragment)
3. Create an event task from within the tasks list of an event (EventEditTaskListFragment)
4. Create a general task from within the overall tasks list (TaskListActivity)

- [ ] create TaskNewActivity and TaskNewFragment with resource fragment_task_create_layout.xml. You can copy TaskEditFragment and fragment_task_edit_layout.xml. Look into SampleNewAcitivity/Fragment for reference
- [ ] TaskNewAcitivty.startActivity needs to have a TaskContext param and the entity uuid (can be case, contact or event)
- [ ] same for the buildBundle method
- [ ] the buildRootEntity method needs to look into the TaskContext and create a new Task using the correct reference entity
- [ ] override isEntryCreateAllowed method of TaskListActivity (is allowed when user has the needed user right)
- [ ] override isShowSaveAction method of CaseEditTaskListFragment, ContactEditTaskListFragment, EventEditTaskListFragment
- [ ] fragment_task_new_layout should not have the button layout with "not-executable" and "done" buttons
- [ ] user should be able to edit the task type and assignee user when creating a task or editing a task that he has created (replace ControlTextReadField with a ControlSpinnerField and initialize in TaskNew- and TaskEditFragment 
- [ ] same for the creator comment
- [ ] check if anything else is missing. Compare to task create in web app

MartinWahnschaffeSymeda:
Merged.

We need a way for the surveillance officer to see the general tasks created. Currenty the overall tasks list only shows tasks assigned to the user, not tasks created by the user.
One solution could be to split the menu filters into "my pending", "my done" and "my not executable" and "others pending", "others done" and "others not executable"

[...]

PR #928, opened 2019-01-11
Feature 831 surveillance officers

faatihi:
#831


MartinWahnschaffeSymeda:
I have fixed the implementation of allowedDaysInFuture and added the property to the fields for dueDate and suggestedStart. The setEnabled calls have been moved to onAfterLayoutBinding.
Everything is added as a commit to this pull request, so please go on making changes based on the pull request.

A few things still need to be changed:

- [ ] taskType should be filtered based on the taskContext (see TaskType.getTaskTypes)
- [ ] assigneeReply should not be visible when creating a new task
- [ ] taskType, dueDate and assigneeUser should be required fields
- [ ] creatorComment should be required when task type is OTHER

Seeing this in action, I think that using the same fragment for an editing user and a executing user (that can comment and change the status) is not ideal - I thought this would work out better. I think we should add a separate TaskExecutionFragment and let the TaskEditActivity pick either one based on the user.

The code itself is very clean and to the point - I like it 👍

faatihi:
Thanks, Martin.
I'll look into the few things.

MartinWahnschaffeSymeda:
We need a way for the surveillance officer to see the general tasks created. Currenty the overall tasks list only shows tasks assigned to the user, not tasks created by the user.
One solution could be to split the menu filters into "my pending", "my done" and "my not executable" and "others pending", "others done" and "others not executable"

Issue #835, opened 2018-10-18
Dashboard changes for Ghana SORMAS version


hzi-braunschweig:
We need to discuss this. 
![our-proposed-dashboard-for-non-technical-stakeholders](https://user-images.githubusercontent.com/19344651/47151596-760bfa00-d2da-11e8-8f34-8d94429dae59.jpg)
[Ghana_District_Codes (1).xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/files/2491659/Ghana_District_Codes.1.xlsx)
@MartinWahnschaffeSymeda @MateStrysewskeSym @SeriousMartin

hzi-braunschweig:
@MartinWahnschaffeSymeda @MateStrysewskeSym @bernardsilenou 
We had a discussion at HZI about the dashboard for SORMAS and below is a workable mockup to give you an idea of what Gerard understands from the ghana dashboard for NON-TECHNICAL STAKEHOLDERS. We are still discussing this but I felt you should all know the progress and see what you can integrate based on this idea. 

Please note, the current dashboard remains but should only be for a filtered disease. The dashboard below is only for an unfiltered view when no disease is selected. Which means an overview of the current situation at any point in time. Let me know if you have any questions and so I can relay it back to the team. 

![sormas ghana dashboard](https://user-images.githubusercontent.com/19344651/47356606-de851d80-d6c4-11e8-86e5-7266e92cffd5.png)

hzi-braunschweig:
@MartinWahnschaffeSymeda @MateStrysewskeSym . So i made some changes to the dashbaord attached. We can discuss details tommorow. 
![image](https://user-images.githubusercontent.com/19344651/47659295-f19c5f80-db94-11e8-9f96-fc1a081a069e.png)

MartinWahnschaffeSymeda:
@hzi-braunschweig Thanks for the mockups. One very important thing to consider: The dashboard needs to be responsive. This basically means that the used elements should have similar widths so they can be displayed vertically on small screens.

hzi-braunschweig:
@MartinWahnschaffeSymeda Yes, we should adapt it to its responsiveness. The mockup is just to give us an idea. I will explain more tommorow.

MartinWahnschaffeSymeda:
Note from last sprint review:
@hzi-braunschweig will discuss the proposal with GHS and the users in Nigeria, so we can have a unified version of it.

MartinWahnschaffeSymeda:
[...]

PR #959, opened 2019-02-06
Feature 835 dashboard changes for ghana

faatihi:
#835


MartinWahnschaffeSymeda:
@faatihi I have made some changes to make the layout responsive (not perfect, but good enough) and have removed some no longer needed code.

Two things that need to be done:

- [ ] "difference in case numbers" diagram order of values does not match the diease caption of the axis when positive and negative values alternate. Since we want to change this with the new layout anyway it's probably easiest to sort the diseases in the diagram by number of cases, from biggest incease to biggest decrease
- [ ] all texts in the dashboard need to be added to the i18n. You can have a look at DashboardContactsStatisticsComponent for reference. Basically you add the needed texts to captions.properties or strings.properties. Then you run the I18nConstantGenerator to update the Strings and Captions classes with the new texts. This can then be used like this: `I18nProperties.getString(Strings.headingFollowUpSituation)`

Issue #966, opened 2019-02-12
Disease burden tile view for dashboard


MartinWahnschaffeSymeda:
Based on #835 

@faatihi Here are updated mockups that address the missing information (as far as I understand).

I have attached screenshots that show the tiles version of the disease burden and the grid version.
My idea is that we could show the tiles version to national users and observers and the grid version to supervisors by default. In addition both user groups are allowed to switch to the other view.

It may also make sense to colorize the differences bar chart with the same colors as used in the tiles.

The last mockup shows what it would look like on a small screen, where you can either scroll down or hide the overview when only the details of a specific disease are relevant.
I’m not sure, but it’s probably better to have the epi curve and the case status map in the same row (instead of below each other, as in the mockup).

In short:

- [x] add alternative tile layout that shows either first 6 or all diseases as tiles
- [x] each tile has the short disease name, number of cases (for the time period), number of events, fatalities (deaths)
- [x] also show last reported sub-district. This can be retrieved by querying for the case with the latest report date that matches the filter criteria
- [x] we should probably associate one fixed color to each disease and use that for the tiles. Try to use colors in the style of the mockup. Dont' use red, yellow, and green
- [x] ~~case numbers diagram should use this color as well~~
- [x] allow users to switch between grid and tiles view
- [x] use grid view for all supervisor roles. All other roles use tiles view by default
- [x] for the disease specific view display the number of fatalities, outbreak districts and last reported sub-district as in the mockup

![dashboardtiles](https://user-images.githubusercontent.com/23217632/52637178-d4cf3200-2ece-11e9-80c0-38db3a9cbd59.jpg)

![dashboardgrid](https://user-images.githubusercontent.com/23217632/52637182-d7ca2280-2ece-11e9-8a4a-358f0a418eba.jpg)

![dashboardtilessmall](https://user-images.githubusercontent.com/23217632/52637187-da2c7c80-2ece-11e9-98ed-04f36ff71246.jpg)

MartinWahnschaffeSymeda:
Remarks from #835

- [ ] DiseaseFacadeEjb.getDiseaseBurdenForDashboard: move the queries for events, outbreaks and case fatalities to the corresponding facades (e.g. getEventCountPerDisease)
- [ ] test results and outbreak districts should just get count from backend and not a list (also remove DashboardOutbreakDto)

MartinWahnschaffeSymeda:
merged #1003

[...]

PR #1003, opened 2019-03-19
Feature 966 disease burden tile view for dashboard

faatihi:
#966


MartinWahnschaffeSymeda:
I haven't tested everything yet. Two things that should be changed:

- [ ] EventService.getEventCountByDisease and getEventCountByStatus: Extend EventCriteria and use that instead of single params
- [ ] Same for OutbreakService

The tiles look very good - much better than the table :-)

Issue #828, opened 2018-10-18
Add infrastructure data for Ghana


MartinWahnschaffeSymeda:
Needed regional data:

- regions with population data, growth rate, epid code
- districts with population data, growth rate, epid code
- communities (in Ghana called sub-districts)
- region shapes
- district shapes
- health facilities (ideally with GPS)
- laboratories

### Implementation

General note: Administrative divisions (regions, districts, communities) are imported on server startup by the StartupShutdownService.startup() method. All data files are located in the sormas-backend project under srv/main/resources.  
The country used during import is defined by the sormas.properties file located in the folder of the sormas server domain.  
The easiest way to test the import is to TRUNCATE .. CASCADE the region and users table. This will also truncate all tables with references to those. Afterwards deploy the sormas-ear (which includes sormas-backend) to your server and StartupShutdownService.startup will automatically be called.

#### Administrative divisions:

- [x] add regions, districts and communities csv files for regions/ghana. This should follow the pattern of the existing files. Make sure to also include population and growth rate.

#### Facilities:

- [x] add health facility csv files for each region. Make sure that names of regions, districts and communities match the imported adminstratve division data.
- [ ] ~~add a Laboratories.csv file with all the laboratories~~

#### Shapes:

Region and district shapes are not stored in the database. Instead they are loaded when GeoShapeProviderEjb initializes and kept in memory for quick access.
See http://mapshaper.org/ for compression
Edit dbf files with https://sourceforge.net/projects/gtkdbfeditor/files/latest/download

- [x] add shape files for regions and districts. The dbf column can be called StateName or REGION. The import uses a very simple heuristic to find similar names that may get things wrong, so make sure names in the shapefiles match those of the actual adminstrative divisions. It would be good to have a more powerful import for this.

MartinWahnschaffeSymeda:
[Health Facilities in Ghana.xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/files/2505967/Health.Facilities.in.Ghana.xlsx)
[Ghana Population.xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/files/2505968/Ghana.Population.xlsx)
[District population by age groups  2018.xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/files/2505969/District.population.by.age.groups.2018.xlsx)

hzi-braunschweig:
[...]

PR #1107, opened 2019-05-06
Fix ghana no shape for districts found

faatihi:
#828


faatihi:
@MartinWahnschaffeSymeda @MateStrysewskeSym

Issue #1109, opened 2019-05-09
Bed-side lab testing


MateStrysewskeSym:
Informants and Officers need to be able to document certain Pathogen Tests on their own without the need of sending the samples to a laboratory first.

Prerequisite: App users need to be able to create and edit pathogen tests #1134

- [x] Add new required field labType to the sampleDto and the backend and app entities. Values: external, internal
- [x] All lab and shipment/received fields should only be visible when external is selected.

Based on that a user can create a sample for internal testing and add a pathogen test to it using rapid test as type (already exists).

MartinWahnschaffeSymeda:
Notes from review:

- [ ] The final result of the sample likely needs to be set based on created pathogen test. Same process that happens when doing this in the web app. 
* What about the lab field for bed-side tests?
* What  about the "verified by supervisor" field?

MartinWahnschaffeSymeda:
@MateStrysewskeSym I have merged this now, but this needs some further work:

The final lab result of the sample needs to be displayed in the read and edit form of the app
The final result of the sample likely needs to be set based on created pathogen test. Same process that happens when doing this in the web app
The fragments still have the most recent pathogen and additional test layouts. At least the pathogen test layout is seemingly no longer used and can be removed. I'm not sure about the additional tests. Since there is currently no way to edit them in the app the "most recent" layout probably still makes sense for them.

PR #1226, opened 2019-07-19
Feature 1134 enable informants officers create edit pathogen tests

opatajoshua:
#1134

PR #1291, opened 2019-09-02
Feature 1109 bed side lab testing

opatajoshua:
#1109


sormas-api/src/main/resources/captions.properties
```diff
@@ -698,6 +698,7 @@ pathogenTestNewResult = New result
 pathogenTestNewTest = New test result
 
 PathogenTest = Pathogen test
+PathogenTests = Pathogen tests
```
MateStrysewskeSym: What exactly is this caption for?

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/sample/PathogenTest.java
```diff
@@ -64,9 +68,21 @@
     @Column(nullable = false)
     private PathogenTestResultType testResult;
 
+    @Column(nullable = true)
```
MateStrysewskeSym: "nullable = true" is redundant, please remove this from all columns

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/sample/PathogenTest.java
```diff
@@ -64,9 +68,21 @@
...
+    private Boolean testResultVerified;
+
+    @Column(length=512, nullable = true)
+    private String testResultText;
+
     @DatabaseField(dataType = DataType.DATE_LONG, canBeNull = false)
```
MateStrysewskeSym: Don't use not null constraints in the app because SQLite doesn't support removing those at a later point in time when they might become obsolete

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/sample/PathogenTest.java
```diff
@@ -64,9 +68,21 @@
...
     private Date testDateTime;
 
+    @DatabaseField(foreign = true, foreignAutoRefresh = true, maxForeignAutoRefreshLevel = 3, canBeNull = false)
```
MateStrysewskeSym: Don't use not null constraints (apart from those unfortunately already there) in the app because SQLite doesn't support removing those at a later point in time when they might become obsolete

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/sample/PathogenTestDtoHelper.java
```diff
@@ -64,6 +70,12 @@ protected void fillInnerFromDto(PathogenTest target, PathogenTestDto source) {
         target.setTestType(source.getTestType());
         target.setTestedDisease(source.getTestedDisease());
         target.setTestedDiseaseDetails(source.getTestedDiseaseDetails());
+
+        //newly added
```
MateStrysewskeSym: Please remove all comments that don't provide some essential information; particularly these "newly added" comments

sormas-app/app/src/main/java/de/symeda/sormas/app/pathogentest/edit/PathogenTestEditFragment.java
```diff
@@ -0,0 +1,123 @@
...
+    private List<Item> diseaseList;
+    private List<Item> testResultList;
+
+    // Instance methods
+
+    public static PathogenTestEditFragment newInstance(PathogenTest activityRootData) {
+        return newInstance(PathogenTestEditFragment.class, null, activityRootData);
+    }
+
+    private void setUpControlListeners(FragmentPathogenTestEditLayoutBinding contentBinding) {
```
MateStrysewskeSym: Remove empty methods

sormas-app/app/src/main/java/de/symeda/sormas/app/pathogentest/edit/PathogenTestEditFragment.java
```diff
@@ -0,0 +1,123 @@
...
+        testResultList = DataUtils.getEnumItems(PathogenTestResultType.class, true);
+        labList = DatabaseHelper.getFacilityDao().getLaboratories(true);
+    }
+
+    @Override
+    public void onLayoutBinding(FragmentPathogenTestEditLayoutBinding contentBinding) {
+        setUpControlListeners(contentBinding);
+
+        contentBinding.setData(record);
+//        contentBinding.setCaze(sourceCase);
```
MateStrysewskeSym: Remove commented code that is no longer used

sormas-app/app/src/main/java/de/symeda/sormas/app/pathogentest/edit/PathogenTestNewActivity.java
```diff
@@ -0,0 +1,177 @@
...
+        if (saveTask != null) {
+            NotificationHelper.showNotification(this, WARNING, getString(R.string.message_already_saving));
+            return; // don't save multiple times
+        }
+
+        final PathogenTest PathogenTestToSave = getStoredRootEntity();
+        PathogenTestNewFragment fragment = (PathogenTestNewFragment) getActiveFragment();
+
+        if (PathogenTestToSave.getLabUser() == null) {
+            PathogenTestToSave.setLabUser(ConfigProvider.getUser());
```
MateStrysewskeSym: This is not a good idea because lab users don't use the mobile app, so the result will always be wrong data; instead, this should result in an error and not allow the data to be saved

sormas-app/app/src/main/java/de/symeda/sormas/app/pathogentest/list/PathogenTestListViewModel.java
```diff
@@ -0,0 +1,140 @@
...
+
+            int offset = params.requestedStartPosition;
+            int count = params.requestedLoadSize;
+            int totalCount = 0;
+            List<PathogenTest> pathogenTests = new ArrayList<PathogenTest>();
+            if(offset==0){
+                pathogenTests = DatabaseHelper.getSampleTestDao().queryBySample(sample);
+                totalCount = pathogenTests.size();
+            }else{
+
```
MateStrysewskeSym: Remove empty else blocks

sormas-app/app/src/main/java/de/symeda/sormas/app/pathogentest/list/PathogenTestListViewModel.java
```diff
@@ -0,0 +1,140 @@
...
+import androidx.paging.PositionalDataSource;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import de.symeda.sormas.app.backend.common.DatabaseHelper;
+import de.symeda.sormas.app.backend.sample.PathogenTest;
+import de.symeda.sormas.app.backend.sample.Sample;
+
+public class PathogenTestListViewModel extends ViewModel {
```
MateStrysewskeSym: Please either use the paged list model properly (i.e. implement the loadInitial and loadRange methods like in the other view models to make the list lazy loading) OR make the list eager in the first place, i.e. don't use a paged list. Currently, this is a lazy list that simulates eager loading which is not a good concept.

sormas-app/app/src/main/java/de/symeda/sormas/app/sample/edit/SampleEditActivity.java
```diff
@@ -57,6 +70,34 @@ protected Sample buildRootEntity() {
...
+    public List<PageMenuItem> getPageMenuData() {
+        List<PageMenuItem> menuItems = PageMenuItem.fromEnum(SampleSection.values(), getContext());
+        return menuItems;
+    }
+
+    @Override
+    protected BaseEditFragment buildEditFragment(PageMenuItem menuItem, Sample activityRootData) {
+//        return SampleReadFragment.newInstance(activityRootData);
+        int menuKey =0;
+        if(menuItem !=null)
```
MateStrysewskeSym: Please try to adhere to the following code style (we will add a code style guide at a later point in time):
- Empty space between statements like "if" and "for" and the following (
- Always use { and }, even if the body contains only one line

sormas-app/app/src/main/java/de/symeda/sormas/app/sample/edit/SampleEditFragment.java
```diff
@@ -244,6 +260,17 @@ public void onChange(ControlPropertyField field) {
         }
     }
 
+    void showLabExternal(FragmentSampleEditLayoutBinding contentBinding, int show){
```
MateStrysewskeSym: This method should have a more general name, e.g. setVisibilitiesBasedOnLabType or something like that; replace the "show" parameter with the actual lab type and set visibilities in a switch statement

sormas-app/app/src/main/java/de/symeda/sormas/app/pathogentest/list/PathogenTestListViewModel.java
opatajoshua: I need clarification

1. "use the paged list model properly" 
Need to add DatabaseHelper.getPathogenTestDao()
Need to implement queryByCriteria() method in PathogenTestDao

2. "make the list eager in the first place"
PathogenTestListAdapter extends BindingPagedListAdapter which is a paged list adapter. So what is the eager loading list adapter class to extend

MateStrysewskeSym: I don't think we use any eager lists anymore in the app, so the first suggestion is probably the easier solution. Please have a look especially at the loadInitial and loadRange methods of other view models. And yes, you're right, you'll need to add queryByCriteria and countByCriteria methods.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleCreateForm.java
```diff
@@ -101,8 +104,16 @@ protected void addFields() {
 		addField(SampleDto.SPECIMEN_CONDITION, ComboBox.class);
 		addField(SampleDto.NO_TEST_POSSIBLE_REASON, TextField.class);
 		addField(SampleDto.COMMENT, TextArea.class).setRows(2);
+		ComboBox labType = addField(SampleDto.LAB_TYPE, ComboBox.class);
 		CheckBox shipped = addField(SampleDto.SHIPPED, CheckBox.class);
 		CheckBox received = addField(SampleDto.RECEIVED, CheckBox.class);
+		
+		//hidding for labType
```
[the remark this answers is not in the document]
opatajoshua: Should the labtype column (which is now samplepurpose) have external as default value in the database?
because the field is required and the old records will be null if no default is set, which will cause unchecked errors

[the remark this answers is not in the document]
MateStrysewskeSym: Yes, all currently existing samples should be external since we didn't have any other sample purpose in the system until now.

opatajoshua:
Getting a runtime error in the mobile app:

Error trying to synchronizing data in mode 'CompleteAndRepull'
    de.symeda.sormas.app.backend.common.DaoException: cannot deserialize class de.symeda.sormas.api.caze.classification.ClassificationCriteriaDto subtype named ClassificationPathogenTestNegativeResultCriteriaDto; did you forget to register a subtype?
        at de.symeda.sormas.app.backend.common.AdoDtoHelper.pullEntities(AdoDtoHelper.java:105)

MateStrysewskeSym:
@opatajoshua You shouldn't receive them anymore once you've updated to the current state of the development branch. That bug has been fixed.

opatajoshua:
@MateStrysewskeSym Done with the changes.

MartinWahnschaffeSymeda:
@opatajoshua Same as with #1227:
Since #1226 is merged now, please update this pull request. The github "resolve conflicts" tool does not do a good job in this case. Instead it would be best if you could switch to your branch https://github.com/GCNETGHANA/SORMAS-Project/tree/feature-when-a-contact-is-converted-to-case and rebase is to the latest development version. Thisway we should get rid of all those conflicts related to the pathogen tests.

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleExportDto.java
```diff
@@ -279,7 +291,7 @@ public void setRequestedOtherPathogenTests(String requestedOtherPathogenTests) {
 		this.requestedOtherPathogenTests = requestedOtherPathogenTests;
 	}
 
-	@Order(20)
+	@Order(21)
```
MartinWahnschaffeSymeda: FYI: There was no need to increment this and the following order entries. We have left some unused order entries (e.g. 18 and 19) to allow for easy addition of new fields without having to update all following order entries.
You can leave it like that now.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/sample/PathogenTestDtoHelper.java
```diff
@@ -69,6 +69,7 @@ protected void fillInnerFromDto(PathogenTest target, PathogenTestDto source) {
         target.setTestTypeText(source.getTestTypeText());
         target.setTestedDisease(source.getTestedDisease());
         target.setTestedDiseaseDetails(source.getTestedDiseaseDetails());
+        target.setLab(DatabaseHelper.getFacilityDao().getByReferenceDto(source.getLab()));
```
MartinWahnschaffeSymeda: Duplicate. See line 75

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/sample/Sample.java
```diff
@@ -99,6 +100,9 @@
     @Column(length = 512)
     private String labDetails;
 
+    @Enumerated(EnumType.STRING)
+    private SamplePurpose samplePurpose;
```
MartinWahnschaffeSymeda: Also needs to be added to the database via DatabaseHelper. See https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/GUIDE_ADD_NEW_FIELD.md#iv-adding-the-field-to-the-sormas-android-app

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleExportDto.java
opatajoshua: Alright

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/sample/Sample.java
opatajoshua: i am getting this error after "Database upgrade failed for version 167: Could not run raw execute statement ALTER TABLE samples ADD COLUMN samplePurpose varchar(255);"

opatajoshua: @MartinWahnschaffeSymeda please comment on this

MartinWahnschaffeSymeda: @opatajoshua Sorry, I have asked for clarification via skype: I will need to see your source code you hava added to the DatabaseHelper.onUpgrade method. The branch does not seem to have it yet: https://github.com/GCNETGHANA/SORMAS-Project/blob/feature-1109-bed-side-lab-testing/sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java

opatajoshua: ok let me push what i have now

MartinWahnschaffeSymeda: Have you pushed it? I don't see any new commits related to this

opatajoshua: Has been pushed now

MartinWahnschaffeSymeda: Ah, I see. The case has to be 166, not 167. 
Explanation: The DATABASE_VERSION ist the new version (167 now - this part is correct).
The number in the switch case is the previous version, so 166. It tells the system what to do when moving from the previous version to the new one.
When both numbers are the same, this will always be called. So your statement will have worked the first time, but no more, once you have started the app the second time.

opatajoshua: changing that

opatajoshua: changed but still had the error. the message is not detailed. i had to uninstall before rerunning the app

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -3648,4 +3648,7 @@ INSERT INTO schema_version (version_number, comment) VALUES (166, 'Add pathogenT
 -- 2019-11-04 Fill samples with accurate pathogenTestResultChangeDate #1349
 UPDATE samples SET pathogentestresultchangedate = (SELECT testdatetime FROM pathogentest WHERE pathogentest.sample_id = samples.id ORDER BY pathogentest.testdatetime DESC LIMIT 1);
 
-INSERT INTO schema_version (version_number, comment) VALUES (167, 'Fill samples with accurate pathogenTestResultChangeDate #1349');
\ No newline at end of file
+INSERT INTO schema_version (version_number, comment) VALUES (167, 'Fill samples with accurate pathogenTestResultChangeDate #1349');
+
+-- 2019-11-06 Bed-side lab testing #1109
+ALTER TABLE samples ADD COLUMN samplepurpose  varchar(255) DEFAULT 'EXTERNAL' NOT NULL;
```
MartinWahnschaffeSymeda: @opatajoshua Remember to add a schema_version entry after the changes. I will take care of this now.

MartinWahnschaffeSymeda:
@opatajoshua Merged now. I have re-enabled the pathogen tests for internal samples, because the users have no other way to create test results. Hope that's the way it's intended.

In addition I have done some clean-up that was not directly related to your implementation, so no need to worry about the amount of changes I made. Still it's likely worth a look - possible that I got something wrong.

Issue #834, opened 2018-10-18
New disease: Human rabies


MartinWahnschaffeSymeda:
Details still need to be specified.

hzi-braunschweig:
[Rabies-ProcessModel_Ghana_gek_181011.xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/files/2491460/Rabies-ProcessModel_Ghana_gek_181011.xlsx)

MartinWahnschaffeSymeda:
@hzi-braunschweig This (and Anthrax) needs a more detailed implementation plan. 

1. We need to clearly define which fields have to be added to the system and which existing fields we use (see https://github.com/hzi-braunschweig/SORMAS-Project/raw/development/sormas-base/doc/SormasDictionary.xlsx).
2. Is there anything that needs to be implemented for the process? E.g. new task types, automatic generated tasks, etc.?

faatihi:
Is this issue ready for implementation?

hzi-braunschweig:
[Anthrax_Rabies_Table_jdo_190116_DTA_adk.xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/files/3426766/Anthrax_Rabies_Table_jdo_190116_DTA_adk.xlsx)
@faatihi You can start taking a look at this document. @MartinWahnschaffeSymeda Will the case classification be handled by GCNET or this is separate from the actual development of the new diseases?

MartinWahnschaffeSymeda:
@hzi-braunschweig This should be part of adding a new disease.

faatihi:
![image](https://user-images.githubusercontent.com/1176692/62534585-064f0780-b839-11e9-9c00-5a0351ed824e.png)

This link is broken. Is it no longer relevant?

MartinWahnschaffeSymeda:
Sorry, the document was moved. Still very (!) relevant:
https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-api/src/main/resources/doc/SORMAS_Data_Dictionary.xlsx?raw=true

faatihi:
thanks

faatihi:
![image](https://user-images.githubusercontent.com/1176692/64181267-ab1f2d80-ce55-11e9-8197-8d3f10842f89.png)

The "Which vaccine", I guess its a new field to be added

faatihi:
[...]

PR #1334, opened 2019-10-09
Feature new disease human rabies

faatihi:
#834


MartinWahnschaffeSymeda:
@faatihi Is this ready to be reviewed?

faatihi:
Yes, it's ready

sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsDto.java
```diff
@@ -163,6 +163,25 @@
...
+	public static final String EXCITATION = "excitation";
+	public static final String DYSPHAGIA = "dysphagia";
+	public static final String AEROPHOBIA = "aerophobia";
+	public static final String HYPERACTIVITY = "hyperactivity";
+	public static final String PARESIS = "paresis";
+	public static final String AGITATION = "agitation";
+	public static final String ASCENDING_FLACCID_PARALYSIS = "ascendingFlaccidParalysis";
+	public static final String ERRATIC_BEHAVIOUR = "erraticBehaviour";
+	public static final String COMA = "coma";
+	public static final String DEATH = "death";
```
MartinWahnschaffeSymeda: @hzi-braunschweig @bernardsilenou Does "death" really make sense as a symptom?

faatihi:
OK. I'll take a look at these issues

sormas-api/src/main/java/de/symeda/sormas/api/epidata/EpiDataDto.java
```diff
@@ -147,12 +170,32 @@
 	private YesNoUnknown tickBite;
 	@Diseases({Disease.PLAGUE,Disease.UNDEFINED,Disease.OTHER})
 	private YesNoUnknown fleaBite;
-	@Diseases({Disease.MONKEYPOX,Disease.UNDEFINED,Disease.OTHER})
+	@Diseases({Disease.RABIES, Disease.UNDEFINED, Disease.OTHER})
+	private YesNoUnknown kindOfExposure;
```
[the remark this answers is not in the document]
faatihi: I'm removing the field from both EpiData and EpiDataDto. 
For EpiDataDto, I'll use a getter to return a calculated value.

I've hit a few hurdles:
1. The field is needed in the dto for annotating Diseases. So, I'm keeping the private field
2. The input is disabled when there's no setter for the field. I've therefore added a setter with empty code body

image

Any suggestions?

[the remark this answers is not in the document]
MartinWahnschaffeSymeda: You could also completely remove it from the DTO and have the logic only in the UI. This means using a field in the form that is not bound to the dto. This field would be visible if any of the "kindOfExposure..." fields is visible (which is defined by their disease annotations).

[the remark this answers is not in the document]
faatihi: OK. I was going for a data-driven approach for the form. Your idea also sounds good.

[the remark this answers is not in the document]
faatihi: @MartinWahnschaffeSymeda Kindly point to me an example of a field that is not bound to the dto. I'm currently using an example from Vaadin's official docs. Maybe sormas has an already established way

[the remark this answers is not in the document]
MartinWahnschaffeSymeda: To get a basic idea you can have a look at StatisticsVisualizationComponent. 
Take the example of visualizationSelect:

1. The OptionGroup is created (in your case you could also use a Checkbox or add the ValoTheme.OPTIONGROUP_HORIZONTAL style). The constructor receives a list of the possible values.
2. The field is added to the layout. You can do that with a placeholder for the location similar to EpiDataForm.ENVIRONMENTAL_LOC-
3. A value change listener is added. In your case that listener would show or hide the kindOfExposure fields. In theory you could create a new variant of FieldHelper.setVisibleWhen that takes a vaadin field as source instead of the sourcePropertyId
4. The inital value of the field is set by calculating if any kindOfExposure is not null. The logic would ideally be put into an EpiDataHelper class in the api project - similar to SymptomsHelper.

Hope that helps.

sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsDto.java
faatihi: image

Death was mentioned as a symptom (mentioned at three places). Unless I didn't understand what they meant.

faatihi: OK. @bernardsilenou has explained to me. I'm dropping death

faatihi:
@MartinWahnschaffeSymeda , this pull is ready for review

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
```diff
@@ -163,17 +168,20 @@
...
-	@Diseases({ Disease.MEASLES, Disease.YELLOW_FEVER, Disease.CSM, Disease.MONKEYPOX, Disease.OTHER })
+	@Diseases({ Disease.MEASLES, Disease.YELLOW_FEVER, Disease.CSM, Disease.MONKEYPOX, Disease.RABIES, Disease.OTHER })
 	@Outbreaks
 	private Date vaccinationDate;
-	@Diseases({ Disease.MEASLES, Disease.YELLOW_FEVER, Disease.CSM, Disease.OTHER })
+	@Diseases({ Disease.MEASLES, Disease.YELLOW_FEVER, Disease.CSM, Disease.RABIES, Disease.OTHER })
 	private VaccinationInfoSource vaccinationInfoSource;
+	@Diseases({ Disease.RABIES, Disease.OTHER })
+	@Outbreaks
+	private String whichVaccine;
```
MateStrysewskeSym: Rename to "vaccine"

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
```diff
@@ -115,6 +117,9 @@
 	@Diseases({ Disease.DENGUE })
 	@Outbreaks
 	private DengueFeverType dengueFeverType;
+	@Diseases({ Disease.RABIES })
+	@Outbreaks
+	private HumanRabiesType humanRabiesType;
```
MateStrysewskeSym: Either rename the disease to HUMAN_RABIES or fields associated with the disease to accord to the disease name, e.g. in this case "rabiesType" instead of "humanRabiesType"; I think I prefer the latter because internally, it's not necessary to call the disease HUMAN_RABIES

sormas-api/src/main/java/de/symeda/sormas/api/epidata/EpiDataHelper.java
```diff
@@ -0,0 +1,32 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *******************************************************************************/
+package de.symeda.sormas.api.epidata;
+
+import de.symeda.sormas.api.utils.YesNoUnknown;
+
+public final class EpiDataHelper {
+
+	public static YesNoUnknown getKindOfExposure (EpiDataDto dto) {
```
MateStrysewskeSym: As far as I understand it, this method is used to determine whether the "kindOfExposure" fields are shown in the UI; please rename it to something like "hadExposureToAnimals"; also, if it's used for this purpose, it's probably better to make this method return a boolean. If I'm wrong, please explain what this method is supposed to do

sormas-api/src/main/resources/captions.properties
```diff
@@ -1146,6 +1162,25 @@ Symptoms.unexplainedBleeding = Bleeding or bruising
...
+Symptoms.excitation = Excitation
+Symptoms.dysphagia = Difficulty in swallowing (Dysphagia)
+Symptoms.aerophobia = Fear of flying (Aerophobia)
+Symptoms.hyperactivity = Hyperactivity
+Symptoms.paresis = Paresis
+Symptoms.agitation = Agitation
+Symptoms.ascendingFlaccidParalysis = Ascending flaccid paralysis
+Symptoms.erraticBehaviour = Erratic behaviour
+Symptoms.coma = Coma
+Symptoms.death = Death
```
MateStrysewskeSym: I didn't see a symptom called "Death" (and it wouldn't make sense to have one anyway); remove this

sormas-api/src/main/java/de/symeda/sormas/api/epidata/EpiDataHelper.java
faatihi: I've put some thought into your suggestion. I think we should keep the name and return type as is. This method is used to tell whether any `kindOfExposure` field has been ticked. 
If none is ticked, we set the `kindOfExposure` field to UNKNOWN and hide all related fields. 
If any field is ticked, we set it to YES and show all related fields.

With your suggestion, `hadExposureToAnimals` will only check whether any exposure field is set to `YES`. 
If so, it will show the `kindOfExposure` field, else it'll hide it.
In this way, when the user sets a field to `NO` or `UNKNOWN`, the system will still hide all the fields although some fields have been ticked (albeit not the `YES` option).

Maybe we can give this method a better name to describe what it does. Like `isAnyKindOfExposureFieldTicked` ;)

faatihi:
It's ready for review

sormas-api/src/main/java/de/symeda/sormas/api/epidata/EpiDataHelper.java
MateStrysewskeSym: Your suggestion makes sense, but makes me wonder whether this whole logic makes sense. Because, in theory, when all the individual `kindOfExposure` fields are set to `NO`, the custom `Kind of exposure` field should be set to `NO` as well; however, we typically hide depending fields when a parent field is set to `NO` and this would break that pattern.

Additionally, it's confusing that the caption of the generated field is `Kind of exposure`, but it's a `YesNoUnknown` field; As far as I understand it, the question whether the patient was exposed to animals in the first place is answered with the fields `dogs`, `cats`, `canidae` etc., right? Maybe it's a better idea to make the visibility status of the individual `kindOfExposure` fields dependent on whether one of these animal fields is set to `YES` and get rid of the generated `kindOfExposure` field completely? What do you think?

MateStrysewskeSym:
@faatihi I've commented on the `kindOfExposure` issue above. Please let me know what you think.

sormas-api/src/main/java/de/symeda/sormas/api/epidata/EpiDataHelper.java
faatihi: The whole idea of visibility toggling is to avoid showing certain options unnecessarily. 
I'm thinking if we get rid of the kindOfExposure logic, we'll have to always show the dependent fields irrespective of their values. This is true, at least, for the create and edit views.
If it is okay to always show all the exposure fields, then we can take this route.

In another perspective, your initial comment is making a comeback. Ideally, "Kind of exposure" is a section that can be expanded or collapsed. It is initially (on loading of the view) collapsed to make room for other fields. If the user wants to fill those fields, he can expand it and tick appropriate fields.
With this logic, dependent fields do not have to clearOnHide because the user is allowed to collapse and expand the section at will. However, on initial loading of the view, we expand it if any field in the section is filled.

MateStrysewskeSym: That sounds like a cool piece of logic! I wouldn't use a `YesNoUnknown` field for the expand/collapse part though, but probably just a label with an arrow to the left. This UI element needs to be created first, which is an effort we probably can't make right now because the top priority is to get the new features into the system by Friday. However, I'll make sure to keep it in mind!

In the meantime, I suggest to get rid of the generated `kindOfExposure` field and make the visibility of the specialised `kindOfExposure` fields dependent on the various fields that determine contact with animals (because it doesn't make sense to specify the kind of exposure if no contact with animals took place; this way, we make sure to always hide these fields if they're not relevant at all). This also means that you can keep the method in the `EpiDataHelper`, but only have to change the fields that are taken into account. The new name you've already given it also makes sense in this context.

Are you okay with that approach for now?

faatihi: Yep. Sounds good.

PR #1440, opened 2020-01-02
Android change label community to subdistrict

faatihi:
This is a fix to a reported issue. On creating a new case, the community field is labeled "Community". Ghana, however, prefers "Sub District".
I looked into the source and found that the Web is able to translate fine using the `community `entry in captions.properties. For the android, it requires an explicit `CaseData.community` entry.
I went ahead to add entries for "Region" and "District".
If SORMAS repo is strictly issue-driven, pls let me know so I open a new issue and reference it from this pull request.


MateStrysewskeSym:
@faatihi I can't reproduce this problem on my development system. When I set the locale to en_gh in my properties, restart the server and re-install the app, the caption is "Sub District".

Did your server run with the en_ng locale at any point and is it possible that the affected tablets have been set up during this time?

faatihi:
I reproduced it on my local. I'll, however, re-install the app to see what happens.

faatihi:
Hi @MateStrysewskeSym , I have also seen the "Sub District" label. 
I don't know what happened, but I think updating sormas on my local and reinstalling the android app might have helped.

I'll pull down this request then.

Thanks.

PR #1732, opened 2020-04-03
Refactor 1597 user UUID

BarnaBartha:


HolgerReiseVSys:
That's definately going in the right direction.
Here's a few points you might want to consider:

* `PathogenTestFacade::getNewTestResultsForDashboard` has missed; is only used in test, maybe it can be removed
* (`PathogenTestFacade::getByUuids` is only used in tests, maybe it can be removed)
* `TaskFacade::getAllByUserForDashboard` is only used in tests, maybe it can be removed
* `TaskFacade::getIndexList` still has `userUuid`
  I did not go looking for other instances of missed userUuid parameters.

* `ContactFacadeEjb::count`, `::getContactFollowUpList` etc.: 
	Checks for `user==null` have been dropped, but that should hopefully be OK.

* Service calls should (usually?) not need the current user as parameter, e.g.
	`AbstractAdoService::createUserFilter`
	`ContactService::getAllActiveUuids`
	`ContactService::getAllActiveContactsAfter`
  Maybe `AbstractAdoService` should have `Instance<CurrentUser`> injected and maybe have some convenience methods.
  I'm not sure how that works out with `UserService`.

* Maybe instead of using `CurrentUserQualifier`, `CurrentUser` could be an interface.

* Instead of having `SessionContext` injected into the `UserService`,
  maybe it could be injected as parameter of `produceCurrentUser`.
  It looks like even just the `Pincipal` can be injected: https://stackoverflow.com/questions/36984056/shiro-cdi-inject-security-principal-works-on-glassfish-but-not-on-wildfly

* Even though it should not be a problem to initialize the user entity in the producer method,
  this means that every call needs an additional transaction. I'm not sure, but maybe `@Transactional` could help.

* Javadoc explaining the purpose of `CurrentUser` would be helpful.

* I have a bit mixed feelings about changing to lambdas instead of anonymous classes as listeners.
  It's great to have that cleaned up, but the mixup of the refactorings complicated the review.

* `de.symeda.sormas.ui.AbstractBeanTest::init`: better use value from `user.getUserName()` instead of `"admin"` (guards against changes to the producer)

HolgerReiseVSys:
I've just looked thru the commits, and didn't try it out.
Except for one point this all looks reasonable:
`UserFacade::getByUuid` realy needs `uuid` parameter, it is also used for user management.
Also the Javadoc of `CurrentUser` states "The class getCurrentUser()."

Issue #1281, opened 2019-08-27
Absence of ''Contacts" attribute in statistics menu 


Iheanacho2027:
There is no "contacts" attribute in statistics menu filtering options. There is a need for this because in future, data management analyst may need to carry out statistical analysis as regards the number of followed-up and unfollowed contacts available on the SORMAS platform. There should be also be sub-attributes for followed-up, unfollowed and lost to follow-up contacts
![image c3](https://user-images.githubusercontent.com/50739609/63804158-71f33480-c90e-11e9-82e7-e8a77aefd3c1.JPG)

MateStrysewskeSym:
@Iheanacho2027 That's because the statistics section right now exclusively supports cases, none of the other entities in SORMAS. So this is more a matter of extending the statistics to also include information about contacts which I absolutely see the reasoning for. I've renamed the issue accordingly.

Iheanacho2027:
Also in addition to this, can we also have information on sample results in the statistics menu

foyiri:
yes the lab forks should be able to carry out sample analysis on the platform like how many sample were received how many were test  etc

faatihi:
users should be able to extract statistics of contacts **per region/district** for a particular disease

faatihi:
An example of statistics that might be needed from contacts

![image](https://user-images.githubusercontent.com/1176692/77935723-5b94c900-72a1-11ea-9921-59b80e719666.png)

faatihi:
I've come up with two UI design proposals

1. Tabbed statistics

![tabbed](https://user-images.githubusercontent.com/1176692/78137476-eb0eb900-7414-11ea-9d8f-852c380af4ba.png)

2. Using grouped-buttons

![grouped-buttons](https://user-images.githubusercontent.com/1176692/78137519-fc57c580-7414-11ea-8e9f-f4a8cbb5e201.png)

I would like second/third opinions

MartinWahnschaffeSymeda:
Clear vote on the first option (tabbed)

MateStrysewskeSym:
Seconding this.

Iheanacho2027:
I prefer the second because it looks more detailed

foyiri:
I think the second looks more better

MateStrysewskeSym:
@Iheanacho2027 They're both exactly the same, the only difference is that in option A), you choose whether to get case or contact statistics via a tab menu while in option B), you make the same choice in a switch field.

[...]

PR #1779, opened 2020-04-08
Contact statistics

faatihi:
#1281


faatihi:
It's left with the legend labels of the map. Review may proceed whiles I figure that out.

MateStrysewskeSym:
@faatihi The TravisCI build failed because apparently you used Lambda expressions in the sormas-api project. Unfortunately, because the api project is also used in the Android app and Android still officially only supports Java 7, using Java 8 features is not allowed there. I'll start the review as soon as that's fixed.

faatihi:
OK. On it

faatihi:
@MateStrysewskeSym , it's ready

MateStrysewskeSym:
@faatihi Thanks! I don't think I'll be able to do the review before Tuesday because Germany has holidays tomorrow and on Monday.

faatihi:
That's ok. We're also off till Tuesday.

Issue #1862, opened 2020-04-20
Dashboard should include all cases by default and not filtered by epiweek


faatihi:
### Feature Description
The current dashboard by default only includes cases that were reported in the current epiweek. User has to adjust the filters to get the overall figures on the dashboard.

### Problem Description
The need to always adjust the filters to get the overall figures is not ideal.

### Proposed Change
The dashboard should show figures similar to the way the online covid-19 dashboards do, that is, include all cases since the onset of the epidemic/pandemic. This way, user does not have to adjust the date filters to get these figures

MartinWahnschaffeSymeda:
@faatihi @bernardsilenou 
We need to find another solution for this.
WHen "since outbreak onset" is selected, the pull request implementation will always show ALL cases of all disease in the top part of the dashboard and have individual time filters for the bottom part of the dashboard.

It is important to understand that the dashboard in SORMAS is not for a single disease only, but to give an overview on the situation for the whole country. 

Besides that you are right, there is a big need to make it as easy for users as possible to look into the time filter that they are currently interested in - and that is define by the onset of COVID these days.

We need to have a consistent time filter here. There are three options that come to mind:

1. Define a default time filter for the dasboard in the sormas.properties for the whole server that can be activated in times like these
2. Store the picked date filter option for the current user
3. Encode the filters in the URL, so users can use the url to always open the dashboard for the same time filter

bernardsilenou:
> WHen "since outbreak onset" is selected, the pull request implementation will always show ALL cases of all disease in the top part of the dashboard and have individual time filters for the bottom part of the dashboard. _(I think this will be too heavy for dashboard and it may crash as in the case of Nigeria )_
> 
> It is important to understand that the dashboard in SORMAS is not for a single disease only, but to give an overview on the situation for the whole country. _( This is the goal but focusing on lower part or hiding the upper part will give us disease specific dashboard only for a particular time, not for all time from onset of used on the tool)_
> 
> Besides that you are right, there is a big need to make it as easy for users as possible to look into the time filter that they are currently interested in - and that is define by the onset of COVID these days.
> 
> We need to have a consistent time filter here. There are three options that come to mind:
> 
> 1. Define a default time filter for the dasboard in the sormas.properties for the whole server that can be activated in times like these _( this will also need to be disease specific and will not make sense for regions that got affected 9  months after the first region in the country, so I will not vote for this option)_
[...]

PR #1883, opened 2020-04-21
Dashboard no date filters by default

faatihi:
#1862


faatihi:
@MateStrysewskeSym I spoke with @bernardsilenou and we think the Outbreak entity is complex for determining outbreak onset of a disease at the national level. I started working on it as you proposed but noticed that if we're to use this entity, we'll have to find the earliest date among all the districts for each disease. I'm not certain we would want that.

Also, as you pointed out, this request could slow down the dashboard so it shouldn't be the default for all users. 

@bernardsilenou  suggested we have a configuration for individual countries to set a date of outbreak for each disease and also a configuration to set this date as the dashboard's default filter for National Observers. This way, individual countries can configure the dashboard for their National Observers whilst the rest of the users keep the dashboard as it is.

bernardsilenou:
@MateStrysewskeSym As Faat already mentioned, the goad is to prevent the national observer to reset the dashboard daily.  Admins should be able to configure fixed filter per disease for them.

We though of having a parameter to set for only National observers and admins to set the onset of when to view their dashboard per diseases. This can be an additional table under configurations. If this is not the best option, then you can suggest another please

PR #1904, opened 2020-04-23
1647 prevent users to edit cases and contacts

sergiupacurariu:
Hi Mate, I also fix an issue I've found on creating Points of entry


MartinWahnschaffeSymeda:
@sergiupacurariu 

- [ ] case region and district are mandatory fields and will always match the health facility/point of entry, so there is no need for the getCaseRegion/getCaseDistrict methods
- [ ] we need to distinuigh between hospital informant and community informant. I have updated the user access doc accordingly
- [ ] For the user role check it would be perfect if we could move the logic into two new method in ConfigProvider, so you can do a call like: ConfigProvider.hasUserRole(UserRole.HOSPITAL_INFORMANT) and ConfigProvider.hasUserRole(UserRole::isInformant)

Issue #1865, opened 2020-04-20
Automatically pick current location coordinates when creating case and contact


faatihi:
### Situation Description
Users usually forget to enter location coordinates when creating a case. Same happens with creating a contact.

### Feature Description
The system should automatically pick current location for the coordinates. Users can modify the coordinates later on. This way, all cases will have coordinates and will show on the map.

MateStrysewskeSym:
@faatihi The mobile app already does that automatically - what exactly is missing here?

faatihi:
We have a lot of cases and contacts without coordinates. We looked into ways to keep that in check. One option is to put the picker on the first page of CreateCase/CreateContact and make it the first field so that users may see it first and pick the coordinates.
The second option is to pick the coordinates automatically and allow user to make changes.
Third is to have a new field that keeps the auto-picked location so that if user forgets to enter the coordinates, the auto-picked one will be used for the maps.

> Picking the current coordinates for the person's address seems not like a good idea to me

Users who remember to pick a location can do so to overwrite the auto-picked one.

MateStrysewskeSym:
@faatihi We obviously can't automatically pick locations for cases created in the web because it's very likely that the device used to create those cases does not have a GPS module. We'll talk about the other suggestions in our next sprint review.

faatihi:
I agree with you. This is about the mobile app.

bernardsilenou:
Faat and I had a call and I suggested that we do not activate automatic picking of coordinates as will significantly affect the epidemiology of the disease in terms of space. The reason is that, the place of creating a case is not always the address of the case nor case person.  
The need for this feature was because many cases without health facilities did not show up on map.
Here are possible solutions we decided:
1, For cases with a HF, we use the geo coordinate of the HF (already implemented )
2, For cases at home, we use the centroid of the district of the case with random error. For example cases at home with coordinate missing will have coordinate d + ei, where d is the centroid of the district and ei is small random error lying in the districts. ei is just to prevent all case from having the same coordinate.
3, Add a reminder for users to update and correct this predicted coordinate, twice or once a week
3, Add coordinate to the list of variable in calculating the completeness of the case. This will remind users to always complete it,

MartinWahnschaffeSymeda:
@faatihi Based on what Bernard is saying: Will you create separate issues for those missing things, please?
I guess the pul request for the automatic location picking can be deleted, right?

faatihi:
[...]

PR #1908, opened 2020-04-23
#1865 App feature - automatic location picking

opatajoshua:
Automatically pick current location coordinates when creating case and contact
app should initially set current address when person address is null
save a contact person's address to current location after follow up


faatihi:
#1865

bernardsilenou:
@faatihi The address of the person is not always where the case is being created. Picking the gps address when a case is created is not prone to errors? How does this work please?

MateStrysewskeSym:
@bernardsilenou I'm also not a fan of this change. However, it's a request from Ghana. I suggest to discuss this in the review next week.

opatajoshua:
> In this state, the feature doesn't make sense to me. First of all, the address location right now is always overridden, even if it is already set, which means that during every change, the potentially correct person address is removed. There needs to be a check whether the person address already has GPS coordinates. Secondly, the code to set the person address should not be executed in the fragment, especially not in a binding method. It's fine to do this during a save, but it should then be done when saving a NEW case, contact or visit, not when saving an existing one because it's extremely likely that those aren't done where the person actually lives.

Will make the necessary changes and submit

MateStrysewskeSym:
@opatajoshua Alright, thanks!

bernardsilenou:
@opatajoshua It is not a good practice to implement automatic pecking of coordinates for cases or contacts. I know this is requested but implementing it this way will affect the epidemiology behind the special spread of the diseases on the map. I commented on #1865  on  how to walk around it.

Issue #1910, opened 2020-04-23
Dashboard map should have case classification and interactive date filters


faatihi:
### Situation Description
Contacts on the dashboard map can be filtered by their classification. Cases, however, can't be filtered that way. 
Also, supervisors would like a more interactive way of visualizing progress (or otherwise) by period.

### Feature Description
Give user an option to filter markers by case classification.
Also, the map could use a date filters aside from the one for the whole dashboard. One that can show progress incrementally as time progresses.

PR #1911, opened 2020-04-23
Case classification options on dashboard map 

faatihi:
#1910


MateStrysewskeSym:
@faatihi Adding the case classification filters makes total sense to me. What exactly is the purpose of the date filters though? The map is supposed to give an overview of the situation at whatever date period has been selected in the top filters, why is there a need for another date filter?

Independent of that, the new date filters currently don't work for me. The second dropdown is always empty.

faatihi:
The period filter is there to provide an interactive way of visualizing progress (or otherwise) by period.
For example, there were 5 cases in January, then it increased to 12 cases in February.

faatihi:
> Independent of that, the new date filters currently don't work for me. The second dropdown is always empty.

This fix should resolve this.

MartinWahnschaffeSymeda:
@faatihi Unfortunately it's still not filling, when the map ony contains health facilities.
In updatePeriodFilters you need to use mapAndFacilityCases instead of mapCaseDtos (there are multiple occurences that need to be replaced).

Besides that it looks fine.

I have to add, that I share Maté's view: Wouldn't it be a lot easier to use the existing time filter component? 
You could add an error button to the left and right of it to easily go back and forth in time.

grafik

faatihi:
Arrow buttons is a great idea. 
With regards to the location, we would like the map to have a similar "grouped periods" as the EpiCurve. This is why it's only implemented in the map. 
Additionally, this implementation doesn't have to refresh the whole dashboard. It doesn't hit the database either. It only provides grouped periods within the dashboard's filtered period.

faatihi:
> In updatePeriodFilters you need to use mapAndFacilityCases instead of mapCaseDtos (there are multiple occurences that need to be replaced)

@MartinWahnschaffeSymeda thanks for pointing this out

faatihi:
I have two more tasks to complete:

- [ ] resolve unexpected result from `int DateHelper.getWeeksBetween(Date, Date)`

- [ ] add an option to **(a)** show only new cases for a selected period **(b)**show all cases upto the selected period

faatihi:
With regards to the getWeeksBetween,

When I give it "2020-04-18 02:15:24.482" and "2020-04-27 12:23:53.207", I expect it to return 3.
That is:
Wk 16, Apr 13 - 19
Wk 17, Apr 20 - 26
Wk 18, Apr 27 - May 3

But it returns 2. That is, Wk 16 and 17. Is that expected?

faatihi:
Also, the option "Show for confirmed cases only" is selected by default.
Is that okay?

MartinWahnschaffeSymeda:
> 
> 
> With regards to the getWeeksBetween,
> 
> When I give it "2020-04-18 02:15:24.482" and "2020-04-27 12:23:53.207", I expect it to return 3.
> That is:
> Wk 16, Apr 13 - 19
> Wk 17, Apr 20 - 26
> Wk 18, Apr 27 - May 3
> 
> But it returns 2. That is, Wk 16 and 17. Is that expected?

You are right. This seems to be a bug. Looks like the method is also not unit tested.
I guess we have the same problem for getMonthsBetween.

The map should show all cases by default. We have color coding to identify the different types of cases. I can't make a decision on this one, but for now we should leave it as is (all) and just offer the option to look into confirmed cases.

faatihi:
Thanks. I'll show all cases by default. We'll tackle the date bug in a separate issue.

Issue #1948, opened 2020-04-29
Adding personAddress field to both add and edit forms of Case and Contact


opatajoshua:
### Situation Description
Users usually forget to enter location coordinates when creating a case. Same happens with creating a contact.

### Feature Description
Add and edit forms of contact and case should include personAddress field

### Additional Information
included edit forms for consistency

PR #1949, opened 2020-04-29
Added personAddress field to both add and edit forms of Case and Contact

opatajoshua:
#1948

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseEditFragment.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseEditFragment.java
@@ -50,11 +50,13 @@ import de.symeda.sormas.app.backend.classification.DiseaseClassificationAppHelpe
 import de.symeda.sormas.app.backend.classification.DiseaseClassificationCriteria;
 import de.symeda.sormas.app.backend.common.DatabaseHelper;
 import de.symeda.sormas.app.backend.config.ConfigProvider;
+import de.symeda.sormas.app.backend.location.Location;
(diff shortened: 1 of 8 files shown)


MateStrysewskeSym:
@opatajoshua Adding the person address to the case or contact form doesn't seem like a good idea to me. The case and contact are semantically different from the person. We'll discuss this with the HZI in our review next week.

faatihi:
@MateStrysewskeSym what was the outcome of the meeting?
The reason for this change is to help officers enter the coordinates.
What happens now is, officers only focus on the first page of the form.
Once they're done with the first page, they do not pay much attention to subsequent pages.
For that reason, we want the most important fields on the first page.

I think its a user-friendly approach to have important fields on the first page, irrespective of the entities those fields belong.

MateStrysewskeSym:
@faatihi I honestly think this is a matter of training. It's basically impossible to fit all important fields into the first page (you could assume that symptoms, for example, are also very important), and the danger of mixing up the case jurisdictions with the person's home address is too big in this case. Users already have problems distinguishing cases and persons because they often think they're the same. This doesn't really make it better.

Unfortunately I didn't have a chance to talk to @bernardsilenou yet, but maybe he can have a quick look at our discussion here and let us know his opinion. The proposed change is that the person's address is added to the case data and contact data pages (for the reasoning see the comment by @faatihi, for my concerns see what I've written above).

bernardsilenou:
@All,
I also think this is a matter of training. Adding person information to the cases and contacts will also distort structure of the system.

faatihi:
OK. We'll drop this and work on adding the person address to case completeness calculation, as suggested by @bernardsilenou

PR #1967, opened 2020-05-04
1847 dom unique ids

leventegal-she:
- introduced AbstractFilterForm(by re-using code from AbstractEditForm) and used on most of the pages. 
- the filters on some pages behave a bit differently, there I just added the id-s manually. I didn't want to change the functionality
- Introduced helper functions for button creation to ensure that id is set for them. Changed all buttons to be created using the helper methods
- introduced helper function for bulk actions dropdown creation


ChristopherRiedelSymeda:
- [ ] add ids to the "tabs"

Issue #1923, opened 2020-04-27
LocationEditForm: ConversionException with alphabetic chars [0.5]


StefanKock:
### Bug Description
When typing manually GPS coordinates I can produce Exceptions when entering invalid numbers (like letters), but that was already an issue before.

### Steps to Reproduce
1. Open view CASE PERSON in a case.
2. Type GPS coordinates with letters and leave field.

### Expected Behavior
- No exception, but a meaningful validation message.

### System Details
* SORMAS version: 1.37,x, 1.38.x

### to be reviewed by
@leventegal-she

PR #2007, opened 2020-05-07
#1923 validate long/lat fields before updating the leaflet map popup on LocationEditForm

leventegal-she:
This way it doesn't throw exception and validation message appears on the input fields


sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -180,6 +181,10 @@ private void triggerGeocoding() {
 		}
 	}
 
+	private boolean areFieldsValid(String... propertyIds){
```
MartinWahnschaffeSymeda, line 184: Can be moved to AbstractEditForm

Issue #2017, opened 2020-05-08
NullPointerException when entering EpiDataView


MateStrysewskeSym:
```
java.lang.NullPointerException: null
at de.symeda.sormas.ui.epidata.EpiDataView.enter(EpiDataView.java:40)
at com.vaadin.navigator.Navigator.performNavigateTo(Navigator.java:778)
at com.vaadin.navigator.Navigator.lambda$navigateTo$9a874efd$1(Navigator.java:702)
at com.vaadin.navigator.Navigator.runAfterLeaveConfirmation(Navigator.java:726)
at com.vaadin.navigator.Navigator.navigateTo(Navigator.java:701)
at com.vaadin.navigator.Navigator.navigateTo(Navigator.java:678)
at com.vaadin.ui.UI.doInit(UI.java:781)
at com.vaadin.server.communication.UIInitHandler.getBrowserDetailsUI(UIInitHandler.java:218)
at com.vaadin.server.communication.UIInitHandler.synchronizedHandleRequest(UIInitHandler.java:76)
at com.vaadin.server.SynchronizedRequestHandler.handleRequest(SynchronizedRequestHandler.java:40)
at com.vaadin.server.VaadinService.handleRequest(VaadinService.java:1602)
at com.vaadin.server.VaadinServlet.service(VaadinServlet.java:445)
at javax.servlet.http.HttpServlet.service(HttpServlet.java:791)
...
at java.lang.Thread.run(Thread.java:748)
```

StefanKock:
@MartinWahnschaffeSymeda please have a look on this: What is the architectual approach when a case dependent view is opened without selecting a case?

MateStrysewskeSym:
@StefanKock @MartinWahnschaffeSymeda @leventegal-she I think the user should be redirected to the case directory (which is already happening) without an error being thrown. I also don't really think we should invest time into displaying a notification message because normally users should not manually fiddle with URLs anyway.

StefanKock:
If the fix of @leventegal-she is appropriate and automatically redirecting to tha case directory, can this be generalised in `AbstractCaseView`? Since other sibling view throw NPEs as well (but not that often recently).

leventegal-she:
I think generalizing in `AbstractCaseView` is not enough because not only the case pages throw NPEs so it needs to be generalized on a higher level, but I think this should not be the scope of this bug ticket. 

@StefanKock @MateStrysewskeSym

StefanKock:
[...]

PR #2022, opened 2020-05-08
#2017 fixed null pointer exception on epi data view

leventegal-she:
the NullPointerException was thrown probably due to having an inexistent case id in the page URL


StefanKock:
Discussed with @leventegal-she: I will do some refactoring to solve this issue for all sibling views with need a case and we will review that.

PR #2028, opened 2020-05-08
1944 open read view for fragments with no user access

sergiupacurariu:
the branch contains only the fix which prevents restricted user to access a contact from within CaseEditContact list. Edit allowed must be kept as the user still needs to be able to add new contacts. I'll move the generic part on a different issue.


MartinWahnschaffeSymeda:
@sergiupacurariu Ah. That's actually a fix for a problem I wasn't aware of yet. That's very good!

The actual problem I meant is: When you are within the contacts list of the case you can switch to the edit view (which is the way it's supposed to be). When you now switch to another form of the case (e.g. symptoms) you enter it in edit view. Instead the system should recognize that it needs to be opened in read view.

PR #2058, opened 2020-05-11
2017 Avoid NPE in case views

StefanKock:
I tested my changes: It redirects every case centric view to the cases list when no case identifier is present in the url or is messed up manually. This problem might have been also happen when calling an url for a deleted case.


StefanKock:
I'm doing a rebase and merging it myself after resolving the conflicts.

PR #2080, opened 2020-05-12
1591 prevent users to edit cases and contacts web

sergiupacurariu:


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseEditAuthorization.java
```diff
@@ -32,7 +32,7 @@ public Boolean caseEditAllowedCheck(Case caze) {
 
 		User user = userService.getCurrentUser();
        
-        if (DataHelper.equal(user.getUuid(), (caze.getUuid()))) {
+        if (DataHelper.equal(user.getUuid(), (caze.getReportingUser().getUuid()))) {
```
MartinWahnschaffeSymeda, line 35: To be safe, let's add a null check to the reporting user here. Same in the ContactEditAuthorization

sergiupacurariu: done

Issue #2128, opened 2020-05-18
Navigation to sample list does not work


BarnaBartha:
### Bug Description
Navigation to sample list does not work when a sample was opened automatically.

### Steps to Reproduce
1. Go to samples directory and adapt the filters such that only one sample results after the search. -> sample automatically opens with a message containing null.

2. Navigate back to the samples list - not possible. Message with null is show again.

### Expected behaviour
Message should not contain "null"
Navigation back to samples directory should be possible. Maybe with reseted filters?

### Screenshots
![image](https://user-images.githubusercontent.com/8166805/82184833-efbdfc80-98f0-11ea-98f2-d148b50e5db8.png)

### System Details
* Device: Browser
* SORMAS version: 1.41.0-SNAPSHOT
* Android version/Browser: Chrome

PR #2134, opened 2020-05-18
Feature 1753 samples to contacts

BarnaBartha:


sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactService.java
```diff
@@ -17,44 +17,9 @@
...
-import de.symeda.sormas.api.contact.ContactClassification;
-import de.symeda.sormas.api.contact.ContactCriteria;
-import de.symeda.sormas.api.contact.ContactLogic;
-import de.symeda.sormas.api.contact.ContactProximity;
-import de.symeda.sormas.api.contact.ContactReferenceDto;
-import de.symeda.sormas.api.contact.ContactStatus;
-import de.symeda.sormas.api.contact.DashboardContactDto;
-import de.symeda.sormas.api.contact.FollowUpStatus;
-import de.symeda.sormas.api.contact.MapContactDto;
+import de.symeda.sormas.api.contact.*;
```
MateStrysewskeSym, line 22: Same as before; please make sure to remove these imports from all classes where you added them

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleFacadeEjb.java
```diff
@@ -212,33 +206,67 @@ public SampleReferenceDto getReferenceByUuid(String uuid) {
...
+
+		final Join<Sample, Contact> contact            = qc.getJoin(Sample.class, Contact.class);
+		final Join<Contact, Person> contactPerson           = qc.getJoin(Contact.class, Person.class);
+		final Join<Contact, Region> contactRegion      = qc.getJoin(Contact.class, Region.class);
+		final Join<Contact, District> contactDistrict  = qc.getJoin(Contact.class, District.class);
+		final Join<Contact, Case> contactCase          = qc.getJoin(Contact.class, Case.class);
+		final Join<Case, Region> contactCaseRegion     = qc.getJoin(Case.class, Region.class, "contactCaseRegion");
+		final Join<Case, District> contactCaseDistrict = qc.getJoin(Case.class, District.class, "contactCaseDistrict");
+
+		final Expression diseaseSelect = qc.addExpression("disease",
```
MateStrysewskeSym, line 234: Never use hard-coded strings here because they might break if the entity fields are renamed; always use the String constants from the entity classes instead (e.g. Case.DISEASE)

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleFacadeEjb.java
```diff
@@ -630,19 +671,31 @@ private void onSampleChanged(SampleDto existingSample, Sample newSample) {
...
-					messagingService.sendMessage(recipient, I18nProperties.getString(MessagingService.SUBJECT_LAB_SAMPLE_SHIPPED), 
-						String.format(I18nProperties.getString(MessagingService.CONTENT_LAB_SAMPLE_SHIPPED_SHORT), 
-							DataHelper.getShortUuid(newSample.getAssociatedCase().getUuid())), 
-							MessageType.EMAIL, MessageType.SMS);
+					String messageContent = null;
+					if (newSample.getAssociatedCase() != null) {
+						messageContent =
+								String.format(I18nProperties.getString(MessagingService.CONTENT_LAB_SAMPLE_SHIPPED_SHORT),
+								DataHelper.getShortUuid(newSample.getAssociatedCase().getUuid()));
+					} else if (newSample.getAssociatedContact() != null){
```
MateStrysewskeSym, line 690: Please make sure to format your files

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -4248,5 +4248,11 @@ ALTER TABLE contact ADD COLUMN additionaldetails varchar(512);
 ALTER TABLE contact_history ADD COLUMN additionaldetails varchar(512);
 
 INSERT INTO schema_version (version_number, comment) VALUES (208, '2020-05-11 Add additionalDetails to contact #1933');
-                                                                                                                        
+
+-- 2020-05-07 Add samples to contacts #1753
+ALTER TABLE samples ADD COLUMN associatedcontact_id bigint;
```
MateStrysewskeSym, line 4253: Add the new column to the history table as well

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -4248,5 +4248,11 @@ ALTER TABLE contact ADD COLUMN additionaldetails varchar(512);
...
+ALTER TABLE samples ADD CONSTRAINT fk_samples_associatedcontact_id FOREIGN KEY (associatedcontact_id) REFERENCES contact (id);
+ALTER TABLE samples ALTER COLUMN associatedcase_id DROP NOT NULL;
+
+INSERT INTO schema_version (version_number, comment) VALUES (209, '-- 2020-05-07 Add samples to contacts #1753');
```
MateStrysewskeSym, line 4257: Remove the dashes

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleSearchType.java
```diff
@@ -0,0 +1,13 @@
+package de.symeda.sormas.api.sample;
```
MateStrysewskeSym, line 1: Rename this to something like SampleAssociationType - SampleSearchType suggests a very specific use case, but we might also need this for something else in the future

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1104,6 +1107,15 @@ public CaseDataDto saveCase(CaseDataDto dto, boolean handleChanges) throws Valid
 
 		caze = fillOrBuildEntity(dto, caze);
 
+		final ContactReferenceDto fromContactReference = dto.getFromContactReference();
```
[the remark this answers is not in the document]
BarnaBartha: The DTOs ContactDto and CaseDataDto don't have a list of samples, and I have not seen any example where we have/map the many to one associations to the DTO level.
That's why I thought it's best if I keep that the same and handle the association of the samples to the new case within the BE.
It is possible to move it to the CaseFaceadeEJb#buildFromContact - but that means i need to add the samples to the ContactDto and map that every time even though in all other cases it is not used.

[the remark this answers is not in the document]
MateStrysewskeSym: I think I'd prefer a separate setSampleAssociations method in the case facade that can be called once during the case creation process which basically does what you're currently doing in the saveCase method, because right now, the new fromContactReference field seems to be practically useless after the case has been created.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1531,6 +1531,34 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				case 201:
 					currentVersion = 201;
 					getDao(Contact.class).executeRaw("ALTER TABLE contacts ADD COLUMN additionalDetails varchar(512);");
+				case 202:
```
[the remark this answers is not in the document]
BarnaBartha: you're right, the dropping of the null constraint on the associated case was not working with sql lite
but it's actually not even there anyway so i will just add the new column

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleIndexDto.java
```diff
@@ -22,6 +22,7 @@
```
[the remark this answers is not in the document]
BarnaBartha: ok

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
```diff
@@ -256,6 +257,7 @@
 	private YesNoUnknown quarantineHomeSupplyEnsured;
 	private String quarantineHomeSupplyEnsuredComment;
 	private ReportingType reportingType;
+	private ContactReferenceDto fromContactReference;
```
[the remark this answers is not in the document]
BarnaBartha: renamed it to sourceContact

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -58,6 +58,9 @@
 import javax.transaction.Transactional.TxType;
 import javax.validation.constraints.NotNull;
 
+import de.symeda.sormas.api.*;
```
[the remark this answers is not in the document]
BarnaBartha: ok

sormas-backend/src/main/resources/sql/sormas_schema.sql
BarnaBartha: ok

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleSearchType.java
BarnaBartha: renamed

sormas-backend/src/main/resources/sql/sormas_schema.sql
BarnaBartha: ok

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleFacadeEjb.java
BarnaBartha: ok, actually here these are the aliases for the joins and expressions
but nonetheless they should not be hard coded extracted them to constants within the sample facade

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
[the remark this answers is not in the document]
BarnaBartha: removed it

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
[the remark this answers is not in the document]
BarnaBartha: ok, extracted it to separate method

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1532,35 +1532,10 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 					currentVersion = 201;
 					getDao(Contact.class).executeRaw("ALTER TABLE contacts ADD COLUMN additionalDetails varchar(512);");
 				case 202:
+					currentVersion = 202;
+					getDao(Sample.class).executeRaw("ALTER TABLE samples ADD COLUMN associatedcontact_id bigint REFERENCES contact (id);");
```
MateStrysewskeSym, line 1536: Please use camel case in the mobile app (so => associatedContact_id) - SQLite requires this

MateStrysewskeSym:
I've pulled and tested the pull request. The following changes and fixes are required:

- [ ] Pathogen tests need a refactoring. For example, when saving a pathogen test, an error is thrown because the method tries to retrieve the associated case from the backend. Please make sure that all logic associated with pathogen tests still works for samples that are exclusively assigned to a contact.
- [ ] Some pathogen test logic doesn't make sense for contacts. For example, when saving a pathogen test for the same disease as a case that is positive and verified, SORMAS asks the user whether the case should be automatically confirmed. This obviously doesn't make sense when the sample is not assigned to a case; we need to re-write this logic, but that should probably take place in another issue.
- [ ] You should fix #2128 with this pull request as well because it seems like this bug has been introduced by your changes (correct me if I'm wrong). At least it doesn't occur on the demo server.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -238,7 +238,8 @@ private void requestSampleCollectionTaskCreation(SampleDto dto, SampleEditForm f
 			@Override
 			public void buttonClick(ClickEvent event) {
 				popupWindow.close();
-				ControllerProvider.getTaskController().createSampleCollectionTask(TaskContext.CASE, dto.getAssociatedCase(), dto);
+				ControllerProvider.getTaskController().createSampleCollectionTask(TaskContext.CASE,
```
MateStrysewskeSym, line 244: The task context also needs to change to TaskContext.CONTACT if the sample is only assigned to a contact

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleGridComponent.java
```diff
@@ -106,12 +106,6 @@ public HorizontalLayout createFilterBar() {
 			if(!samplesView.navigateTo(criteria, false)){
 				filterForm.updateResetButtonState();
 				grid.reload();
-
```
MateStrysewskeSym, line 114: Please also removed the no longer needed string from the properties file and the getFirstItem() method from the FilteredGrid class if it's no longer used elsewhere

BarnaBartha: ok, done

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -241,8 +241,16 @@ private void requestSampleCollectionTaskCreation(SampleDto dto, SampleEditForm f
...
+				final CaseReferenceDto associatedCase = dto.getAssociatedCase();
+				if (associatedCase != null) {
+					ControllerProvider.getTaskController().createSampleCollectionTask(TaskContext.CASE,
+							associatedCase, dto);
+				}
+				final ContactReferenceDto associatedContact = dto.getAssociatedContact();
+				if (associatedContact != null) {
+					ControllerProvider.getTaskController().createSampleCollectionTask(TaskContext.CONTACT,
+							associatedContact, dto);
+				}
```
MateStrysewskeSym, line 253: I don't think two tasks should be generated because they're basically redundant. If a sample is assigned to both a case and a contact, only create a sample collection task for the case association.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/AbstractSampleView.java
```diff
@@ -48,13 +50,20 @@ protected AbstractSampleView(String viewName) {
...
 		menu.addView(SamplesView.VIEW_NAME, I18nProperties.getCaption(Captions.sampleSamplesList));
+
+		final SampleDto sampleByUuid = FacadeProvider.getSampleFacade().getSampleByUuid(params);
+		final CaseReferenceDto caseRef = sampleByUuid.getAssociatedCase();
 		if (caseRef != null && UserProvider.getCurrent().hasUserRight(UserRight.CASE_VIEW)) {
 			menu.addView(CaseDataView.VIEW_NAME, I18nProperties.getString(Strings.entityCase), caseRef.getUuid(), true);
 		}
+
+		final ContactReferenceDto contactRef = sampleByUuid.getAssociatedContact();
+		if (contactRef != null && UserProvider.getCurrent().hasUserRight(UserRight.CASE_VIEW)) {
```
MateStrysewskeSym, line 64: Wrong user right, should be CONTACT_VIEW

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/PathogenTestFacadeEjb.java
```diff
@@ -68,274 +72,316 @@
...
+        if (caze != null) {
+            region = caze.getRegion();
+            disease = caze.getDisease();
+            messageRecipients = userService.getAllByRegionAndUserRoles(region,
+                    UserRole.SURVEILLANCE_SUPERVISOR, UserRole.CASE_SUPERVISOR);
+        } else {
+            region = contact.getRegion();
+            disease = contact.getDisease();
+            messageRecipients = userService.getAllByRegionAndUserRoles(region,
+                    UserRole.SURVEILLANCE_SUPERVISOR, UserRole.CONTACT_SUPERVISOR);
```
MateStrysewskeSym, line 335: The recipients should not be if-else - if a sample is assigned to a case and a both contact, both case supervisors and contact supervisors should receive the message, not only one of them

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/PathogenTestFacadeEjb.java
```diff
@@ -68,274 +72,316 @@
...
+        final Disease disease;
+        final List<User> messageRecipients;
+
+        if (caze != null) {
+            region = caze.getRegion();
+            disease = caze.getDisease();
+            messageRecipients = userService.getAllByRegionAndUserRoles(region,
+                    UserRole.SURVEILLANCE_SUPERVISOR, UserRole.CASE_SUPERVISOR);
+        } else {
+            region = contact.getRegion();
```
MateStrysewskeSym, line 332: Might be null, you have to use the source case region and district as a fallback

sormas-backend/src/test/java/de/symeda/sormas/backend/sample/PathogenTestFacadeEjbTest.java
```diff
@@ -1,45 +1,155 @@
...
+        assertNotNull(updatedPathogen);
+        assertEquals(Disease.EVD, updatedPathogen.getTestedDisease());
+        assertEquals(PathogenTestType.OTHER, updatedPathogen.getTestType());
+        assertEquals(PathogenTestResultType.POSITIVE, updatedPathogen.getTestResult());
+        assertEquals("all good!", updatedPathogen.getTestResultText());
+        assertEquals(testDateTime, updatedPathogen.getTestDateTime());
+        assertTrue(updatedPathogen.getTestResultVerified());
+    }
+
+    private PathogenTestDto createPathogenTestDto(RDCFEntities rdcf, UserDto user, SampleDto sample, Disease disease) {
```
MateStrysewskeSym, line 140: Move this to TestDataCreator and change the method name to "buildPathogenTestDto" to hint at it not being saved to the database automatically

sormas-ui/src/test/java/de/symeda/sormas/ui/TestDataCreator.java
```diff
@@ -253,7 +253,8 @@ public PathogenTestDto createPathogenTest(SampleReferenceDto sample, PathogenTes
 		PathogenTestDto sampleTest = PathogenTestDto.build(sample, labUser);
 		sampleTest.setTestType(testType);
 		sampleTest.setTestDateTime(testDateTime);
-		sampleTest.setLab(lab);
+		sampleTest.
+				setLab(lab);
```
MateStrysewskeSym, line 257: How did this line break happen? :D

BarnaBartha: was testing if you catch me :)
joking.. i have no clue :D , reverted it

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/PathogenTestFacadeEjb.java
```diff
@@ -319,62 +318,65 @@ private void onPathogenTestChanged(PathogenTestDto existingPathogenTest, Pathoge
...
-            messageRecipients = userService.getAllByRegionAndUserRoles(region,
+            sendMessageOnPathogenTestChanged(existingPathogenTest, newPathogenTest, caze, contact, disease, messageRecipients);
+        }
+
+        if (contact != null){
+            final Region region = contact.getRegion() != null ? contact.getRegion() : contact.getCaze().getRegion();
+            final Disease disease = contact.getDisease() != null ? contact.getDisease() : contact.getCaze().getDisease();
+            final List<User> messageRecipients = userService.getAllByRegionAndUserRoles(region,
                     UserRole.SURVEILLANCE_SUPERVISOR, UserRole.CONTACT_SUPERVISOR);
+            sendMessageOnPathogenTestChanged(existingPathogenTest, newPathogenTest, caze, contact, disease, messageRecipients);
```
MateStrysewskeSym, line 335: With this solution, surveillance supervisors will receive the mail twice. The message recipients should only be built once and contain all user roles that should receive the message.

PR #2168, opened 2020-05-25
Feature 1638 simplify sample creation

BarnaBartha:


sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleDto.java
```diff
@@ -106,6 +106,8 @@
 	private String requestedOtherPathogenTests;
 	private String requestedOtherAdditionalTests;
 
+	private PathogenTestDto defaultTest;
```
MateStrysewskeSym, line 109: Please remove this field; it doesn't serve any purpose after the sample has been initially created. We should save the pathogen test independently instead during sample creation instead.

sormas-backend/src/main/java/de/symeda/sormas/backend/util/DtoHelper.java
```diff
@@ -70,7 +70,11 @@ public static void fillDto(EntityDto dto, AbstractDomainObject entity) {
 
 				Object targetValue = pd.getReadMethod().invoke(target);
 				Object sourceValue = pd.getReadMethod().invoke(source);
-				
+
+				if (sourceValue == null) {
```
MateStrysewskeSym, line 74: This is really, REALLY bad! It essentially makes it impossible to override target values with null in case they have been removed from the source, which is perfectly possible and viable. This change would result in really problematic bugs.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -135,6 +137,23 @@ public void onDiscard() {
 		VaadinUiUtil.showModalPopupWindow(createView, I18nProperties.getString(Strings.headingReferSample));
 	}
 
+	private void fillPathogenFields(SampleCreateForm createForm, SampleDto dto) {
```
MateStrysewskeSym, line 140: Rename to "fillPathogenTest" and add a PathogenTestDto as a parameter (since the SampleDto won't have the PathogenTestDto as a field itself); the current method name indicates that the fields in the UI are filled while it's actually the other way around.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -143,22 +162,19 @@ public void onDiscard() {
...
-					if (dto.getSpecimenCondition() != originalDto.getSpecimenCondition() &&
-							dto.getSpecimenCondition() == SpecimenCondition.NOT_ADEQUATE &&
-							UserProvider.getCurrent().hasUserRight(UserRight.TASK_CREATE)) {
-						requestSampleCollectionTaskCreation(dto, form);
-					} else {
-						Notification.show(I18nProperties.getString(Strings.messageSampleSaved), Type.TRAY_NOTIFICATION);
-					}
+		editView.addCommitListener(() -> {
+			if (!form.getFieldGroup().isModified()) {
+				SampleDto dto1 = form.getValue();
```
MateStrysewskeSym, line 167: Not a good naming pattern; "changedDto" or "updatedDto" might be an appropriate name, but just adding a number to it in a place where the code logic is actually relevant makes it hard to read the code.

sormas-backend/src/main/java/de/symeda/sormas/backend/util/DtoHelper.java
BarnaBartha: was added as a fix for some errors i didn't understand at the moment and i forgot it there
will revert

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
BarnaBartha: i've refactored a bit the whole save sample part - this method is now integrated there

PR #2303, opened 2020-06-11
Refactor 2065 fix open view with bad UUID exception

BarnaBartha:


StefanKock:
I have some concerns how you did it:

1. You removed the `redirectSimpleModeToCaseDataView` from case-dependent views without explanation.
2. You altered views that do not use UUIDs as parameters (`AbstractConfigurationView`, `AbstractStatisticsView` and their subclasses).
3. It feels still quite smelly how often uuids are read out of the parameters.

We should talk about this (I think tomorrow).

Issue #2223, opened 2020-05-29
Filters to add to the event directory


bernardsilenou:
### Situation Description
- The event directory is lacking filters making it difficult for users to work with it.
### Feature Description
1, Add the defaults filter method we have for cases to events:
 - time, disease, region, id, name, description, responsible SO, 
2, Add additional filter option at the "more tab": for case uuid, person uuid.  I do not think we need to include contact uuid because it is already covered by person Id.

bernardsilenou:
@markusmann-vg  Can we add this to the excel  table pls, I got many users requesting to have it.
@MateStrysewskeSym  do you see any other refinement to be done? To me its ready for implementation. The event entity if being used often now and there is a need to have such filters.

MateStrysewskeSym:
@bernardsilenou I don't think we need any further refinement, no.

oleuci:
Hello, do you know when it will be delivered in SORMAS version? BR

MateStrysewskeSym:
@oleuci I can't say that for sure right now. We've got a lot of extremely pressing issues that we need to take care of. It's basically a matter of priorisation from HZI.

Concerning your second request, you should've received a detailed answer by my colleague @MartinWahnschaffeSymeda.

Iheanacho2027:
@bernardsilenou pls this ticket should be reopened. Region and district filters are missing on the events directory

bernardsilenou:
@Iheanacho2027  It was added in 1,46. We will update the symeda server to 1,46 so that you can test.

PR #2314, opened 2020-06-12
#2223: Add filters to event directory

ChristopherRiedelSymeda:


leventegal-she:
#### Review: 
The code look good but the date filter doesn't seem to work.
- [ ] Missing "Apply" button on the date filter
- [ ] The more filters section is always expanded even if there is nothing changed in the date filter

Issue #2237, opened 2020-06-02
Make sure deletion of entities is bound to user rights, not user roles [0.5]


MartinWahnschaffeSymeda:
### Problem Description
In the deletion of the following entities the system does not check for the existing ..._DELETE user right, but for the user role of the ADMIN.

### Proposed Change
Update the logic to use the user right:

- [x] ClinicalVisitFacadeEjb.deleteClinicalVisit
- [x] Add EVENTPARTICIPANT_DELETE user right (give it to ADMIN only) and use it in EventParticipantEjb.deleteEventParticipant and EventParticpantsController.editEventParticipant
- [x] AdditionalTestFacadeEjb.deleteAdditionalTest
- [x] PrescriptionFacadeEjb.deletePrescription
- [x] TreatmentFacadeEjb.deleteTreatment

PR #2315, opened 2020-06-12
#2237 - update delete methods to check the user rights before deleting

BarnaBartha:


sormas-api/src/main/java/de/symeda/sormas/api/user/UserRight.java
```diff
@@ -628,6 +628,9 @@
 			SURVEILLANCE_OFFICER,
 			EVENT_OFFICER
 	),
+	EVENTPARTICIPANT_DELETE(
+			ADMIN
```
MartinWahnschaffeSymeda, line 632: Just saw that we also allow the NATIONAL_USER to delete events, so that user should probably also be allowed to delete participants.

PR #2321, opened 2020-06-15
Feature 2238 jurisdiction level to roles

BarnaBartha:


sormas-api/src/main/java/de/symeda/sormas/api/utils/jurisdiction/CaseJurisdictionHelper.java
```diff
@@ -17,39 +17,40 @@
...
 
 import de.symeda.sormas.api.caze.CaseJurisdictionDto;
+import de.symeda.sormas.api.user.JurisdictionLevel;
 import de.symeda.sormas.api.user.UserRole;
 import de.symeda.sormas.api.utils.DataHelper;
 
 public class CaseJurisdictionHelper {
 
-	public static Boolean isInJurisdiction(RoleCheck roleCheck, UserJurisdiction userJurisdiction, CaseJurisdictionDto caseJurisdictionDto) {
+	public static Boolean isInJurisdiction(Set<UserRole> userRoles, UserJurisdiction userJurisdiction, CaseJurisdictionDto caseJurisdictionDto) {
```
MartinWahnschaffeSymeda, line 29: Instead of passing the user roles directly pass the JurisdictionLevel and use a switch statement.

sormas-api/src/main/java/de/symeda/sormas/api/utils/jurisdiction/ContactJurisdictionHelper.java
```diff
@@ -17,33 +17,34 @@
...
 
 import de.symeda.sormas.api.contact.ContactJurisdictionDto;
+import de.symeda.sormas.api.user.JurisdictionLevel;
 import de.symeda.sormas.api.user.UserRole;
 import de.symeda.sormas.api.utils.DataHelper;
 
 public class ContactJurisdictionHelper {
 
-	public static boolean isInJurisdiction(RoleCheck roleCheck, UserJurisdiction userJurisdiction, ContactJurisdictionDto contactJurisdiction) {
+	public static boolean isInJurisdiction(Set<UserRole> userRoles, UserJurisdiction userJurisdiction, ContactJurisdictionDto contactJurisdiction) {
```
MartinWahnschaffeSymeda, line 29: Same here

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
```diff
@@ -774,46 +775,35 @@ public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<?,
...
 
 			// allow case access based on user role
-			if (currentUser.hasAnyUserRole(
-				UserRole.SURVEILLANCE_SUPERVISOR,
-				UserRole.CONTACT_SUPERVISOR,
-				UserRole.CASE_SUPERVISOR,
-				UserRole.POE_SUPERVISOR,
-				UserRole.EVENT_OFFICER,
-				UserRole.STATE_OBSERVER) && currentUser.getRegion() != null) {
+			if (jurisdictionLevel == JurisdictionLevel.REGION && currentUser.getRegion() != null) {
```
MartinWahnschaffeSymeda, line 786: I'd suggest to use a switch statement instead. Also in the other createUserFilter methods.

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactService.java
```diff
@@ -824,7 +825,8 @@ public Predicate createUserFilterWithoutCase(CriteriaBuilder cb, CriteriaQuery c
 
 		// National users can access all contacts in the system
 		User currentUser = getCurrentUser();
-		if (currentUser.hasAnyUserRole(UserRole.NATIONAL_USER, UserRole.NATIONAL_CLINICIAN, UserRole.NATIONAL_OBSERVER, UserRole.REST_USER)) {
+		final JurisdictionLevel jurisdictionLevel = UserRole.getJurisdictionLevel(currentUser.getUserRoles());
+		if (jurisdictionLevel == JurisdictionLevel.NATION || currentUser.hasAnyUserRole(UserRole.REST_USER)) {
```
MartinWahnschaffeSymeda, line 829: You have to exclude POE users here.

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java
```diff
@@ -83,8 +84,10 @@ public Predicate createCriteriaFilter(
 	public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<AggregateReport, AggregateReport> from) {
 
 		User currentUser = getCurrentUser();
+		final JurisdictionLevel jurisdictionLevel = UserRole.getJurisdictionLevel(currentUser.getUserRoles());
 		if (currentUser == null
-			|| currentUser.hasAnyUserRole(UserRole.NATIONAL_USER, UserRole.NATIONAL_CLINICIAN, UserRole.NATIONAL_OBSERVER, UserRole.REST_USER)) {
+			|| (jurisdictionLevel == JurisdictionLevel.NATION && !currentUser.hasAnyUserRole(UserRole.POE_NATIONAL_USER))
```
MartinWahnschaffeSymeda, line 89: For the POE check I'd use UserRole.isPortHealthUser

sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportService.java
```diff
@@ -209,8 +208,9 @@ public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<Wee
 			return usersStream;
 		}
 
+		final JurisdictionLevel jurisdictionLevel = UserRole.getJurisdictionLevel(user.getUserRoles());
```
MartinWahnschaffeSymeda, line 211: Maybe we should add a transient getJurisdictionLevel method to the user that wraps this and could also cache it? What do you think?

sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserEditForm.java
```diff
@@ -168,15 +169,15 @@ protected void addFields() {
...
-		boolean isLabUser = UserRole.isLabUser(userRoles);
-		boolean isPortHealthUser = UserRole.isPortHealthUser(userRoles);
-		boolean isStateObserver = userRoles.contains(UserRole.STATE_OBSERVER);
-		boolean isDistrictObserver = userRoles.contains(UserRole.DISTRICT_OBSERVER);
+		final OptionGroup userRolesField = (OptionGroup) getFieldGroup().getField(UserDto.USER_ROLES);
+		final Set<UserRole> userRoles = (Set<UserRole>) userRolesField.getValue();
+
+		final JurisdictionLevel jurisdictionLevel = UserRole.getJurisdictionLevel(userRoles);
+
+		final boolean isInformant = UserRole.hasAssociatedOfficer(userRoles);
```
MartinWahnschaffeSymeda, line 177: rename to hasAssociatedOfficer

sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserEditForm.java
```diff
@@ -220,15 +221,15 @@ private void updateFieldsByUserRole() {
 		}
 
 		ComboBox region = (ComboBox) getFieldGroup().getField(UserDto.REGION);
-		boolean useRegion = isSupervisor || isInformant || isOfficer || isStateObserver || isDistrictObserver;
+		boolean useRegion = jurisdictionLevel == JurisdictionLevel.REGION || isInformant || jurisdictionLevel == JurisdictionLevel.DISTRICT;
```
MartinWahnschaffeSymeda, line 224: Rule should be JurisdictionLevel.REGION || useDistrict (as calcualted below).

sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserEditForm.java
```diff
@@ -220,15 +221,15 @@ private void updateFieldsByUserRole() {
...
 		region.setVisible(useRegion);
 		setRequired(useRegion, UserDto.REGION);
 		if (!useRegion) {
 			region.clear();
 		}
 
 		ComboBox district = (ComboBox) getFieldGroup().getField(UserDto.DISTRICT);
-		boolean useDistrict = isInformant || isOfficer || isDistrictObserver;
+		boolean useDistrict = isInformant || jurisdictionLevel == JurisdictionLevel.DISTRICT;
```
MartinWahnschaffeSymeda, line 232: Additional options are: JursidictionLevel.COMMUNITY and health facility and point of entry.

sormas-api/src/main/java/de/symeda/sormas/api/utils/jurisdiction/CaseJurisdictionHelper.java
BarnaBartha: ok

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
BarnaBartha: but there is also the checks for not null if i add the switch statement we'd still remain also with those if statements

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactService.java
BarnaBartha: ok

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java
BarnaBartha: here it's only about ignoring the poe national user

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
MartinWahnschaffeSymeda: Yes, it's more about code logic and readability: The most important part is the jurisdiction and a switch statement would reflect that. The null check would be within the case block.

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java
MartinWahnschaffeSymeda: True, but from the logic side it's sufficient to check POE in general and the goal is to use as many explicit user roles as possible.

sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportService.java
BarnaBartha: ok, good idea

sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserEditForm.java
```diff
@@ -168,72 +169,70 @@ protected void addFields() {
...
+		final JurisdictionLevel jurisdictionLevel = UserRole.getJurisdictionLevel(userRoles);
+
+		final boolean hasAssociatedOfficer = UserRole.hasAssociatedOfficer(userRoles);
+		final boolean hasOptionalHealthFacility = UserRole.hasOptionalHealthFacility(userRoles);
+		final boolean isLabUser = UserRole.isLabUser(userRoles);
+		final boolean isPortHealthUser = UserRole.isPortHealthUser(userRoles);
+
+		final boolean usePointOfEntry = (isPortHealthUser && hasAssociatedOfficer) || jurisdictionLevel == JurisdictionLevel.POINT_OF_ENTRY;
+		final boolean useHealthFacility = jurisdictionLevel == JurisdictionLevel.HEALTH_FACILITY;
+		final boolean useCommunity = jurisdictionLevel == JurisdictionLevel.COMMUNITY || useHealthFacility || usePointOfEntry;
```
MartinWahnschaffeSymeda, line 184: health facility and point of entry don't rely on community.

sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportService.java
```diff
@@ -168,12 +169,11 @@ public WeeklyReport getByEpiWeekAndUser(EpiWeek epiWeek, User user) {
...
 		User currentUser = getCurrentUser();
-		if (currentUser == null) {
-			return null;
-		}
-
 		// National users can access all reports in the system
-		if (currentUser.hasAnyUserRole(UserRole.NATIONAL_USER, UserRole.NATIONAL_CLINICIAN, UserRole.NATIONAL_OBSERVER, UserRole.REST_USER)) {
+		final JurisdictionLevel jurisdictionLevel = currentUser.getJurisdictionLevel();
+		if (currentUser == null
+				|| (jurisdictionLevel == JurisdictionLevel.NATION && !currentUser.hasAnyUserRole(UserRole.POE_NATIONAL_USER))
```
MartinWahnschaffeSymeda, line 175: I'd still suggest to check for POE in general and not the specific POE_NATIONAL... user role.

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java
```diff
@@ -93,9 +96,7 @@ public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<Agg
 		Predicate filter = cb.equal(reportingUser, currentUser);
 
 		// Allow access based on user role
-		if (currentUser
-			.hasAnyUserRole(UserRole.SURVEILLANCE_SUPERVISOR, UserRole.CONTACT_SUPERVISOR, UserRole.CASE_SUPERVISOR, UserRole.STATE_OBSERVER)
-			&& currentUser.getRegion() != null) {
+		if (jurisdictionLevel == JurisdictionLevel.REGION && currentUser.getRegion() != null) {
```
MateStrysewskeSym, line 99: Probably need to check for isPortHealthUser here as well (to be consistent with the check above)

sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportService.java
MateStrysewskeSym: In which way was this resolved? I don't see any code changes here; Martin's suggestion makes sense to me because it shouldn't be possible to have two user roles from different jurisdictions in the first place

sormas-backend/src/main/java/de/symeda/sormas/backend/task/TaskService.java
```diff
@@ -114,7 +115,10 @@ public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<Tas
 
 		// National users can access all tasks in the system
 		User currentUser = getCurrentUser();
-		if (currentUser.hasAnyUserRole(UserRole.NATIONAL_USER, UserRole.NATIONAL_CLINICIAN, UserRole.NATIONAL_OBSERVER, UserRole.REST_USER)) {
+		final JurisdictionLevel jurisdictionLevel = currentUser.getJurisdictionLevel();
+		if (currentUser == null
+				|| (jurisdictionLevel == JurisdictionLevel.NATION && !currentUser.hasAnyUserRole(UserRole.POE_NATIONAL_USER))
```
MateStrysewskeSym, line 120: Same as in the conversation above, check for point of entry jurisdiction instead of specific user role

sormas-backend/src/main/java/de/symeda/sormas/backend/user/User.java
```diff
@@ -315,4 +316,9 @@ public void setLanguage(Language language) {
 	public boolean hasAnyUserRole(UserRole... userRoles) {
 		return Arrays.stream(userRoles).anyMatch(getUserRoles()::contains);
 	}
+
+	@Transient
+	public JurisdictionLevel getJurisdictionLevel(){
```
MateStrysewskeSym, line 321: Pretty sure you didn't use code formatting in this class

sormas-api/src/main/java/de/symeda/sormas/api/user/UserRole.java
```diff
@@ -238,105 +223,52 @@ public void addAssignableRoles(Collection<UserRole> collection) {
...
+	 * @param roles
+	 * @return
+	 */
+	public static JurisdictionLevel getJurisdictionLevel(Collection<UserRole> roles) {
 
 		for (UserRole role : roles) {
-			if (role.isSupervisor()) {
-				return true;
+			JurisdictionLevel jurisdictionLevel = role.getJurisdictionLevel();
+			if (jurisdictionLevel != JurisdictionLevel.NONE && jurisdictionLevel != JurisdictionLevel.LABORATORY) {
```
MateStrysewskeSym, line 240: I don't quite get why this method returns NONE if the jurisdiction level is LABORATORY. Can you explain that to me?

sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportService.java
BarnaBartha: it's a mistake this was also for aggregate report service and changed only there

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java
BarnaBartha: ok, but that means i'm changing the existing functionality

sormas-backend/src/main/java/de/symeda/sormas/backend/task/TaskService.java
BarnaBartha: i cannot check for that as only the POE_INFORMANT has pointOfEntry jurisdiction
what i can use instead is !UserRole.isPortHealthUser(currentUser.getUserRoles())

sormas-api/src/main/java/de/symeda/sormas/api/user/UserRole.java
BarnaBartha: it is a mistake or maybe a shortcoming of not having what we need, was added because laboratory is combinable with other roles
for LAB_USER -> should return LABORATORY jurisdiction - this is clear and will change it
for {LAB_USER, NATIONAL_USER} -> it should return NATION - is clear and works like this
for {LAB_USER,ADMIN} -> it should return NONE or LABORATORY jurisdiction?

sormas-backend/src/main/java/de/symeda/sormas/backend/user/User.java
BarnaBartha: indeed, formatted it, space is there now

BarnaBartha:
@MateStrysewskeSym 
To test every user role what should I use as guide what to expect is there a summary of what user roles can do same as for the user rights in SORMAS_User_rights?

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java
MateStrysewskeSym: Didn't you already change it a few lines above though?

sormas-backend/src/main/java/de/symeda/sormas/backend/task/TaskService.java
MateStrysewskeSym: Oh, you're right! Okay, maybe that's the way to go then. :)

sormas-api/src/main/java/de/symeda/sormas/api/user/UserRole.java
MateStrysewskeSym: I'm not exactly sure; since you've created the logic, maybe you can answer the following question:

* If LABORATORY is used, will the admin part of the user have any problems? 
* If NONE is used, will the lab user part of the user have any problems?

If the answers to both questions are "no", then I'd probably say that having a jurisdiction should have priority over having no jurisdiction. What do you say?

MateStrysewskeSym:
@BarnaBartha SORMAS_User_Rights could actually be a good reference. You could test whether all of this functionality still works for them.

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java
BarnaBartha: no, i changed the if's using the new enum, but the actual checking done by the if should remain the same, at least that is the intention

Issue #2285, opened 2020-06-09
Unique identifiers needed for UI automation testing


raghupola-vg:
Unique identifiers are required at below screens for UI test automation.

Contacts- Follow up visits--> **Previous, Next, From** and **To**  elements.
Contacts- Open a contact ---> **Case ID** field on contact information screen.

PR #2322, opened 2020-06-15
#2285 - add identifiers to vaadin component UI elements needed for UI…

BarnaBartha:
… testing


sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactsView.java
```diff
@@ -463,11 +463,14 @@ public HorizontalLayout createStatusFilterBar() {
 				scrollLayout.setMargin(false);
 
 				DateField toReferenceDate = new DateField(I18nProperties.getCaption(Captions.to), LocalDate.now());
+				toReferenceDate.setId("toReferenceDateField");
```
leventegal-she, line 466: Please use ButtonHelper for creating buttons, that will assign the id to the button

BarnaBartha: ok

PR #2349, opened 2020-06-18
Feature 2065 handle unappropiate parameters

StefanKock:
I half started new from the input @BarnaBartha delivered in his pull request.


sormas-ui/src/main/java/de/symeda/sormas/ui/utils/AbstractDetailView.java
```diff
@@ -0,0 +1,89 @@
...
+	 */
+	protected abstract R getReferenceByUuid(String uuid);
+
+	/**
+	 * @return The fallback View when no object is found to display.
+	 */
+	protected abstract String getRootViewName();
+
+	/**
+	 * We be called by {@link #enter(ViewChangeEvent)}, when an object is selected and the view shall show its specific content.
```
BarnaBartha, line 83: should be 'Will be called..'

PR #2351, opened 2020-06-19
2130 pseudonymize sensitive data

leventegal-she:


sormas-ui/src/test/java/de/symeda/sormas/ui/FacadeProviderMock.java
```diff
@@ -92,7 +92,7 @@
 		} else if (HospitalizationFacade.class == clazz) {
 			return (P) bm.getBean(HospitalizationFacadeEjbLocal.class);
 		} else if (EpiDataFacade.class == clazz) {
-			return (P) bm.getBean(EpiDataFacadeEjbLocal.class);
+			return (P) bm.getBean(EpiDataFacadeEjb.EpiDataFacadeEjbLocal.class);
```
MateStrysewskeSym, line 95: This change should be reverted

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseExportDto.java
```diff
@@ -1226,47 +1231,68 @@ public void setNumberOfClinicalVisits(int numberOfClinicalVisits) {
...
+		public CaseExportSampleDto() {
+		}
+
+		public CaseExportSampleDto(Date dateTime, String lab, PathogenTestResultType result) {
+			this.dateTime = dateTime;
+			this.lab = lab;
+			this.result = result;
+		}
+
+		public String stringFormat() {
```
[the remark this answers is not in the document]
leventegal-she: I wouldn't override toString method because because for example if there will be needed an extra function argument and also other generic code will rely on that toString method it will be hard to locate the places where to use the updated toString method. I faced the same issue during the implementation of locale based date formatting.

I can rename to `asString` if you think it's a better name.

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleExportDto.java
```diff
@@ -787,6 +715,48 @@ public ContactJurisdictionDto getAssociatedContactJurisdiction() {
...
+
+	public void setPathogenTest3(SamplePathogenTest pathogenTest3) {
+		this.pathogenTest3 = pathogenTest3;
+	}
+
+	public List<SamplePathogenTest> getOtherPathogenTests() {
+		return otherPathogenTests;
+	}
+
+	public static class Material {
```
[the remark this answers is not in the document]
leventegal-she: It Is necessary because the `sampleMaterialDetails` should be pseudonymized but it looks like I forgot to add the annotation :)

I will rename these classes

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/AbstractCaseView.java
```diff
@@ -164,12 +167,12 @@ public void refreshMenu(SubMenu menu, Label infoLabel, Label infoLabelSub, Strin
...
 			}
 			if (UserProvider.getCurrent().hasUserRight(UserRight.THERAPY_VIEW)
-				&& !caze.isUnreferredPortHealthCase()
+//				&& !caze.isUnreferredPortHealthCase()
 				&& !FacadeProvider.getFeatureConfigurationFacade().isFeatureDisabled(FeatureType.CLINICAL_MANAGEMENT)) {
 				menu.addView(TherapyView.VIEW_NAME, I18nProperties.getPrefixCaption(CaseDataDto.I18N_PREFIX, CaseDataDto.THERAPY), params);
 			}
 			if (UserProvider.getCurrent().hasUserRight(UserRight.CLINICAL_COURSE_VIEW)
-				&& !caze.isUnreferredPortHealthCase()
+//				&& !caze.isUnreferredPortHealthCase()
```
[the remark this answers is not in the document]
leventegal-she: I did just for testing purpose, I uncomment them

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/AbstractCaseView.java
```diff
@@ -251,9 +254,13 @@ public ViewMode getViewMode() {
 
 	public void setCaseEditPermission(Component component) {
 
-		Boolean isCaseEditAllowed = FacadeProvider.getCaseFacade().isCaseEditAllowed(getReference().getUuid());
+		Boolean isCaseEditAllowed = isCaseEditAllowed();
```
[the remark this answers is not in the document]
leventegal-she: It was the same before my changes, I think the reason for the if is that if `component.setEnabled` is called with true it will enable all the fields also the ones should always be disabled.

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleExportDto.java
```diff
@@ -119,7 +119,7 @@ public SampleExportDto(long id, String uuid, String labSampleId, String epidNumb
 		this.epidNumber = epidNumber;
 
 		if (caseUuid != null) {
-			this.associatedCase = new AssociatedCase(
+			this.samplAssociatedCase = new SampleExportAssociatedCase(
```
MateStrysewskeSym, line 122: Typo

sormas-api/src/main/java/de/symeda/sormas/api/utils/jurisdiction/SampleJurisdictionHelper.java
```diff
@@ -1,39 +1,45 @@
...
+		case HEALTH_FACILITY:
+		case POINT_OF_ENTRY:
+			return false;
+		case LABORATORY:
+		case EXTERNAL_LABORATORY:
+			return sampleJurisdiction.getLabUuid() != null
+				&& DataHelper.equal(sampleJurisdiction.getLabUuid(), userJurisdiction.getHealthFacilityUuid());
+		}
+
+		if (sampleJurisdiction.getCaseJurisdiction() != null) {
```
MateStrysewskeSym, line 35: If I'm not mistaken, this code is never reached because the switch above will always return, which would very likely lead to bugs (because users should be allowed to edit samples of cases that they can edit no matter their jurisdiction level as long as they have the SAMPLE_EDIT user right)

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/facility/FacilityDao.java
```diff
@@ -145,6 +149,18 @@ public Date getLatestChangeDateByRegion(Region region) {
 			where.and().eq(InfrastructureAdo.ARCHIVED, false);
 			where.and().eq(AbstractDomainObject.SNAPSHOT, false);
 			where.and().ne(Facility.UUID, FacilityDto.OTHER_LABORATORY_UUID).query();
+
+			User currentUser = ConfigProvider.getUser();
+			UserRole.getJurisdictionLevel(currentUser.getUserRoles());
+
+			if (currentUser.getCommunity() != null) {
```
MateStrysewskeSym, line 156: I'm pretty sure that this code is problematic. Laboratories usually don't have a district or community, and app users normally are on district, community or facility level, so this will almost always return an empty list and lead to null pointers or empty mandatory fields. Do we really need this logic? If the user is allowed to edit the laboratory field, they should also be allowed to see the names of all laboratories; if they don't, maybe we actually have to do the pseudonymization in the user interface. The alternative (which I would actually prefer right now) would be to get rid of the pseudonymization of laboratory names for now because I don't really think it's as important. So maybe just remove this code for now and we'll revisit it later.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/user/UserDao.java
```diff
@@ -73,6 +75,34 @@ public User getByUsername(String username) {
 		}
 	}
 
+	public List<User> getAllInJurisdiction() {
+		try {
+			QueryBuilder builder = queryBuilder();
```
MateStrysewskeSym, line 80: Again, I think we'll run into problems here. We can't restrict the list of users that are retrieved because this will lead to empty mandatory fields or empty values where a value was supposed to be. I think we need a solution that actually pseudonymises the data here instead of filtering the list of possible options.

sormas-backend/src/main/java/de/symeda/sormas/backend/facility/FacilityService.java
```diff
@@ -103,20 +104,33 @@ public FacilityService() {
...
 
 	public List<Facility> getAllActiveLaboratories(boolean includeOtherLaboratory) {
+		return getAllActiveLaboratories(includeOtherLaboratory, null);
+	}
+
+	public List<Facility> getAllActiveLaboratoriesInJurisdiction(boolean includeOtherLaboratory) {
+		return getAllActiveLaboratories(includeOtherLaboratory, this::createJurisdictionFilter);
+	}
+
+	private List<Facility> getAllActiveLaboratories(
```
MateStrysewskeSym, line 114: Same problem as in the app; I suggest that we also remove this for now.

sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserService.java
```diff
@@ -82,34 +85,50 @@ public User getByUserName(String userName) {
...
 
 	public List<User> getAllByRegionAndUserRoles(Region region, UserRole... userRoles) {
+		return getAllByRegionAndUserRoles(region, Arrays.asList(userRoles), null);
+	}
+
+	public List<User> getAllByRegionAndUserRolesInJurisdiction(Region region, UserRole... userRoles) {
+		return getAllByRegionAndUserRoles(region, Arrays.asList(userRoles), this::createJurisdictionFilter);
+	}
+
+	private List<User> getAllByRegionAndUserRoles(
```
MateStrysewskeSym, line 102: Again, same problem as in the app; this could potentially lead to unexpected and undesired behaviour. We probably need actual pseudonymisation instead.

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskEditForm.java
```diff
@@ -166,9 +166,6 @@ protected void addFields() {
 				users = FacadeProvider.getUserFacade().getUserRefsByDistrict(district, true);
 			} else if (region != null) {
 				users = FacadeProvider.getUserFacade().getUsersByRegionAndRoles(region);
-			} else {
```
MateStrysewskeSym, line 169: I think we still need this code because having an empty user list would be a problem. With pseudonymized user names it would be weird, but functional.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
```diff
@@ -887,7 +887,7 @@ private Predicate createNewCaseFilter(CriteriaBuilder cb, From<?, Case> caze, Da
 		return newCaseFilter;
 	}
 
-	public Predicate inInJurisdiction(CriteriaBuilder cb, CaseJoins<Case> joins) {
+	public Predicate isInInJurisdiction(CriteriaBuilder cb, CaseJoins<Case> joins) {
```
MateStrysewskeSym, line 890: Still a typo :) now you've got "In" twice there

PR #2353, opened 2020-06-19
2112 make events tasks and sample forms and fragments read only based on their jurisdiction

sergiupacurariu:


sergiupacurariu:
> Please make sure that all your code is formatted. I'd suggest to enable auto-format on save because it's easy to forget to do it manually.

I reformat the entire branch.

PR #2364, opened 2020-06-22
Feature 1838 add epidemiological data to contacts

ChristopherRiedelSymeda:


sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -428,6 +453,29 @@ public void deleteContact(String contactUuid) {
 						.setLastCooperativeVisitSymptomatic(lastCooperativeVisit.getSymptoms().getSymptomatic() ? YesNoUnknown.YES : YesNoUnknown.NO);
 				}
 
+				if (travels != null) {
```
BarnaBartha, line 456: it would be nice to have a test case that covers this extra part int he query and the logic to create the travelHistory

PR #2365, opened 2020-06-22
Feature 2312 large epid number validated

StefanKock:
I do now a more specific exception when a too large epid number is checked. 
@MateStrysewskeSym 
- please review and merge this.
- Consider whether, how and by whom the length validiation can be made in `sormas-ui` and `sormas-app`.


MateStrysewskeSym:
I just realised that the UI still throws an undefined error when the user is trying to enter an epid number that is too long because doesEpidNumberExist is called when the field loses focus, not only when the user saves the form. I think we should catch this exception in the frontend and set a meaningful error to the epid number field (e.g. "The entered epid number is invalid" - I don't think that more details are necessarily needed) instead of forcing an error popup to the user.

PR #2374, opened 2020-06-23
2112 make events tasks and sample forms and fragments read only based on their jurisdiction

sergiupacurariu:


sormas-api/src/main/java/de/symeda/sormas/api/utils/jurisdiction/SampleJurisdictionHelper.java
```diff
@@ -21,22 +21,11 @@ public static boolean isInJurisdiction(
...
 		case LABORATORY:
-			return sampleJurisdiction.getLabUuid() != null
-				&& DataHelper.equal(sampleJurisdiction.getLabUuid(), userJurisdiction.getHealthFacilityUuid());
+			return sampleJurisdiction.getLabUuid() != null && DataHelper.equal(sampleJurisdiction.getLabUuid(), userJurisdiction.getLabUuid());
 		case EXTERNAL_LABORATORY:
 			return sampleJurisdiction.getOtherLabUuid() != null
-				&& DataHelper.equal(sampleJurisdiction.getOtherLabUuid(), userJurisdiction.getHealthFacilityUuid());
-		case POINT_OF_ENTRY:
-			return false;
+				&& DataHelper.equal(sampleJurisdiction.getOtherLabUuid(), userJurisdiction.getLabUuid());
```
MateStrysewskeSym, line 28: Should this really be "otherLabUuid"? The "otherLab" that is still in the database is not actually used in the app, so I think that this field should be removed from the sample jurisdiction as well.

sergiupacurariu: removed otherLabUuid from SampleJurisdictionDto and EXTERNAL_LAB check from SampleJurisdictionHelper

Issue #2135, opened 2020-05-18
Performance impact of RequestResponseLogger


HolgerReiseVSys:
### Bug Description
`RequestResponseLogger` helps with debugging REST calls.
This comes with considerable overhead as the complete content of the response is duplicated in memory before being logged.
At the moment this is the case even when the data is ot logged at all.

Also, as a side effect, the filter sets the default response encoding to UTF-8.

### Steps to Reproduce
1. make sure logging is not set to TRACE
2. debug SORMAS with a breakpoint e.g. in the `HttpServletResponseCopier` constructor 
3. call a REST endpoint

### Expected Behavior
The breakpoint should not be triggered

### Additional Information
The following patch factors in that other code may rely on the default UTF-8 encoding:
```diff
diff --git "a/sormas-rest/src/main/java/de/symeda/sormas/rest/logging/RequestResponseLogger.java" "b/sormas-rest/src/main/java/de/symeda/sormas/rest/logging/RequestResponseLogger.java"
index 986425f40..b5ca7cd61 100644
--- "a/sormas-rest/src/main/java/de/symeda/sormas/rest/logging/RequestResponseLogger.java"
+++ "b/sormas-rest/src/main/java/de/symeda/sormas/rest/logging/RequestResponseLogger.java"
@@ -19,6 +19,7 @@ package de.symeda.sormas.rest.logging;
 
 import java.io.IOException;
 import java.util.Map;
+import java.util.Map.Entry;
 
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
@@ -60,26 +61,24 @@ public class RequestResponseLogger implements Filter {
 	 */
 	public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
 			throws IOException, ServletException {
-// 		There is an issue, when both criteria are enabled
-//		if (logger.isDebugEnabled()) {
+		if (logger.isDebugEnabled()) {
 
 			// request logging
 			HttpServletRequest httpRequest = (HttpServletRequest) request;
 			logger.debug("requestUri = {}", httpRequest.getRequestURI());
 
 			Map<String, String[]> params = httpRequest.getParameterMap();
-			for (String s : params.keySet()) {
-				// logger.debug(" " + s + " = " + req.getParameter(s));
-				logger.debug("  {} = {}", s, httpRequest.getParameter(s));
+			for (Entry<String, String[]> e : params.entrySet()) {
[...]

PR #2399, opened 2020-06-26
#2135 do not copy response body and log it only if the log level is trace

leventegal-she:


StefanKock:
Looks better, but:
1. Typo in `LoggingServletOutpuStream` -> ...`Output`...
2. I do not get why you made the other classes to inner classes.

StefanKock:
There some differences:
Before:
```
chain.doFilter(request, responseCopier);
responseCopier.flushBuffer();
```

After (only on Trace level):
```
chain.doFilter(request, wrapper);
httpServletResponse.getOutputStream().write(wrapper.getContentAsBytes());
```

1. I wonder if `chain.doFilter(...)` is nessassary
2. I do not understand the difference in the second line

PR #2433, opened 2020-07-03
Feature 2240 unify grid itemclicklistener for navigate and avoid NPE

StefanKock:
I was not able to reproduce the NPE, but refactored all `ItemClickListener`s to the new `ShowDetailsListener` which does not execute the `itemHandler` if `event.getItem() == null`.

### DIFF
--- a/sormas-ui/src/test/java/de/symeda/sormas/ui/utils/ShowDetailsListenerTest.java
+++ b/sormas-ui/src/test/java/de/symeda/sormas/ui/utils/ShowDetailsListenerTest.java
@@ -1,0 +1,63 @@
+package de.symeda.sormas.ui.utils;
+
+import static org.mockito.Mockito.mock;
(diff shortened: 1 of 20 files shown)


sormas-ui/src/test/java/de/symeda/sormas/ui/utils/ShowDetailsListenerTest.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @see ShowDetailsListener
+ */
+public class ShowDetailsListenerTest {
+
+	@Test
+	@SuppressWarnings({
+		"rawtypes",
+		"unchecked" })
+	public void testItemClickDoesNotCallHandlerWhenItemIsNull() {
```
BarnaBartha, line 28: really nice unit test

PR #2435, opened 2020-07-06
Improve field « EVENT DESCRIPTION»

### LINKED ISSUE #2415: Ergonomics improvement of the field « EVENT DESCRIPTION»
### Feature Description

Within the EVENT module, the current SORMAS version includes a text field “Event description” to formalize useful data concerning event description, comprehension and investigation.

### Problem Description

Users are informed of this size limitation only once entry is validated. Number of characters entered can’t be controlled throughout data entry.
Furthermore, as text is displayed on two lines only, users have to play with the lift to complete and read this zone.

### Proposed Change

It would thus be necessary to:
1)	Implement dynamic display of this zone to make it correspond to the text’s size.
2)	Indicate size limitation  by :

- Blocking data entry when the user reaches the maximum capacity of the field. This modification should be first implemented for the event formulary. It could further on be generalized to all fields of this type within SORMAS application ;
- Displaying the number of characters already entered or the number of characters left.

### Additional Information
screenshot 

image

xavier-calland:
Closes #2415

### DIFF
--- a/sormas-api/src/main/resources/captions.properties
+++ b/sormas-api/src/main/resources/captions.properties
@@ -41,6 +41,7 @@ total=Total
 unknown=Unknown
 creationDate=Creation date
 notAvailableShort=NA
+numberOfCharacters=Number of characters: %d / %d
 
 # About
 aboutSormasVersion=SORMAS Version
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/TextFieldWithMaxLengthWrapper.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/TextFieldWithMaxLengthWrapper.java
@@ -1,0 +1,92 @@
+/*******************************************************************************
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2018 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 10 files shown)


sormas-api/src/main/resources/captions.properties
```diff
@@ -41,6 +41,7 @@ total=Total
 unknown=Unknown
 creationDate=Creation date
 notAvailableShort=NA
+numberOfCharacters=Number of characters %d / %d
```
MateStrysewskeSym, line 44: I'd add a colon after "Number of characters"

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/TextFieldWithMaxLengthWrapper.java
```diff
@@ -0,0 +1,92 @@
...
+
+	private String buildLabelMessage(String text, T textField) {
+		return String.format(I18nProperties.getCaption(Captions.numberOfCharacters), Strings.nullToEmpty(text).length(), textField.getMaxLength());
+	}
+
+	/**
+	 * Set number of rows in textareas to the number of lines of text + 1.
+	 * Min: {@link #MIN_ROWS}, Max: {@link #MAX_ROWS}
+	 */
+	private void ajustRows(T textField, String text) {
```
MateStrysewskeSym, line 87: Typo: adjustRows

sormas-api/src/main/resources/captions.properties
xavier-calland: Done

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/TextFieldWithMaxLengthWrapper.java
xavier-calland: Done

MateStrysewskeSym:
@xavier-calland We actually use TravisCI for this repository, but for some reason it didn't run for your initial pull request. It apparently does now after your changes though.

Issue #2511, opened 2020-07-22
Redeployment of sormas-ear does not work with Payara 5.2020.2 


StefanKock:
### Bug Description
In the running Payara domain the redeployment or un- and deployment of sormas-ear does not work since #2453 was done.

### Steps to Reproduce
1. Start Payara Domain
2. Deploy `sormas-ear` (by IDE or als ear file into `autodeploy/`) is successful.
3. Undeploy `sormas-ear`.
4. Deploy `sormas-ear` -> fails.

### Expected Behavior
Redeployment with running domain worked with Payara 5.192.

### System Details
* SORMAS version: 1.45, ..., 1.67
* Payara version: 5.2020.2, 5.2021.6, 5.2021.10

### Additional Information
Workaround: Undeploy sormas artifacts, stop Payara domain and maybe clear not deleted `sormas-ear` in directories `generated`, `eclipseApps` or `applications`.

Current status with payara-5.2021.10: The redeployment of `sormas-ear` causes the access from outside to fail because of mismatching classes from different `EarClassLoader` instances (debug break point on `UnsafeFieldAccessorImpl.java:58` to see this). I did not find any issues concerning this problem on the payara repo.
<details>
<summary>StackTrace when (re-)deploying sormas-ui after 2nd deployment of sormas-ear</summary>

```
2021-09-09T16:31:44.842+0200|SCHWERWIEGEND: Exception while loading the app : javax.ejb.CreateException: Initialization failed for Singleton StartupShutdownService
javax.ejb.CreateException: Initialization failed for Singleton StartupShutdownService
	at com.sun.ejb.containers.AbstractSingletonContainer.createSingletonEJB(AbstractSingletonContainer.java:542)
	at com.sun.ejb.containers.AbstractSingletonContainer.access$000(AbstractSingletonContainer.java:81)
	at com.sun.ejb.containers.AbstractSingletonContainer$SingletonContextFactory.create(AbstractSingletonContainer.java:687)
	at com.sun.ejb.containers.AbstractSingletonContainer.instantiateSingletonInstance(AbstractSingletonContainer.java:421)
	at org.glassfish.ejb.startup.SingletonLifeCycleManager.initializeSingleton(SingletonLifeCycleManager.java:219)
	at org.glassfish.ejb.startup.SingletonLifeCycleManager.initializeSingleton(SingletonLifeCycleManager.java:180)
[...]

PR #2529, opened 2020-07-23
#2511 Refrain from using FacadeProvider in backend

HolgerReiseVSys:


HolgerReiseVSys:
Unfortunately this only works until it is accessed remotely.
WHich is the case with sormas-ui and sormas-rest...

StefanKock:
I reviewed the 2nd commit and merged it manually to `development`.

Issue #2544, opened 2020-07-24
Symptom onset date is not copied to symptom tap when converting event participant or contact to a case.


bernardsilenou:
### Steps to Reproduce
1. 
2.

### Expected Behavior
- Chen symptom onset date is provided when converting event participant to a case, this value should be assigned to symptoms.onsetdate also.
### System Details
* Device:web
* SORMAS version: 1,44,3
* Android version/Browser:

BarnaBartha:
@bernardsilenou Can you please confirm that these are the fields from these forms? - as i don't see any problem here

![image](https://user-images.githubusercontent.com/8166805/88522220-bbf10880-cffe-11ea-9ab7-398dbaa1f500.png)

![image](https://user-images.githubusercontent.com/8166805/88522316-db883100-cffe-11ea-84da-3c666322cf93.png)

bernardsilenou:
@BarnaBartha your screen shot has no issue and it is ok. The issue exist only when converting an event participant or a contact to a case.

BarnaBartha:
Hi @bernardsilenou 
the screenshot is part of the process when creating (converting) a case from event participant
I just rechecked this on https://test.sormas.netzlink.com/sormas-ui/ and the 'Date of symptom on set' i specify on the case form is shown in the created case symptoms tab.

But indeed converting a contact to case in the 'Date of symptom on set' date field not being saved. I will address this.

bernardsilenou:
@BarnaBartha I also tested and it worked for me, so only contact is left.
 Thanks

PR #2560, opened 2020-07-28
#2544 - fix filling of symptom on set date for case converted from co…

BarnaBartha:
…ntact


BarnaBartha:
- [ ] refactor duplicated code for setting symptom on set date for case conversions

PR #2568, opened 2020-07-29
Feature 1637 import and use new facility types

ChristopherRiedelSymeda:
The feature is not yet finished. But the web part can be reviewed.


sormas-ui/src/main/java/de/symeda/sormas/ui/contact/importer/ContactImporter.java
```diff
@@ -356,7 +356,12 @@ private void insertColumnEntryIntoData(ContactDto contact, PersonDto person, Str
 						Pair<DistrictReferenceDto, CommunityReferenceDto> infrastructureData =
 							ImporterPersonHelper.getPersonDistrictAndCommunity(pd.getName(), person);
 						List<FacilityReferenceDto> facility = FacadeProvider.getFacilityFacade()
-							.getByName(entry, infrastructureData.getElement0(), infrastructureData.getElement1(), false);
+							.getByNameAndType(
```
MartinWahnschaffeSymeda, line 359: From my understanding this would be the facility of the person (occupation), so it does not make sense to use the facility type of the case. Also the case is most likely not available.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/importer/CaseImporter.java
```diff
@@ -445,7 +447,12 @@ private void insertColumnEntryIntoData(CaseDataDto caze, PersonDto person, Strin
 						Pair<DistrictReferenceDto, CommunityReferenceDto> infrastructureData =
 							ImporterPersonHelper.getDistrictAndCommunityBasedOnFacility(pd.getName(), caze, person, currentElement);
 						List<FacilityReferenceDto> facility = FacadeProvider.getFacilityFacade()
-							.getByName(entry, infrastructureData.getElement0(), infrastructureData.getElement1(), false);
```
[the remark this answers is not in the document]
MartinWahnschaffeSymeda: I just notice that there is a second problem here. The imported facility could also be the occupation facility of the person. This has to be handled separately.

sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/map/DashboardMapComponent.java
```diff
@@ -684,7 +684,7 @@ private VerticalLayout createLegend() {
 		if (showCases) {
 			if (mapCaseDisplayMode == MapCaseDisplayMode.HEALTH_FACILITY
```
MartinWahnschaffeSymeda, line 685: HEALTH_FACILITY -> FACILITY

sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
```diff
@@ -100,55 +105,58 @@
 	private TextField occupationFacilityDetails;
 	private final ViewMode viewMode;
 	private ComboBox birthDateDay;
+	private ComboBox occupationFacilityType;
```
MartinWahnschaffeSymeda, line 108: Needs to be persisted for OTHER_FACILITY

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
```diff
@@ -197,6 +199,7 @@
 	@Outbreaks
 	@PersonalData
 	private CommunityReferenceDto community;
+	private FacilityType facilityType;
```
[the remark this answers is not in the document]
ChristopherRiedelSymeda: I wasn't sure about that because we won't rename Case.healthFacility.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/importer/CaseImporter.java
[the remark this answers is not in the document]
ChristopherRiedelSymeda: The facilityType is already before the facility and of course the column is also in the template, I checked that. Using Hospital as fallback is not a good idea, first of all it would be quite magical behaviour and secondly it will lead to problems in the UI.

sormas-ui/src/main/java/de/symeda/sormas/ui/contact/importer/ContactImporter.java
ChristopherRiedelSymeda: You're right, this isn't correct. But I don't really the what would be correct here.

ChristopherRiedelSymeda:
@MartinWahnschaffeSymeda Adjust import so that only reasonable facilities are allowed (Case and Person)

Issue #2525, opened 2020-07-23
Campaign diagram definition [3]


MartinWahnschaffeSymeda:
### Situation Description

For campaigns data is collected in forms with custom fields. The data collected needs to be visualized in a diagram on the dashboard.

Mockups and information on example indicators that need to be visualized can be found in #2004.

This ticket is created based on the concept in #2005 (I'll repeat the relevant aspects of it below).

### Feature Description

- [x] Create a CampaignDiagramDefinition entity that will hold:
  * unique id of the diagram (e.g. missedChildrenChart)
  * a type of the diagram. For now only 'column' - in the long run we should support more and more of the types supported by highcharts https://api.highcharts.com/highcharts/series
  * a list of CampaignDiagramSeries as defined below Should be stored as a json field matching the examples provided below.
- [x] Create a CampaignDiagramSeries class
  * fieldId: id of the field to be used as data source
  * formId: id of the form that contains it
  * fieldValue: optional, a value that is used to count the number of entries that match this value (e.g. "YES" for a yes/no field)
  * stack: optional, a string that is used to stack similar elements. For now this is not translated
- [x] Create Services, Facades, DTOs that allow to access this definition.
- [x] Create a unit test that takes the json in the example and correctly loads it into dtos.

Example for a series

```
[{
	"fieldId": "recordedAbsentDuring",
	"formId": "tallySheetTotalsForm",
	"stack": "days 1-3"
}, {
	"fieldId": "recordedAbsentAfter",
	"formId": "tallySheetTotalsForm",
	"stack": "days 1-3"
}, {
	"fieldId": "recordedNewbornSleepSick",
	"formId": "tallySheetTotalsForm",
	"stack": "days 1-3"
}, {
	"fieldId": "recordedRefusal",
	"formId": "tallySheetTotalsForm",
	"stack": "days 1-3"
}, {
	"fieldId": "missedStillAbsent",
	"formId": "monitoringRevisitForm",
	"stack": "revisit"
}, {
	"fieldId": "missedRefusals",
	"formId": "monitoringRevisitForm",
	"stack": "revisit"
}, {
	"fieldId": "missedTeamNeglience",
	"formId": "monitoringRevisitForm",
	"stack": "revisit"
}]
```

PR #2574, opened 2020-07-30
#2525 - campaign diagram definition - entity, json mapping, DTOs, fac…

BarnaBartha:
…ade, service + test


sormas-api/src/main/java/de/symeda/sormas/api/user/UserRight.java
```diff
@@ -1047,6 +1047,32 @@
 	),
 	CAMPAIGN_FORM_DATA_DELETE(
 			ADMIN
+	),
+	CAMPAIGN_DIAGRAM_VIEW(
```
MartinWahnschaffeSymeda, line 1051: Let's remove those user rights until we have a user interface that allows to view, edit and delete the diagrams. For now all of this will be done directly in the database.

Issue #2104, opened 2020-05-14
Add responsible community to contact [3]


MartinWahnschaffeSymeda:
As discussed here https://github.com/hzi-braunschweig/SORMAS-Project/issues/2021#issuecomment-627890951_

- [x] add "responsible community" drop down to contact (not mandatory)
- [x] rename caption of region, district and community of case to "responsible ..."
- [x] same for the filters used in case directory of app and UI
- [x] add a filter to the contact directory

max-hzi:
This issue does not cover the need for the community filter. Content-wise, I ask for it in the initial issue #2021. So we're covered, right? Just to be sure. Thanks!

MartinWahnschaffeSymeda:
Ah. Good point. I have added the filter to the list above.

Issue #1637, opened 2020-03-23
Extend the facility types [5+3]


MateStrysewskeSym:
### Feature Description
We need to allow users to import lists of:

* General practitioners
* Treatment centers
* Elderly care facilities

The best approach is probably to add these as facility types and then decide for individual facility fields which types of facilities should be added to the dropdown. Also, we probably have to think about changing the captions of these fields.

In addition, general practitioners probably need information like address and phone number, so we might need to add these as fields to facilities.

Before upload a selection of type is mandatory
- [x] Update facility types for old facilities in the db
- [x] In places where a health facility is selected, the system first asks for the type of facility and then displays a list of facilities of that type. This list shall also contain the option "Other facility". If this is selected, the "Additional Details" will also be displayed.
- [x] Consider the facility type during import
- [x] hospitalization should only be changed if transfer is related to hospital

### Review 
@MateStrysewskeSym

max-hzi:
I received a couple of requests for a directory for GPs/ doctors offices. By having a directory, are no longer required to type in address and other info every time. One request was to have these doctors offices separate from bigger health facilities like hospitals.

markusmann-vg:
Facility types:
Arztpraxen
Labore
Krankenhäuser
Betreuungseinrichtungen: Altenheim, Obdachlosenunterkunft, 
Einrichtungen zur gemeinschaftlichen Unterbringung von Asylbewerbern, vollziehbar Ausreisepflichtigen, Flüchtlingen und Spätaussiedlern; sonstige Massenunterkünfte, Justizvollzugsanstalten, ambulante Pflegedienste) 
Gemeinschaftseinrichtungen: Kindertagesstätte, -Horte, Kindertagespflege, Schulen und sonstige Ausbildungseinrichtungen, Heime, Ferienlager)

bernardsilenou:
@markusmann-vg  The list you provided has mixture of health facilities, lab and other facilities is in the current  system. 
In issues #2340 #2211, # 2093 we have the variable typeofplace which will contain a similar list including facilities that are not health facilities. We need to maybe consider implementing one methods to handle both issues. Don't know how it sounds to you

ChristopherRiedelSymeda:
Facility types:
GENERAL_PRACTITIONERS_OFFICE
LABORATORY (was ther before)
HOSPITAL
ELDERLY_CARE_FACILITY
HOMELESS_SHELTER
REFUGEE_HOSTEL
[...]

PR #2615, opened 2020-08-10
#2104 add responsible community to contact

FredrikSchaeferVitagroup:


MartinWahnschaffeSymeda:
See #2104

sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactIndexDto.java
```diff
@@ -40,6 +40,7 @@
 	public static final String DISEASE = "disease";
 	public static final String REGION_UUID = "regionUuid";
 	public static final String DISTRICT_UUID = "districtUuid";
+	public static final String COMMUNITY_UUID = "communityUuid";
```
MartinWahnschaffeSymeda, line 43: Please move these constants to the ContactJurisdictionDto. It's also used in getPrefixCaption in some places. There I'd suggest to do e.g. `I18nProperties.getPrefixCaption(ContactJurisdictionDto.I18N_PREFIX, ContactJurisdictionDto.REGION_UUID)`. ContactJurisdictionDto can use the same I18n_PREFIX as ContactIndexDto.

sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactDataForm.java
```diff
@@ -101,7 +101,8 @@
                     fluidRowLocs(ContactDto.UUID, ContactDto.EXTERNAL_ID) +
                     fluidRowLocs(ContactDto.REPORTING_USER, ContactDto.REPORT_DATE_TIME) +
                     fluidRowLocs(ContactDto.REGION, ContactDto.DISTRICT) +
-                    fluidRowLocs(ContactDto.CASE_ID_EXTERNAL_SYSTEM, "") +
```
MartinWahnschaffeSymeda, line 104: Let's put the contact in one line with region and district instead. We are doing the same in the case in #1637

FredrikSchaeferVitagroup:
Merged after discussion with @MartinWahnschaffeSymeda.

PR #2649, opened 2020-08-14
 Solved Issue2332 Document read-only query hint

john1496:


HolgerReiseVSys (changes requested):
I am a bit confused by the first commit. The commit message does not relate to the ticket. Many of the changes also seem to be unrelated.
Usuall the commt messages start with referencing the ticked number.

Also, there seems to be an edit war concerning the order of imports. Please have a look at the [IDE Setup Guide](/hzi-braunschweig/SORMAS-Project/blob/development/DEVELOPMENT_ENVIRONMENT.md) concerning Import order.

Please use html markup to help with the readability of the javadoc for ModelConstants.HINT_READ_ONLY.

As far as I can see, the query hint is still used with queries not returning entities.

I guess the easiest way forward would be to start anew from the current development branch.

john1496:
Thanks for the review, so i have taken @HolgerReiseVSys advice and started anew with a checkout from the development branch to make the changes better.

Issue #2621, opened 2020-08-11
[SORMAS2SORMAS] Creation of certificates to send and receive data [1]


markusmann-vg:
### Feature Description
Guide + Script should be delivered/implemented

possible approach:
use openssl 
reuse functionality of the Sormas to SurvNet implementation 
automatic generation of a certificate should be considered

PR #2659, opened 2020-08-17
#2621 Add certificate generation guide and script

vlad-ciucescu:


sormas-base/setup/server-setup.sh
```diff
@@ -331,6 +335,12 @@ ${ASADMIN} create-javamail-resource --mailhost localhost --mailuser user --froma
 ${ASADMIN} create-custom-resource --restype java.util.Properties --factoryclass org.glassfish.resources.custom.factory.PropertiesFactory --property "org.glassfish.resources.custom.factory.PropertiesFactory.fileName=\${com.sun.aas.instanceRoot}/sormas.properties" sormas/Properties
 
 cp sormas.properties "${DOMAIN_DIR}"
+SORMAS_PROPERTIES=${DOMAIN_DIR}/sormas.properties
+if [[ ${LINUX} = true ]]; then
+  export SORMAS_PROPERTIES
+else
+  setx SORMAS_PROPERTIES ${SORMAS_PROPERTIES}
```
HolgerReiseVSys, line 342: This seems to persistently add the environment variable to the user environment, which is probably not intended

sormas-base/setup/generate-cert.sh
```diff
@@ -0,0 +1,91 @@
...
+	LINUX=false
+fi
+
+# DIRECTORIES
+if [[ ${LINUX} = true ]]; then
+	ROOT_PREFIX=
+else
+	ROOT_PREFIX=/c
+fi
+SORMAS2SORMAS_DIR=${ROOT_PREFIX}/opt/sormas2sormas
```
HolgerReiseVSys, line 38: I would prefer to only set the variable when it has not been set before.
We could also ask for it if the directory doesn't exist.
Also shouldn't it be `/opt/sormas/sormas2sormas` ?

sormas-base/setup/generate-cert.sh
```diff
@@ -0,0 +1,91 @@
...
+
+# generate Private & Public Key & CSR
+openssl req -passout pass:"${SORMAS_S2S_CERT_PASS}" -sha256 -newkey rsa:4096 -keyform PEM -new -keyout "${PEM_NAME}" -out "${P10_NAME}" -subj "${CERT_SUBJ}"
+
+# generate self signed certifcate
+openssl x509 -passin pass:"${SORMAS_S2S_CERT_PASS}" -req -days 1095 -sha256 -in "${P10_NAME}" -keyform PEM -signkey "${PEM_NAME}" -out "${CRT_NAME}"
+
+# encrypt
+openssl pkcs12 -export -inkey "${PEM_NAME}" -out "${P12_NAME}" -passin pass:"${SORMAS_S2S_CERT_PASS}" -password pass:"${SORMAS_S2S_CERT_PASS}" -in "${CRT_NAME}"
+
```
HolgerReiseVSys, line 75: Shouldn't the .pem and .p10 be deleted?

sormas-base/setup/server-setup.sh
```diff
@@ -403,3 +413,4 @@ if [[ ${DEV_SYSTEM} != true ]]; then
 	echo "  - Configure the apache web server according to the server setup guide"
 fi
 	echo "  - Execute the r-setup.sh file to enable disease network diagrams"
+  echo "  - To enable SORMAS to SORMAS communication, see the related guide and execute the generate-cert.sh and import-to-truststore.sh scripts"
```
HolgerReiseVSys, line 416: I think I wouldn't refer to the scripts here

sormas-base/setup/server-setup.sh
```diff
@@ -80,6 +80,7 @@ GENERATED_DIR=${ROOT_PREFIX}/opt/sormas/generated
 CUSTOM_DIR=${ROOT_PREFIX}/opt/sormas/custom
 PAYARA_HOME=${ROOT_PREFIX}/opt/payara5
 DOMAINS_HOME=${ROOT_PREFIX}/opt/domains
+SORMAS2SORMAS_DIR=${ROOT_PREFIX}/opt/sormas2sormas
```
HolgerReiseVSys, line 83: I think it shold be `/opt/sormas/sormas2sormas`

GUIDE_CREATE_CERTIFICATE.md
```diff
@@ -0,0 +1,48 @@
+# How to create a new certificate?
+
+This guide explains how to create a new self signed certifcate, used for SORMAS to SORMAS communication or 
+for accessing the SurvNet interface.
```
HolgerReiseVSys, line 4: This cert won't be used for SurvNet

GUIDE_CREATE_CERTIFICATE.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Important:** This is the first version of this guide. Please get in contact if anything is not clear or 
+you have suggestions on how to improve the guide, the source code or the underlying architecture.
```
HolgerReiseVSys, line 7: I think we don't need this disclaimer. It would probably never be removed.

GUIDE_CREATE_CERTIFICATE.md
```diff
@@ -0,0 +1,48 @@
...
+   
+### Using the certificate generation script
+
+1. **Important**: If you customized the install paths during server installation, please open ``generate-cert.sh`` in a text editor and 
+   set the ``ROOT_PREFIX`` variable to the same value which was used during the installation. <br/>
```
HolgerReiseVSys, line 12: Probably not needed when SORMAS2SORMAS_DIR can be specified

GUIDE_CREATE_CERTIFICATE.md
```diff
@@ -0,0 +1,48 @@
...
+   The generated certificate has a validity of 3 years. 
+   The certificate files will be available in the root SORMAS directory, in the folder ``/sormas2sormas``.
+4. The generated ``.p12`` file should not be shared with third parties. <br/>
+   The generated ``.crt`` file will be verified and shared with other SORMAS instances, from which this instance
+   will be able to request data. Conversely, in order to enable other SORMAS instances to request data from this 
+   instance, their certificate files should be obtained and added to the local truststore. More details can be found
+   in the next section.
+5. If the ``SORMAS_PROPERTIES`` environment variable is available, the relevant properties will be 
+    automatically set by the script.
+    * If the ``SORMAS_PROPERTIES`` environment variable is not available, the properties which need to be added will be
```
HolgerReiseVSys, line 35: `Else` instead of `If the ``SORMAS_PROPERTIES`` environment variable is not available,`?

HolgerReiseVSys, line 35: Else

HolgerReiseVSys (changes requested):
The main use case of creating a certificate will be when or before setting up a docker container.
So the setup should not depend on sormas being installed.

GUIDE_CREATE_CERTIFICATE.md
vlad-ciucescu: duplicate

sormas-base/setup/server-setup.sh
vlad-ciucescu: This was actually my intention. I probably misunderstood this part in our discussion. I thought I should set an environment variable with the sormas.properties path so that it is easily accessible in the other scripts.

sormas-base/setup/generate-cert.sh
vlad-ciucescu: - How can this variable be set before? Should it be an environment variable?
- Do you mean if the root directory doesn't exist, ask for a path to the sorms2sormas directory?

HolgerReiseVSys: Maybe something like that?
```
if [[ -z "${SORMAS2SORMAS_DIR}" ]]; then
  DEFAULT_SORMAS2SORMAS_DIR="${ROOT_PREFIX}/opt/sormas2sormas"
  if [[ -d "${DEFAULT_SORMAS2SORMAS_DIR}" ]]; then
    SORMAS2SORMAS_DIR="${DEFAULT_SORMAS2SORMAS_DIR}"
  else
    	while [[ ! -d "${SORMAS2SORMAS_DIR}" ]]; do
		read -r -p "Please specify a valid sormas2sormas directory: " SORMAS2SORMAS_DIR
	done
  fi
fi
```

sormas-base/setup/server-setup.sh
HolgerReiseVSys: Under windows this will probably only affect developers.
Who usually are not really happy when their environment is meddled with.
And `export` only sets the variable for SUB-processes, so with Linux this will not be available after the script has returned.
For this to work it has to be called via [source](https://ss64.com/bash/source.html).

vlad-ciucescu: I considered the option of preserving the variables in the shell, but I thought that the scripts can be run at different times in different shells. However, if this is sufficient I will do it like this.

HolgerReiseVSys (approved):

Issue #2669, opened 2020-08-19
Adjust Covid-19 Symptoms for Switzerland 


markusmann-vg:
### Situation Description

The exisitng symptoms for COVID-19 need to be adjusted for CH systems.

### Feature Description

Adjust the symptoms for SORMAS CH to match the following definitions:

SORMAS CH German | SORMAS CH English | SORMAS symptoms – Data dictionary
-- | -- | --
Fieber | Fever | fever
Allgemeines   Krankheitsgefühl | General signs   of disease | generalSignsOfDisease
Fiebergefühl | Feverish   feeling | Missing
Schüttelfrost   oder Schwitzen | Chills | chillsSweats
Gesteigerte   Müdigkeit | Increased   fatigue | Missing
Allgemeine   Schwäche | General weakness | ~~Fatigue/general weakness~~ **Missing**
Kopfschmerzen | Headache | Headache
Muskel- oder   Gliederschmerzen | Muscle pain | Muscle pain
Verwirrtheit | Confused/disoriented | confusedDisoriented
Schnupfen (laufende oder verstopfte Nase) | Runny nose | runnyNose
Halsschmerzen | Sore throat | soreThroat
Trockener Husten ohne Auswurf | Dry cough without sputum | Missing
Husten mit Auswurf | Cough with sputum | coughWithSputum
Husten mit blutigem Auswurf | Coughing with heamoptysis| **coughWithHeamoptysis**
Rasche Atmung | Rapid breathing | rapidBreathing
Schwierigkeiten beim Atmen | Difficulty breathing | difficultyBreathing
Kurzatmigkeit | Difficulty in brealthing | difficultyBreathing
Atemnot in Ruhe oder bei Anstrengung | Breathlessness   at rest or during exertion | Missing
Druck auf der Brust | Pressure on   the chest | Missing
Brustschmerzen | Chest pain | chestPain
Blaue Lippen | Blue lips | Missing
Allgemeine Kreislaufprobleme | General   blood circulation problems | Missing
Erhöhte Herzfrequenz (Herzrasen) | Tachycardia | **Fast heart rate (Tachycardia)**
Herzstolpern | Palpitations | Missing
Schwindel (beim Aufstehen aus dem Sitzen und Liegen) | Dizziness   (when standing up from a sitting or lying position) | Missing
Blutdruck zu hoch oder zu tief (gemessen) | Blood   pressure too high or too low (measured) | Missing
[...]

PR #2680, opened 2020-08-21
2669 ch covid symptoms

MartinWahnschaffeSymeda:
#2669


MateStrysewskeSym:
This can't work if the server is configured correctly. "CH" is not a language, but a country, and the current logic only checks the language part of the locale. Will be fixed in a separate issue.

Issue #2608, opened 2020-08-06
Display person uuid on person  forms 


bernardsilenou:
### Proposed Change
- add on:
- case person,
- contact person
- event participant entity: here we show both person id and event participant id
### Additional Information
![image](https://user-images.githubusercontent.com/12996724/89536143-bca44e80-d7f7-11ea-8992-bae6f943f419.png)

JaquM:
this is already done, so I close this Issue

PR #2693, opened 2020-08-24
Added UUID to Person Edit Form which is read-only

fuatse:
Add the display of the UUID as requested in issue #2608

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
@@ -83,6 +83,7 @@ public class PersonEditForm extends AbstractEditForm<PersonDto> {
 
 	private static final long serialVersionUID = -1L;
 
+	private static final String PERSON_UUID_LOC = "personUUIDLoc";
 	private static final String PERSON_INFORMATION_HEADING_LOC = "personInformationHeadingLoc";
 	private static final String OCCUPATION_HEADER = "occupationHeader";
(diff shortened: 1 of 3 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
```diff
@@ -83,6 +83,7 @@
 
 	private static final long serialVersionUID = -1L;
 
+	private static final String PERSON_UUID_LOC = "personUUIDLoc";
```
MateStrysewskeSym, line 86: Should be unneccessary

MateStrysewskeSym (changes requested):
First of all, thanks for your submission and for supporting the project!

I'd prefer the UUID field to be read-only instead of not editable to make it appear as a label instead of a disabled text field. Have a look at the CaseDataForm to see how the case UUID is handled there. Also, please make sure to add an entry to the captions.properties file for Person.UUID and use "Person ID" as the value.

sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
fuatse: Very true.

MateStrysewske (approved):
Thank you for implementing the changes. Please make sure to reference the issue number in every commit next time you're contributing!

fuatse:
> Thank you for implementing the changes. Please make sure to reference the issue number in every commit next time you're contributing!

Noted!

Issue #2584, opened 2020-08-01
"set cleared to unknown" button on symptom information page


Iheanacho2027:
### Feature Description
Just like there is a "set cleared to no" button on symptom information page, there should also "set cleared to unknown" button to accomodate for situations where symptom variables are unknown

AndyBakcsy-she:
Verified issue on SORMAS-Version: 1.50.0-SNAPSHOT

AndyBakcsy-she:
linking new issue #3121

PR #2726, opened 2020-08-27
added setcleared to unknown button #2584

john1496:


sormas-api/src/main/resources/captions.properties
```diff
@@ -1381,6 +1381,7 @@ statisticsDataDisplayed=Data displayed
 symptomsLesionsLocations=Localization of the lesions
 symptomsMaxTemperature=Maximum body temperature in \u00B0 C
 symptomsSetClearedToNo=Set cleared to No
+symptomsSetClearedToUnknown=Set to Unknown
```
MateStrysewske, line 1384: Should be "Set cleared to Unknown"

sormas-api/src/main/resources/captions_ar.properties
```diff
@@ -1134,6 +1134,7 @@ statisticsDataDisplayed = Data displayed
 symptomsLesionsLocations = Localization of the lesions
 symptomsMaxTemperature = Maximum body temperature in ° C
 symptomsSetClearedToNo = Set cleared to No
+symptomsSetClearedToUnknown=Set to Unknown
```
MateStrysewske, line 1137: Revert changes to all captions_XX files that aren't the original captions.properties file, please. Those are handled via the translation tool.

sormas-ui/src/main/java/de/symeda/sormas/ui/symptoms/SymptomsForm.java
```diff
@@ -788,6 +789,35 @@ protected void addFields() {
 			}
 		}, ValoTheme.BUTTON_LINK);
 
+
+		//setting symptoms to unknown button
```
MateStrysewske, line 793: Please extract the code to a dedicated method an reuse it for the two functionalities (setting fields to no and settings fields to unknown) to reduce the amount of duplicate code.

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/AbstractEditForm.java
```diff
@@ -79,7 +79,8 @@ protected AbstractEditForm(
 		this.fieldAccessCheckers = fieldAccessCheckers;
 
 		getFieldGroup().addCommitHandler(this);
-		setWidth(900, Unit.PIXELS);
+		setWidth(1100, Unit.PIXELS);
```
MateStrysewske, line 82: Revert this change because it might have unintended side effects (the width is set to 900 for a reason). Instead, please move the buttons to clear the symptoms and set the cleared symptoms to no/unknown to a new line to make sure that they don't interfer with the description. They should still be right-aligned though.

MateStrysewske (changes requested):
Just like mentioned in your pull request #2741, please use automatic code formatting on all files that you've changed.

Issue #2427, opened 2020-07-02
Unify install scripts


HolgerReiseVSys:
### Feature Description
Integrate `/sormas-base/dev-setup/server-setup-dev-docker.sh` back into `/sormas-base/setup/server-setup.sh`
and clearify setup documentation

### Problem Description
https://github.com/hzi-braunschweig/SORMAS-Project/pull/2331 has duplicated the `server-setup.sh` script.
Having duplicate install scripts for different purposes is error prone when it comes to maintainace.
Also, the installation procedures must be well-documented to facilitate developer onboarding and rollout.
Requiring the latest release artifacts in order to setup a development system has always been cumbersome.

### Proposed Change
Incoroprate use cases that lead to the secession of `/sormas-base/dev-setup/server-setup-dev-docker.sh` back into `/sormas-base/setup/server-setup.sh`.
If needed, `server-setup-dev-docker.sh` could call `server-setup.sh` and control it via environment variables.

Setting up development servers directly from the workspace should be the preferred approach.

The documentation needs to be clearified.
For development systems, only the setup directly from the workspace should be described
(`DEVELOPMENT_ENVIRONMENT.md` should not mention `SERVER_SETUP.md`).
`SERVER_SETUP.md#postgres-database` should give specifics concerning Docker.
Maybe a docker-compose file could be helpful.

### Possible Alternatives
Place a reminder in `server-setup.sh` that `server-setup-dev-docker.sh` has to be kept in sync

#### Aspects to consider:
- SORMAS deployment
  - Machine
  - Container
- System type
  - Productive
  - Demo
  - Development
- OS
  - Win
  - Linux ( /Docker)
- Source
  - Workspace
  - Release
- Postgres
  - Local
  - Remote (/Docker)

jabrandes:
#2375 should provide a fix for most of that :)

MartinWahnschaffe:
It happened as expected - server-setup-dev-docker.sh is no outdated, because it's not actively being used.
sormas-base/server-setup.sh is already taking care of the following things:

1. Local dev system vs. actual server
2. Demo/test vs. production system
3. Linux vs. Windows
4. Authentication provider: SORMAS vs. keycloak on docker (including keycloak-setup.sh) (#2848)
5. Postgres: Existing or new database (#4601)

What remains:

* For dev systems it should be possible to run the server-setup.sh from a local workspace instead of having to download the latest release
[...]

PR #2375, opened 2020-06-23
Enhance development setup for contributors

jabrandes:
This pull request serves a similar cause like the PR hzi-braunschweig/SORMAS-Project#2331 in that it tries to lower the bar for new SORMAS contributors by making initial development server setup easier. In order to achieve this, this PR supplies the following features:

- Extensions of the `sormas-base/setup/server-setup.sh` script
    - The user will now be asked about custom values for the various parameters of server setup like installation directories and database setup; values of the pre-PR script are provided as default values the user may apply by just hitting `Enter`
    - Choices for custom values will be remembered for quick recovery in case the setup script fails
    - `{start,stop}-payara-sormas.sh`,  `payara-sormas.sh` and `sormas.properties` are automatically adjusted to respect the setup choices of the user
    - Additional adjustments like allowing dev systems with demo functionality and optional `sudo` support
- Creation of a new `dev/server-setup.sh` script to replace the `sormas-base/dev-setup/server-setup-with-docker.sh` setup script
    - Checks whether required applications are installed
    - Provides the option to configure a PostgreSQL docker container (see below)
    - Writes environment variables used by development tools (e.g. M2_HOME) to `dev/config/env.sh` file such that these can be quickly set up when starting a development session
    - Configures `sormas-base/build.properties` for the installed server environment
    - Creates a documentation of the configuration values of the development server located at `dev/config/SERVER_SPECS.md`
- Inclusion of a PostgreSQL docker container recipe fit for SORMAS (`dev/postgres-docker`)
- Ant tasks `deploy`, `deploy-serverlibs` and `deploy-artifacts` for deploying builds to the development server
- Adoption of the development server setup documentation to the new dev setup process

#### Related Issues
- #813 (made easier by generating a script setting the M2_HOME environment variable appropriately when sourced)
- #2297 (evaded by providing a preconfigured PostgreSQL Docker container setup)
- #1336 (setup procedure and documentation with command line users in mind)

**Important:**
Since this PR comes with heavy modification to the standard server setup script, some additional testing with test machines should be performed before merging to make sure that everything works as expected. I have only tested the setup procedure on my local Linux machine where it worked as promised. I am not quite sure whether it also works on Windows WSL systems.

One might consider to wait with merging until an amount of new contributors ran the setup procedure on their local machines without critical problems.

PR #2734, opened 2020-08-28
[IMIS] Local dev setup using maven-cargo (fixes #2695)

### LINKED ISSUE #2695: [IMIS] Local dev setup using maven-cargo
### Feature Description
Setup of a local development
 server using maven-cargo-plugin. Setup is reduced to two steps:
- docker-compose up (database
 setup using the docker image)
- mvn cargo:run (install
 payara server, deploy artifacts start server)
The
 implementation of this task includes the integration of serverlib dependency artifacts in the maven build process, generating a dependency jar as well as a zip artifact.

syntakker:
Added modules:
- **sormas-serverlibs**: server dependencies, provides both zip and jar artifacts (to be discussed: integrate sormas-serverlibs zip into server setup to avoid code duplication of `sormas-base/dependencies/serverlibs.pom`)
- **sormas-devserver**: local dev server setup using maven-cargo


syntakker:
The definition of Travis lifecycle phases prior to commit `515a82b` executes `mvn test -B ...` as defined for the `install` phase, followed by an execution of another `mvn test -B` from Travis' default `script` phase. Thus
- `mvn install ...` is never executed, i.e., the maven lifecycle phases `package` and `install` are not covered by a Travis run
- `mvn test -B ...` is executed twice, which substantially increases the time of a Travis run

The Travis log reads (e.g., https://travis-ci.org/github/hzi-braunschweig/SORMAS-Project/builds/721680110):
```
$ mvn test -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn (install) (726.43s)
$ mvn test -B (537.68s)
```

By skipping instead Travis' `install` phase and have `mvn install ...` executed in the `script` phase, Travis runs a complete maven install covering maven's lifecycle phases `test` - as before - as well as `package` and `install`. Now the Travis log reads (https://travis-ci.org/github/hzi-braunschweig/SORMAS-Project/builds/721954371):
```
Skipping the install step, as specified in the configuration.
$ mvn install -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn (803.49s)
```

The elapsed time for the Travis run is reduced from 21 min 27 sec in the above example to 13 min 48 sec in the latter run.

MartinWahnschaffeSymeda:
@syntakker Thanks a lot for the PR!

Starting with a general comment first: The dev/server setup of SORMAS is currently quite messy and needs to be streamlined (PR #2375 will improve that quite a bit). We have the SERVER_SETUP, the SERVER_DEV_SETUP and also the SORMAS-Docker Project. All of this comes with some redundancy (see #2427), so we are a bit hesitant when it comes to adding even more. 

On the other side the approach you have chosen here makes it really easy to quickly start a server to do some testing and development, so we should definitely go for it and have it here as an alternative way.

In detail @StefanKock and I have a few change requests:

- [ ] Since the domain.xml is completely checked in, it's difficult to change the ports. Especially the postgres 5432 is likely to interfere with an existing postgres server on the local system. Either the port configuration needs to be done in another way (e.g. providing the port in an .env file and setting up the database connections with script commands) or we should at least use a different port here (e.g. 5434).
- [ ] Instead of creating a new module for the serverlibs, please use the existing pom in sormas-base/dependencies. The artifact xml files can also be moved there.
- [ ] There are several placeholder.txt files, for this please use the .gitkeep files instead
- [ ] Please change the name of the devserver module to something like quickstart/quickserver (or hopefully you come up with something better). The reason for this request is, that there will be two ways for developers to setup their local server. The solution provided here will be a quick start for people who want to contribute to a few issues. For mid- and long-term developers the "setup" way will for now stay the default, because it's faster to startup, allows easy full customization and a permanent database (the postgres database here will be lost when a mvn clean is done - which is perfect for this use case).
  #2375 is improving the workflow for the dev server to make that 
- [ ] Postgres docker version 2.7.2 does not seem to exist

syntakker:
@MartinWahnschaffeSymeda the requested changes are implemented:

- Cargo server configuration: the database port is configurable, so the server can connect against an existing (permanent) database now. I could also abstract the database host to connect against a remote database, if there is a use case for that. Configuration is done via the `.env` file an described in the `README.md`. The setup does not change, the configurations in `.env` are used for both running the docker image and filtering the `domain.xml` file.
- Serverlibs: I've moved the assembly to the existing `sormas-base/dependencies` module. However, to me both the location of the module (it is the only module not in the toplevel directory) and the naming of the pom seem slightly unnatural (it is not possible to `cd` into the directory and run `mvn install`). The zip assembly is unused for now, but it could replace the ant script to extract the serverlibs. This way the serverlibs are a proper artifact and can be distributed via the maven repository.

syntakker:
For the serverlibs: to avoid any code duplication at all, it might also be possible to extract dependencies with scope `provided` from the deployed artifacts, i.e., `sormas-ear`, `sormas-rest` and `sormas-ui`. I can have a look at this, if this is of interest.

syntakker:
@MartinWahnschaffeSymeda As discussed, I have added some configurability. Additional properties can be defined by adding a file `custom.env` (for docker-compose and domain.xml configuration) and/or `custom.properties` (to be merged in `sormas.properties`), see `sormas-cargoserver/README.md`.
We have also set the tag `latest` for the `sormas-postgres` docker image.

MartinWahnschaffeSymeda:
@syntakker Looks great! 

Shouldn't the custom example files be located in the cargo root directory instead of the target subdirectory, though? At least it works fine for me to have a custom env in the module root (sormas-cargoserver).

Also I'd still suggest to use 5434 as the default port. It would make a conflict with an existing postgres installation by default less likely and from my understanding it would not have any negative side-effects.

syntakker:
Yes, example files in the target folder was by mistake :-)
Default postgres port is now changed to 5434.

Issue #2684, opened 2020-08-21
Linking a case to an event breaks when there is no event in the db


bernardsilenou:
### Proposed Change
-  When user click on "link event" present an option to create an event
- Inform user that there is no existing even that they are allowed to see or edit

PR #2741, opened 2020-08-28
solved Linking a case to an event breaks when there is no event in the db issue #2684

john1496:
… to create event #2684

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/eventLink/EventListComponent.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/eventLink/EventListComponent.java
@@ -20,8 +20,6 @@
 
 package de.symeda.sormas.ui.caze.eventLink;
 
-
-
 import com.vaadin.icons.VaadinIcons;
 import com.vaadin.ui.Alignment;
 import com.vaadin.ui.Button;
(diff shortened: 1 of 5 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/eventLink/EventListComponent.java
```diff
@@ -37,14 +43,29 @@
...
 	private EventList list;
 	private Button createButton;
 
 	public EventListComponent(CaseReferenceDto caseRef) {
 
-		createEventListComponent(new EventList(caseRef), e -> ControllerProvider.getEventController().selectOrCreateEvent(caseRef));
+		createEventListComponent(new EventList(caseRef), e -> {
+
+			//check if there are active events in the database
+			List<String> events = FacadeProvider.getEventFacade().getAllActiveUuids();
```
MateStrysewske, line 58: Please use a "count" method instead for performance reasons.

MateStrysewske (changes requested):
Please have a look at the guide to set up your local dev environment again. There's a section on code formatting there. Please configure code formatting in your IDE as specified in the guide and reformat all files that you've changed.

Also, I don't think that displaying a message when there's no event in the database is necessary. This will only happen once on the server (per user) anyway and the information isn't really relevant for the user in the first place.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/eventLink/EventListComponent.java
john1496: ok will do

MateStrysewske (changes requested):
Please use code formatting as described in the development environment guide.

MateStrysewske (approved):

Issue #2735, opened 2020-08-28
Don't re-use symptoms, epi data and health conditions when converting contacts to cases


MateStrysewske:
Currently, when converting a contact to a case, we simply set the existing epi data and health conditions from the contact to the case as well (+ the symptoms of the last cooperative visit), i.e. we use the same entity. This is very problematic because these are embedded entities and it leads to side-effects in the synchronization to the mobile app.

Instead of the current logic, we need to:

- [x] Clone epi data, symptoms and health conditions when converting contacts to cases

Additionally, we need to migrate the existing data to make sure that we don't run into problems on existing servers:

- [x] Write SQL script that creates new database entries for epi data and symptoms that are used in both a case and a contact/visit, copy the values from the existing epidata/symptoms entity and make sure that both entities are only referencing one contact/visit OR case, but not both; this also needs to be done for gatherings, travels and burials

bernardsilenou:
@xavier-calland this means our solution discussed today on the followup of cases may be affected since it was based on the old design. After the release on Monday, i suggest we have a second look,

PR #2747, opened 2020-08-31
#2735 - clone child entities when converting contact to case

BarnaBartha:


MateStrysewskeSym:
- [ ] Locations of burials and gatherings also need to be assigned a new UUID.

PR #2750, opened 2020-08-31
#2678 - Case follow-up adjustments

### LINKED ISSUE #2678: [FRENCH TEAM] Changes to be made for followup of cases
### Feature Description
- These are points we identified after testing the followup module.
### Proposed Change

1. Use date of report as reference date to determine the start and end of followup always. Because people can get sick and go to hospital 6 days latter, thus follow up will be less. In addition, when the symptom onset date is latter registered, all visits are excluded from the followup chart. 

2. In the followup tab of case directory, filter by investigation status does not work right. When you chose pending, cases with “investigation done” shows up.

3. On case information form, move reception date of paper form below follow-up status

xavier-calland:
Fixes #2678


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
```diff
@@ -192,6 +189,8 @@
 					fluidRowLocs(CaseDataDto.FOLLOW_UP_COMMENT);
 
 	private static final String COMMENTS_HTML_LAYOUT =
+			loc(PAPER_FORM_DATES_LOC) +
```
MateStrysewske, line 192: This is semantically wrong; the paper form fields shouldn't be in something that's called "COMMENTS_HTML_LAYOUT". Please either find a new name for this field or split it into two fields (COMMENTS_LAYOUT and PAPER_FORM_DATES_LAYOUT).

MateStrysewske (changes requested):
I think what's missing here is the change to the follow-up until calculation. You've only adjusted how it's displayed in the CaseFollowUpGrid, but if I understand Bernard's comment correctly, this needs to be changed in the backend as well.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
xavier-calland: Yes, you're right.

xavier-calland:
> I think what's missing here is the change to the follow-up until calculation. You've only adjusted how it's displayed in the CaseFollowUpGrid, but if I understand Bernard's comment correctly, this needs to be changed in the backend as well.

After discussion with Bernard, it seems that it only needs to be changed in `CaseFollowUpGrid`.

@bernardsilenou, can you confirm?

bernardsilenou:
> > I think what's missing here is the change to the follow-up until calculation. You've only adjusted how it's displayed in the CaseFollowUpGrid, but if I understand Bernard's comment correctly, this needs to be changed in the backend as well.
> 
> After discussion with Bernard, it seems that it only needs to be changed in `CaseFollowUpGrid`.
> 
> @bernardsilenou, can you confirm?

- I suggest we change both: Backend and  CaseFollowUpGrid
- I also suggest we call the "follow-up until calculation" method each time we save the entity if the date parameters used in the calculation has been updated

xavier-calland:
> * I suggest we change both: Backend and  CaseFollowUpGrid

Does that mean removing `CaseLogic.getStartDate` (always use `reportDate`) and similar logic in `CaseService` db queries ?

`CaseLogic.getStartDate` is used in :
- `CaseFacade.updateCaseAge`
- `CaseFacade.updateCaseVisitAssociations`
- `ClassificationNotInStartDateRangeCriteriaDto.eval`
- `CaseService.updateFollowUpUntilAndStatus`
- `CaseDataForm.addFields`
- `CaseFollowUpGrid.setVisitColumns`
- `VisitEditForm.addFields`

`Symptoms.ONSET_DATE` is used in `CaseService` :
- `createCaseRelevanceFilter`
- `createCaseRelevanceFilter`
- `buildRelevantCasesFilterForFollowUp`

Does that mean to remove the `onsetDate` parameter in `CaseLogic.getEndDate` ?

> * I also suggest we call the "follow-up until calculation" method each time we save the entity if the date parameters used in the calculation has been updated

`CaseService.updateFollowUpUntilAndStatus` is already call each time a visit attached to a case is saved (in `CaseService.onVisitChanged`).
Is there more to do ?

bernardsilenou:
Its @MateStrysewske  who can confirm this

MateStrysewske:
@xavier-calland You should only change code associated with calculating the follow-up period for cases, nothing else, so the case start age is still relevant for other logic. But yes, for these calculations, you don't need the start age anymore since you can simply use the report date.

MateStrysewske (approved):

PR #2752, opened 2020-08-31
Feature 2111 keycloak authentication

vidi42:


sormas-ui/src/main/java/de/symeda/sormas/ui/security/config/DefaultOpenIdAuthenticationDefinition.java
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	@Override
+	public boolean tokenAutoRefresh() {
+		return false;
+	}
+
+	@Override
+	public int tokenMinValidity() {
+		return 10 * 1000;
```
BarnaBartha, line 174: shouldn't these values be extracted somewhere in a properties file?

sormas-backend/src/main/java/de/symeda/sormas/backend/user/KeycloakService.java
```diff
@@ -0,0 +1,249 @@
...
+
+	public void handlePasswordResetEvent(@Observes PasswordResetEvent passwordResetEvent) {
+		Optional<Keycloak> keycloak = getKeycloak();
+		if(!keycloak.isPresent()) {
+			logger.warn("Cannot obtain keycloak instance. Will not reset user password in keycloak");
+			return;
+		}
+
+		User user = passwordResetEvent.getUser();
+		Optional<UserRepresentation> userRepresentation = getUserByUsername(keycloak.get(), user.getUserName());
```
BarnaBartha, line 135: for the update you are logging in case there is no user with this name - shouldn't it be logged here as well?
also this method to get the keycloak user, just for me to understand does keycloak allow duplicate user names for the same realm ? or is it a configuration ?

BarnaBartha:
code changes look good, clean and organised
i've added some questions but my bigger suggestion is to extend the development environment documentation for what one needs if wants to use sormas with keycloak auth locally

BarnaBartha (approved):

sormas-ui/src/main/java/de/symeda/sormas/ui/security/config/DefaultOpenIdAuthenticationDefinition.java
vidi42: The values in this class are just the default values copied from `OpenIdAuthenticationDefinition` and they can be overwritten trough MicroProfile Config Properties.

In this case the `tokenMinValidity` can be overwritten by setting the `payara.security.openid.token.minValidity` property.

sormas-backend/src/main/java/de/symeda/sormas/backend/user/KeycloakService.java
vidi42: Good point, I will add a log here as well.

It's a big strange because Keycloak doesn't allow duplicate usernames for the same realm, but it only offers a search endpoint to retrieve users by username which can be used to also search for substrings of the username, that's why the result is always returned as a list.

In our case however we are using `exact=true` which means that it searches for the exact username, but API still returns the response as a list.

Issue #1970, opened 2020-05-04
Allow PIA to inform SORMAS about the status of a contact person in PIA


MartinWahnschaffeSymeda:
### Situation Description

Extension for #1669

### Feature Description

- [x] ReST interface should receive the following information:
   - UUID of the person
   - Status: PIA account created, account deleted
- [x] Store the status of the PIA account for the contact person

Do we really need account created? This could also be handled within the iframe #1669 
Do we need any other information (e.g. reason for the account deletion)?

markusmann-vg:
Where do we show the PIA status in SORMAS?

MartinWahnschaffeSymeda:
I think we should show it within each contact. The same place we want to display the PIA iframe.

o-frietsch:
Final interface specification:

HTTP-Request
------------
	Endpoint: /sormas-rest/visits-external/contact/[contactUuid]/status
	Method: POST

	URL example: http://localhost:6080/sormas-rest/visits-external/contact/S3CJF2-Y7SCJZ-NQVERG-43CGCDL4/status

	Body example:

		{
			status: “REGISTERED”,
			statusDateTime: "2020-04-15T12:55:00.000+02:00" // datetime format yyyy-MM-dd'T'HH:mm:ss.SSSZ
		}

HTTP-Response:
--------------
	returns boolean literal

		"true" - if contactUuid was valid and status was set
		"false" - otherwise

Possible status codes:
----------------------
	UNREGISTERED: User has not yet sent any state (never sent by PIA, just internal state in SORMAS)
	REGISTERED: After succesfull registration in PIA
	ACCEPTED: User has accepted the In-App confirmation
	REJECTED: User has rejected (declined) the In-App confirmation
	DELETED: User was deleted in PIA (=UNREGISTERED?)

PIA ensures following status changes:
-------------------------------------
	UNREGISTERED -> REGISTERED -> DELETED
	UNREGISTERED -> REGISTERED -> ACCEPTED or REJECTED -> DELETED

o-frietsch:
See https://github.com/hzi-braunschweig/SORMAS-Project/issues/2083 for UI extensions.

o-frietsch:
The above spec has also been transmitted by email some days ago, this is the same spec.

FredrikSchaeferVitagroup:
I have a question regarding the specification: Is it correct that the contactUuid shall be used here? Don't we use the personUuid?

o-frietsch:
[...]

PR #2756, opened 2020-09-01
#1970 Allow PIA to inform SORMAS about the status of a contact person

FredrikSchaeferVitagroup:
rebased state

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/backend/person/Person.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/backend/person/Person.java
@@ -40,6 +40,7 @@ import de.symeda.sormas.api.person.EducationType;
 import de.symeda.sormas.api.person.OccupationType;
 import de.symeda.sormas.api.person.PresentCondition;
 import de.symeda.sormas.api.person.Sex;
+import de.symeda.sormas.api.person.SymptomJournalStatus;
 import de.symeda.sormas.app.backend.common.PseudonymizableAdo;
 import de.symeda.sormas.app.backend.facility.Facility;
 import de.symeda.sormas.app.backend.location.Location;
@@ -173,6 +174,8 @@ public class Person extends PseudonymizableAdo {
 	private String passportNumber;
 	@Column
 	private String nationalHealthId;
+	@Enumerated(EnumType.STRING)
+	private SymptomJournalStatus symptomJournalStatus;
 
 	public Person() {
 	}
@@ -575,5 +578,13 @@ public class Person extends PseudonymizableAdo {
 
 	public void setOccupationFacilityType(FacilityType occupationFacilityType) {
 		this.occupationFacilityType = occupationFacilityType;
+	}
+
+	public SymptomJournalStatus getSymptomJournalStatus() {
+		return symptomJournalStatus;
+	}
+
+	public void setSymptomJournalStatus(SymptomJournalStatus symptomJournalStatus) {
+		this.symptomJournalStatus = symptomJournalStatus;
 	}
 }
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/backend/person/Person.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/backend/person/Person.java
@@ -174,8 +174,6 @@ public class Person extends PseudonymizableAdo {
 	private String passportNumber;
 	@Column
 	private String nationalHealthId;
-	@Enumerated(EnumType.STRING)
-	private SymptomJournalStatus symptomJournalStatus;
 
 	public Person() {
 	}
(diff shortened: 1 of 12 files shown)


MateStrysewske (changes requested):
Please remove the symptomJournalStatus field from the app; it's not needed there and thus shouldn't be synchronised in the first place.

sormas-rest/src/main/java/de/symeda/sormas/rest/ExternalVisitsResource.java
```diff
@@ -30,6 +35,23 @@ public Boolean isValidPersonUuid(@PathParam("personUuid") String personUuid) {
 		return FacadeProvider.getPersonFacade().isValidPersonUuid(personUuid);
 	}
 
+	@POST
+	@Path("/person/{personUuid}/status")
+	public boolean postSymptomJournalStatus(@PathParam("personUuid") String personUuid, Map<String, Object> body) {
```
BarnaBartha, line 40: why not create here a specific transfer object for the body?

sormas-rest/src/main/java/de/symeda/sormas/rest/ExternalVisitsResource.java
```diff
@@ -30,6 +35,23 @@ public Boolean isValidPersonUuid(@PathParam("personUuid") String personUuid) {
...
+		try {
+			PersonDto person = FacadeProvider.getPersonFacade().getPersonByUuid(personUuid);
```
BarnaBartha, line 42: usually it is desirable not to have any kind of logic in the resource layer - i would advise to expose a specific method in the facade

sormas-rest/src/main/java/de/symeda/sormas/rest/ExternalVisitsResource.java
```diff
@@ -30,6 +35,23 @@ public Boolean isValidPersonUuid(@PathParam("personUuid") String personUuid) {
...
+			SymptomJournalStatus statusValue = SymptomJournalStatus.valueOf(body.get("status").toString());
+			person.setSymptomJournalStatus(statusValue);
+			FacadeProvider.getPersonFacade().savePerson(person);
+			return true;
+		} catch (Exception e) {
+			Logger logger = LoggerFactory.getLogger(ExternalVisitsResource.class);
```
BarnaBartha, line 48: in the parent class there is a logger defined, please use that one

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/person/Person.java
```diff
@@ -173,6 +174,8 @@
 	private String passportNumber;
 	@Column
 	private String nationalHealthId;
+	@Enumerated(EnumType.STRING)
+	private SymptomJournalStatus symptomJournalStatus;
```
BarnaBartha, line 178: @MartinWahnschaffeSymeda is this something we expect to receive from other 3rd party systems as well? or only from PIA ? if it is only from PIA i suggest we move this to a separate table, for example PiaRegistration and do a join with this table when needed
this way we could keep the domain entity clean from 3rd party app influences

BarnaBartha (changes requested):

sormas-rest/src/main/java/de/symeda/sormas/rest/ExternalVisitsResource.java
FredrikSchaeferVitagroup: Conventic gave us [some API specification](https://github.com/hzi-braunschweig/SORMAS-Project/issues/1970) with one field that we don't need (statusDateTime). Instead of putting this into a transfer object, I preferred this more flexible approach. Imagine that, in the future, some other symptom journals want to give us more or other additional information. With this more generic approach, we are better prepared for that. Is that convincing?

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/person/Person.java
FredrikSchaeferVitagroup: I guess we may, in the future, also get a status from other symptom journals, for example from climedo. I don't know if they will always have the same status types (ACCEPTED, REJECTED.. as defined in the enum), though. I would also like Martins opinion about it.

sormas-rest/src/main/java/de/symeda/sormas/rest/ExternalVisitsResource.java
BarnaBartha: not really, we should be having one standardised API and not expect everything
nonetheless with a TO this expectation is also doable as it can be extended and at least if the body doesn't correspond to the object is handled by javax rs and no need to catch a possible nullpointer

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/person/Person.java
FredrikSchaeferVitagroup: @MateStrysewske also thinks that this should be generic.

sormas-rest/src/main/java/de/symeda/sormas/rest/ExternalVisitsResource.java
FredrikSchaeferVitagroup: As you suggested, I have moved the logic from the ExternalVisitRessource to the PersonFacade. The Exception is now thrown from in there, and automatically logged. The ExternalVisitRessource appears in the stack trace. Is that also acceptable?
When I forward the exception and log it from within the ExternalVisitRessource with the parent's logger, the server log looks messier...

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/person/Person.java
MartinWahnschaffeSymeda: Yes, I agree - should be generic for now.

Issue #2235, opened 2020-06-02
Quarantine end field to suggest end of follow-up date [1]


max-hzi:
### Feature Description
When the quarantine field is activated, the “quarantine end” field should suggest the same end date as the follow-up end. If the quarantine end date is changed manually, a pop up window should ask if the follow-up end date should be adjusted as well. If “yes” is clicked, the follow-up end date should be changed to the quarantine end date. In case of "no", the follow-up date should remain untouched (similar to #2233).

markusmann-vg:
Do together with #2233

vlad-ciucescu:
As per @bernardsilenou 's comment in #2233 

>  4.1 If quarantine end date is reduced from its default followup until date, show dialog box to ask user if they want to reduce it from it expected end date. if yes, then set "quarantine period modified" to "reduced" and do not modify followup until date.

> 4.2 If quarantine end date is increased from its default followup until date, show dialog box to ask user if they want to increase it from it expected end date. if yes, then set "quarantine period modified" to "extended" AND also ask if they want to extend the followup until date, if yes then extend accordingly.

The quarantine extended checkbox implemented in #2264 now shows either "quarantine period reduced" or "quarantine period extended" based on how the date was modified.
This also implies changes to the database to store the new "quarantine period reduced" status, and a new filter.

AndyBakcsy-she:
Verified issue on SORMAS Version: 1.49.0-SNAPSHOT

PR #2760, opened 2020-09-01
#2235 quarantine end field

vlad-ciucescu:


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
```diff
@@ -864,32 +870,108 @@ protected void addFields() {
...
 					if (confirmed) {
-						if (!originalCase.isQuarantineExtended()) {
-							quarantineExtendedCheckbox.setValue(true);
-						}
-					} else {
-						quarantineEndField.setValue(oldQuarantineEnd);
+						originalCase.setFollowUpUntil(quarantineEnd);
+						followUpUntil.setReadOnly(false);
+						followUpUntil.setValue(quarantineEnd);
+						followUpUntil.setReadOnly(true);
```
leventegal-she, line 943: followUpUntil is not always readonly, here you force it to readonly.

See the checkbox  "Overwrite follow-up until date" besides the date field

sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactDataForm.java
```diff
@@ -638,29 +651,120 @@ protected String createHtmlLayout() {
...
 					if (confirmed) {
-						if (!originalContact.isQuarantineExtended()) {
-							quarantineExtendedCheckbox.setValue(true);
-						}
-					} else {
-						quarantineEndField.setValue(oldQuarantineEnd);
+						originalContact.setFollowUpUntil(quarantineEnd);
+						followUpUntil.setReadOnly(false);
+						followUpUntil.setValue(quarantineEnd);
+						followUpUntil.setReadOnly(true);
```
leventegal-she, line 724: followUpUntil is not always readonly, here you force it to readonly

leventegal-she (changes requested):
- [x] You should provide **only English** translations, the rest will be done by the ones who provide the translations for all other languages. Please remove the non English ones
- [ ] The followup feature is optional on cases, followup related question should be shown only if case followups are enabled, also the default value of quarantine end date should be empty when the followups are not enabled.
- [x] "Only contacts with reduced quarantine" filter is not visible on the Case directory page

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseExportDto.java
```diff
@@ -681,6 +683,16 @@ public boolean isQuarantineExtended() {
 		return quarantineExtended;
 	}
 
+	@Order(137)
```
leventegal-she, line 686: also the @Order should be changed to `42` and increase the ones following

@Order annotation sets the position of the column in the exported CSV. 

`quarantineExtended`, `quarantineReduced` fields should go one after another.

leventegal-she (changes requested):
Looks like the mobile app implementation is missing but the ticket has the `sormas-app` flag.

sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactExportDto.java
```diff
@@ -403,6 +405,11 @@ public boolean isQuarantineExtended() {
 		return quarantineExtended;
 	}
 
+	@Order(59)
```
leventegal-she, line 408: the order should be 38

leventegal-she (changes requested):

leventegal-she (changes requested):
Some bugs found in the web app:
- [ ] When I first set the quarantine to be Institutional, the quarantine end date is not set to the followup until date
- [ ] When I first set the quarantine end date the system asks "Are you sure you want to reduce the quarantine?"
- [ ] When I open a case that has followup until before the quarantine end date, the system asks "Are you sure you want to reduce the quarantine?"
- [ ] when I open a case I see both extended and reduced checkboxes but when I change the end date the unchecked one disappears
- [ ] I would hide the "quarantine extended/reduced" checkboxes until they are not checked. It looks wierd when you see the unchecked "quarantine extended" and when you reduce the quarantine it turns into quarantine reduced

leventegal-she:
On mobile:

- [ ] When editing I see quarantine reduced checkbox checked even if I extend it. Looks like it checks the extended checkbox but it gets unchecked right after the confirmation dialog disappears
- [ ] On a contact created in the app, I get a NullPointerException when trying to extend the quarantine, probably due to followup until date is not set
- [ ] Both extended end reduced checkboxes are always visible, I think it should be similar to web

leventegal-she (approved):

Issue #2769, opened 2020-09-01
Add "Unknown" as category of present condition to person entity


bernardsilenou:

AndyBakcsy-she:
Verified Issue on SORMAS Version: 1.50.0-SNAPSHOT

PR #2780, opened 2020-09-02
added unknown field to presentCondition #2769

john1496:


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1724,6 +1722,13 @@ private void updatePersonAndCaseByOutcome(CaseDataDto existingCase, Case newCase
 					PersonDto existingPerson = PersonFacadeEjb.toDto(newCase.getPerson());
 					newCase.getPerson().setPresentCondition(PresentCondition.ALIVE);
 					// attention: this may lead to infinite recursion when not properly implemented
+					personFacade.onPersonChanged(existingPerson, newCase.getPerson());
+				}
+			}else if(newCase.getOutcome()==CaseOutcome.UNKNOWN) {
```
MateStrysewske, line 1727: This code doesn't make much sense to me. Just because the outcome of a case is unknown doesn't mean that the present condition also necessarily has to be unknown. It could just mean that it's not known whether the person has already recovered or not, but is definitely alive. Please remove this change.

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -327,6 +325,17 @@ private void cleanUp(Person person) {
 			person.setBurialDate(null);
 			person.setBurialPlaceDescription(null);
 			person.setBurialConductor(null);
+
+		}if(person.getPresentCondition()==PresentCondition.UNKNOWN){
```
MateStrysewske, line 329: You can simply extend the first line in this method like this:

`if (person.getPresentCondition() == null || person.getPresentCondition() == PresentCondition.ALIVE || person.getPresentCondition() == PresentCondition.UNKNOWN) {`

instead of adding all these new code lines.

MateStrysewske (changes requested):

Issue #2233, opened 2020-06-02
Changes in follow-up date triggers pop-up window to ask for changes for quarantine end date [1]


max-hzi:
### Feature Description
When the follow-up time is changed manually a pop-up window should ask if the quarantine end time shall also be adjusted automatically. When ”yes” is clicked, the quarantine end date should be changed to the same end date as the follow-up end date. In case of "no", quarintine end date should remain untouched.

markusmann-vg:
reducing and extending the follow-up date, which might change the quarantine end date is possible
When the quarantine end date is also extended, then the checkbox (quarantine end date extended) is ticked and stored in the db.

bernardsilenou:
@markusmann-vg  This suggested refinement is related to issues #2426 #2235 . I think its good to see the complete story while implementing any of these issues.
1. Use the followup until date as reference date for quarantine end date. Cases will also have followup until date based on issue #2488 
2. Set default value of quarantine end date == followup until date
3. Followup until date is not possible to be reduced, thus can only be extended. If over written and extended, ask is user want to extend the quarantine end date, if yes, then match it to followup until date and set "quarantine period extended" which we can now call "quarantine period modified" to "Extended"
4. Contrary to followup until date, quarantine end date can be reduced and also increased from its default and expected value. 
4.1 If quarantine end date is reduced  from its default followup until date, show dialog box to ask user if they want to reduce it from it expected end date. if yes, then set "quarantine period modified" to "reduced" and do not modify followup until date.

4.2 If quarantine end date is increased  from its default followup until date, show dialog box to ask user if they want to increase it from it expected end date. if yes, then set "quarantine period modified" to "extended" AND also ask if they want to extend the followup until date, if yes then extend accordingly.

PR #2783, opened 2020-09-02
#2233: Follow-up end change triggers change request

ChristopherRiedel:


MateStrysewske (changes requested):
The popup that asks the user to change the quarantine end date only pops up AFTER the case has been saved. If the user clicks "Yes", the change is never actually saved. Everytime the user navigates back to the case afterwards (probably also when "No" is selected), the dialog pops up again immediately.

ChristopherRiedel:
> The popup that asks the user to change the quarantine end date only pops up AFTER the case has been saved. If the user clicks "Yes", the change is never actually saved. Everytime the user navigates back to the case afterwards (probably also when "No" is selected), the dialog pops up again immediately.

Thats just not true.

a-bakcsy-she:
Verified this on SORMAS Version: 1.48.0-SNAPSHOT

It is NOT reproducible. 
The pop-up is displayed immediately after the user changes the end date. 
Clicking on No does not save any changes, as well, clicking on Yes and afterwards on Discard reverts all changes.

vlad-ciucescu:
Also checked and cannot reproduce. I got the same results as Andy.

vlad-ciucescu (approved):

PR #2785, opened 2020-09-02
WIP - 2758 cough with heamoptysis is misspelled

tabotkevin:


sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1337,7 +1337,7 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 			case 200:
 				currentVersion = 200;
 				getDao(Symptoms.class).executeRaw("ALTER TABLE symptoms ADD COLUMN coughWithSputum varchar(255);");
-				getDao(Symptoms.class).executeRaw("ALTER TABLE symptoms ADD COLUMN coughWithHeamoptysis varchar(255);");
+				getDao(Symptoms.class).executeRaw("ALTER TABLE symptoms ADD COLUMN coughWithHemoptysis varchar(255);");
```
MateStrysewske, line 1340: Revert this change because it will break existing servers. You need to write a migration script instead that renames the column. Since SQLite unfortunately doesn't support renaming columns, you'll probably have to re-create the whole symptoms table by creating a temporary table as a copy of the existing symptoms table and transfering the values between those tables. We've done this already before in the mobile app for other tables, e.g. in version number 140 for samples.

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -4177,9 +4177,9 @@ ALTER TABLE symptoms_history ADD COLUMN lossofsmell varchar(255);
 
 INSERT INTO schema_version (version_number, comment) VALUES (205, 'Added symptoms loss of taste and loss of smell #1936');
 
--- 2020-05-05 Add new symptoms and health conditions #1824
+-- 2020-09-02 changed column name from coughWithHeamoptysis to coughWithHemoptysis #2758
```
MateStrysewske, line 4180: Revert these changes because they will break existing servers. You need to write a migration script instead that renames the column. NEVER change existing SQL scripts.

MateStrysewske (changes requested):
Thanks for your contribution. Please make sure to notify us when you want to take care of an issue directly inside the issue. In this case, one of our developers has already started working on it because he wasn't aware that you're doing the same.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
tabotkevin: Well noted @MateStrysewske, should I leave this issue for your developer to finish?

MateStrysewske: We don't really have resources to work on this right now; since it's not an extremely urgent issue, I'd appreciate if you could work on it; otherwise, we'll take care of it at some point in the future.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1578,6 +1578,65 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				getDao(Case.class).executeRaw("ALTER TABLE contacts ADD COLUMN quarantineOfficialOrderSent SMALLINT DEFAULT 0;");
 				getDao(Case.class).executeRaw("ALTER TABLE contacts ADD COLUMN quarantineOfficialOrderSentDate timestamp;");
 
+			case 226:
```
MateStrysewske, line 1581: You also have to increase the DATABASE_VERSION by 1; please make sure to test your database migration script afterwards.

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -5269,4 +5269,7 @@ ALTER TABLE campaignformmeta_history ADD COLUMN campaignformtranslations json;
 
 INSERT INTO schema_version (version_number, comment) VALUES (250, 'Campaign diagram visualization refinement #2753');
 
+-- 2020-09-11 changed column name from coughWithHeamoptysis to coughWithHemoptysis #2758
+ALTER TABLE symptoms RENAME COLUMN coughWithHeamoptysis TO coughWithHemoptysis;
```
MateStrysewske, line 5273: The INSERT INTO command is missing

MateStrysewske (changes requested):

MateStrysewske (commented):
The database script you've written for the app doesn't run through. It's an absolute necessity that you test your code before submitting it, which apparently hasn't happened here.

Also the script in the sormas_schema has the wrong schema version number assigned which leads to test failures. Please also make sure that the tests succeed before submitting your code.

We will, at some point, move to another database architecture in the mobile app that makes code like this unnecessary. Renaming a column in the app right now is very tedious, and it's not worth the effort that we've already spent on it and the risk involved (e.g. by forgetting to copy a column which would result in data loss). Because of that, I'm gonna close this pull request and we'll take it on again in the future once we've gotten rid of SQLite.

tabotkevin:
@MateStrysewske, thank you very much for the information. The pull request title starts with "WIP" to mean "Work in Progress". I didn't push the code as a submission worthy code.

MateStrysewske:
@tabotkevin Oh, I see! Sorry, I missed that. Thank you for your contribution anyway - but I think both your and our time is better spent on more pressing matters! :)

Issue #2626, opened 2020-08-11
[SORMAS2SORMAS] Add encryption [2]


markusmann-vg:
### Feature Description
- Encryption/Decryption is needed for the binary data which is sent from one Sormas Instance to another one.
- After setting up the local keystore (#2621) and truststore (#2622), encrypted data can be shared between SORMAS instances which have agreed to communicate (#2622, #2623)
- Provide an encryption/decryption service for encrypting/decrypting data sent to/received from a certain SORMAS instance

HolgerReiseVSys:
I have [added some crypto utils](/hzi-braunschweig/SORMAS-Project/compare/feature-2626-SORMAS2SORMAS-crypto)

Issue #2623, opened 2020-08-11
[SORMAS2SORMAS] Create list with server access data  [1]


markusmann-vg:
### Feature Description
- Create a list with the following data about servers: ID, Name, URL.
- This should be generated together with the self signed certificate (as done in #2621 )
- This list should be passed along together with the certificate when connecting to other SORMAS instances (as done in #2622 )
- Update the available SORMAS to SORMAS guide (written in #2621) with the steps related to handling this list.
- Create a Java class for reading the locally available list.

### Additional Information
How to maintain the list for Admins

PR #2786, opened 2020-09-02
#2623 #2626 sormas to sormas server list & encryption

vlad-ciucescu:


sormas-backend/src/main/java/de/symeda/sormas/backend/sormas2sormas/Sormas2SormasEncryptionService.java
```diff
@@ -0,0 +1,81 @@
...
+		PrivateKey privateKey = (PrivateKey) keystore.getKey(config.getKeyAlias(), config.getKeystorePass().toCharArray());
+
+		Path truststorePath = Paths.get(config.getFilePath(), config.getTruststoreName());
+		KeyStore truststore = KeyStore.getInstance("pkcs12");
+		truststore.load(new FileInputStream(truststorePath.toFile()), config.getTruststorePass().toCharArray());
+		X509Certificate recipientCertificate = (X509Certificate) truststore.getCertificate(instanceID);
+
+		byte[] encryptedBytes =
+			CmsCreator.signAndEncrypt(data.getBytes(StandardCharsets.UTF_8), signerCertificate, privateKey, recipientCertificate, true);
+		return new String(encryptedBytes, StandardCharsets.UTF_8);
```
HolgerReiseVSys, line 61: I don't see why the encrypted data should be a valid UTF-8 String.

sormas-backend/src/main/java/de/symeda/sormas/backend/sormas2sormas/Sormas2SormasEncryptionService.java
```diff
@@ -0,0 +1,81 @@
...
+import com.google.common.collect.Lists;
+
+import de.symeda.sormas.api.Sormas2SormasConfig;
+import de.symeda.sormas.backend.common.ConfigFacadeEjb;
+import de.symeda.sormas.backend.crypt.CmsCreator;
+import de.symeda.sormas.backend.crypt.CmsReader;
+
+@Stateless
+@LocalBean
+public class Sormas2SormasEncryptionService {
```
HolgerReiseVSys, line 41: This service seems not to be covered by tests.

sormas-backend/src/main/java/de/symeda/sormas/backend/sormas2sormas/Sormas2SormasEncryptionService.java
```diff
@@ -0,0 +1,81 @@
...
+		byte[] encryptedBytes =
+			CmsCreator.signAndEncrypt(data.getBytes(StandardCharsets.UTF_8), signerCertificate, privateKey, recipientCertificate, true);
+		return new String(encryptedBytes, StandardCharsets.UTF_8);
+	}
+
+	public String decrypt(String data, String instanceID) throws Exception {
+		Sormas2SormasConfig config = configFacade.getSormas2SormasConfig();
+		Path keystorePath = Paths.get(config.getFilePath(), config.getKeystoreName());
+		KeyStore keystore = KeyStore.getInstance("pkcs12");
+		keystore.load(new FileInputStream(keystorePath.toFile()), config.getKeystorePass().toCharArray());
```
HolgerReiseVSys, line 68: Handling of keystores should be extracted

sormas-backend/src/main/java/de/symeda/sormas/backend/sormas2sormas/Sormas2SormasEncryptionService.java
```diff
@@ -0,0 +1,81 @@
...
+import de.symeda.sormas.backend.crypt.CmsReader;
+
+@Stateless
+@LocalBean
+public class Sormas2SormasEncryptionService {
+
+	@EJB
+	private ConfigFacadeEjb.ConfigFacadeEjbLocal configFacade;
+
+	public String encrypt(String data, String instanceID) throws Exception {
```
HolgerReiseVSys, line 46: Should the service encrypt byte arrays instead?

GUIDE_SORMAS2SORMAS_CERTIFICATE.md
```diff
@@ -31,18 +33,25 @@ the path to the ``sormas.properties`` file.
...
+   file will also have to be shared so that the access data of this instance is known to other instances. 
+   More details can be found in the next section.
+8. The relevant properties will be automatically set by the script in the ``sormas.properties`` file.
 
 ### Adding a new certificate to the Truststore
 
 To enable other SORMAS instances to request data from this instance, their certificate must be added to the 
-truststore of this instance. After obtaining their certificate file, which should be a ``.crt`` file, please
-follow the next steps:
+truststore of this instance. Furthermore, the access data of other instances must be added to the local server
```
HolgerReiseVSys, line 53: As fa as I can see, "request data from this instance" is not supported. The truststore is required for sending and receiving data. The `server-list.csv` is used for sending data to other instances.

HolgerReiseVSys (changes requested):
When the whole transport is working, we should probably have another look at the scripts.
To me this all feels a bit too cumbersome, e.g. that the csv lists have to be handled manually.

vlad-ciucescu:
> 
> 
> When the whole transport is working, we should probably have another look at the scripts.
> To me this all feels a bit too cumbersome, e.g. that the csv lists have to be handled manually.

Of course, after we integrate the changes we will have a look at how we can automatise the handling of the CSVs

sormas-backend/src/main/java/de/symeda/sormas/backend/sormas2sormas/Sormas2SormasEncryptionService.java
vlad-ciucescu: I tried to write tests for this service but I realised that I would just be testing whether methods from the CmsCreator and the CmsReader are called, because in the rest of the methods I am basically just collecting configuration data to pass to those other classes.
If you have a suggestion as to how I should test this service, please tell me.

sormas-base/setup/sormas.properties
```diff
@@ -58,6 +58,10 @@ app.url=
 # default: "/opt/sormas/custom/"
 #custom.path=/opt/sormas/custom/
 
+# Path where SORMAS to SORMAS related certificates and files are stored
+# default: "/opt/sormas/sormas2sormas/"
+#sormas2sormas.path=/opt/sormas/sormas2sormas/
+
```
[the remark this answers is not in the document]
vlad-ciucescu: Those are automatically generated and filled by the scripts, should I also add the commented version? In that case I must modify the scripts to remove the commented properties too before generating the new ones.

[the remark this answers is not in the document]
HolgerReiseVSys: By convention all properties need to be added there.
The default case will be anyways that the properties are being set by docker,
see [commit "added sormas2sormas "](/hzi-braunschweig/SORMAS-Docker/commit/392b8507d9f9ec127b547004c4a468c5c45b46ee)

sormas-backend/src/main/java/de/symeda/sormas/backend/sormas2sormas/Sormas2SormasEncryptionService.java
HolgerReiseVSys: I guess we can do without when the handling of the keys etc. is simplified

sormas-backend/src/main/java/de/symeda/sormas/backend/sormas2sormas/Sormas2SormasEncryptionService.java
vlad-ciucescu: I talked with Levi and he said that for the integration he would prefer Strings

sormas-base/setup/sormas.properties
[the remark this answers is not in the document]
vlad-ciucescu: I will add them in this case

sormas-backend/src/main/java/de/symeda/sormas/backend/sormas2sormas/Sormas2SormasEncryptionService.java
HolgerReiseVSys: Then we have to rely on e.g. Base64. I'm just not sure if this is the right layer for that.
Also, the service should propably better accept the data as byte array, so that the encoding is left to the caller. This also allows for supporting formats that are not string-based.

vlad-ciucescu: It makes sense, I replaced String with byte[]

sormas-base/setup/sormas.properties
```diff
@@ -62,6 +62,15 @@ app.url=
 # default: "/opt/sormas/sormas2sormas/"
 #sormas2sormas.path=/opt/sormas/sormas2sormas/
 
+# Key data for the generated SORMAS to SORMAS certificate";
+#sormas2sormas.keyAlias=
+#sormas2sormas.keystoreName=
+#sormas2sormas.keystorePass=
+
+# SORMAS to SORMAS truststore data";
+#sormas2sormas.truststoreName=${TRUSTSTORE_FILE_NAME}";
```
HolgerReiseVSys, line 71: I guess the value should be removed? The training `"}` is also present in the two headers above.

HolgerReiseVSys (approved):

Issue #2653, opened 2020-08-14
Add data of follow-up visits to case export [0.5]


MartinWahnschaffeSymeda:
### Situation Description
#1886 has added follow-up visits to cases.

### Feature Description
We want to have the same information that is included in a contacts export also for the case export:

- number of visits
- last cooperative visit was symptomatic
- date of last cooperative visit
- symptoms of last cooperative visit

This has already been implemented, but is commented out, since pseudonymization is not implemented for it yet.

bernardsilenou:
@MartinWahnschaffeSymeda  In terms of "number of visits", do we also count visits when case is converted from contact or just visits while a case, It will be good to have 2 types: all visits and visits while the person is a case

bernardsilenou:
Since many visits is often done in one day and the expected number of visit is 1 per day, In order not to have false statistics, we should add a column of: all visits and another for sum of visits per day with only the one visit selected per day.

markusmann-vg:
- [ ] uncomment the code
- [ ] check the code for sensitive data (comment and staff)

PR #2810, opened 2020-09-04
#2653 added data of follow-up visits to case export

leventegal-she:


BarnaBartha (approved):
good that you moved those utility methods to the helper class

Issue #2743, opened 2020-08-28
Add phone number and email to entity creation forms (case, contact, event participants) on web only [1]


bernardsilenou:
### Proposed Change
- Add on forms for: case, contact, event participant
- Only on web app
![image](https://user-images.githubusercontent.com/12996724/91593938-58415e80-e961-11ea-8839-9b15b0e53269.png)

john1496:
Helllo, I want to find out if the fields are going to be compulsory

bernardsilenou:
@john1496 Not compulsory please, Thanks

john1496:
@bernardsilenou Thanks, got it.

AndyBakcsy-she:
Should these fields also have verifications in the backend?
From a QA point of view: the Phone Number Filed should only accept phone numbers and the Email field should only accept emails.

MateStrysewske:
Valid point. I don't think it's a priority right now (given the huge list of priorities), but definitely something we should do in the future. We even have some kind of phone number validation already in the user form if I'm not mistaken.

AndyBakcsy-she:
Linking new feature request for the validation of these fields: https://github.com/hzi-braunschweig/SORMAS-Project/issues/2910

bernardsilenou:
Test result:
Feature works fine for cases and contact creation but it is missing for event participants

AndyBakcsy-she:
@bernardsilenou , isn't this the right place to add phone and email? in the Edit Event Participant field?
![image](https://user-images.githubusercontent.com/70562142/93765289-4cbe1c00-fc1d-11ea-8f7d-41b6d7e201ee.png)

AndyBakcsy-she:
Verified issue on SORMAS Version: 1.49.0-SNAPSHOT

bernardsilenou:
@AndyBakcsy-she I think for now, this is not a big deal. The initial need for this request was that the users wanted to add these information in the initial dialog box (create new person) when creating an event participant. For now, they have to save and go to next tab that takes time.

In addition, we will soon have different tabs for event participant entity and the event participant person entity as we introduce followup and pseudonimization to event participants. At this point it will be valuable to have it at the initial creation.

PR #2837, opened 2020-09-07
Added phone number to create forms #2743

john1496:


ChristopherRiedel:
Unfortunately you have misunderstood the task. The email and phone number are already there in the person. These should be able to be specified in the create forms analogous to the name. So i will close this pull request.

Issue #2724, opened 2020-08-27
Add Task Creation on contact creation [1]


markusmann-vg:
### Situation Description
Problem: For the cases, tasks are automatically created during import (one per case). For contacts, tasks are not created automatically.

### Feature Description
Requirement: For contact tracing, the automatic creation of tasks for the contacts is also necessary during import. Is there a possibility to execute the import in a way that one task is created per contact? Here it is even more important because more than 100 return travelers have to be imported as contacts every day and with the help of the tasks these can then be processed.

markusmann-vg:
Current situation:
=> contact follow-up tasks are automatically created at 00:00:01 and not directly after import.

"If Switzerland would like to have a contact investigation task, this would be needed to be implemented"

bernardsilenou:
I think what we need to do now is to:
- keep the contact followup task creation at 00:00:01
- Activate case/contact investigation task during importation. This should only happen for new entities, not when they are merged. 
- If report date of the entity is say >30 days in the past of the creation date, then we do not need to generate the automatic task.

MateStrysewske:
Actual task:
- [ ] Create a contact investigation task when a new contact is saved and its report date is within the last 30 days

AndyBakcsy-she:
Verified issue on SORMAS Version: 1.49.0-SNAPSHOT

PR #2887, opened 2020-09-15
#2724 Add task creation on contact creation

vlad-ciucescu:


sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -271,9 +276,24 @@ public ContactDto saveContact(ContactDto dto, boolean handleChanges) {
 		//		}
 
 		Contact entity = fromDto(dto);
-
 		contactService.ensurePersisted(entity);
 
+		if (configFacade.isConfiguredCountry(CountryHelper.COUNTRY_CODE_SWITZERLAND) && existingContact == null) {
```
MateStrysewske, line 281: This code should be moved to a dedicated method to keep the saveContact method as clean as possible; also, it shouldn't be restricted to Switzerland because it makes sense in general to create this task.

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -1274,22 +1259,26 @@ public void generateContactFollowUpTasks() {
...
-			if (contact.isHighPriority()) {
-				task.setPriority(TaskPriority.HIGH);
-			}
-
+			Task task = getContactTask(TaskType.CONTACT_FOLLOW_UP, fromDateTime, toDateTime, contact, assignee);
 			taskService.ensurePersisted(task);
 		}
 	}
 
+	private Task getContactTask(TaskType taskType, LocalDateTime fromDateTime, LocalDateTime toDateTime, Contact contact, User assignee) {
```
MateStrysewske, line 1267: Rename to createContactTask (because "get" suggests that the method returns an existing task)

sormas-backend/src/main/java/de/symeda/sormas/backend/common/FollowUpTaskCreationException.java
```diff
@@ -0,0 +1,9 @@
+package de.symeda.sormas.backend.common;
+
+@SuppressWarnings("serial")
+public class FollowUpTaskCreationException extends Exception {
```
MateStrysewske, line 4: I'd suggest to give this a more generic name (TaskCreationException) that we can then re-use in the future (since this is already not only used for follow-up tasks, but also for contact investigation tasks)

MateStrysewske (changes requested):

PR #2926, opened 2020-09-21
Add event title

### LINKED ISSUE #2904: Creation of an "event title" field in the form for each event
### Situation Description
To date, only the description field of the event makes it possible to identify an event. However, this is a free field of up to more than 4000 characters. It is important to have a short field (512 characters) which will be used in different places of the application to filter cases / contacts on an event, or to restrict the information in the dashboard to that of a given event.

### Feature Description
For each event sheet, add a "Event title" field allowing users to define a very brief name to identify an event.
This field is not mandatory, there is no initialization required for existing events.
Modification on the events directory:
- use this field as a column instead of the "description" field
- take this field into account in the filter “ID, description” → “ID, title, description”
- add this field in the detailed export

This field will be included in the case / contact directories to filter the cases / contacts who participated in an event (this will be detailed in another issue). It will also be used to create an HTML page with information about an event and associated actions that the user can print (this will be detailed in another issue). 

### Additional Information

@xavier-calland

xavier-calland:
Fixes #2904


MateStrysewske:
@xavier-calland Looks good. Please resolve the merge conflicts and I'll merge it!

MateStrysewske (approved):

PR #2974, opened 2020-09-24
Add event filter in case directory

### LINKED ISSUE #2908: Have a filter in the case directory to target cases linked to an event
### Situation Description
The case directory can be filtered by region, department, identifier, etc. Conversely, no filter can be used to filter on the possible involvement in a given event.

### Feature Description
In the directory of cases, it is proposed to
- Create two columns: event ID and event description (first 20 characters) (or "event title" when #2904 will be done)
- add a filter allowing to filter the cases according to the possible implication to a given event (via ID event or keyword of the description field) It is therefore a text field (like the filter "ID, name, EPID number, institution, person's telephone / postal code / city ”)

The filter works as follows: For a case to be identified as linked to an event, the case (its ID) must be present in "EventParticipant.resultingCase" of the event in question.

In the following example, a filter on Event 1 would display Case 1, not Case 2.
image

The following columns are to be added in the detailed view, the detailed export ("event" export group):

- title of the event (if field added)
- event code
- event status

### Additional Information

xavier-calland:
Fixes #2908

Draft because relying on PR #2926


xavier-calland:
@MateStrysewske any remarks are welcome, especially for: 
- Refactor caseService.createCriteriaFilter add parameter joins
- Add EventParticipant in CaseJoins
- Adds distinct when retrieving CaseIndexDto

MateStrysewske (changes requested):
This generally looks very good to me. No complaints so far, but I haven't tested it yet (because of the dependencies to the other pull requests). I'll do that as soon as the Event Title pull request is merged.

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseCriteria.java
```diff
@@ -92,6 +94,8 @@
 	private Boolean withReducedQuarantine;
 	private Boolean deleted = Boolean.FALSE;
 	private String nameUuidEpidNumberLike;
+	private String eventLike;
+	private Boolean eventAny = Boolean.FALSE;
```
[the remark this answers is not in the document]
xavier-calland: Done

xavier-calland:
> This generally looks very good to me. No complaints so far, but I haven't tested it yet (because of the dependencies to the other pull requests). I'll do that as soon as the Event Title pull request is merged.

@MateStrysewske did you can test it?

MateStrysewske:
@xavier-calland Will do so today. I have to take care of some potential problems with the automatic app update first.

xavier-calland:
@MateStrysewske Thanks

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseIndexDto.java
```diff
@@ -98,6 +98,8 @@
 	private FollowUpStatus followUpStatus;
 	private Date followUpUntil;
 	private Integer visitCount;
+	private Date changeDate;
```
MateStrysewske, line 101: I couldn't find any section in the code where these two new fields are used; are they actually needed?

MateStrysewske (changes requested):
- [ ] Move the "Only cases with events" checkbox to the filter row that contains the other checkboxes
- [ ] I18n for the new "Event" export group is missing

Also, I'm thinking about whether it would be better to include the event text filters into the already existing search field; having two separate fields may be a bit confusing (and also takes up more space). What's your opinion on that matter?

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseIndexDto.java
xavier-calland: The property is not used.
We need the parameter in the constructor because the field is used in the select part of the query (field has to be in the select part of the query to allow sorting on it, as it is a distinct select).
I can remove the property and keep the parameter with an "unused" annotation.

xavier-calland:
>     * [ ]  Move the "Only cases with events" checkbox to the filter row that contains the other checkboxes
> 
>     * [ ]  I18n for the new "Event" export group is missing
> 
> 
> Also, I'm thinking about whether it would be better to include the event text filters into the already existing search field; having two separate fields may be a bit confusing (and also takes up more space). What's your opinion on that matter?

I think having separate fields to search on different data (event, case) is a good approach.

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseIndexDto.java
MateStrysewske: Ah, I see, missed that. Alright!

MateStrysewske:
@xavier-calland Alright, I'm fine with that.

MateStrysewske (approved):

PR #3019, opened 2020-09-29
Add action view in the events directory

### LINKED ISSUE #2907: Addition of an event actions directory
### Situation Description
To date, SORMAS does not offer a synthetic view of the events and the actions taken as part of their investigation.

### Feature Description

**Development of a detailed view for monitoring all the actions of the events.**

Like what exists for cases (default view / detailed view), the proposed development concerns the events directory which would be supplemented by a synthetic view of all the actions associated with the different events. 

This view will allow users to access all events and their associated actions:
- Event identifier 
- Title of the event (cf issue #2904)
- date of signal 
- event status: cluster, signal, event, dropped 
- Action: "action title" field ( cf issue #2905 )
- Action: date of creation of the action and date of last modification (min/max)
- Action: status of the action 

**Filters to be provided in the detailed view "Actions"**

Like all the other tabs, some "Filters" help users to select information based on different criteria:
- Region (of the event)
- Department (of the event)
- Municipality (of the event)
- Identifier or title of the event
- Signal date (min / max) (of the event)
- Status of the event
- Status of the investigation
- Status of the action
- Date of modification of the action

**Features to be provided in the detailed and default view "Actions"**

For reporting purposes, from the events directory (default view and detailed view), provide the function "basic export" allowing to export the only information displayed after selection of filters

### Additional Information

xavier-calland:
Closes #2907

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/action/ActionFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/action/ActionFacade.java
@@ -22,6 +22,10 @@ import java.util.List;
 
 import javax.ejb.Remote;
 
+import de.symeda.sormas.api.event.EventActionIndexDto;
+import de.symeda.sormas.api.event.EventCriteria;
+import de.symeda.sormas.api.utils.SortProperty;
+
 @Remote
 public interface ActionFacade {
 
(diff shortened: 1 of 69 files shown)


MateStrysewske (changes requested):
* The default event view now has two date filters; I guess the filter to filter by action date isn't really useful there, so I suggest to remove it

Also, is there any reason to remove the event status filter from the action list?

xavier-calland:
> The default event view now has two date filters; I guess the filter to filter by action date isn't really useful there, so I suggest to remove it
> Also, is there any reason to remove the event status filter from the action list?

Damn, I renamed `showEventStatusFilter` and `showActionFilters` parameters to `hideEventStatusFilter` and `hideActionFilters` to have simpler tests in `EventsFilterForm`.
I forgot to change the values when calling the constructor.

MateStrysewske (approved):

Issue #2959, opened 2020-09-23
Cases > Minimal Essential Data (MED) for Switzerland [3]


Candice-Louw:
### Situation Description
In order to fulfill the Minimal Essential Data (MED) requirement for the BAG in Switzerland, additional fields and variables need to be introduced to SORMAS: https://github.com/hzi-braunschweig/SORMAS-Switzerland/issues/102

This is part 1.  Part 2 is for CONTACTS: https://github.com/hzi-braunschweig/SORMAS-Project/issues/2960

### Feature Description

The following is required for CASES:

NOTE:
In the Swiss context-
Cases go into Isolation
Contacts go into Quarantine

---

1. 
Cases > Open Case > Case Information:

New field (CH only) -
"Case-ID ISM" (en) | "Fall-ID ISM" (de-CH)
Next to or under the existing "External ID" field.
Data Type: Integer

---

2.  
Cases > Open Case > Case Information:

New field (CH only) -
"Reason for COVID-19-Test" (en) | "Grund für COVID-19-Test" (de-CH)
Data Type: Dropdown/enum
Values (en | de-CH):
- Presence of symptoms | COVID-19-Test durchgeführt aufgrund von Symptomen.
- Screening during outbreak investigation | COVID-19-Reihentest im Rahmen einer Ausbruchsuntersuchung.
- Cohort Screening | COVID-19-Test durchgeführt aufgrund eines COVID-19-Kohorten-Screenings.
- Requirement of employer | COVID-19-Test durchgeführt auf Wunsch des Arbeitgebers.
- During quarantine, as requested by the Canton | COVID-19-Test durchgeführt während der Quarantäne auf Wunsch des Kantons.
- After first contact with Contact Tracing (already symptomatic) | COVID-19-Tests durchgeführt nach Erstkontakt mit klassischem Contact Tracing (bereits symptomatisch).
- Due to SwissCovidApp notification | COVID-19-Tests durchgeführt aufgrund einer SwissCovidApp-Meldung.
- Other reason | Anderer Grund für den COVID-19-Test

If "Other reason | Anderer Grund für den COVID-19-Test" > Display free-text field.

---

3.  
Cases > Open Case > Case Information:

(All CH only)

New section/heading -
"First contact with Contact Tracing" (en) | "Erstkontakt mit kantonaler Contact-Tracing-Stelle" (de-CH)

New field -
"Type of contact" (en) | "Art des Kontaktes" (de-CH)
[...]

PR #3023, opened 2020-09-30
#2959 added case fields for Switzerland

leventegal-she:


sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
```diff
@@ -390,6 +403,28 @@
 	private Date followUpUntil;
 	private boolean overwriteFollowUpUntil;
 
+	@HideForCountriesExcept(countries = "ch")
```
vidi42, line 406: Just a suggestion, but you could use `CountryHelper.COUNTRY_CODE_SWITZERLAND` for easier tracing later if needed instead of the "hardcoded" country code.

vidi42 (approved):

PR #3040, opened 2020-10-01
Feature 2855 extend campaign form

BarnaBartha:


sormas-api/src/main/resources/strings.properties
```diff
@@ -515,6 +517,7 @@ infoContactsViewRegionDistrictFilter = When you select a region and/or district
 infoDeveloperOptions = You can use the controls below to generate dummy cases and contacts based on the selected restraints. Please note that generating a lot of data at once might take some time.
 infoCreateNewSampleDiscardsChanges = Creating a new sample will discard all unsaved changes made to this case
 infoCreateNewSampleDiscardsChangesEventParticipant = Creating a new sample will discard all unsaved changes made to this event participant
+infoUsageOfEditableGrid = Editable grids allow users to change their content and reorder them by dragging & dropping
```
MateStrysewske, line 520: I'd like this to be more directed towards the user, something like "You can edit the campaign data and dashboard definitions by clicking inside one of the cells in the grid, and you can reorder the dashboard elements by dragging and dropping the grid rows."

MateStrysewske (changes requested):
* Order seems to be a bit messed up on initial load; I have six diagrams in my definition and the orders displayed are "1, 1, 2, 2, 3, 4". If I start to drag them around, the order is properly changed to "0, 1, 2, 3, 4, 5".
* There should be a confirmation popup when removing diagrams from the grid
* Add some spacing above the "Discard" and "Save" buttons as well as above and below the hint

Issue #2894, opened 2020-09-16
Archiving of many cases at once (3000+) does not work


bernardsilenou:
### Bug Description
- Under 1000 cases, archiving time is short and all works fine. For about 50000 cases, system takes works for some time and return error message.
### Steps to Reproduce
1.
2.

### Expected Behavior
- Archiving should be possible in max 3 min for 100000 cases.
- Maybe we archive cases immediately and latter archive the associated entities like contacts at back-end so users do  not need to wait for log
### System Details
* Device:
* SORMAS version:1,48
* Android version/Browser:

bernardsilenou:
This belong to the same improvement story with #2889

StefanKock:
The cause for the problem: The archiving by selection in the UI calls _CaseFacade.archiveOrDearchiveCase(String caseUuid, boolean archive)_ for every case in the same JTA transaction. Internally only _Case.archived:boolean_ is changed but by loading the JPA entity. This results for too much entities in a blown up transaction with entities in the hibernate cache.

Possible solution: Reuse the approach of _CaseFacadeEjb.archiveAllArchivableCases_ by enable to build another _where_ clause by _caseUuids_ (maybe batched and not 1000+ at once and beware of parameter limit, see #2193). But the implicit change would be that the _Case.changeDate_ will not by updated anymore.

MateStrysewske:
Adding a _CaseFacade.archiveOrDearchiveCases(List<String> caseUuids, boolean archive)_ method that does this with a where clause should still update the changeDates while getting rid of the individual transactions, shouldn't it?

StefanKock:
Clarified with @MateStrysewske what to do:
- [ ] Move the CriteriaUpdate-Logic of _CaseFacadeEjb.archiveAllArchivableCases_ to _CaseService_ to enable JTA Tx separation per batch.
- [ ] Enhance this new method: set _Case.changeDate = now()_
- [ ] Introduce a layer before this method to separate the uuids into batches with batch size less than parameter limit.
- [ ] Change the archiving feature of the UI to sent all Case uuids at once.
- [ ] Change the single archiving method also to new pattern.

StefanKock:
I did the work but was not able to test it in the sormas-ui since I did not find how to archive cases.
Also I started on start after 1.49.0 since I had problems with development HEAD.

StefanKock:
Newly considered improvements:

1. Log execution times to debug the performance
[...]

PR #3061, opened 2020-10-05
Feature 2894 case archiving with batches

StefanKock:
Batching, JTA Transactions and logging is implemented. Performance is documented in #2894.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/utils/DateHelper.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/utils/DateHelper.java
@@ -1004,4 +1004,48 @@ public final class DateHelper {
 		timestamp.setNanos(timestamp.getNanos() + 999999);
 		return timestamp;
 	}
+
+	/**
+	 * Anzahl Millisekunden pro Sekunde.
+	 */
+	private static final int MILLIES_PER_SECOND = 1000;
+
+	/**
+	 * @return Current time in milliseconds.
+	 */
+	public static long now() {
+
+		return System.currentTimeMillis();
+	}
+
+	/**
+	 * Redundant to {@link #now()} to propose {@code startTime} as variable name.
+	 * 
+	 * @return Current time in milliseconds.
+	 */
+	public static long startTime() {
+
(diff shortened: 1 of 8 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/utils/DateHelper.java
```diff
@@ -1004,4 +1004,48 @@ public static Timestamp toTimestampUpper(Date date) {
 		timestamp.setNanos(timestamp.getNanos() + 999999);
 		return timestamp;
 	}
+
+	/**
+	 * Anzahl Millisekunden pro Sekunde.
```
MateStrysewske, line 1009: Translate to English; Also I'm not sure whether "Millies" is a common abbreviation in English, I'd prefer to use "Milliseconds" everywhere

sormas-backend/src/main/java/de/symeda/sormas/backend/util/ModelConstants.java
```diff
@@ -23,4 +23,9 @@
 	String PERSISTENCE_UNIT_NAME_AUDITLOG = "auditlogPU";
 	String HINT_HIBERNATE_READ_ONLY = "org.hibernate.readOnly";
 
+	/**
+	 * Hard limit how much parameter can be allowed in a query.<br />
```
MateStrysewske, line 27: how much parameter => how many parameters
to big => too big
SQL-Statement => SQL statement

MateStrysewske (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/utils/DateHelper.java
StefanKock: fixed

sormas-backend/src/main/java/de/symeda/sormas/backend/util/ModelConstants.java
StefanKock: fixed

Issue #2111, opened 2020-05-15
User Authentication with Keycloak (create users, login- logout users, forgot password) [5+5]


markusmann-vg:
### Situation Description
France: 2f Authentication
Current situation of password generation & transmitting password
Change & Forgot password methods
Password Policy
Keycloak is required (open source) 

### Feature Description
Deliverable:
New issues that descibe the next steps
- one ticket needs to contain a Migration Guide

APP & Web

BarnaBartha:
Within the timeframe I setup a local keycloak and tried integrating the payara openid connect support into sormas-ui, unfortunately I've ran into some issues:
- authorization issue:
[2020-06-04T10:02:28.793+0300] [Payara 5.192] [WARNING] [] [fish.payara.security.openid.OpenIdAuthenticationMechanism] [tid: _ThreadID=46 _ThreadName=http-thread-pool::http-listener-1(5)] [timeMillis: 1591254148793] [levelValue: 900] [[
  Error occurred in validating Authorization Code : unknown_error caused by Unknown]]
- infinte redirect redirect between keycloak authentication and sormas-ui.

Documentation and example I've used for the changed that can be found in the branch:
- https://docs.payara.fish/docs/5.201/documentation/payara-server/public-api/openid-connect-support.html
- https://github.com/bcgov/moh-iam/wiki/How-to-secure-a-Jakarta-EE-application-deployed-to-Payara-with-Keycloak
- https://github.com/bcgov/moh-iam/tree/master/keycloak-javaee-example

For installing local keycloak I've used:
- https://www.keycloak.org/getting-started/getting-started-docker
The guide is quite ok but the docker image referenced does not work I've used jboss/keycloak (https://hub.docker.com/r/jboss/keycloak/)

Another example worth checking is:
- https://blog.payara.fish/openid-connect-in-the-payara-platform-5.183
- https://github.com/javaee-samples/vendoree-samples/tree/master/payara/openid

@markusmann-vg 
At this point I think we need more time to continue with investigation or ideally somebody with in depth knowledge of keycloak and oauth2/openid connect to debug and continue with this.

BarnaBartha:
@HolgerReiseVSys @MartinWahnschaffeSymeda please feel free to update these steps
Steps to do: 
- Have a correctly configured Keycloak
- Integrate app-auth into sormas-app
- Integrate a Keycloak client with sormas-ui (possible starting point code from existing branch - feature-2111-keycloak)

[...]

PR #3074, opened 2020-10-06
#2111 Add explicit Payara maven repository

vidi42:


vidi42:
Payara dependencies will be downloaded from https://github.com/payara/Payara_PatchedProjects
```bash
Downloading from payara-patchedprojects: https://raw.github.com/payara/Payara_PatchedProjects/master/org/glassfish/soteria/javax.security.enterprise/1.1-b01.payara-p3/javax.security.enterprise-1.1-b01.payara-p3.jar
Downloaded from payara-patchedprojects: https://raw.github.com/payara/Payara_PatchedProjects/master/org/glassfish/soteria/javax.security.enterprise/1.1-b01.payara-p3/javax.security.enterprise-1.1-b01.payara-p3.jar (130 kB at 153 kB/s)
```

StefanKock:
I am not sure if we need this:

1. It is redundant to what is provided in _security-connector-oidc-client_. If it changes we will not notice and cause a hard problem then.
2. It does not solve the problem when you have a catch all mirror defined in your local Maven setup.

vidi42:
> I am not sure if we need this:
> 
> 1. It is redundant to what is provided in _security-connector-oidc-client_. If it changes we will not notice and cause a hard problem then.
> 2. It does not solve the problem when you have a catch all mirror defined in your local Maven setup.

The idea was to make it visible that this repository is also needed. However you are right it doesn't solve the issue. Maybe we can include a entry in the setup documentation if we see that more people are having this issue.

Issue #2989, opened 2020-09-25
Change "home description" to optional, when home is selected [2]


markusmann-vg:
### Feature Description
Currently the home description is mandatory.
With this change we change it to optional. 
There are many locations, where we have to do a NULL-check and check if there is another impact.

AndyBakcsy-she:
Verified issue on SORMAS Version: 1.50.0-SNAPSHOT

PR #3081, opened 2020-10-06
#2989 - Change "home description" to optional, when home is selected

sergiupacurariu:


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
```diff
@@ -1193,7 +1193,9 @@ private void updateFacilityDetails(ComboBox cbFacility, TextField tfFacilityDeta
 			boolean visibleAndRequired = otherHealthFacility || noneHealthFacility;
 
 			tfFacilityDetails.setVisible(visibleAndRequired);
-			tfFacilityDetails.setRequired(visibleAndRequired);
+			if (otherHealthFacility) {
+				tfFacilityDetails.setRequired(visibleAndRequired);
```
BarnaBartha, line 1197: this is a bit confusing, the variable name is a bad one as it represents two things and is already defined by the boolean in the if statement.. please consider cleaning this up

sergiupacurariu: updated, removed if clause and linked setRequired to "otherHealthFacility"

Issue #2545, opened 2020-07-24
Add a global symptomatic variable to a case 


bernardsilenou:
### Situation Description
Currently, asymptomatic cases can only be inferred from all symptoms == NO. Users will want to explicitly register that a case is asymptomatic in one click.

### Feature Description
The symptomatic variable can have categories: Symptomatic & asymptomatic
-when asymptomatic == T, then all symptom ==NO,
-when symptomatic == T, then all symptom are presented on form for user to choose,

### Possible Alternatives
- Add a check box button beside "set all clear to NO" and if set to T, the symptoms == NO and if set to yes, then symptoms == NA and user then have to choose option for each symptom.

MateStrysewskeSym:
@bernardsilenou I don't really understand why they can't just use the existing buttons for this purpose. Do you need this so that they can see at a glance whether a case is asymptomatic without having to look at the list of symptoms?

carolinverset:
@bernardsilenou @benchaabaner @Emond-Fuseau @OlivierLeuci 

I agree with you : For asymptomatic cases, we could click on no for all the symptoms.
But for symptomatic cases, in France, we do not know the symptom from which the case is suffering. We just need to know whether the case has symptoms or not.

Another advantage is that we will thus have a synthetic variable which would make it possible to add a filter in the case directory to visualize and export only symptomatic or asymptomatic patients.

Best regards
Carolin

MateStrysewskeSym:
@carolinverset There's already a hidden "symptomatic" variable that is automatically updated based on whether any symptom has been set to "yes". I still don't really understand the need for making it available for the user though. Adding such a filter would not be a problem.

bernardsilenou:
@MateStrysewskeSym I think it goes both ways, we still need to show this hidden variable for users to be able to choose yes/no right?
-If a case is symptomatic and you do not know the symptoms, then this may affect the relationship between the hidden symptomatic variable and the symptoms because symptomatic can be "yes" and all symptoms are NA. Maybe we need to add a new symptomatic satiable that does not depend on the symptoms?
-The benefit i see with this variable is in a situation that we know a case is symptomatic but we do not know the exact symptoms. In the current setting, we can't record this information.

MateStrysewskeSym:
@bernardsilenou How exactly can a situation like this even happen? If you know for a fact that a case is symptomatic, you'd also need to know about at least one symptom, right? How else would you verify that the person is indeed symptomatic?

[...]

PR #3093, opened 2020-10-07
#2545 symptomatic

mouloud-mala:
#2545

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/caze/GermanCaseClassificationValidator.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/caze/GermanCaseClassificationValidator.java
@@ -18,6 +18,7 @@ package de.symeda.sormas.api.caze;
 import de.symeda.sormas.api.Disease;
 import de.symeda.sormas.api.symptoms.SymptomsDto;
 import de.symeda.sormas.api.symptoms.SymptomsHelper;
+import de.symeda.sormas.api.utils.YesNoUnknown;
 
 public class GermanCaseClassificationValidator {
 
@@ -66,7 +67,7 @@ public class GermanCaseClassificationValidator {
 	private static boolean hasCoronavirusSymptom(CaseDataDto caseDataDto) {
 		final SymptomsDto symptoms = caseDataDto.getSymptoms();
 		return caseDataDto.getDisease() == Disease.CORONAVIRUS
-			&& symptoms.getSymptomatic()
+			&& symptoms.getSymptomatic() == YesNoUnknown.YES
 			&& SymptomsHelper.atLeastOnSymptomTrue(
 				symptoms.getPneumoniaClinicalOrRadiologic(),
 				symptoms.getDifficultyBreathing(),
--- a/sormas-api/src/main/java/de/symeda/sormas/api/caze/GermanCaseClassificationValidator.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/caze/GermanCaseClassificationValidator.java
@@ -67,7 +67,7 @@ public class GermanCaseClassificationValidator {
 	private static boolean hasCoronavirusSymptom(CaseDataDto caseDataDto) {
 		final SymptomsDto symptoms = caseDataDto.getSymptoms();
 		return caseDataDto.getDisease() == Disease.CORONAVIRUS
-			&& symptoms.getSymptomatic() == YesNoUnknown.YES
+			&& YesNoUnknown.YES.equals(symptoms.getSymptomatic())
 			&& SymptomsHelper.atLeastOnSymptomTrue(
 				symptoms.getPneumoniaClinicalOrRadiologic(),
 				symptoms.getDifficultyBreathing(),
--- a/sormas-api/src/main/java/de/symeda/sormas/api/i18n/I18nProperties.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/i18n/I18nProperties.java
@@ -17,10 +17,6 @@
  *******************************************************************************/
 package de.symeda.sormas.api.i18n;
 
-import de.symeda.sormas.api.Language;
-import de.symeda.sormas.api.ResourceBundle;
-import org.apache.commons.lang3.StringUtils;
-
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
--- a/sormas-api/src/main/java/de/symeda/sormas/api/i18n/I18nProperties.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/i18n/I18nProperties.java
@@ -33,6 +33,11 @@ import java.util.Map;
 import java.util.PropertyResourceBundle;
 import java.util.ResourceBundle.Control;
 
+import org.apache.commons.lang3.StringUtils;
+
+import de.symeda.sormas.api.Language;
+import de.symeda.sormas.api.ResourceBundle;
+
 public final class I18nProperties {
 
 	public static final String FULL_COUNTRY_LOCALE_PATTERN = "[a-zA-Z]*-[a-zA-Z]*";
--- a/sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsHelper.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsHelper.java
@@ -28,7 +28,6 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 
-import de.symeda.sormas.api.utils.YesNoUnknown;
 import org.apache.commons.lang3.StringUtils;
 
 import de.symeda.sormas.api.EntityDto;
@@ -37,6 +36,7 @@ import de.symeda.sormas.api.i18n.I18nProperties;
 import de.symeda.sormas.api.i18n.Strings;
 import de.symeda.sormas.api.utils.DataHelper;
 import de.symeda.sormas.api.utils.DateHelper;
+import de.symeda.sormas.api.utils.YesNoUnknown;
 
 public final class SymptomsHelper {
 
@@ -200,8 +200,7 @@ public final class SymptomsHelper {
 			dto.setSymptomatic(YesNoUnknown.UNKNOWN);
 		} else if (dto.getSymptomatic() == YesNoUnknown.NO) {
 			dto.setSymptomatic(YesNoUnknown.NO);
-		}
-		else {
+		} else {
 			dto.setSymptomatic(null);
 		}
 	}
@@ -234,15 +233,11 @@ public final class SymptomsHelper {
 					if (result == SymptomState.YES) {
 						pd.getWriteMethod().invoke(targetSymptoms, result);
 					}
-				} else if (pd.getReadMethod().getReturnType() == YesNoUnknown.class) {
+				} else if (pd.getReadMethod().getReturnType() == Boolean.class) {
 					// Booleans are carried over when they are TRUE
-					if (pd.getName().equals(SymptomsDto.SYMPTOMATIC)) {
-						continue;
-					} else {
-						YesNoUnknown result = (YesNoUnknown) pd.getReadMethod().invoke(sourceSymptoms);
-						if (result == YesNoUnknown.YES) {
-							pd.getWriteMethod().invoke(targetSymptoms, result);
-						}
+					YesNoUnknown result = (YesNoUnknown) pd.getReadMethod().invoke(sourceSymptoms);
+					if (result == YesNoUnknown.YES) {
+						pd.getWriteMethod().invoke(targetSymptoms, result);
 					}
 				} else if (pd.getName().equals(SymptomsDto.TEMPERATURE)) {
 					// Temperature is carried over when it's higher than the targetSymptoms temperature
--- a/sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsHelper.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsHelper.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 
+import de.symeda.sormas.api.utils.YesNoUnknown;
 import org.apache.commons.lang3.StringUtils;
 
 import de.symeda.sormas.api.EntityDto;
@@ -186,7 +187,7 @@ public final class SymptomsHelper {
 			for (Method method : SymptomsDto.class.getDeclaredMethods()) {
 				if (method.getReturnType() == SymptomState.class) {
 					if (method.invoke(dto) == SymptomState.YES) {
-						dto.setSymptomatic(true);
+						dto.setSymptomatic(YesNoUnknown.YES);
 						return;
 					}
 				}
@@ -195,7 +196,14 @@ public final class SymptomsHelper {
 			throw new RuntimeException(e);
 		}
 
-		dto.setSymptomatic(false);
+		if (dto.getSymptomatic() == YesNoUnknown.UNKNOWN) {
+			dto.setSymptomatic(YesNoUnknown.UNKNOWN);
+		} else if (dto.getSymptomatic() == YesNoUnknown.NO) {
+			dto.setSymptomatic(YesNoUnknown.NO);
+		}
+		else {
+			dto.setSymptomatic(null);
+		}
 	}
 
 	/**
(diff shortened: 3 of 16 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsHelper.java
```diff
@@ -195,7 +196,14 @@ public static void updateIsSymptomatic(SymptomsDto dto) {
 			throw new RuntimeException(e);
 		}
 
-		dto.setSymptomatic(false);
+		if (dto.getSymptomatic() == YesNoUnknown.UNKNOWN) {
```
MateStrysewske, line 199: This code block doesn't really make sense; all you're doing is setting the same value that was already set before. Should be removed.

sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsHelper.java
```diff
@@ -226,13 +234,13 @@ public static void updateSymptoms(SymptomsDto sourceSymptoms, SymptomsDto target
 					if (result == SymptomState.YES) {
 						pd.getWriteMethod().invoke(targetSymptoms, result);
 					}
-				} else if (pd.getReadMethod().getReturnType() == Boolean.class) {
+				} else if (pd.getReadMethod().getReturnType() == YesNoUnknown.class) {
```
MateStrysewske, line 237: You're breaking existing logic here. Please revert this change. All you have to do is to remove the special condition for SymptomsDto.SYMPTOMATIC.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -870,7 +870,11 @@ public long count(CaseCriteria caseCriteria) {
 						pseudonymizer.pseudonymizeDto(SymptomsDto.class, visitSymptoms, inJurisdiction, null);
 
 						exportDto.setLastCooperativeVisitSymptoms(SymptomsHelper.buildSymptomsHumanString(visitSymptoms, true, userLanguage));
-						exportDto.setLastCooperativeVisitSymptomatic(visitSymptoms.getSymptomatic() ? YesNoUnknown.YES : YesNoUnknown.NO);
+						if (visitSymptoms.getSymptomatic() == YesNoUnknown.YES)
```
MateStrysewske, line 873: Always use curly brackets please, even for one-line statements

sormas-ui/src/main/java/de/symeda/sormas/ui/symptoms/SymptomsForm.java
```diff
@@ -104,8 +105,9 @@
 					loc(SIGNS_AND_SYMPTOMS_HEADING_LOC) +
 					fluidRowCss(VSPACE_3,
 							//XXX #1620 fluidColumnLoc?
-							fluidColumn(8, 0, loc(SYMPTOMS_HINT_LOC))) +
+							fluidColumn(8, 0, loc(SYMPTOMS_HINT_LOC)) +
 					fluidRow(fluidColumn(6,6, locCss(CssStyles.ALIGN_RIGHT,BUTTONS_LOC)))+
+					fluidRowLocs(SYMPTOMATIC, ONSET_DATE) +
```
MateStrysewske, line 110: Onset date and onset symptom shouldn't be split because they're related to each other. Please move onset symptom to the top as well.

sormas-ui/src/main/java/de/symeda/sormas/ui/visit/VisitGrid.java
```diff
@@ -69,7 +70,7 @@ public VisitGrid(VisitCriteria criteria) {
 
 		((Column<VisitIndexDto, Date>) getColumn(VisitIndexDto.VISIT_DATE_TIME))
 			.setRenderer(new DateRenderer(DateHelper.getLocalDateTimeFormat(I18nProperties.getUserLanguage())));
-		((Column<VisitIndexDto, String>) getColumn(VisitIndexDto.SYMPTOMATIC)).setRenderer(new BooleanRenderer());
+		((Column<VisitIndexDto, String>) getColumn(VisitIndexDto.SYMPTOMATIC)).setRenderer(new TextRenderer());
```
MateStrysewske, line 73: You shouldn't need a renderer for this column because enums are automatically rendered according to their toString method.

sormas-ui/src/main/java/de/symeda/sormas/ui/symptoms/SymptomsForm.java
```diff
@@ -488,6 +493,8 @@ protected void addFields() {
 
 		// Initialize lists
 
+		symptomaticFieldId = SYMPTOMATIC;
```
MateStrysewske, line 496: This looks like a pretty redundant variable for me; please remove it and explicitely use SYMPTOMATIC instead

sormas-ui/src/main/java/de/symeda/sormas/ui/symptoms/SymptomsForm.java
```diff
@@ -990,6 +1003,36 @@ private void addListenerForOnsetFields(ComboBox onsetSymptom, DateField onsetDat
 		onsetDateField.setEnabled(false); // will be updated by listener if needed
 	}
 
+	// event for the symptomatic field
+	private void addListenerForOnsetSymptomaticField(DateField onsetDateField) {
```
MateStrysewske, line 1007: => addListenerForSymptomaticField (the word "onset" isn't appropriate here)

sormas-ui/src/main/java/de/symeda/sormas/ui/symptoms/SymptomsForm.java
```diff
@@ -990,6 +1003,36 @@ private void addListenerForOnsetFields(ComboBox onsetSymptom, DateField onsetDat
...
+		Field<YesNoUnknown> sourceFieldSymptomatic = (Field<YesNoUnknown>) getFieldGroup().getField(symptomaticFieldId);
+		List<String> allPropertyIds =
+				Stream.concat(unconditionalSymptomFieldIds.stream(), conditionalBleedingSymptomFieldIds.stream()).collect(Collectors.toList());
+		allPropertyIds.add(LESIONS_THAT_ITCH);
+
+		sourceFieldSymptomatic.addValueChangeListener(event -> {
+			for (Object sourcePropertyId : allPropertyIds) {
```
MateStrysewske, line 1014: The visibility status of the onset date field should only be changed once here, not for every symptom that is processed. Additionally, the onset symptom field should also be taken care of with the same logic as the onset date.

It's probably also a good idea to clear the onset date and onset symptom fields when NO or UNKNOWN are selected because if there's been a value entered before and the field is disabled, users can't remove the value anymore.

MateStrysewske (changes requested):
The whole mobile app part is missing. We can't accept this pull request without the app being taken care of because it will stop working otherwise.

Also, please apply code formatting as described in the development enviroment guide to all classes you're working on.

sormas-ui/src/main/java/de/symeda/sormas/ui/symptoms/SymptomsForm.java
mouloud-mala: I also add a clear on onset date when we click on clearAllButton

sormas-api/src/main/java/de/symeda/sormas/api/caze/GermanCaseClassificationValidator.java
```diff
@@ -66,7 +67,7 @@ public static boolean isValidGermanCaseClassification(
 	private static boolean hasCoronavirusSymptom(CaseDataDto caseDataDto) {
 		final SymptomsDto symptoms = caseDataDto.getSymptoms();
 		return caseDataDto.getDisease() == Disease.CORONAVIRUS
-			&& symptoms.getSymptomatic()
+			&& symptoms.getSymptomatic() == YesNoUnknown.YES
```
MateStrysewske, line 70: Will throw a NullPointerException if "symptomatic" is null. Instead use:
`YesNoUnknown.YES.equals(symptoms.getSymptomatic())`

sormas-api/src/main/java/de/symeda/sormas/api/i18n/I18nProperties.java
```diff
@@ -33,6 +33,11 @@
 import java.util.PropertyResourceBundle;
 import java.util.ResourceBundle.Control;
 
+import org.apache.commons.lang3.StringUtils;
+
+import de.symeda.sormas.api.Language;
```
MateStrysewske, line 38: Remove unused imports

sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsHelper.java
```diff
@@ -228,13 +235,9 @@ public static void updateSymptoms(SymptomsDto sourceSymptoms, SymptomsDto target
...
 				} else if (pd.getReadMethod().getReturnType() == Boolean.class) {
 					// Booleans are carried over when they are TRUE
-					if (pd.getName().equals(SymptomsDto.SYMPTOMATIC)) {
-						continue;
-					} else {
-						Boolean result = (Boolean) pd.getReadMethod().invoke(sourceSymptoms);
-						if (Boolean.TRUE.equals(result)) {
-							pd.getWriteMethod().invoke(targetSymptoms, result);
-						}
+					YesNoUnknown result = (YesNoUnknown) pd.getReadMethod().invoke(sourceSymptoms);
```
MateStrysewske, line 238: Revert this change. This doesn't make sense, you're asking whether the return type is Boolean and then process the field as if it were a YesNoUnknown field. Please just restore the former code, get rid of the if condition and only use the code in the else block.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -870,7 +870,13 @@ public long count(CaseCriteria caseCriteria) {
 						pseudonymizer.pseudonymizeDto(SymptomsDto.class, visitSymptoms, inJurisdiction, null);
 
 						exportDto.setLastCooperativeVisitSymptoms(SymptomsHelper.buildSymptomsHumanString(visitSymptoms, true, userLanguage));
-						exportDto.setLastCooperativeVisitSymptomatic(visitSymptoms.getSymptomatic() ? YesNoUnknown.YES : YesNoUnknown.NO);
+						if (visitSymptoms.getSymptomatic() == YesNoUnknown.YES) {
```
MateStrysewske, line 873: Why not just use `exportDto.setLastCooperativeVisitSymptomatic(visitSymptoms.getSymptomatic())`?

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1826,7 +1832,7 @@ private float calculateCompleteness(Case caze) {
 		if (sampleService.getSampleCountByCase(caze) > 0) {
 			completeness += 0.15f;
 		}
-		if (Boolean.TRUE.equals(caze.getSymptoms().getSymptomatic())) {
+		if ((caze.getSymptoms().getSymptomatic()) == YesNoUnknown.YES) {
```
MateStrysewske, line 1835: Will throw a NullPointerException when symptomatic is null, see above

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -537,7 +537,11 @@ public void deleteContact(String contactUuid) {
 
 					exportContact.setLastCooperativeVisitDate(lastCooperativeVisit.getVisitDateTime());
 					exportContact.setLastCooperativeVisitSymptoms(SymptomsHelper.buildSymptomsHumanString(symptoms, true, userLanguage));
-					exportContact.setLastCooperativeVisitSymptomatic(symptoms.getSymptomatic() ? YesNoUnknown.YES : YesNoUnknown.NO);
+					if (lastCooperativeVisit.getSymptoms().getSymptomatic() == YesNoUnknown.YES) exportContact
```
MateStrysewske, line 540: See above

sormas-ui/src/main/java/de/symeda/sormas/ui/symptoms/SymptomsForm.java
```diff
@@ -994,6 +1002,43 @@ private void addListenerForOnsetFields(ComboBox onsetSymptom, DateField onsetDat
...
+						symptom.setValue(SymptomState.NO);
+						symptom.setEnabled(false);
+					}
+				} else {
+					symptom.setEnabled(true);
+				}
+			}
+			if (sourceFieldSymptomatic.getValue() != null) {
+				if (sourceFieldSymptomatic.getValue().equals(YesNoUnknown.NO) || sourceFieldSymptomatic.getValue().equals(YesNoUnknown.UNKNOWN)) {
+					onsetDateField.setVisible(false);
```
MateStrysewske, line 1028: I'd strongly recommend to use setEnabled(false) instead of setVisible(false) because otherwise users will be confused why the fields are suddenly missing

sormas-ui/src/main/java/de/symeda/sormas/ui/symptoms/SymptomsForm.java
```diff
@@ -994,6 +1002,43 @@ private void addListenerForOnsetFields(ComboBox onsetSymptom, DateField onsetDat
...
+						symptom.setEnabled(true);
+					} else {
+						symptom.setValue(SymptomState.NO);
+						symptom.setEnabled(false);
+					}
+				} else {
+					symptom.setEnabled(true);
+				}
+			}
+			if (sourceFieldSymptomatic.getValue() != null) {
```
MateStrysewske, line 1026: You can combine these two if statements because both else blocks use identical code

sormas-ui/src/main/java/de/symeda/sormas/ui/visit/VisitGrid.java
```diff
@@ -69,7 +68,7 @@ public VisitGrid(VisitCriteria criteria) {
 
 		((Column<VisitIndexDto, Date>) getColumn(VisitIndexDto.VISIT_DATE_TIME))
 			.setRenderer(new DateRenderer(DateHelper.getLocalDateTimeFormat(I18nProperties.getUserLanguage())));
-		((Column<VisitIndexDto, String>) getColumn(VisitIndexDto.SYMPTOMATIC)).setRenderer(new BooleanRenderer());
+		getColumn(VisitIndexDto.SYMPTOMATIC);
```
MateStrysewske, line 71: Pointless line of code

sormas-api/src/main/java/de/symeda/sormas/api/symptoms/SymptomsHelper.java
```diff
@@ -195,7 +196,13 @@ public static void updateIsSymptomatic(SymptomsDto dto) {
 			throw new RuntimeException(e);
 		}
 
-		dto.setSymptomatic(false);
+		if (dto.getSymptomatic() == YesNoUnknown.UNKNOWN) {
```
MateStrysewske, line 199: * See above concerning possible NullPointerExceptions
* You can write this way more efficiently like this:
```
if (!YesNoUnknown.UNKNOWN.equals(dto.getSymptomatic()) && !YesNoUnknown.NO.equals(dto.getSymptomatic())) {
  dto.setSymptomatic(null);
}
```

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactsView.java
```diff
@@ -268,25 +270,24 @@ public ContactsView() {
...
-//							caption += " (" + DateFormatHelper.getDateFormatPattern() + ")";
-//						}
-//						return caption;
-//					},
-//					createFileNameWithCurrentDate("sormas_BAG_contacts_", ".csv"),
-//					null);
-//
-//				addExportButton(bagExportResource, exportButton, exportLayout, VaadinIcons.FILE_TEXT, Captions.BAGExport, Strings.infoBAGExport);
-//			}
+			if (FacadeProvider.getConfigFacade().isConfiguredCountry(CountryHelper.COUNTRY_CODE_SWITZERLAND)
```
MateStrysewske, line 273: Revert this change please, it's not associated to this issue

MateStrysewske (changes requested):
I just realised that there are no changes related to the mobile app here yet; this also needs to be taken care of or the mobile app will stop working. Unfortunately that also means that you'll have to re-create the whole symptoms table thanks to SQLite being a _wonderful_ technology...

MateStrysewske:
@mouloud-mala Actually, let's go with another approach: Please rename the current "symptomatic" field in the mobile app to something like "oldSymptomatic" and set it to `@Deprecated`. Create a new YesNoUnknown field called "symptomatic" and a new column for the symptoms table called "newSymptomatic". Add an annotation to the "symptomatic" field in Symptoms.java: `@DatabaseField(dataType = DataType.ENUM_STRING, columnName = "newSymptomatic")`.

Have a look at the "feelingIll" field for reference, we already went with this approach there.

MateStrysewske:
Closing this pull request because it hasn't been updated since October.

Issue #2910, opened 2020-09-18
Add phone number and email validation forms  (Case, Contact, Event Participant)


AndyBakcsy-she:
### Feature Description
Add on forms for: case, contact, event participant
Only on web app
![phoneNumber_email_Validation](https://user-images.githubusercontent.com/70562142/93588547-3e20fc00-f9b4-11ea-8b72-da18bbf08684.jpg)

Phone Forms should only accept phone number alike entries while Email Forms should only accept Email-alike entries.

AndyBakcsy-she:
Verified issue on SORMAS Version: 1.50.0-SNAPSHOT

MateStrysewske:
- [ ] Use validation in case and contact creation forms

AndyBakcsy-she:
Verified the issue on Case and Contact creation forms on SORMAS Version: 1.51.0-SNAPSHOT

AdrianMichaelis:
With minors we often use the phone number field as follows:  

`Mother: 1234 567 890 Father: 9876 5432` 

Not being able to enter values like that from now on is **not** a problem. 
However, I am worried about what will happen with **existing data** in this field once 1.50.1 goes live. Phone numbers disappearing from our records would be catastrophic. 

I'm unable to confirm the outcome of retroactive validation on the test server. Can anyone enlighten me?

ChristopherRiedel:
> With minors we often use the phone number field as follows:
> 
> `Mother: 1234 567 890 Father: 9876 5432`
> 
> Not being able to enter values like that from now on is **not** a problem.
> However, I am worried about what will happen with **existing data** in this field once 1.50.1 goes live. Phone numbers disappearing from our records would be catastrophic.
> 
> I'm unable to confirm the outcome of retroactive validation on the test server. Can anyone enlighten me?

I can assure you, there will be no data disappearing with the new feature. However, the feature will cause that if an existing person is opened with entered data as described by you, it will not be possible to save the person without changing the data in the phone number field.

AdrianMichaelis:
@ChristopherRiedel Thank you for the quick reply! And for setting my mind at ease

PR #3097, opened 2020-10-07
#2910 - Added phone number and email address validations

tabotkevin:
Screen Shot 2020-10-08 at 12 14 26


sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
```diff
@@ -403,6 +408,18 @@ protected void addFields() {
 				false,
 				I18nProperties.getValidationError(Validations.afterDate, burialDate.getCaption(), deathDate.getCaption())));
 
+		phoneNumber.addValidator(
+			new PhoneNumberValidator(
+				I18nProperties.getValidationError(Validations.validPhoneNumber)));
+
+		phoneOwner.addValidator(
```
MateStrysewske, line 415: The "Phone Owner" is not a phone number. It's the owner of the phone (used e.g. in African countries where there are remote communities with only one or a few phones that a specific person is responsible for).

sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
```diff
@@ -403,6 +408,18 @@ protected void addFields() {
 				false,
 				I18nProperties.getValidationError(Validations.afterDate, burialDate.getCaption(), deathDate.getCaption())));
 
+		phoneNumber.addValidator(
```
MateStrysewske, line 411: The PhoneNumberValidator we've written isn't really a good idea here; its main purpose is to ensure that the phone number is formatted like an international phone number which is necessary for the phone number added to users because those are then used for automatic SMS notifications. In retrospection, we should've probably named that class "UserPhoneNumberValidation" to make clear that it has a pretty small scope of usage.

For phone numbers of persons, we shouldn't enforce this. Instead, we should make sure that the phone number doesn't include characters that are invalid in phone numbers (e.g. letters I suppose). This isn't trivial because we, e.g., can't forbid characters like "(", ")", "-", "/" etc. because those are commonly used in different countries to format phone numbers.

We haven't clearly defined this yet which is why the issue was in the product backlog and not really "ready for development". If you have a proposal on how to do proper phone number validation, I'd like to hear it.

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
tabotkevin: @MateStrysewske thanks for the clarification, I think we can try one of these two things.

1. We can extend the existing PhoneNumberValidator to exclude invalid characters like letters and maybe other common things that phone numbers shouldn't have.
 
1. Write a new PhoneNumberValidator that can at least cover common validation formats for countries that Sormas supports, I know this process will be a long short, but we can pull it off.

MateStrysewske: The first approach would only be possible if there's an easy way to "configure" the validator to ignore the existing rule that phone numbers have to start with a "+" sign for all fields except the phone number of a user.

In general, I think having some basic validation (e.g. no letters) is enough for now. Explicitely covering the validation formats of specific countries is a very volatile approach because as SORMAS is an Open Source project, we technically don't even know which countries are using it.

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/PhoneNumberValidator.java
```diff
@@ -28,7 +28,7 @@ public PhoneNumberValidator(String errorMessage) {
 
 	@Override
 	protected boolean isValidValue(String phoneNumber) {
-		return phoneNumber == null || phoneNumber.isEmpty() || phoneNumber.startsWith("+");
+		return (phoneNumber == null || phoneNumber.isEmpty() || phoneNumber.startsWith("+")) && !phoneNumber.matches(".*[a-zA-Z].*");
```
MateStrysewske, line 31: Like written above, we can't go with this approach; this will force users to always use international phone numbers for persons and I don't think that's a good idea. Please:

* Revert the change to the PhoneNumberValidator and rename it to UserPhoneNumberValidator
* Create a new PhoneNumberValidator class that includes the check you've written, but does NOT check for "startsWith("+")" and use that for the person's phone number

MateStrysewske (changes requested):

MateStrysewske (changes requested):
The way you did it now is exactly the other way around. The UserPhoneNumberValidator should check whether the phone number starts with a + and the PhoneNumberValidator should ensure that no letters are used.

MateStrysewske (approved):

PR #3135, opened 2020-10-12
Add event info and filter in contact directory #2922 #3034

### LINKED ISSUE #2922: Have a filter in the contact directory to target contact linked to an event
@ftavin @xavier-calland @bernardsilenou  @olivierleuci

### Situation Description
The contact directory can be filtered by region, department, identifier, etc. Conversely, no filter can be used to filter on the possible involvement in a given event.

### Feature Description
In the directory of contact, it is proposed to
- Create two columns: event ID and event description (first 20 characters) (or "event title" when #2904 will be done)
- add a filter allowing to filter the contact according to the possible implication to a given event (via ID event or keyword of the description field) It is therefore a text field (like the filter "ID, name, EPID number, institution, person's telephone / postal code / city ”)

Like the cases (issue #2908 )

### LINKED ISSUE #3034: From the events directory, access the filtered contacts directory on involvement in a given event
### Situation Description

To date, the directory of contacts cannot be filtered on a possible involvement in an event. Issue #2922 makes possible to filter contacts based on the title of the event. To complete these developments, it would be useful to have a function allowing access to the list of contacts involved in an event from the "event" screen. The user could then very easily access to the exactly list of contacts involved in an event (via a filter on the id)

### Feature Description

It is proposed to create a link allowing the user to access, from the event file, directly to the contact directory which will then be automatically filtered on the only contacts of the event. This link would be represented by a button located below the “task” and “actions” blocks.

It complements the function allowing the filtering of the contact directory according to the involvement in an event. This offers greater flexibility in terms of managing the events and contacts involved.

Finally, this feature would have the advantage of not having to search, copy and paste the event ID it in the search block of the contact directory.

### Additional Information

This issue is the same mecanism as #3021 but applied to contacts

@carolinverset @xavier-calland @bernardsilenou

alexandre-gille:
* Add event id and title in contact directory
* Add event filter in contact directory
* Add a link to the contact directory automatically filtered by event
* Add a link to the filtered contacts involved in a given event

fixes #2922 
fixes #3034


ftavin:
@bernardsilenou @MateStrysewske 

Do you think this could be merged in time for 1.50 ?

MateStrysewske:
@ftavin Unfortunately not. We had a feature freeze on Friday in preparation of the release tomorrow because we're trying to improve the release stability. We'll have a look at it in the next days.

ftavin:
ok, thanks for your answer !

sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactCriteria.java
```diff
@@ -464,4 +468,31 @@ public YesNoUnknown getReturningTraveler() {
 	public void setReturningTraveler(YesNoUnknown returningTraveler) {
 		this.returningTraveler = returningTraveler;
 	}
+
+	public void setEventLike(String eventLike) {
+		this.eventLike = eventLike;
+	}
+
+	public String getEventLike() {
```
MateStrysewske, line 476: Add `@IgnoreForUrl` here as well to prevent personal data being leaked to the URL

sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactExportDto.java
```diff
@@ -596,6 +601,28 @@ public void setReturningTraveler(YesNoUnknown returningTraveler) {
...
+	@Order(72)
+	public String getLatestEventTitle() {
+		return latestEventTitle;
+	}
+
+	public void setLatestEventTitle(String latestEventTitle) {
+		this.latestEventTitle = latestEventTitle;
+	}
+
+	public long getEventCount() {
```
MateStrysewske, line 622: Order annotation missing

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -563,6 +586,20 @@ public void deleteContact(String contactUuid) {
 					});
 				}
 
+				// Load latest events info
+				// Adding a second query here is not perfect, but selecting the last event with a criteria query
+				// doesn't seem to be possible and using a native query is not an option because of user filters
+				List<ContactEventSummaryDetails> eventSummaries = eventService.getEventSummaryDetailsByContacts(resultContactsUuids);
```
MateStrysewske, line 592: Move this line to above the for loop; right now this is executed once for every contact in the list

MateStrysewske (changes requested):
Thanks for your submission, this looks really good in general! Only some minor change requests.

sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactCriteria.java
alexandre-gille: Unfortunately the link to contacts already filtered by events (#3034) won't work with this annotation

alexandre-gille:
> Thanks for your submission, this looks really good in general! Only some minor change requests.

Thanks for the review again ! Requested changes applied, aside from the  `@IgnoreForUrl` as I mentioned it in another comment

sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactCriteria.java
MateStrysewske: I'd suggest a workaround then; add another property to the ContactCriteria that is simply called "eventUuidLike" and use that if coming from a contact; since this field will not be visible in the UI, users can't accidentally leak personal data with it

MateStrysewske:
@alexandre-gille See my comment on the `@IgnoreForUrl` problem

sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactCriteria.java
alexandre-gille: That would do it, but it would be a bit wobbly in terms of user experience. 
Since there's no access to this field in the UI, you can't really know that this filter is applied without peeking at the URL and you're not able to turn it off directly from within the UI. You'd have to apply another filter, and only then use the `Reset filters` button.

I already made the changes, but this behaviour is bugging me a little bit. If you still think this is okay, i'll push it as is

MateStrysewske: You're right - that will probably lead to confusion. Still I think it's not okay to go with the approach of just putting the information into the URL, so we need to find some kind of solution.

A VERY hacky approach could be to automatically also apply the UUID to the filter field that is visible to the user. So the UUID filter would actually be applied twice, but that should be fine. Clicking on "Reset filters" would then also get rid of the "hidden" filter applied by clicking the button in the event. Do you think that makes sense?

alexandre-gille: We could add :

    if (criteria.getEventUuid() != null) {
	    criteria.eventLike(criteria.getEventUuid());
	    criteria.eventUuid(null);
    }

In the enter method of the contact view. It is indeed hacky, but it works.

It's either that or we add another field to filter on the UUID but that would be tantamount to the newly added eventLike field, and rather useless to the UX.

So, do I just go for the "hacky" way ?

MateStrysewske: I'm fine with approaches that are a bit hacky. Vaadin won't be our UI framework forever anyway. :)

Your suggestion would work for me!

alexandre-gille: Done !

MateStrysewske (approved):

Issue #3141, opened 2020-10-12
Inform Climedo about changes of FollowUpEndDates [0.5]


FredrikSchaeferVitagroup:
### Situation Description

 Climedo is informed about changes of person data, but not about changes of the latest FollowUpEndDate assigned to persons. This has to be implemented according to the procedure already implemented in #2925 .
### Possible Alternatives

None

### Additional Information

None

PR #3145, opened 2020-10-12
#3141 added update of external diaries for FollowUpUntil

FredrikSchaeferVitagroup:


sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
```diff
@@ -137,10 +144,35 @@ private String getPatientDiaryAuthTokenInternal() {
...
 
+	/**
+	 * Notify external journals that a followUpUntilDate has been updated
+	 *
+	 * @param contact
+	 *            a contact assigned to a person already available in the external journal
+	 * 
+	 */
+	public void notifyExternalJournalFollowUpUntilUpdate(ContactDto contact) {
+		SymptomJournalStatus savedStatus = personFacade.getPersonByUuid(contact.getPerson().getUuid()).getSymptomJournalStatus();
```
vlad-ciucescu, line 155: Is it possible that `getPersonByUuid` returns `null` and we get a NPE?

sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
```diff
@@ -137,10 +144,35 @@ private String getPatientDiaryAuthTokenInternal() {
...
+		if (SymptomJournalStatus.REGISTERED.equals(savedStatus) || SymptomJournalStatus.ACCEPTED.equals(savedStatus)) {
+			Date latestSavedFollowUpUntilDate = personFacade.getLatestFollowUpEndDateByUuid(contact.getPerson().getUuid());
+			if (contact.getFollowUpUntil().after(latestSavedFollowUpUntilDate)) {
```
vlad-ciucescu, line 158: Is it possible that `getFollowUpUntil` returns `null` and we get a NPE?

sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
FredrikSchaeferVitagroup: I think that is unlikely, because we call this for an existing contact. A contact is always associated with a person. And even if it returns null, the error should already be logged.
Are my reasons convincing enough to leave it like that? If not, what would you like me to do?

sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
FredrikSchaeferVitagroup: I think that is unlikely, because a contact automatically gets a FollowUpUntil date at creation. And even if it returns null, the error should already be logged.
Are my reasons convincing enough to leave it like that? If not, what would you like me to do?

sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
vlad-ciucescu: Yes, it is ok like this

sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
vlad-ciucescu: Yes, it is ok like this

PR #3148, opened 2020-10-13
3110 incorrect behavior on person forms related to facility fields

sergiupacurariu:


sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -215,9 +214,17 @@ protected void addFields() {
 		facilityTypeGroup.addValueChangeListener(e -> {
 			FieldHelper.removeItems(facility);
 			FieldHelper.updateEnumData(facilityType, FacilityType.getTypes((FacilityTypeGroup) facilityTypeGroup.getValue()));
+			if (facilityTypeGroup.getValue() != null) {
```
MateStrysewske, line 217: Can be simplified: `facilityType.setRequired(facilityTypeGroup.getValue() != null);`

MateStrysewske (changes requested):
- [ ] Facility name & description is not saved (select "Other facility" in the facility field, type in a name and save; the value is erased)

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
sergiupacurariu: updated

MateStrysewske (approved):

PR #3155, opened 2020-10-13
Add an investigation status field to events #2992

### LINKED ISSUE #2992: Addition of an investigation status field for events
### Situation Description
Each event is the subject of an in-depth investigation by one or more ARS teams. Many events are entered every day by the teams. Each of them is subject to a thorough investigation which can take several days.

It now appears essential to have a variable relating to the level of the survey.

This information will allow :

- filter the directory of events on the level of investigation: uninitiated, in progress, to be closed, closed

- to give overall consistency: we have this variable for the different actions implemented for a given event. Having the same variable for the event itself increases readability for the SORMAS user, especially in the detailed view relating to actions which is under development (issue # 2907)

- to give greater flexibility to users in the management of the investigation. To date, only administrators have the ability to archive or unarchive a closed event. It is not conceivable to give the profile of administrator to all the people in charge of the investigation. However, it is essential to be able to allow them to indicate whether the investigation has ended and whether, in his opinion, the event should be closed. The final decision will be up to the supervisor who would switch the status “to be closed” to “closed”. It is only after a period of time to be defined by the different regions that the event can then be archived for greater readability of the directory of events.

### Feature Description
Develop a "investigation status" variable to qualify the event. This variable should take the form of a drop-down list with the following modalities:

- Investigation to start
- Ongoing investigation
- Investigation to be closed
- Investigation closed

This variable would be included in the detailed view of actions and in the events directory : creation of a column and a search field allowing the events to be filtered according to the status of the investigation.

@ftavin @xavier-calland  @bernardsilenou

alexandre-gille:
Fixes #2992


sormas-api/src/main/java/de/symeda/sormas/api/event/EventDto.java
```diff
@@ -66,6 +69,10 @@
 
 	@Required
 	private EventStatus eventStatus;
+	@Required
```
MateStrysewske, line 72: Can't be required because it's possible that other countries don't want to use this feature.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1705,6 +1704,12 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				currentVersion = 236;
 				getDao(Contact.class).executeRaw("ALTER TABLE contacts ADD COLUMN returningTraveler varchar(255);");
 
+			case 237:
```
MateStrysewske, line 1707: You also need to increase the `DATABASE_VERSION` constant by one

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1705,6 +1704,12 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
...
+				currentVersion = 237;
```
MateStrysewske, line 1708: Please remove the not null constraint; SQLite doesn't support the removal of constraints in retrospection so we try to avoid it in the mobile app. Also please remove the default value because I suspect that this is also not editable afterwards

sormas-app/app/src/main/res/layout/fragment_event_edit_layout.xml
```diff
@@ -114,6 +115,29 @@
                 app:value="@={data.endDate}" />
         </LinearLayout>
 
+            <de.symeda.sormas.app.component.controls.ControlSwitchField
+                android:id="@+id/event_eventInvestigationStatus"
+                style="@style/ControlSingleColumnStyle"
+                app:enumClass="@{eventInvestigationStatusClass}"
+                app:required="true"
```
MateStrysewske, line 122: Remove the required attribute

sormas-backend/src/main/java/de/symeda/sormas/backend/event/Event.java
```diff
@@ -126,6 +133,34 @@ public void setEventStatus(EventStatus eventStatus) {
 		this.eventStatus = eventStatus;
 	}
 
+	@Enumerated(EnumType.STRING)
+	@Column(nullable = false)
```
MateStrysewske, line 137: Next line to remove :)

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -5494,4 +5494,14 @@ ALTER TABLE users ADD COLUMN hasConsentedToGdpr boolean default false;
 ALTER TABLE users_history ADD COLUMN hasConsentedToGdpr boolean default false;
 INSERT INTO schema_version (version_number, comment) VALUES (267, 'Add gdpr popup to user');
 
+-- 2020-10-12 Add event investigation status
+ALTER TABLE events ADD COLUMN eventInvestigationStatus varchar(255) not null default 'PENDING';
```
MateStrysewske, line 5498: Also no not null constraints here; default should also be removed because it might lead to the value in the database being changed without the change date being adjusted, which could result in synchronization issues with the mobile app; we're already making sure that events are initialized as pending with the build methods

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -5494,4 +5494,14 @@ ALTER TABLE users ADD COLUMN hasConsentedToGdpr boolean default false;
...
+ALTER TABLE events_history ADD COLUMN eventInvestigationStatus varchar(255) not null default 'PENDING';
```
MateStrysewske, line 5499: You don't need any constraints or default values for the history tables

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
```diff
@@ -155,6 +158,15 @@ protected void addFields() {
 		initEventDateValidation(startDate, endDate, multiDayCheckbox);
 
 		addField(EventDto.EVENT_STATUS, OptionGroup.class);
+		addField(EventDto.EVENT_INVESTIGATION_STATUS, OptionGroup.class);
```
MateStrysewske, line 161: Since we need to remove the requirement, please use the new NullableOptionGroup here

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
```diff
@@ -209,7 +221,14 @@ protected void addFields() {
 			Collections.singletonList(EventDto.DISEASE_DETAILS),
 			Collections.singletonList(Disease.OTHER));
 
-		setRequired(true, EventDto.EVENT_STATUS, EventDto.UUID, EventDto.EVENT_DESC, EventDto.REPORT_DATE_TIME, EventDto.REPORTING_USER);
+		setRequired(
+			true,
+			EventDto.EVENT_INVESTIGATION_STATUS,
```
MateStrysewske, line 226: You know it ;)

MateStrysewske (changes requested):
Looks good! I'd only ask you to remove the required constraint from the field because some countries might not want to use it.

alexandre-gille:
> Looks good! I'd only ask you to remove the required constraint from the field because some countries might not want to use it.

Thanks for the review ! I just applied the requested changes :+1:

MateStrysewske (changes requested):
Looks like the date fields in the app aren't working; I guess you forgot to set them up (have a look at how other date fields are handled in the fragments)

alexandre-gille:
@MateStrysewske 
> Looks like the date fields in the app aren't working; I guess you forgot to set them up (have a look at how other date fields are handled in the fragments)

I think I found the solution, unfortunately, I wasn't able to run the mobile app on my dev environment following the android studio section of the 'DEVELOPMENT_ENVIRONMENT.md' guide. If you don't mind checking it please, that would be nice !

MateStrysewske:
@alexandre-gille Can you have a look at the test failures?

alexandre-gille:
@MateStrysewske I fixed it !

MateStrysewske (approved):

Issue #2923, opened 2020-09-21
When the case investigation is set to done, then the investigation task should be set to done, not removed.


bernardsilenou:
### Bug Description
When the case investigation is set to done, then the investigation task is set to removed. (this should stay as done)
But when you set the case classification to something apart from unclassified, save the case and latter open the task and set it to done,  it remains done. (this is fine like this)
- Setting a case investigation task to done, does not set the investigation status of the case to done. (investigation status should be bet to done also. The condition that we must classify the case before setting the task to done has already taken case of this, so we can assume that when the task is set to done, then the case must have been investigated and classified)
- When investigation status of the case is set to done, after setting the task to done, error message appears
### Steps to Reproduce
1.
2.

### Screenshots
![image](https://user-images.githubusercontent.com/12996724/93743183-d8b94f00-fbef-11ea-97bf-00409d728355.png)

![image](https://user-images.githubusercontent.com/12996724/93743396-4796a800-fbf0-11ea-92b1-513bdf5c897c.png)

### System Details
* Device: web
* SORMAS version: 1,48
* Android version/Browser:

MateStrysewske:
One of the problems still exists:
* Open a case with investigation status pending
* Open the investigation task and set it to done
* Save the case form -> "changed in the meantime" warning appears

Solution:
When opening a task from within a case, display a hint in the task edit window, directly above the buttons:
"Saving this task will discard any unsaved changes made to the case."
When the task is saved, reload the case page.

AndyBakcsy-she:
Verified issue on SORMAS Version: 1.51.0-SNAPSHOT.

PR #3160, opened 2020-10-13
#2923 - reload case view after editing a task + add info to task edit…

BarnaBartha:
… form

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskEditForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskEditForm.java
@@ -82,11 +82,15 @@ public class TaskEditForm extends AbstractEditForm<TaskDto> {
 	//@formatter:off
 
 	private UserRight editOrCreateUserRight;
+	private boolean editedFromTaskGrid;
 
-	public TaskEditForm(boolean create) {
+	public TaskEditForm(boolean create, boolean editedFromTaskGrid) {
 
 		super(TaskDto.class, TaskDto.I18N_PREFIX);
+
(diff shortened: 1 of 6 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskList.java
```diff
@@ -87,7 +87,7 @@ protected void drawDisplayedEntries() {
 			if (UserProvider.getCurrent().hasUserRight(UserRight.TASK_EDIT)) {
 				listEntry.addEditListener(
 					i,
-					(ClickListener) event -> ControllerProvider.getTaskController().edit(listEntry.getTask(), TaskList.this::reload));
+					(ClickListener) event -> ControllerProvider.getTaskController().edit(listEntry.getTask(), TaskList.this::reload, true));
```
MateStrysewske, line 90: TaskList is also used in contacts and events, so this can't be generically "true"

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskEditForm.java
```diff
@@ -104,6 +114,14 @@ protected void addFields() {
 		taskContext.setImmediate(true);
 		taskContext.addValueChangeListener(event -> updateByTaskContext());
 
+		final HorizontalLayout saveInfoLayout = new HorizontalLayout(
```
MateStrysewske, line 117: This hint is currently always displayed, even when entering a task from the task directory

MateStrysewske (changes requested):

MateStrysewske (approved):

Issue #3162, opened 2020-10-13
Updating person data/followup date sends the older data to CLIMEDO


vlad-ciucescu:
### Bug Description
For persons registered in CLIMEDO, changing person data or the followup end date in a contact should trigger an automatic update of the related person in CLIMEDO. This happens, but CLIMEDO receives the old data from SORMAS.

### Steps to Reproduce
1. Open a contact from the contact directory with the REGISTERED status.
2. Open the person data.
3. Update person data such as name or birthdate.
4. Save changes
5. Open the CLIMEDO eDiary by clicking the button in the contact page.
6. The data in CLIMEDO is the person data which was available in SORMAS before updating

### Additional Information
Updating the person a 2nd time will send the data from the 1st update to CLIMEDO, meaning that they are 1 update behind SORMAS.

PR #3175, opened 2020-10-14
#3162 execute notify external journal methods in separate thread with delay

vlad-ciucescu:


FredrikSchaeferVitagroup (approved):

FredrikSchaeferVitagroup:
I was informed that @MateStrysewske said we can in this case merge with the travis build failing. Locally everything works fine.

Issue #3136, opened 2020-10-12
[SORMAS2SORMAS] Share cases and contacts in bulk [3]


leventegal-she:
### Situation Description
Functionality for sharing cases and contacts in bulk.
### Feature Description
- [x] add "Share" bulk action on the directory pages
- [x] for selecting target and share options use the same dialog as when sharing one object from the details page
- [x] implement backend for sending/receiving a list of cases or contacts
- [x] add right to share single data to other SORMAS instances to surveillance officer
- [ ] add right to share bulk data to other SORMAS instances to surveillance officer
- [x] Hide the share option for cases / contacts where you don't have the right to edit it or see the person's data

PR #3219, opened 2020-10-19
#3136 sormas to sormas bulk

leventegal-she:


sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/SormasToSormasFacadeEjb.java
```diff
@@ -165,88 +172,179 @@ public SormasToSormasFacadeEjb() {
...
+
+			ValidationErrors contactErrors = validateContact(contact);
+			if (contactErrors.hasError()) {
+				validationErrors.put(buildContactValidationGroupName(contact), contactErrors.getErrors());
+
+				continue;
+			}
 
-		validateContact(contact);
+			ValidationErrors contactErrorsErrors = new ValidationErrors();
```
vidi42, line 260: I think `contactValidationErrors` or simply `contactErrors` would be enough 😃

sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasException.java
```diff
@@ -15,9 +15,27 @@
 
 package de.symeda.sormas.api.sormastosormas;
 
+import java.util.List;
+import java.util.Map;
+
 public class SormasToSormasException extends Exception {
 
+	private Map<String, Map<String, List<String>>> errors;
```
vidi42, line 23: Same here, `ValidationErrors` or a more generic class could be used.

sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasErrorResponse.java
```diff
@@ -16,23 +16,25 @@
 package de.symeda.sormas.api.sormastosormas;
 
 import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
 
 public class SormasToSormasErrorResponse implements Serializable {
 
-	private String message;
+	private Map<String, Map<String, List<String>>> errors;
```
vidi42, line 24: Since there is the `ValidationErrors` class this could be changed to private final `Map<String, ValidationErrors> errors;` to make it more easier to read and use. `ValidationErrors` can be named more generic if it helps in other places.

vidi42 (commented):
Added some suggestions,

sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasErrorResponse.java
leventegal-she: I don't see where ValidationErrors could be reused right now, maybe later we can generalize it.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
```diff
@@ -1135,4 +1141,12 @@ public void updateArchived(List<String> caseUuids, boolean archived) {
 
 		em.createQuery(cu).executeUpdate();
 	}
+
+	public boolean isCaseEditAllowed(Case caze) {
+		if (caze.getSormasToSormasOriginInfo() != null) {
+			return caze.getSormasToSormasOriginInfo().isOwnershipHandedOver();
+		}
+
+		return caseJurisdictionChecker.isInJurisdictionOrOwned(caze) && !sormasToSormasShareInfoService.isCaseOwnershipHandedOver(caze);
```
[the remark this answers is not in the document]
leventegal-she: s2s origin is on the receiver system and sormasToSormasShareInfo is on the sender system, so need to check both.

vidi42 (approved):

PR #3234, opened 2020-10-20
 #3217 #3231 provide more detailed validation messages, update validation logic for external persons

### LINKED ISSUE #3217: [CLIMEDO Integration] Adapt messages [0.5]
### Problem Description

There are two messages that are not up to date with the state of the interface. 

- One shows when you succesfully created a climedo patient from within SORMAS. It says:
`Successfully registered person in the patient diary.
 Person mit ID: TMH2D2-F3ZZUM-X63TE4-UXGHSITE erfolgreich in das Climedo Symptomtagebuch aufgenommen!`
 The first sentence comes from SORMAS, the second now comes from Climedo. I think we should remove the first one.

- The other shows when you click on the climedo button from a contact without e-mailaddress and phone number. The message says:
`Cannot export the person to the patient diary. The person needs a valid birthdate and either a valid phone number or email address.` 
That is actually wrong. A birthdate is not necessary. it's rather that IF you provide a birthdate for that contactPerson, it has to be a complete one (not just birthyear without birthmonth and -date). We should remove the birthdate part from this message, and create another message for incomplete birthdates.

### LINKED ISSUE #3231: [CLIMEDO Interface] Change logic in ExternalJournalService.isPersonExportable(...) [0.5]
### Feature Description

The ExternalJournalService.isPersonExportable(...) checks if a person has enough valid data to be exported to climedo.

### Problem Description

The method returns true if a valid phone number and/or a valid eMail address exists. So far so good. It also returns true if an invalid phone number and a valid eMail or vice versa exist.

### Proposed Change

That is not desired, as it may lead to errors in the export and should be changed. Also the relating user messages should be adapted.

### Additional Information

May relate to #3217

vlad-ciucescu:
closes #3217 
closes #3231


vlad-ciucescu:
Should be merged after the climedo_integration branch is merged into development

FredrikSchaeferVitagroup:
I have found a behavior that I consider undesired.
1. Open a Contact without birthdate.
2. Enter an invalid birthdate (e.g. only birthyear).
3. Click on Save
4. An error message shows up saying: Invalid birthdate.

The state the ContactPersonView is in after that is what I consider undesired. The change is shown (e.g. the specified year is still visible) but when I correct my mistake (e.g. specify a birthmonth and birthday) and click on save again, I get a popup message saying "The data could net be saved because it has been changed in the meantime". When I reload the view, the invalid birthdate (e.g. only the year) is still shown.
I consider this undesired because it may irritate users. A solution might be to only warn that the invalid birthdate can not be exported to symptom journals (the exclusion of invalid birthdates has to be made sure of in the personFacadeEjb.getPersonForJournal(...)) and reload the view, so that it can directly be edited.
Another solution might be to not allow to save invalid birthdates. I don't know I some users save incomplete birthdates for data protection reasons, so that might not be a good way to go.

FredrikSchaeferVitagroup:
It also seems that I can call climedo to register persons that have neither an email nor a phone number.

vlad-ciucescu:
I fixed the email/phone validation issue.
The other issue with "data changed in the meantime" was fixed by performing the validation before the entity is persisted in the database.

FredrikSchaeferVitagroup (approved):

PR #3270, opened 2020-10-23
2782 optimize person similarity duplication check

sergiupacurariu:
changes:
- the existing 3 similarity queries were transformed into a single query
- a trigram index on person was added
- the similarity function was changed to "%" operator (the tests between these 2 approaches returned the same results). The reason for this switch is that the operator "%" works better with the above index.
- similarity check improvement is 9 - 10x
- overall import process 3 - 4x


sormas-api/src/main/java/de/symeda/sormas/api/person/PersonFacade.java
```diff
@@ -57,6 +57,8 @@
 	 */
 	List<PersonNameDto> getMatchingNameDtos(UserReferenceDto user, PersonSimilarityCriteria criteria);
 
+	boolean checkMatchingNameDtos(UserReferenceDto userRef, PersonSimilarityCriteria criteria);
```
MateStrysewske, line 60: If I don't misunderstand anything, this method is only used in the import to determine whether there are any duplicate results; however, it still uses the same method in PersonService as getMatchingNameDtos, only with a limit; what I'm wondering is whether the limit actually makes a difference? Because the query still needs to do all the joins and that kind of stuff. Wouldn't it maybe be better to already use getMatchingNameDtos at that point and then store the result in the CaseImportEntities so that we don't have to do the same query again in the PersonSelectionGrid?

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/caseimport/CaseImportFacadeEjb.java
```diff
@@ -654,7 +654,11 @@ private boolean isPersonSimilarToExisting(PersonDto referencePerson) {
 			.passportNumber(referencePerson.getPassportNumber())
 			.nationalHealthId(referencePerson.getNationalHealthId());
 
-		return personFacade.getMatchingNameDtos(userFacade.getCurrentUser().toReference(), criteria).stream().findAny().isPresent();
+//		boolean present = personFacade.getMatchingNameDtos(userFacade.getCurrentUser().toReference(), criteria).stream().findAny().isPresent();
```
MateStrysewske, line 657: Remove commented line

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/caseimport/CaseImportFacadeEjb.java
```diff
@@ -654,7 +654,11 @@ private boolean isPersonSimilarToExisting(PersonDto referencePerson) {
...
+
+		boolean present = personFacade.checkMatchingNameDtos(userFacade.getCurrentUser().toReference(), criteria);
```
MateStrysewske, line 659: No need for a variable declaration here, just return the result immediately

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/caseimport/CaseImportFacadeEjb.java
sergiupacurariu: removed

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/caseimport/CaseImportFacadeEjb.java
sergiupacurariu: removed

sormas-api/src/main/java/de/symeda/sormas/api/person/PersonFacade.java
sergiupacurariu: the method "checkMatchingNameDtos" was renamed to "checkMatchingNameInDatabase"

Issue #3069, opened 2020-10-05
Make campaign selector for campaign data entry/import more prominent [1]


MartinWahnschaffeSymeda:
### Feature Description
In the campaign data view it's possible to filter by campaign. When this is done the campaign is pre-selected for import of data and entry of new data forms.

### Problem Description
It would be better to make users select a campaign before being able to import and create data. This will allow us to only show campaign forms that are assigned to the campaign (see #2855).

### Proposed Change
- [x] Move the campaign selection to the top right of the screen - on the left side of the export/new form/import buttons
- [x] Exchange order of Campaigns and Campaign Data tabs in Campaigns view. Campaign Data should be shown first 
- [x] When opening the campaign data view pre-select the last campaign that has started (see dashboard). Only exception to this is when the user selects a specific campaign in the campaigns list view
- [x] Limit campaign data forms for "new form" and "import" to those available for the campaign.
- [x] Add header to the "edit" and "open" icon columns in the campaigns list

PR #3281, opened 2020-10-26
#3069 modify campaign selector for campaign data view

danRosca:


MateStrysewske (changes requested):
* The "New Form" and "Import" selectors aren't working properly. I had one campaign in my system that included the only form in the system. I then created a new campaign with the same form. When I select the new campaign in the header, "New Form" and "Import" are empty. If I select the first campaign that I already had in the database, the form appears twice in both selectors.
* The headings for the edit and view columns are automatically shrinked and only display a single letter; they probably need a minimum width.

sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/campaigns/CampaignGrid.java
```diff
@@ -85,7 +86,10 @@ public CampaignGrid(CampaignCriteria criteria) {
 
 		for (Column<?, ?> column : getColumns()) {
 			column.setCaption(I18nProperties.getPrefixCaption(CampaignIndexDto.I18N_PREFIX, column.getId(), column.getCaption()));
+			column.setMinimumWidth(500);
```
MateStrysewske, line 89: This now makes almost all columns way too wide. Only the two icon columns need a minimum width to make sure that the heading is still shown because the icon is too small to ensure that.

Also, please rename the "Open" column to "View" because it's not technically used to open the campaign, but to view its data.

MateStrysewske (changes requested):
Please correct your import settings and re-organise the imports in all files you've edited.

sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/campaigns/CampaignGrid.java
danRosca: @MartinWahnschaffeSymeda told me to name the column "Open"

MartinWahnschaffeSymeda: Yes, I'd prefer open, because we don't want to view the campaign itself, but to open it's campaign data :-D

MateStrysewske: Okay, fine by me then!

MateStrysewske (approved):

PR #3316, opened 2020-10-28
Modifies a filter's label #3314

### LINKED ISSUE #3314: Wrong caption for a checkbox in contacts directory filters
### Bug Description

The restricting checkbox in the development of #2922 was badly labelled by the developper. The checkbox effect is good and correspond to the issue but the caption is not exact.

### Steps to Reproduce
1. Go to Contacts Directory
2. Develop the filters with "show more filters"
3. The caption is "Only contacts whose source case is linked to their latest event"

### Expected Behavior

The caption should be "only contacts who have an event in common with their source case"

### Screenshots

Capture

### Additional Information

@bernardsilenou @carolinverset @alexandre-gille

alexandre-gille:
Fixes #3314


MateStrysewske (changes requested):
Hey! Please revert the changes to all captions.properties file except the default one and fr; everything else should be automatically handled by Crowdin because it's not only a renaming, but also a rephrasing, and I don't necessarily think it's a good idea to adjust the translations for other languages (e.g. the word "Quellfall" isn't really used and there's a grammatical error in the German translation) anyway.

alexandre-gille:
Hi, no problem, I removed them !

MateStrysewske (approved):

PR #3321, opened 2020-10-28
Feature 2890 campaign data by jurisdiction

BarnaBartha:


sormas-api/src/main/resources/validations.properties
```diff
@@ -34,6 +34,9 @@ importEntryDoesNotExist = Invalid value %s in column %s; Entry does not exist in
 importEntryDoesNotExistDbOrCommunity = Invalid value %s in column %s; Entry does not exist in the database or in the specified community or is archived (Or maybe a wrong facility type was specified)
 importEntryDoesNotExistDbOrDistrict = Invalid value %s in column %s; Entry does not exist in the database or in the specified district or is archived (Or maybe a wrong facility type was specified)
 importEntryDoesNotExistDbOrRegion = Invalid value %s in column %s; Entry does not exist in the database or in the specified region or is archived
+importEntryRegionNotInJurisdiction = Invalid value %s in column %s; Specified region is not within you're user's jurisdiction
```
MateStrysewske, line 37: "you're user's jurisdiction" -> "your jurisdiction" (for all three strings)

Also, rename "importEntryRegionNotInJurisdiction" to "importEntryRegionNotInUsersJurisdiction" to differentiate it from the other "notInJurisdiction" messages that are used to specify that the district of the imported entity does not belong to the region of the imported entity

sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/campaigndata/CampaignFormDataEditForm.java
```diff
@@ -113,6 +115,21 @@ protected void addFields() {
 				}
 			});
 		}
+
+		final UserDto currentUser = UserProvider.getCurrent().getUser();
+
+		if (currentUser.getRegion() != null) {
+			cbRegion.setValue(currentUser.getRegion());
+			cbRegion.setReadOnly(true);
```
MateStrysewske, line 123: use setEnabled(false) instead, read only leads to the field disappearing and only the caption being displayed which looks weird

sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/importer/CampaignFormDataImporter.java
```diff
@@ -213,4 +238,20 @@ private CampaignFormDataDto insertColumnEntryIntoData(CampaignFormDataDto campai
 		}
 		return currentElement;
 	}
+
+	@Override
+	protected boolean executeDefaultInvokings(PropertyDescriptor pd, Object element, String entry, String[] entryHeaderPath)
+		throws InvocationTargetException, IllegalAccessException, ParseException, ImportErrorException {
+		final boolean returnBoolean = super.executeDefaultInvokings(pd, element, entry, entryHeaderPath);
```
MateStrysewske, line 245: If you override executeDefaultInvokings anyway (without calling super), you could also get rid of the call in the above method and just add the region code there

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/importer/CampaignFormDataImporter.java
BarnaBartha: what i want to achieve here is to execute the jurisdiction check for region before the import get's to parse the district
super is called - and if no exception is thrown i return the value from super here

MateStrysewske: Okay, I see - fine by me then :)

MateStrysewske (approved):

Issue #3259, opened 2020-10-22
Error when editing campaign dashboard grid


BarnaBartha:
### Bug Description
Error when editing campaign dashboard grid on https://test-afghanistan.sormas.netzlink.com/
### Steps to Reproduce
1. Go to https://test-afghanistan.sormas.netzlink.com/sormas-ui/#!campaign/data/UT5HKR-3GDBCA-MYIOXL-J35PSLR4
2. Edit the campaign dashboard grid

### Expected Behavior
Grid is editable.

### Screenshots
![image](https://user-images.githubusercontent.com/8166805/96875105-87ff7500-147f-11eb-96f2-8b93b80fde92.png)

### System Details
* Device: 
* SORMAS version: 1.51.1-SNAPSHOT
* Android version/Browser: Chrome on MacOS
* Server URL: https://test-afghanistan.sormas.netzlink.com
* User Role: National coordinator

### Logs
[sormas-21102020.log](https://github.com/hzi-braunschweig/SORMAS-Project/files/5422955/sormas-21102020.log)

MartinWahnschaffeSymeda:
@BarnaBartha This is the log:

[...]

PR #3345, opened 2020-10-29
#3259 - add validation also when fetching dashboard elements to avoid…

BarnaBartha:
… providing null to vaadin fields


sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/campaigns/CampaignDashboardElementsGridComponent.java
```diff
@@ -33,6 +33,7 @@ public CampaignDashboardElementsGridComponent(List<CampaignDashboardElement> sav
 
 		final List<String> existingDiagramIds = allElements.stream()
 			.map(campaignDiagramDefinitionDto -> campaignDiagramDefinitionDto.getDiagramId())
+			.filter(s -> StringUtils.isNotEmpty(s))
```
MartinWahnschaffeSymeda, line 36: We need the same filter in line 50

MartinWahnschaffeSymeda (changes requested):

MartinWahnschaffeSymeda (approved):

PR #3358, opened 2020-11-02
Feature 3195 unit tests for the external visits ressource

### LINKED ISSUE #3195: Unit tests for the ExternalVisitsRessource [2]
### Situation Description

Climedo and PIA are depending on the API defined in the ExternalVisitsRessource. Whenever that API changes, theis developers need to be informed. 

### Feature Description

As this was not noticed previously, unit tests shall help to notify whenever the API changes. It would be good to directly advise to notify depending developer teams in case of a test fail.

### Additional Information

To a small extend, there may be tests for the ExternalVisitsRessource exisitng already (I remember writing some for the getLatestFollowUpUntil).

FredrikSchaeferVitagroup:
closes #3195


FredrikSchaeferVitagroup:
I'm really not proud of the extractPathsOfController(...) in the ExternalVisitsResourceTest.java . If you have a better idea how to do it, please let me know.

sormas-rest/src/test/java/de/symeda/sormas/rest/ExternalVisitsResourceTest.java
```diff
@@ -0,0 +1,150 @@
...
+	 * @param topLevelMap
+	 *            Nested Map from which to extract the information. It's supposed to be a mapped swagger.json
+	 * @param controller
+	 *            The name of the controller, e.g. External Visits Controller
+	 * @param list
+	 *            Extracted information is stored in this list
+	 * @return Documentation about any path found for the specified controller (e.g. /visits-external/person/{personUuid} for the External
+	 *         Visits Controller). This includes parameter names for that path, but not information about related enums.
+	 */
+	private static ArrayList extractPathsOfController(Map topLevelMap, String controller, ArrayList list) {
```
vlad-ciucescu, line 86: I agree this is a bit convoluted. Maybe instead of searching for the tags, it would be easier to match the keys against a regex (to find objects whose key start with `/visits-external/`). It might be easier to parse the json using Jackson specific objects instead of maps/strings and then comparing those objects. See [here](https://www.baeldung.com/jackson-compare-two-json-objects). It may also be possible to use [json-path](https://github.com/json-path/JsonPath).
Unfortunately, due to the structure of the generated json, I can't find better ideas now.

sormas-rest/src/test/java/de/symeda/sormas/rest/ExternalVisitsResourceTest.java
```diff
@@ -0,0 +1,150 @@
...
+	 *            Nested Map from which to extract the information. It's supposed to be a mapped swagger.json
+	 * @param detailName
+	 *            The name of the detail, e.g. JournalPersonDto.
+	 * @param list
+	 *            Extracted information is stored in this list
+	 * @return
+	 *         Documentation found about the detail. The Map is searched for a key equal to detailName, an it, plus the according value is
+	 *         added to the list.
+	 */
+	private static ArrayList extractDetail(Map topLevelMap, String detailName, ArrayList list) {
```
vlad-ciucescu, line 125: same as above

vlad-ciucescu (changes requested):

sormas-rest/src/test/java/de/symeda/sormas/rest/ExternalVisitsResourceTest.java
FredrikSchaeferVitagroup: Thank you for the advice. I did give Jackson a try, but it turned out to be complicated, too.
json-path may be an alternative. It will probably result in less messy code, be might be a little less generic and vulnerable to changes in the structure of the json.
As the code I wrote works and we shall have this in development soon, let's just leave it like that, okay?

sormas-rest/src/test/java/de/symeda/sormas/rest/ExternalVisitsResourceTest.java
FredrikSchaeferVitagroup: Also same as above. At least, the code here is simpler.

sormas-backend/src/test/java/de/symeda/sormas/backend/externaljournal/ExternalJournalServiceTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+		// Create two person with those properties
+		PersonDto person = new PersonDto();
+		PersonDto updatedPerson = new PersonDto();
+		person.setUuid(DataHelper.createUuid());
+		for (Map.Entry<String, Object> property : relevantProperties.entrySet()) {
+			setPersonProperty(person, property.getKey(), property.getValue());
+			setPersonProperty(updatedPerson, property.getKey(), property.getValue());
+		}
+		assertFalse(getExternalJournalService().notifyExternalJournalPersonUpdate(person, updatedPerson));
```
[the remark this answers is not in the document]
FredrikSchaeferVitagroup: I my system, that does not happen. My reasoning behind is that the tests are executed before the deploy is finished. Thus, the sormas.properties holding the config for a real external journal should be out of reach for the config facade. Right?

sormas-rest/src/test/java/de/symeda/sormas/rest/ExternalVisitsResourceTest.java
vlad-ciucescu: I agree we can can leave this like this for now.

sormas-backend/src/test/java/de/symeda/sormas/backend/externaljournal/ExternalJournalServiceTest.java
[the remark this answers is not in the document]
vlad-ciucescu: Yes, when I tested the config facade was returning only nulls in this test so in that way it should be not possible to make an actual call. If that is always the case then I am fine with this, I don't have more insight into how the tests relate to the properties.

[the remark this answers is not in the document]
FredrikSchaeferVitagroup: I'm not 100% sure either.
@MateStrysewske: Are we right in assuming that the configFacade does not have access to actually used sormas.properties during tests? We want to make sure that the tests I wrote don't make real calls to climedo instances configured in production sormas.properties files.

sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
```diff
@@ -312,19 +312,20 @@ public PatientDiaryRegisterResult registerPatientDiaryPerson(PersonDto person) {
...
 	 * @param person
 	 *            the person to validate
+	 * @param calledForTest
+	 *            defines if method makes a call to a patient diary to check for availability of phone and mail addresses.
+	 *            In tests, this call is not made.
+	 *            This boolean should be false unless it's called from tests.
 	 * @return the result of the validation
 	 */
-	public PatientDiaryPersonValidation validatePatientDiaryPerson(PersonDto person) {
+	public PatientDiaryPersonValidation validatePatientDiaryPerson(PersonDto person, boolean calledForTest) {
```
vlad-ciucescu, line 328: I don't think we should add different functionality specifically for tests, since the tested method should work exactly as if it would be called outside a test. If the intent is not to call the actual CLIMEDO API, the you can mock the `queryPatientDiary` method.

vlad-ciucescu (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
FredrikSchaeferVitagroup: Implemented.

vlad-ciucescu (approved):

Issue #3360, opened 2020-11-02
[REST] handle missing UUID in EntityDTOs gracefully [3]


JonasCir:
### Bug Description
This problem exists for each API DTO. The problem is that the `uuid` field for each entity is constrained with `NOT NULL` on database level, however, this is not properly handled/indicated in the API. For every entity which omits the uuid key in a POST request, this will result in an exception in the DB transaction b/c of the constraint violation.

Please also not that the abstract `ReferenceDTO` is also missing a `@Required` annotation for uuid.

### How to resolve it
- [x] ~Add `@Required` annotation for uuid to `ReferenceDTO`~ done in https://github.com/hzi-braunschweig/SORMAS-Project/pull/3366
- [ ] Regarding the entity DTOs: Auto-generate uuids for entities where the key is missing. This aligns with the behavior of the CSV import where we also generate the uuid automatically. In cases where the uuid is provided, it is copied over from the request.

JonasCir:
CC @benchaabaner

JonasCir:
I'll provide a PR which adds `@Required` annotation to `ReferenceDTO.uuid` shortly

JonasCir:
and CC @RafaelKluender b/c of my discussion with him

PR #3366, opened 2020-11-02
[IMIS] 3360 reference dto required UUID

JonasCir:
Ticks partially one box of #3360 
This fix makes the `uuid` fields of all reference DTOs required. -> this makes the `uuid` fields in the openAPI description required -> clients generated from this api spec validate automatically that uuids are set correctly in the request -> this solves (partially) my problem

Please note that this does not add any server side validation, providing empty `uuid` values in references still triggers an internal server error (500) and is not gracefully handled (i.e., 400). This is left as future work, but having this change included in the repo would tremendously help me to debug the API.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/ReferenceDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/ReferenceDto.java
@@ -17,6 +17,8 @@
  *******************************************************************************/
 package de.symeda.sormas.api;
 
+import de.symeda.sormas.api.utils.Required;
+
 import java.io.Serializable;
 
 import org.apache.commons.lang3.ObjectUtils;
@@ -26,6 +28,7 @@ public abstract class ReferenceDto implements Serializable, HasUuid, Comparable<
 
 	public static final String CAPTION = "caption";
 
+	@Required
 	private String uuid;
 	private String caption;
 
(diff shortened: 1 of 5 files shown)


JonasCir:
@syntakker somehow your commits ended up in my PR :D

JonasCir:
Does not work for some reason

JonasCir:
@MateStrysewske This is ready for review, please reassign at your discretion.

This PR indicates that `*ReferenceDtos` should carry an `uuid` property, at least on API level. This does not add any backend checks, providing empty references still blows up the resulting DB transaction as these fields are mapped as `NOT NULL`. However, this allows clients which are generated from the openAPI specs to check the invariants locally which makes development much more pleasant. 

Please also note, that simply adding `@Required ` to `ReferenceDto` wouldn't work. I rewrote the logic such that `@Required` on a field in a super class now gets reliable populated to sub classes. The changes introduced are the following:

1. Make `ReferenceDto.uuid` required
1. Move the annotation handling logic from `SormasSwaggerExtension ` to `AttributeConverter`
  1. The handling for `@Required ` was moved to `applyBeanValidatorAnnotations` (it's the same function which handles `@NotNull` and such)
  1. rest of `resolve` is the same, just the call to super was added

MartinWahnschaffeSymeda:
@JonasCir Looks good to me and makes a lot of sense :-)
Also thanks a lot for cleaning up the code formatting and comments.

Issue #3276, opened 2020-10-23
Complete the event directory with two columns relating to the number of cases and contacts involved in the event


carolinverset:
### Situation Description

To date, the event directory presents information characterizing the event, including the number of people participating in the event. It would be relevant to supplement these data with the number of cases and contacts involved in the event.

### Feature Description

In the event directory, create two columns relating to the number of cases on the one hand and the number of contacts on the other. In terms of the number of cases, the only contacts whose source case is also involved in the event will be counted.
This information will also be integrated into the export file (basic and detailed)

### Additional Information
@ftavin @olivierleuci @bernardsilenou

Seal33:
Just so I understand this right, column 1 is supposed to show how many of the event's participants are cases, while column 2 is supposed to tell how many contacts those cases have in total?

ftavin:
Ok with the first column.

The 2nd column should tell how many "contacts" (in general) have their "person" involved in this event.

A checkbox in the filters should be added to add the constraint "source case of the contact is also involved in the event" for the calculation of the 2nd column (that's what @bernardsilenou usually ask for this type of event-contact relationship count)

Seal33:
> Ok with the first column.
> 
> The 2nd column should tell how many "contacts" (in general) have their "person" involved in this event.
> 
> A checkbox in the filters should be added to add the constraint "source case of the contact is also involved in the event" for the calculation of the 2nd column (that's what @bernardsilenou usually ask for this type of event-contact relationship count)

Thank you for your help! I guess each person should only be counted once though, correct? Assuming a scenario where Person A, B and C are part of an event. A and B are cases, and C is a contact of person A and B. A and B are not listed as contacts of one another. In this scenario, would the total number of contacts be 1? Or 2, as C is a contact of both A and B?

bernardsilenou:
@Seal33 @ftavin Thanks for the insight. Here are my suggestions:
1. Since in the other sections of sormas (contact repository, dashboard, etc) we count contacts rather than contact persons, I would recommend we use 2.
2. @Seal33  an extension of your scenario can be that,  resulting case A has a contact with person D and D is not an event participant of the event in question. In this case, D should not contribute to the sum of number of contacts by default. The logic can be:
Number of contact = sum of  unique contacts of all resulting cases condition on the fact that the contact person is in the set of all event participant persons
[...]

Issue #3035, opened 2020-09-30
Information concerning contacts in event participant screens


ftavin:
### Situation Description

In the event participant screens (list of participants + details of an event participant), there are no informations on the fact that the person is or not a contact regarding this particular event.

### Feature Description

This feature consists in three new elements :
- a new column in the list of participant of an event, after the case id, which indicates the number of time the person is a contact involved in this event (could be more than 1 if there are multiples cases in the same event)
- the list of the contacts in the detail of an event participant with the following informations for each :
-> **still to be determined via the issue comments**
- a button in the detail of an event participant that gives the hability to create a contact concerning the current person
-> this contact will be automaticaly considered involved in he current event
-> the cases suggested as "source case" will be limited to the cases of the current event

@carolinverset @xavier-calland @bernardsilenou

bernardsilenou:
The need for this request is similar to issue #2727. Among the list of event participants, users want to know which of the event participant persons are linked to contacts.

bernardsilenou:
> ### Feature Description
> This feature consists in three new elements :
> 
In this first phase of the implementation, we decided to limit the relationship to only entities that belong to the event in question. In a later issue, we would generalize it to include contacts or cases that are not also in the event in question.

> * add a new column in the "event participants page",  after the case id, which indicates the number of time the person is a contact involved in this event (could be more than 1 if there are multiples cases in the same event)

This is computed as: sum of contacts of this event participant person WHERE source case of contact == any of the resulting cases of the event. In a future issue we will drop the where constrain and include all contacts and add a parameter for this condition.

> * the list of the contacts in the detail of an event participant with the following informations for each :
>   -> **still to be determined via the issue comments**

At the right of the event participant form,  we can add an  information box called "Contact" + blue icon on the right hand side called '"Create contact". This is similar to what we have on the case form, that permit us to link the case to an event. 
It will be nice to also have a similar box for the case information, similar to what we have on the "contact information page" for a contact that has a source case. An event participant can be linked to 3 contacts and later result to a case ( from an event participant rather then any of the contacts)
For every method that leads to linking 2 entities, we should also consider to include the option to unlink them. I added a suggested mockup on how to unlink cases from events in issue #2984

I have attached a suggested mockup based on  the new event participant entity mentioned here #3003 

![image](https://user-images.githubusercontent.com/12996724/95140521-65771780-076f-11eb-8594-87b4d9d06904.png)

bernardsilenou:
[...]

Issue #2016, opened 2020-05-08
Trying to sort the sample list by the "edit" column results in an error


MateStrysewskeSym:
The edit column in the sample directory seems to be sortable, but the backend obviously doesn't know how to do that.

- [x] Set sortable to false
- [x] Check the edit columns of all other grids and make sure that those aren't editable as well

```
javax.ejb.EJBTransactionRolledbackException: null
at de.symeda.sormas.api.sample._SampleFacade_Wrapper.getIndexList(de/symeda/sormas/api/sample/_SampleFacade_Wrapper.java)
at de.symeda.sormas.ui.samples.SampleGrid.lambda$setLazyDataProvider$7f6cf0df$1(SampleGrid.java:140)
at com.vaadin.data.provider.CallbackDataProvider.fetchFromBackEnd(CallbackDataProvider.java:137)
...
at com.sun.ejb.containers.BaseContainer.postInvoke(BaseContainer.java:2147)
... 101 more
Caused by: java.lang.IllegalArgumentException: edit
at de.symeda.sormas.backend.sample.SampleFacadeEjb.getIndexList(SampleFacadeEjb.java:284)
at sun.reflect.GeneratedMethodAccessor811.invoke(Unknown Source)
...
at com.sun.corba.ee.impl.presentation.rmi.codegen.CodegenStubBase.invoke(CodegenStubBase.java:226)
at de.symeda.sormas.api.sample.__SampleFacade_Remote_DynamicStub.getIndexList(de/symeda/sormas/api/sample/__SampleFacade_Remote_DynamicStub.java)
at de.symeda.sormas.api.sample._SampleFacade_Wrapper.getIndexList(de/symeda/sormas/api/sample/_SampleFacade_Wrapper.java)
at de.symeda.sormas.ui.samples.SampleGrid.lambda$setLazyDataProvider$7f6cf0df$1(SampleGrid.java:140)
at com.vaadin.data.provider.CallbackDataProvider.fetchFromBackEnd(CallbackDataProvider.java:137)
...
at com.sun.ejb.containers.BaseContainer.postInvoke(BaseContainer.java:2147)
... 101 more
Caused by: java.lang.IllegalArgumentException: edit
at de.symeda.sormas.backend.sample.SampleFacadeEjb.getIndexList(SampleFacadeEjb.java:284)
at sun.reflect.GeneratedMethodAccessor811.invoke(Unknown Source)
...
[...]

PR #3367, opened 2020-11-02
[IMIS] List case and contact counts in event overview #3276

DavidBaldsiefen:
This PR implements #3276 
It would be especially interesting to see how it performs on systems with a large database (many events & contacts)

Edit: #3562 is also implemented with this PR.


MateStrysewske (changes requested):
The approach taken in this PR will lead to massive performance problems because it requires two queries per list entry; one of those even involves multiple additional queries. The correct way to do this is to extend the selection in the getIndexList with counts.

MateStrysewske:
The other option is to do two additional queries after the index list has been retrieved and build a map. That allows a more complex query, but still limits the amount of additional queries to two instead of two per shown event. There are quite a lot examples in the existing getIndexList methods for this approach.

ftavin:
@Seal33 do you know when you will be able to adjust your PR ?

Seal33:
@ftavin I am still on it, and I hope to do so this week.

Seal33:
@bernardsilenou @ftavin @carolinverset

All the base functionalities of #3276 and #3562 are implemented now. I will still have to do some cleanup, but you could try out the changes already and tell me if they behave as expected (especially both calculation methods for contact counts)

Seal33:
@carolinverset @bernardsilenou 
The PR is ready for merging now. I haven't been able to implement Sorting of the columns yet, as it is a little bit more complicated than one might think (see https://github.com/hzi-braunschweig/SORMAS-Project/issues/2016).
My suggestion would be to move that to another issue, so that we can make sure this PR will be merged for 1.53

Also, another remark would be that one might consider changing the column titles:
- "Participant Count" -> "Participants"
- "Case Count" -> "Cases"
- "Contacts among Participants" -> "Contacts"

This would improve the overall readability of the table. What do you think?

Seal33:
Checks may fail after latest merge as current development branch fails tests.

MateStrysewske:
@Seal33 Just a quick heads-up - reviewing this pull request will take a bit longer because we need to have a deeper look into the queries. We recently had serious performance problems that were caused by some subqueries and we'll have to make sure that we won't run into similar issues with the new subqueries introduced in this PR. It's definitely A LOT better than having an additional query per list entry, but we still need to analyse the performance impact.

Seal33:
@MateStrysewske Thank you for the headsup! Yes, take your time. The performance impact would be especially interesting on systems with large Contact and Event databases.

It would be really nice if this PR would be merged for 1.53

MateStrysewske:
@Seal33 The query that this code results in will very likely have a very negative impact on performance; not necessarily in the directory because that's lazy-loaded by default, but at least for the export. We need a different solution for this. One approach could be to collect the information after the initial query with one (or more) additional queries, as it's done e.g. in the detailed case export. This would result in more than one query, but those queries would ideally be more lightweight and therefore have a less negative impact on performance overall. If you find a different solution, that would also be fine.

This is the resulting query:

```
  select 
  event0_.uuid as col_0_0_, 
  event0_.eventStatus as col_1_0_, 
  event0_.eventInvestigationStatus as col_2_0_, 
  
  (select count(eventparti7_.id) 
  from EventParticipant eventparti7_ 
  where eventparti7_.event_id=event0_.id and eventparti7_.deleted=false) as col_3_0_, 
  
  (select count(eventparti8_.id) 
  from EventParticipant eventparti8_ 
  where eventparti8_.event_id=event0_.id and eventparti8_.deleted=false and (eventparti8_.resultingCase_id is not null)) as col_4_0_, 
  
  (select count(eventparti9_.id) 
  from EventParticipant eventparti9_ 
  left outer join cases case10_ on eventparti9_.resultingCase_id=case10_.id 
  where eventparti9_.event_id=event0_.id and eventparti9_.deleted=false and (eventparti9_.resultingCase_id is not null) and case10_.outcome=?) as col_5_0_, 
  
  (select count(eventparti11_.id) 
  from EventParticipant eventparti11_ 
  inner join Person person12_ on eventparti11_.person_id=person12_.id 
  inner join Contact contacts13_ on person12_.id=contacts13_.person_id 
  where eventparti11_.event_id=event0_.id and eventparti11_.deleted=false and contacts13_.deleted=false) as col_6_0_, 
  
  (select count(contact14_.id) 
  from Contact contact14_ 
  inner join Person person15_ on contact14_.person_id=person15_.id 
  inner join EventParticipant eventparti16_ on person15_.id=eventparti16_.person_id 
  inner join cases case17_ on contact14_.caze_id=case17_.id 
  inner join EventParticipant eventparti18_ on case17_.id=eventparti18_.resultingCase_id 
  where contact14_.deleted=false and eventparti16_.deleted=false and eventparti16_.event_id=event0_.id and eventparti18_.event_id=event0_.id) as col_7_0_, 
  
  event0_.disease as col_8_0_, 
  event0_.diseaseDetails as col_9_0_, 
  event0_.startDate as col_10_0_, 
  event0_.endDate as col_11_0_, 
  event0_.eventTitle as col_12_0_, 
  region2_.uuid as col_13_0_,
  region2_.name as col_14_0_, 
  district3_.uuid as col_15_0_, 
  district3_.name as col_16_0_, 
  community4_.uuid as col_17_0_, 
  community4_.name as col_18_0_, 
  location1_.city as col_19_0_, 
  location1_.street as col_20_0_, 
  location1_.houseNumber as col_21_0_, 
  location1_.additionalInformation as col_22_0_, 
  event0_.srcType as col_23_0_, 
  event0_.srcFirstName as col_24_0_, 
  event0_.srcLastName as col_25_0_, 
  event0_.srcTelNo as col_26_0_, 
  event0_.srcMediaWebsite as col_27_0_, 
  event0_.srcMediaName as col_28_0_, 
  event0_.reportDateTime as col_29_0_,
  user5_.uuid as col_30_0_, 
  user6_.uuid as col_31_0_ 
  
  from events event0_ 
  left outer join Location location1_ on event0_.eventLocation_id=location1_.id 
  left outer join Region region2_ on location1_.region_id=region2_.id 
  left outer join District district3_ on location1_.district_id=district3_.id 
  left outer join Community community4_ on location1_.community_id=community4_.id 
  left outer join users user5_ on event0_.reportingUser_id=user5_.id 
  left outer join users user6_ on event0_.surveillanceOfficer_id=user6_.id 
  
  where (event0_.archived=? or event0_.archived is null) and event0_.deleted=? order by event0_.changeDate desc limit ?]]
```

And this is a screenshot of the query analysis which visualises how the result is built:

query_analysis

MateStrysewske (changes requested):
See comment above

Seal33:
Thank you for the review! I've just now seen a lot of groundwork has been done in parallel with #3035 (PR #3344 )

That means there is now a lot of duplicate code in my approach. I will try to find a solution that uses the changes implemented with PR #3344 ASAP so that the code and approach is more uniform.

Seal33:
@MateStrysewske 
I have moved the calculations to a separate query as asked. However I can not really guarantee or tell if the performance increase would be significant enough. Do we have a test system with demo event-data somewhere?

MateStrysewske:
@Seal33 The new implementation doesn't solve the issue, but basically only shifts it to another place. It's still multiple subqueries with multiple FROMs (most of them using the same table) that will result in a very very complex query that will probably take a lot of time to run through. That's why my suggestion was to move this to individual queries that might be simpler.

However, since these changes have the potential to slow down the system, I'd like to postpone it to the next release (given that we have feature freeze today). We were pretty busy with the hotfix this sprint which is why it took me so long to pick up the review of this PR again, so I'm really sorry for that because we could've otherwise made sure that it can be pushed to this release.

Is that okay for you anyway?

Seal33:
@MateStrysewske Thank you for the quick review, I already thought so. However, I've based my solution on the present approaches in CaseFacadeEjb and EventParticipantEjb.

Could you tell me how/where you created the query analysis from your comment above? Would it help in any way if I move the subqueries into complete standalone-queries, or does that not change a thing as the queries themselves stay the same?

For me it's fine if this PR get's pushed back, however I am not the consumer of this product ;)

Another thing I could offer is create a reduced PR today which just includes the Case and Death Count queries. Those are less complex than the contact queries, and would already offer some value to the health offices. What do you think?

MateStrysewske:
@Seal33 Yeah, that's a possibility, similar to e.g. the EventSummaryInformation in the case export. 

I created the query analysis in pgAdmin. It definitely helps to check whether an individual query is way too complex. I'm no database expert myself though, so I don't know for sure that moving the subqueries to individual queries will help, but I guess it definitely won't make the matter worse. :)

That would be fine by me, yes!

Seal33:
@MateStrysewske I've created PR #3760 , which is essentially this PR reduced to case and death count queries.
The queries themselves are unchanged, however they should be light enough performance-wise.

It would be really great if #3670 can still be merged for 1.53.
Functionality-Wise the PR works as far as I could test it, and all tests passed locally.

Meanwhile, I will try and find a way to calculate the contact counts more efficiently, so that the full feature can finally be implemented for 1.54. If I find a better way I will also adjust the calculation methods for Participant, Case and Deathcount accordingly, too.

Seal33:
@MateStrysewske 

I think i found a much better query which should be massively better for performance. However, while it works flawlessly in pgAdmin, I simply cannot get it to work in the code. Could you tell me how you were able to extract the generated query from the code above?

MateStrysewske:
@Seal33 Open the persistence.xml in sormas-backend and set the property `hibernate.show_sql` to true; that will print every query that's performed to the console.

DavidBaldsiefen:
Closed in favor of #3981 .

PR #3371, opened 2020-11-03
Add an event component in contact data view #3133 

### LINKED ISSUE #3133: Attach a contact to an event (from the contacts directory)
### Situation Description
To date, users can link a case to an event from the case directory. This feature does not exist for contacts.

### Feature Description
Like the cases, develop in the case file a data block to include the list of events in which the contact is involved. 
Same information as the block integrated in the case sheet: description of the event, region, district, municipality, date of the event. I only propose to replace the description of the event with the title of the event which will be available from v 1.50.
All events in which the contact has been involved will be displayed. A check mark will restrict the events to those for which the contact and its zero case is involved.

This block will also integrate a button to attach a contact to an event, or conversely, to break this link (in the event of improper handling for example)

### Additional Information
image

@ftavin @olivierleuci @bernardsilenou

alexandre-gille:
Fixes #3133


alexandre-gille:
The Travis build failed due to a network issue. If someone with the required access doesn't mind to re-trigger it, that'd be much appreciated !

Thank you beforehand

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
```diff
@@ -446,6 +446,14 @@ public EventParticipantReferenceDto getReferenceByUuid(String uuid) {
 		return new EventParticipantReferenceDto(eventParticipant.getUuid());
 	}
 
+	@Override
+	public EventParticipantReferenceDto getReferenceByEventAndPerson(String eventUuid, String personUuid) {
+		Event event = eventService.getByUuid(eventUuid);
+		EventParticipant eventParticipant =
+			event.getEventPersons().stream().filter(ep -> !ep.isDeleted() && ep.getPerson().getUuid().equals(personUuid)).findFirst().orElse(null);
+		return new EventParticipantReferenceDto(eventParticipant.getUuid());
```
MateStrysewske, line 454: I suspect this throws a NullPointerException if no event participant is found above because you explicitely set it to null in that case.

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
```diff
@@ -528,6 +528,16 @@ public Predicate buildCriteriaFilter(EventCriteria eventCriteria, CriteriaBuilde
 
 			filter = and(cb, filter, cb.isFalse(eventParticipantJoin.get(EventParticipant.DELETED)));
 		}
+		if (eventCriteria.getPerson() != null) {
+			Join<Event, EventParticipant> eventParticipantJoin = from.join(Event.EVENT_PERSONS, JoinType.LEFT);
```
MateStrysewske, line 532: I think it would be a good idea to do the join from event to event participant at the start of the method because it's also used when the event criteria has a case, like this:

```
Join<Event, EventParticipant> eventParticipantJoin = null;
if (eventCriteria.getCaze() != null || eventCriteria.getPerson() != null) {
  eventParticipantJoin = from.join(Event.EVENT_PERSONS, JoinType.LEFT);
}
```

That makes sure that we don't do the join twice which would probably slow down the query.

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventController.java
```diff
@@ -207,6 +255,37 @@ public void onCommit() {
...
+		final CommitDiscardWrapperComponent<EventDataForm> editView = new CommitDiscardWrapperComponent<>(
+			eventCreateForm,
+			UserProvider.getCurrent().hasUserRight(UserRight.EVENT_CREATE),
+			eventCreateForm.getFieldGroup());
+
+		editView.addCommitListener(() -> {
+			if (!eventCreateForm.getFieldGroup().isModified()) {
+				EventDto dto = eventCreateForm.getValue();
+				FacadeProvider.getEventFacade().saveEvent(dto);
+				Notification.show(I18nProperties.getString(Strings.messageEventCreated), Type.WARNING_MESSAGE);
```
MateStrysewske, line 273: Show a tray notification instead of a warning message

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventController.java
```diff
@@ -207,6 +255,37 @@ public void onCommit() {
...
+
+				if (contact.getCaze() != null) {
+					EventReferenceDto createdEvent = new EventReferenceDto(dto.getUuid());
+
+					createEventParticipantWithContact(createdEvent, contact);
```
MateStrysewske, line 278: What's the reason for only doing this when the contact has a source case? What about contacts without a source case?

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsView.java
```diff
@@ -225,7 +226,7 @@ public EventsView() {
 			createButton = ButtonHelper.createIconButton(
 				Captions.eventNewEvent,
 				VaadinIcons.PLUS_CIRCLE,
-				e -> ControllerProvider.getEventController().create(null),
+				e -> ControllerProvider.getEventController().create((CaseReferenceDto) null),
```
MateStrysewske, line 229: What's the purpose of adding this cast?

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventController.java
alexandre-gille: Should I change this in the method alter ego for cases too ? This is copy-pasted from it.

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsView.java
alexandre-gille: I overloaded the create method, and therefore he can't choose between the two if the argument isn't typed.

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventController.java
alexandre-gille: Thanks for pointing that out ! That's on me, it's an artifact of when I first c/c the equivalent method for Cases and quickly changed thinks to get a compiling one for Contacts. I forgot to adapt this part. Fixed it !

It's there because it's Cases' method that's used (with null as case param) even for the base event creation (hence the https://github.com/hzi-braunschweig/SORMAS-Project/pull/3371#discussion_r519666907). But it feels a bit cranky. I think I should refactor this and add a proper create method for it if you don't mind it.

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
alexandre-gille: It does ! It didn't take care of it because it there's the same thing with the getReferenceByUuid right above and without really thinking about it I was telling myself it shouldn't happen (that, and maybe lazyness :tongue:). But it is indeed going to happen (another user already deleted the EP in the meantime => :boom:).
I changed that and added a proper error handling.

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
alexandre-gille: This isn't the most sustainable thingy thing. While I was at it, I added an EventJoins class and refactored the uses of event joins.

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventController.java
MateStrysewske: No, you don't have to do that :)

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
MateStrysewske: Great, thank you!

MateStrysewske (changes requested):
The "Only events in which the contact's source case is involved" filter somehow doesn't work properly for me; I have a case for the person "Frodo Baggins" in my system who is also an event participant in an event, and the case is set as the resulting case for that event participant. I also have a contact for the person "Meriadoc Brandybock" in my system, and this person also is an event participant in that event, and its source case is the case of "Frodo Baggins". If I know check the checkbox in the event list in Meriadoc's contact, the event disappears even though Meriadoc's source case is part of the event.

TL;DR - can you check the filter again, please? Everything else seems fine to me!

alexandre-gille:
It's actually because of the EventJoins, we can't combine the two filters using the same join (person & case).
I just reverted the EventJoins commit and it works just fine.

MateStrysewske (approved):

PR #3378, opened 2020-11-03
[IMIS] sticky button panel for long forms (fixes #3377)

### LINKED ISSUE #3377: "Save" and "discard" buttons are hard to find in lengthy forms
### Bug Description
In all forms that are too long to be shown in total in the browser, the user has to scroll to reach the "save" and "discard" buttons.
This makes it unnecessary hard to do quick updates and - especially to unexperienced users - to find these central elements at all.

### Steps to Reproduce
This applies to a number of forms, including those for case data, case person, contact data, contact person and event.

### Expected Behavior
The "save" and "discard" buttons should always be visible. This can be solved via a sticky button panel.

### System Details
* Device: PC
* SORMAS version: 1.51.0-SNAPSHOT
* Android version/Browser:
* Server URL:
* User Role:

syntakker:
This fix makes the commit/discard button panel sticky at the bottom for long forms. "Save" and "discard" buttons are always visible now. This has no visible effect for shorter forms.
image
image

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/CommitDiscardWrapperComponent.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/CommitDiscardWrapperComponent.java
@@ -148,6 +148,8 @@ public class CommitDiscardWrapperComponent<C extends Component> extends Vertical
 		buttonsPanel.setComponentAlignment(commitButton, Alignment.BOTTOM_RIGHT);
 		buttonsPanel.setExpandRatio(commitButton, 0);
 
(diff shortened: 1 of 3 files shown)


syntakker:
Arrow added to indicate wether there is more content to scroll to:
image

syntakker:
@ChristopherRiedel @MateStrysewske How do we proceed here? Do you have further change requests or a better suggestion? Or just "Let'em scroll" and search for the hidden save button for now?
(see discussion #3377)

syntakker:
Won't fix (see discussion #3377).

PR #3395, opened 2020-11-05
Location filters and columns in event directory #2914

### LINKED ISSUE #2914: Authorize all user roles to be able to filter events by region, district, community in  event directory
### Problem Description

Need to have a filter to select the regions in the directories of cases, contacts and events. Only departments are available as a filter

### Proposed Change

Add region filter

### Additional Information
@bernardsilenou

alexandre-gille:
Closes #2914


alexandre-gille:
:stop_sign: This PR shouldn't be merged yet. A commit is missing. This was developed by another dev and I didn't notice it was in two parts, my bad !

alexandre-gille:
:green_circle: This PR is now ready.

sormas-api/src/main/java/de/symeda/sormas/api/location/LocationReferenceDto.java
```diff
@@ -100,4 +100,35 @@ public static String buildCaption(
...
+			}
+			stringBuilder.append(city);
+		}
+		if (!DataHelper.isNullOrEmpty(street)) {
+			if (stringBuilder.length() > 0) {
+				stringBuilder.append(", ");
+			}
+			stringBuilder.append(street);
+		}
+		if (!DataHelper.isNullOrEmpty(houseNumber)) {
```
MateStrysewske, line 120: I know it's the same as in the other buildCaption method, but there shouldn't be a comma in between street and house number if both are present; would be great if you could also fix that for the above method

MateStrysewske (changes requested):
District filter doesn't work properly for region-level users and community filter doesn't properly work for district-level users. Their content is never updated when the region/district filter is selected. The current logic that the district/community filters are pre-selected based on the user's region/district probably doesn't make sense in this case because all filters are displayed (and none of them is pre-selected).

MateStrysewske (approved):

PR #3432, opened 2020-11-09
Warns if the assigned user is missing contact info at task creation/update if notifications are enabled #3273

### LINKED ISSUE #3273: Develop the mechanism that alerts the absence of an email in the task function
### Situation Description
To date, if the sending email notification setting is activated in the properties file, the user designated to perform a task receives an email on the date and time specified in the task.
When, by mistake, the email address of this user is not entered, no error message specifies that the email does not exist. Therefore, user A who created the task thinks he has notified a task to person B but person B does not receive the notification by email. Therefore, there is a risk that the task will not be performed since the user has not been informed.

### Feature Description
It is proposed to develop a mechanism that alerts the person who creates a task on the absence of the email address of the user designated to perform a task. This message could take the following form: "Please note, the user designated to perform the task does not have an email address and, therefore, will not receive a notification."

### Additional Information
@ftavin @bernardsilenou @MateStrysewske @olivierleuci

alexandre-gille:
Warns if the assigned user is missing contact info at task creation/update if notifications are enabled #3273

Fixes #3273


MateStrysewske (changes requested):
* Please also take into account SMS, because users can also be notified via SMS; this of course only makes sense if the SMS_AUTH_KEY and SMS_AUTH_SECRET properties aren't blank (see ConfigFacade); maybe add a `isSmsServiceSetUp` helper method to the config facade; if it returns true, the hint should only be displayed if the user has neither an email address nor a mobile number, and it should then also be extended (=> "... does not have an e-mail address or mobile number provided...")
* The hint should also be displayed when already existing tasks are edited because notifications aren't only sent when a task is supposed to be started, but also when it's due
* Please remember to also add the French translation to Crowdin; else it will be overridden with the default English one because Crowdin seems to have priority when the translations are automatically merged

alexandre-gille:
>     * Please also take into account SMS, because users can also be notified via SMS; this of course only makes sense if the SMS_AUTH_KEY and SMS_AUTH_SECRET properties aren't blank (see ConfigFacade); maybe add a `isSmsServiceSetUp` helper method to the config facade; if it returns true, the hint should only be displayed if the user has neither an email address nor a mobile number, and it should then also be extended (=> "... does not have an e-mail address or mobile number provided...")
> 
>     * The hint should also be displayed when already existing tasks are edited because notifications aren't only sent when a task is supposed to be started, but also when it's due

Done

>     * Please remember to also add the French translation to Crowdin; else it will be overridden with the default English one because Crowdin seems to have priority when the translations are automatically merged

So, we gotta wait till our PR is merged and then add the french translation on Crowdin (We'll need an access), that's it ?

MateStrysewske:
@alexandre-gille You can request access in the Crowdin project (don't know where exactly) and then someone from HZI is going to have a look at it and grant you the access. And yes, all you need to do is to translate the strings again there after the PR is merged, correct. I know that it's a bit annoying; once we have the time for it, we'll have a look at the automatic merging process and see whether we can tweak it.

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskEditForm.java
```diff
@@ -221,6 +228,39 @@ protected void addFields() {
 		});
 	}
 
+	private void checkIfAssigneeEmailIsProvided(UserReferenceDto assigneeRef) {
```
MateStrysewske, line 231: Should be called `checkIfAssigneeEmailOrPhoneIsProvided` now

MateStrysewske (changes requested):
The label that's displayed when both an email address and a phone number are missing is too long for the window and the text is cropped; you should be able to fix this by giving the label a width of 100 %.

alexandre-gille:
@MateStrysewske Done !

MateStrysewske (changes requested):
Hey - looks like you didn't test the change concerning the problems with the label length. You set the width of the surrounding layout, which does nothing; you need to adjust the width of the label instead.

alexandre-gille:
Sorry about that. Done.

MateStrysewske (changes requested):
Looks like the layout also needs to be set to full width for this to work. Wasn't aware of that myself, but I just tested it.

Issue #3100, opened 2020-10-07
Separate follow-up until date calculation configurability for cases and contacts [2]


kwa20:
### Problem Description
Health departments face the problem that the legislator determined different quarantine/isolation periods for cases and contacts. Since this is connected to the follow-up period and its calculation, the configured calculation in the database doesn't fit either the contacts quarantine period or case isolation period. When users try to reduce the follow-up period, they are limited to the minimal configured period. They then either have to extend every follow-up manually for each case or drop the follow-up for every contact after the quarantine end.
### Proposed Change
Separate the configurability for the follow-up until date calculation for cases and contacts (e.g. 10 days for contacts and 14 days for cases). These values can then be adapted depending on what the legislator determines the minimal quarantine or isolation period to be, without giving the user too much freedom.
### Possible Alternatives
Users could also be given the option to overwrite the follow-up until date and set it to any desired time, meaning that they could reduce it to their liking.

bernardsilenou:
@kwa20 Please can we further refine this in any of our next meetings?

kwa20:
@bernardsilenou Can we remove the "refine" label? I think this is fine after our discussion, isn't it?

bernardsilenou:
1. Use the incubation period of the disease to calculate the followup until date as currently in the system. 
2. Add a new variable to disease configuration called " case follow-up duration" and assign it by default to the incubation period of the disease
3 Add a parameter "case follow-up duration" to server configuration and if set to a date other than the incubation period of the disease, then use the server value in all calculations
3. For all followup other than for cases (eg event participant), use the "contact follow-up duration" in the calculations

ChristopherRiedel:
@bernardsilenou  I would suggest to store the case follow-up duration as the modified follow-up duration in the database in the disease configuration.

markusmann-vg:
Split the configurable follow-up default (minimum days) to
a.) cases
b.) contacts
c.) event participants

Candice-Louw:
Step 2: https://github.com/hzi-braunschweig/SORMAS-Project/issues/3243

PR #3435, opened 2020-11-09
#3100: Split follow-up duration

ChristopherRiedel:


BarnaBartha (approved):
looks good
one note is that in my opinion we should rename also the existing followUpDuration to contactFollowUpDuration

bernardsilenou:
> looks good
> one note is that in my opinion we should rename also the existing followUpDuration to contactFollowUpDuration
 
The existing followUpDuration is used by contacts and event participant, that whey its good to have the name generic for now

ChristopherRiedel:
> looks good
> one note is that in my opinion we should rename also the existing followUpDuration to contactFollowUpDuration

We had explicitly decided in the planning not to rename the followupduration to contactfollowupduration.

Issue #3168, opened 2020-10-13
Introduce duplicate case detection when converting contacts into cases [1]


kwa20:
### Bug Description
When converting a contact into a case, while a case for the same person already exists, duplication recognition is not triggered. The result is two cases being created even though these two might describe the same case.
### Steps to Reproduce
1. create a case and then create a contact for the same person (duplication recognition should be triggered when creating the contact and show the person you have created beforehand).
2. Convert the contact into a case and check whether or not duplication recognition is triggered.
3. Check the case directory for a case duplicate.

### Expected Behavior
Converting a contact into a case should trigger duplication recognition if there is a similar case already existing in the database.
### Screenshots
<img width="1048" alt="Screen Shot 2020-10-13 at 10 14 23 PM" src="https://user-images.githubusercontent.com/70317594/95912292-ee5f0600-0da2-11eb-806a-8f6065a2f73b.png">

### System Details
* Device:
* SORMAS version: 1.49.1
* Android version/Browser: Chrome
* Server URL: test1401.sormas-oegd.de
* User Role: admin

kwa20:
As part of this issue, please also add duplication recognition when linking cases to events. So far, event participant duplicates are created when linking cases to events with the same person already existing as an event participant. We might also consider asking for the involvement description since this field stays empty, even though it is a compulsory entry field.

Note that only the event participant ID differs.

<img width="1110" alt="Screen Shot 2020-10-27 at 5 14 27 PM" src="https://user-images.githubusercontent.com/70317594/97330989-27a27600-1879-11eb-8fd6-94d6c79e4b9d.png">
<img width="1105" alt="Screen Shot 2020-10-27 at 5 15 07 PM" src="https://user-images.githubusercontent.com/70317594/97331010-2c672a00-1879-11eb-8e47-b77489b83dfd.png">

bernardsilenou:
@kwa20  Your additional comment would resolve issue #3250

StefanKock:
@sergiupacurariu please continue on this ticket, it is now added to Sprint 96.

kwa20:
I also noticed the following: When an event participant has a sample with a positive test result, the user gets asked if he wants to create a case for that participant. When a similar case already exists, duplicate recognition triggers. When an existing case is selected, the sample doesn't get adopted.

This might also be taken into consideration when contacts are converted to cases that already exist. When there is no case yet, the sample already gets adopted.

StefanKock:
It was already merged

PR #3437, opened 2020-11-09
#3168 - Introduce duplicate case detection when converting contacts i…

sergiupacurariu:
…nto cases


sergiupacurariu:
this ticket was merged with a different pull request (#3537 ) and a different implementation

PR #3442, opened 2020-11-10
#3392 Added duplicate detection for linking cases

### LINKED ISSUE #3392: Add event participant duplicate detection when linkinig a case to an event [0.5]
### Situation Description

Situation: a person is event participant. Now that person also is a case. When you link that case to the event the person already is event participant of, what currently happens is that a new event participant is created and added to the event, even though the person already is event participant. After that the person appears as an event participant twice. Once as only event participant (this one was there first), and once as an event participant related to a case (this one was created when linking the case to the event).

### Feature Description

There should be a duplicate detection that automatically uses the already existing event participant for that person, if any exists.

FredrikSchaeferVitagroup:
Closes #3392


MateStrysewske (changes requested):
I'd suggest a different approach:

* Add PersonReferenceDto to EventParticipantCriteria that can be used to check whether the person of the event participant matches the given person
* Add a dedicated "getEventParticipantUuid" method to EventParticipantFacade that receives an EventReferenceDto and a PersonReferenceDto; the method should then manually build an EventParticipantCriteria with these two properties and then query for the uuid of a possible match (should probably still expect a list as the result and then just return the first result if there is any, because it's possible that there are multiple event participants for the same person in the same event in older servers)
* When checking for an existing event participant in the event, use the new method instead of getIndexList

MateStrysewske:
@FredrikSchaeferVitagroup Pull request #3371 actually adds such a method; you might want to wait with the further development on this PR until that one is merged into development.

FredrikSchaeferVitagroup:
Once more taken be the infinite wisdom of more experiened developers, I implemented the requested changes.
Please have a look...

MateStrysewske (approved):

Issue #3389, opened 2020-11-04
Use Jsoup instead of StringEscapeUtils to sanitize inputs and prevent XSS attacks


Seal33:
### Feature Description
Currently, html inputs and data fields are sanitized using `StringEscapeUtils.escapeHtml4`. 

### Problem Description
However in some cases html formatting is required, and thus certain tags need to be unescaped. This isn't supported by `StringEscapeUtils.escapeHtml4`, and our current local solution implemented with PR #3220 is not extensive enough to include all tags used.

### Proposed Change
Instead of `StringEscapeUtils.escapeHtml4` use `Jsoup.clean` with `Whitelist.basic()` as argument. This allows certain tags while still preventing XSS. See [here](https://jsoup.org/apidocs/org/jsoup/safety/Whitelist.html).

### Additional Information
This approach was discussed in #3193, an issue that would be solved by this change.

Seal33:
As a sidenote, wherever required the Whitelist can be set to induvidual levels of "strictness".
Example whitelists are `none`, `simpleText`, `basic`, `basicWithImages` and `relaxed`. In the future it would also be thinkable to implement custom whitelists.

Issue #3220, opened 2020-10-19
Broken formatting on classification rules


Seal33:
### Bug Description
Formatting of Classification Rules and "Classified-By"-Label is broken. The popup containing classification rules is also needlessly large.

### Steps to Reproduce
1. Go to any Case overview, with a disease which requires Classification (e.g. Cholera, Dengue-Fever)
2. When clicking on an Info-Button for Classification Rules, html is not correctly displayed (see second screenshot)
3. After classifying the case, a linebreak after "Classified by" is missing (see first screenshot)

### Expected Behavior
1. Correct display of html
2. Linebreak after "Classified By"
3. Info-Popup only as large as necessary

### Screenshots
![image](https://user-images.githubusercontent.com/4655486/96448832-3ef7b880-1214-11eb-9e3e-a95743a068a5.png)
![image](https://user-images.githubusercontent.com/4655486/96448861-47e88a00-1214-11eb-8d16-251997abbc32.png)

### System Details
* Device: Ubuntu
* SORMAS version: 1.51.0-SNAPSHOT
* Android version/Browser: Firefox 81.0.2
* Server URL: local
* User Role: admin

Seal33:
@MartinWahnschaffeSymeda @syntakker @JonasCir 

The issue seems to be caused by numerous (redundant) calls of `StringEscapeUtils.escapeHtml4()` in `ClassificationHtmlRenderer.java`. Logically it is only necessary to escape strings on the lowest level, in this case only i18n strings would need to be escaped. However, there are numerous i18n strings containing html-tags themselves.
How should we solve this issue? Two possible ideas would be to

1. Allow some tags (like `<br> <b> <i>`) to be ignored by `escapeHtml()`
2. Disallow usage of tags in i18n. This would require larger code changed in several places

JonasCir:
Ignoring HTML tags in in  the `escapeHtml()`function implies that you cannot anymore rely on it to properly handle all untrusted content. Meddling with HTML-Encoders or writing custom implementations _will_ [fail](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#you-need-a-security-encoding-library) and ultimately leave you open to injection or scripting attacks. 

[...]

Issue #3193, opened 2020-10-15
EVENT : List of actions unreadable (code elements displayed in the middle of the information entered)


carolinverset:
### Bug Description
In the event part, when we use the editor component to formalize the different stages of the investigation around the event, there is a bug on the screen rendering. Pieces of code appear in the middle of the information entered, affecting the readability and understanding of the formalized actions.

### Steps to Reproduce
1. In event, create an action
2. Change the font or underline elements
3. Save the information
4. Look at the list of actions

### Screenshots
![image](https://user-images.githubusercontent.com/65449450/96085292-e7c6b080-0ec0-11eb-8310-f54ec374c9af.png)

### System Details
* Device:
* SORMAS version: 1.50 
* Android version/Browser:
* Server URL: plateforme https://test.sormas.netzlink.com/
* User Role: admin

### Additional Information

@bernardsilenou  @ftavin @olivierleuci @xavier-calland

xavier-calland:
it seems to come from https://github.com/hzi-braunschweig/SORMAS-Project/commit/3e619bb5bab8fd6478f7cc4d010ea3cad1d99326

https://github.com/hzi-braunschweig/SORMAS-Project/blob/3e619bb5bab8fd6478f7cc4d010ea3cad1d99326/sormas-ui/src/main/java/de/symeda/sormas/ui/action/ActionListEntry.java#L81-L85

MateStrysewske:
@xavier-calland You're right. We'll have to discuss how to handle this specific situation in terms of XSS prevention.

Seal33:
This seems to be related to #3220 . The problem is likely that the html is escaped at multiple levels. A helper class `htmlhelper.java `has been added with PR #3240 , which allows the programmer to explicitly unescape specific tags (like `<b>` and `<i>`). Calling this would probably be enough to have the html displayed correctly again. I can test this solution.

Seal33:
I've looked closer at the issue, is there a spefic reason why users even get the ability to format Event Action descriptions to such a degree? I haven't seen anything like it and don't really understand why selecting fonts, background colour and text size is necessary here. It seems a little out of place @MateStrysewske

ftavin:
@Seal33 the user story for actions is here : #2434

ftavin:
@bernardsilenou @MateStrysewske do you still plan to have this bug corrected for 1.51 ? (I'm not sure why you added the "fr" tag to the issue)

Seal33:
[...]

PR #3448, opened 2020-11-11
[IMIS] Use Jsoup instead of StringEscapeUtils to sanitize inputs and prevent XSS attacks #3389

Seal33:
More info can be found in #3389 and #3193 

Jsoup allows the use of custom whitelists to allow certain tags, while still preventing XSS attacks reliably. This is a cleaner and more standardized approach than that of PR #3220 , which will be reverted by this PR. Jsoup was already used in several spots throughout the code.


MateStrysewske:
@Seal33 Please have a look at #3464; I had to create that issue because apparently `Jsoup.clean` can't deal with null values. I guess that we could and would run into a lot of problems with the current implementation in this PR; my suggestion would be to add some kind of a wrapper method to the API that calls Jsoup.clean if the given string is not null or just returns null again otherwise, and use that everywhere where you've added the call to Jsoup.clean.

Seal33:
> @Seal33 Please have a look at #3464; I had to create that issue because apparently `Jsoup.clean` can't deal with null values. I guess that we could and would run into a lot of problems with the current implementation in this PR; my suggestion would be to add some kind of a wrapper method to the API that calls Jsoup.clean if the given string is not null or just returns null again otherwise, and use that everywhere where you've added the call to Jsoup.clean.

Oh, I didn't know that, I'm sorry we didn't catch that during our internal review. I think a wrapper method is a good idea. This would also allow us to create predefined Whitelists in the wrapper.

syntakker:
@MateStrysewske I think it would make sense to do the sanitizing in the getters and setters of the Dto classes. This way sanitizing would be implemented in one central place and cannot be forgotten in the UI.

MateStrysewske:
@syntakker That approach could make sense, although it would require developers to remember to add the sanitisation to every getter and setter they're adding. @MartinWahnschaffeSymeda do you have a preference here?

MateStrysewske:
@syntakker @Seal33 Discussed this with Martin and we came to the conclusion that doing the sanitising in the getters and setters could make sense, but it's too big of a refactoring to be considered right now. We'll got with the solution that you've already implemented for now. I'll review this pull request after the next release.

sormas-api/src/main/java/de/symeda/sormas/api/utils/HtmlHelper.java
```diff
@@ -1,35 +1,53 @@
...
-		res = res.replaceAll("&lt;/ul&gt;", "</ul>");
-		// <li> Tags
-		res = res.replaceAll("&lt;li&gt;", "<li>");
-		res = res.replaceAll("&lt;/li&gt;", "</li>");
-		// <br> Tags
-		res = res.replaceAll("&lt;br&gt;", "<br>");
-		res = res.replaceAll("&lt;/br&gt;", "<br>");
-		res = res.replaceAll("&lt;br/&gt;", "<br>");
+	public static String cleanHtml(String string) {
+		if (string == null)
```
MateStrysewske, line 30: We're using brackets even for one-line statements; please refactor your code accordingly; also, you could use a shorter syntax for the return statements in this class (e.g. `return string == null ? "" : Jsoup.clean(string, Whitelist.none());`

MateStrysewske (changes requested):

MateStrysewske:
@Seal33 Can you solve the merge conflicts again? I'll approve the PR afterwards.

Seal33:
@MateStrysewske The merge conflicts are now resolved. However, some tests fail during building. This is most likely caused by development failing tests right now, but I cannot guarantee that the merge itself did not introduce any new problems too. The branch did pass all tests before merging today.

MateStrysewske:
@Seal33 Should be fine. We'll run manual tests on our system anyway. I don't think that anything that you've done in this PR should result in test failures.

Sorry, you'll have to do some conflict resolving again - I'll try to merge the PR asap afterwards so that you don't have to do it another time.

Seal33:
@MateStrysewske Done. I also ran the tests locally with 0 failures.

MateStrysewske (approved):

PR #3454, opened 2020-11-11
#2993 countries import

### LINKED ISSUE #2993: Import of countries should be possible [5]
### Situation Description
Lists from RKI, BAG, WHO will be provided
From these examples we should create a template which can be imported as e.g. the facilities can be imported right now.

vlad-ciucescu:
closes #2993


sormas-api/src/main/resources/countries_ar.properties
```diff
@@ -0,0 +1,179 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+###############################################################################
+
+# Descriptions used to detail captions
+
+# CaseData
```
MartinWahnschaffeSymeda, line 21: @vlad-ciucescu Guess this file needs to be removed.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1747,6 +1751,9 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				getDao(DiseaseConfiguration.class).executeRaw("UPDATE diseaseConfiguration SET caseFollowUpDuration = followUpDuration;");
 				getDao(DiseaseConfiguration.class).executeRaw("UPDATE diseaseConfiguration SET eventParticipantFollowUpDuration = followUpDuration;");
 
+			case 242:
+				currentVersion = 242;
+				getDao(Country.class).executeRaw("ALTER TABLE country ADD COLUMN archived SMALLINT DEFAULT 0;");
```
MartinWahnschaffeSymeda, line 1756: @vlad-ciucescu I think the actual creation of the country table is missing when upgrading an existing mobile app. This needs to be added here. No idea why you have added the archived column to the table here. Maybe this is related to a state you had during development?

sormas-ui/src/main/java/de/symeda/sormas/ui/configuration/AbstractConfigurationView.java
```diff
@@ -71,6 +72,11 @@ public void refreshMenu(SubMenu menu, Label infoLabel, Label infoLabelSub, Strin
 					null,
 					false);
 			}
+			menu.addView(
```
MartinWahnschaffeSymeda, line 75: Move this one up - between outbreaks and area

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -5552,4 +5552,21 @@ UPDATE diseaseconfiguration SET casefollowupduration = followupduration;
...
 
 INSERT INTO schema_version (version_number, comment) VALUES (275, 'Split follow-up duration #3100');
+CREATE TABLE country (
+    id bigint NOT NULL,
+    uuid varchar(36) not null unique,
+    creationdate timestamp without time zone NOT NULL,
+    changedate timestamp not null,
+    archived boolean not null default false,
+    defaultname varchar(255),
+    externalid varchar(4) unique,
```
MartinWahnschaffeSymeda, line 5562: The externalId should not be limited to 4 characters. Use 255 instead

sormas-backend/src/main/resources/sormas_import_all_countries.csv
```diff
@@ -0,0 +1,202 @@
+## All WHO countries
+"defaultName","externalId","isoCode","unoCode","archived"
```
MartinWahnschaffeSymeda, line 2: This won't work on systems that use another csv delimiter (e.g. ";").
You will probably have to explicitly define the delimiter when importing this.

MartinWahnschaffeSymeda (changes requested):
@vlad-ciucescu Overall it looks really great. There are a lot of parts I would simply have forgotten.

I found a few things and added comments to the related code sections.

One important thing: The externalId is currently limited to 4 characters, but should allow 255.

When importing the file below, I ran into this problem. In this case the imported did not add an error to the error template though, but just crashed.

I'm not sure what is the best way to deal with this. I think whenever saving an imported row fails we should probably catch the related exception and put it into the error result file for that row.

This is also relevant for the other fields. E.g. it's likely that a user mistakenly enters an iso code that has more than 3 letters. 


[countries.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/5535459/countries.txt)

sormas-api/src/main/resources/countries_ar.properties
vlad-ciucescu: Copy paste error oopsie

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
vlad-ciucescu: I don't know what happened here either. I will add the table.

sormas-ui/src/main/java/de/symeda/sormas/ui/configuration/AbstractConfigurationView.java
vlad-ciucescu: But inside the `INFRASTRUCTURE_VIEW `block?

MartinWahnschaffeSymeda: Yes

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
vlad-ciucescu: Update: I did this like for the Region table
the table is created at line 253 `TableUtils.createTable(connectionSource, Country.class);`
and then I added the `archived` column, like it was added for Region at line 1219

sormas-api/src/main/java/de/symeda/sormas/api/region/CountryDto.java
```diff
@@ -0,0 +1,76 @@
...
+	public static final String I18N_PREFIX = "Country";
+	public static final String DEFAULT_NAME = "defaultName";
+	public static final String EXTERNAL_ID = "externalId";
+	public static final String ISO_CODE = "isoCode";
+	public static final String UNO_CODE = "unoCode";
+
+	private String defaultName;
+	private String externalId;
+	@Size(min = 3, max = 3)
+	private String isoCode;
```
MartinWahnschaffeSymeda, line 21: We should also allow 2-letter iso codes

MartinWahnschaffeSymeda (changes requested):

MartinWahnschaffeSymeda:
@vlad-ciucescu You can merge this after the one change I requested.

sormas-backend/src/main/java/de/symeda/sormas/backend/region/CountryFacadeEjb.java
```diff
@@ -0,0 +1,223 @@
...
+
+	@Override
+	public String saveCountry(CountryDto dto) throws ValidationRuntimeException {
+		if (StringUtils.isBlank(dto.getIsoCode())) {
+			throw new EmptyValueException(I18nProperties.getValidationError(Validations.importCountryEmptyIso));
+		}
+
+		Country country = countryService.getByUuid(dto.getUuid());
+
+		if (country == null
```
MartinWahnschaffeSymeda, line 126: @vlad-ciucescu It seems unnecessary and unperformant to do another query to the DB here. There are DB constraints in place that should do the check.

MartinWahnschaffeSymeda (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/region/CountryFacadeEjb.java
vlad-ciucescu: Normally I would agree but

- we will at most have ~200 countries to import, unlike cases for example where there can be thousands. When there are so few I think the impact will be insignificant.
- this is also done in other `save` methods, for example for regions or even in cases, in the `validate` method
- this is also useful for specifically throwing the `ValidationRuntimeException` which can then be caught in the Importer. Otherwise, it gets much clumsier with the `TransactionRolledbackException` or whatever is thrown.

I would prefer to leave it like this.

PR #3467, opened 2020-11-12
Feature 3368 sub tab

### LINKED ISSUE #3368: Add second hierarchy layer for campaign chart tabs [3]
### Situation Description

The campaign dashboard currently allows to group charts in different tabs. Based on feedback we realized that we will need another hierarchy layer to do this grouping.

### Feature Description

- [x] Add on optional subTabId to the CampaignDashboardElement
- [x] When a sub tab is defined for a tab, the element shown in the picture below should be displayed (it's the SubMenu we are using to select the disease sub-dashboard on the surveillance dashboard)
- [x] add the field to the campaign edit form
- [x] The CampaignDashboardDataProvider should only load the actual diagram data for the active tab / sub-tab (currently it seems to be loading everything whenever it's opened).

grafik

BarnaBartha:
Closes #3368


sormas-backend/src/main/java/de/symeda/sormas/backend/campaign/CampaignFacadeEjb.java
```diff
@@ -222,6 +232,18 @@ private void validate(CampaignDto campaignDto) {
 							.getValidationError(Validations.campaignDashboardChartValueNull, CampaignDashboardElement.TAB_ID, campaignDto.getName()));
 				}
 
+				if (cde.getSubTabId() == null || cde.getSubTabId().isEmpty()) {
+					if (oneSubTabIsNotNullOrEmptyMap.get(cde.getTabId())) {
```
MartinWahnschaffeSymeda, line 236: The check doesn't work like that, when only the last sub tab id is not empty. It would go through the previous ones and be ok, because until then all sub tabs are empty and only for the last row the sub tab id would be added to the map.

MartinWahnschaffeSymeda (changes requested):
The changed logic in the CampaignDashboardDataProvider is great and I only had one finding in the code (see my other comment).

Besides that I ran into a few bugs, when testing this. I'm not sure whether those bugs are new or have been there before.

- [ ] Edit an existing campaign
Change something about the campaign data charts, so validation fails, when saving (eg. empty tab name)
Fix the problem and try to save again
"The data could not be saved, because it has been changed in the meantime"
- [x] Adding an empty row to the campaign data forms and then saving just gives an error, no clear message
- [x] Try to add the same campaign data form twice to the campaign -> leads to an error without explanation on what is wrong
- [ ] When I add two rows to the campaign data forms and try to define them, something strange happens. It looks like one of the row gets selected (blue background) and then I can't delete them anymore and going to the dropdown of the other row pre-selects the value that I have selected for the other row.

Also I'd suggest to 

- [x] Give the new campaign window the same width as the edit view
- [x] put the sub tabs menu into a full width layout with style class highlighted-statistics-component

Issue #3294, opened 2020-10-27
[External Journal] Inform External Journal about canceled or completed follow-ups [2]


FredrikSchaeferVitagroup:
### Situation Description

SORMAS has a followUpStatus. This can for example be set to CANCELLED. If SORMAS is used alongside climedo, the latter probably needs to be informed about such a change. There should not be a reason to send further questionnaires. This might also be true if a contact is converted to a case, but this is not confirmed yet.

### Feature Description

The idea is to add a followUpStatus to the JournalPersonDto. It can be
```
FOLLOW_UP,
COMPLETED,
CANCELED,
LOST,
NO_FOLLOW_UP; 
```
This should be FOLLOW_UP if any contact of the related person is has this status. 
If there are just CANCELLED contact follow ups for the person, the contactStatus should also be CANCELLED. The same applies for any other possible follow up status values.
If there are various follow up status values for one person, but none of  them is FOLLOW_UP, the status for the external symptom journals should be NO_FOLLOW_UP.
If a changed contactStatus is saved, SORMAS should check if a notification of a related climedo instance is necessary and trigger it in case.

This feature should be verified with HZI before implementation. Also climedo should be informed about this.

FredrikSchaeferVitagroup:
In the contactView, there also is a button to cancel follow up. The same as above should be considered for this button.

FredrikSchaeferVitagroup:
There is a followUpStatus for Contacts and for Cases.  It's an enum with the following values:
`FOLLOW_UP,
COMPLETED,
CANCELED,
LOST,
NO_FOLLOW_UP;`
We should use this to synchronize with climedo, and make sure that the followUpStatus is updated in any reasonable situation 
- [ ] when a Contact is dropped
- [ ] when a Contact is converted to a Case in a system that has Case follow up disabled
- [ ] when a Case classification is set to 'Not a case' in a system with Case follow up enabled. Here, we need to make sure that we don't unreasonably cancel Person's follow up. Imagine a Person has a Contact. The Contact is the converted to a Case. Because of that, the Contact follow up is disabled. Now the Case classification is set to 'Not a case' which disables the person's case follow up as well. We should not have a Person who had a valid (yet converted) Contact and no follow up at all.

FredrikSchaeferVitagroup:
Climedo has something like a followUpStatus, as well as a way to update it via API. They sent it to us.
[...]

PR #3470, opened 2020-11-12
#3265 Add CLIMEDO functionalities to cases

### LINKED ISSUE #3265: [CLIMEDO Integration] Add the CLIMEDO Functionality to cases
### Situation Description
Currently the symptom diary can only be used for contacts. With this story we extend the functionality to cases

### Feature Description
Add the functionality of creating users, sync personal data and endfollowupdate, and receive data from Symptom Journal for cases and the follow-up for cases

vlad-ciucescu:
closes #3265 
**NOTE:** uses commit from branch `feature-3294-Inform-External_Journal-about-canceled-or-completed-follow-ups`. This should probably be merged after #3294 is merged.


sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -392,8 +399,82 @@ public Date getLatestFollowUpEndDateByUuid(String uuid) {
 		cq.orderBy(cb.desc(contactRoot.get(Contact.FOLLOW_UP_UNTIL)));
 
 		List<PersonFollowUpEndDto> resultlist = em.createQuery(cq).getResultList().stream().distinct().collect(Collectors.toList());
-		return resultlist.get(0).getLatestFollowUpEndDate();
```
FredrikSchaeferVitagroup, line 395: This method needs to also consider FollowUpEndDates of cases here, if i'm not mistaken...

FredrikSchaeferVitagroup (dismissed):

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
vlad-ciucescu: So should the latest follow up end date from both contacts and cases related to the person be returned?
What about the method above, `getLatestFollowUpEndDates`? Should we return a list with dates combined from both cases and contact?

FredrikSchaeferVitagroup: Yes, good point. Each of those methods should return the latest follow up end date for persons, no matter if they come from a contact or a case.

FredrikSchaeferVitagroup:
If we add cases to external journals, this should be reflected in unit tests, for example because they change logic for latestFollowUpEndDates of persons. Relating tests are implemented in #3195.

vlad-ciucescu:
> 
> 
> If we add cases to external journals, this should be reflected in unit tests, for example because they change logic for latestFollowUpEndDates of persons. Relating tests are implemented in #3195.

I updated the tests in `PersonFacadeEjbTests` which were related to the followup date. The tests in `ExternalJournalServiceTest` do not contain any reference to contacts.

FredrikSchaeferVitagroup:
The code looks good to me now. Please resolve merge conflicts.

FredrikSchaeferVitagroup:
I talked to the climedo developers about it, they said this issue needs further refinement.

FredrikSchaeferVitagroup:
In a diskussion with climedo it came out that they changed their app. It seems that it happens to just fit to what we implemented here (they removed the different study for cases, all cases and contacts use the same study now).
Please update the branch to the latest state of development!

vidi42 (approved):

PR #3476, opened 2020-11-12
[IMIS] Fix #3462 by adding onsetDate caption

### LINKED ISSUE #3462: Faulty error message when symptom onset date is set in the future
### Bug Description
When symptom onset date is set in the future, error message mentions 'null'.
### Steps to Reproduce
1. Create a new case.
2. Try saving a new case with the symptom onset date set in the future.

### Expected Behavior
Error message should say 'Date of symptom onset cannot be in the future'.
### Screenshots
Screen Shot 2020-11-11 at 7 26 42 PM

### System Details
* Device:
* SORMAS version: 1.51.0
* Android version/Browser: Chrome, Firefox

Seal33:
Description of the solution can be found in #3462 , in case identical issues pop up anywhere else


MateStrysewske (changes requested):
I'm not really happy with this solution. The onset date is not part of the case, and simply adding a caption for the case therefore isn't appropriate. The solution should rather make sure that the validation can handle the caption of fields that are not part of the main DTO of the form. This would also help in other places where this could be a problem as well.

Seal33:
> The solution should rather make sure that the validation can handle the caption of fields that are not part of the main DTO of the form. This would also help in other places where this could be a problem as well.

Yes, I've thought about that too. However, I am not exactly sure how to go on about that.
For example if the code can't find `caseData.onsetDate` nor just `onsetDate`, how should it know whicht onsetDate it is supposed to choose instead? In this case it would be easy, as `onsetDate` only appears in one other part (`Symptoms.onsetDate`). However that's not always the case (for example, `Person.firstName` and `Contact.firstName` are very different strings)

The only solutions I see would involve somewhat larger code changes (for example adding additional parameters to "addField").
If you have any other idea please tell me, I will try to come up with something better too. It would indeed be nice to have a more general solution that might work in other places too, however one has to be very careful not to break all the other code parts involving adding Fields and auto-captioning them.

Seal33:
I have overloaded the `addCustomField` method with an additional parameter that allows the user to set custom captions. For now, this is by far the easiest and 'safest' solution, which can easily be applied anywhere the same issue arises.

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/AbstractEditForm.java
```diff
@@ -240,6 +240,29 @@ protected ComboBox addInfrastructureField(String fieldId) {
 		field.setWidth(100, Unit.PERCENTAGE);
 	}
 
+	@SuppressWarnings("rawtypes")
+	@Override
+	protected <T extends Field> void formatField(T field, String propertyId, String customCaption) {
```
MateStrysewske, line 245: A lot of duplicated code; this method can stay as it is, but the already existing formatField method should be changed. Basically remove everything but the line that builds the caption and then use that to call the new method:

```
formatField(field, propertyId, I18nProperties.getPrefixCaption(propertyI18nPrefix, propertyId, field.getCaption()));
```

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
```diff
@@ -163,8 +163,7 @@ protected void addFields() {
 		ComboBox presentCondition = addCustomField(PersonDto.PRESENT_CONDITION, PresentCondition.class, ComboBox.class);
 		presentCondition.setCaption(I18nProperties.getCaption(Captions.Person_presentCondition));
 
-		DateField onsetDate = addCustomField(SymptomsDto.ONSET_DATE, Date.class, DateField.class);
-		onsetDate.setCaption(I18nProperties.getCaption(Captions.Symptoms_onsetDate));
+		addCustomField(SymptomsDto.ONSET_DATE, Date.class, DateField.class, I18nProperties.getCaption(Captions.Symptoms_onsetDate));
```
MateStrysewske, line 166: Use `I18nProperties.getPrefixCaption` instead

MateStrysewske (changes requested):
I like that solution! Two more change requests, but the approach is totally fine.

PR #3484, opened 2020-11-14
Update SERVER_CUSTOMIZATION.md

victorricardo:
Translater to spanish


MateStrysewske:
Hey! SORMAS is an international project, and thus English is the language we're using throughout this repository, including the guides. Replacing the existing English guide with a Spanish one is not an option at all, and we'll also not add translated versions to the repository in general because we'd have to do that for a lot of languages.

PR #3490, opened 2020-11-16
Add new tasks columns and filters  #3279 #3278

### LINKED ISSUE #3278: Develop additional filters in the TASKS Management
### Situation Description

The task directory has very few filters and columns. Its use is currently limited: it’s for example not possible to filter the tasks over a given period, neither on a given region, nor on the SORMAS environment concerned by the task (case / contact / event).
It is proposed to complete the existing filters on the task directory in order to facilitate navigation and the identification of tasks relating to a person, to an event, to a given spatio-temporal area.

### Feature Description

The additional filters are:
-	Dates or period of the task 
-	Region / department of the case / contact / event concerned by the task
-	SORMAS environment concerned by the task: case/contact/event/general
-	Case or contact ID search / Event title
-  "Assigned to" and "Created by" 

### Possible Alternatives
A free-text filter that allows searching for keywords in/across all fields. 

### Additional Information

@ftavin @olivierleuci @bernardsilenou

### LINKED ISSUE #3279: Add additional variables and filters to task directory
### Situation Description

To date, the number of columns available in the task dashboard is quite limited and does not offer reading of information by geographic level for example

### Feature Description

For a better use of the task dashboard, additional columns are needed, namely: 
- The geographical level to which the task refers (region, district) 
- The SORMAS  entity  associated to the task, ie "task context" (case / contact / event / general) 
- The title of the event concerned by the task (empty for others)

### Additional Information

alexandre-gille:
Closes #3279 
Closes #3278

@ftavin


sormas-api/src/main/java/de/symeda/sormas/api/task/TaskCriteria.java
```diff
@@ -26,14 +26,20 @@
...
 
 public class TaskCriteria extends BaseCriteria implements Serializable {
 
 	private static final long serialVersionUID = -9174165215694877625L;
 
+	public static final String CONTEXT_ENTITY_FREE_TEXT = "contextEntityFreeText";
+
 	private TaskStatus taskStatus;
+	private TaskContext taskContext;
+	private String contextEntityFreeText;
```
MateStrysewske, line 42: The name is a bit misleading for me because it's not about the context entity; I'd rather just call it "freeText"

sormas-api/src/main/java/de/symeda/sormas/api/task/TaskIndexDto.java
```diff
@@ -58,6 +59,9 @@
 	public static final String TASK_STATUS = "taskStatus";
 	public static final String TASK_TYPE = "taskType";
 	public static final String CONTEXT_REFERENCE = "contextReference";
+	public static final String REGION = "region";
+	public static final String DISTRICT = "district";
+	public static final String CONTEXT_ENTITY_NAME = "contextEntityName";
```
MateStrysewske, line 64: I don't quite get the purpose of this property (and the associated column in the task grid); the entity's name is already displayed in the "Associated link" column, so it's basically redundant and takes away space for other columns that provide new information. The new information that it provides is the event title, but that doesn't really match the column header ("Associated entity name") and it's really only useful for tasks that are linked to events. For that reason, I'd ask you to remove this column and property. If you need the event title information in the task grid, my suggestion would be to write an additional method that builds an event caption based on the event title and date (removing the disease and event investigation status) and use that in the "Associated link" column.

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskGridFilterForm.java
```diff
@@ -1,27 +1,71 @@
...
+
+			final ComboBox districtField = addDistrictField();
+
+			regionField.addValueChangeListener(e -> {
+				RegionReferenceDto region = (RegionReferenceDto) e.getProperty().getValue();
+				FieldHelper
+					.updateItems(districtField, region != null ? FacadeProvider.getDistrictFacade().getAllActiveByRegion(region.getUuid()) : null);
+			});
+		} else {
+			final ComboBox districtField = addDistrictField();
```
MateStrysewske, line 55: District filter should also only be displayed when the user does not have a district

MateStrysewske (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/task/TaskIndexDto.java
alexandre-gille: Well, that's what i've been asked to do in #3279 
I do agree with you and I already proposed to add the event title to the associated link caption but the idea hasn't been retained.

ftavin: @MateStrysewske I confirm that this new column is the result of the specification process between @carolinverset and @bernardsilenou 

The review step shouldn't question the user story or, at least, the comment should be done in the issue and not the review (because @alexandre-gille is not the product owner and can only do what he was asked to do...)

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskGridFilterForm.java
alexandre-gille: Done

sormas-api/src/main/java/de/symeda/sormas/api/task/TaskIndexDto.java
MateStrysewske: @ftavin The problem is that we currently don't have the capacities to review every issue that is created before external developers are picking it up; because those issues aren't part of our internal plannings, we don't really have another process in place than reviewing what's been done and checking whether it makes sense in the first place at the same time. This is definitely not a good process and we should revise it, but for that we also need the time to think about a better solution - you get the point. ;)

In this case, I've reviewed the pull request and realised that part of the specifications don't make a lot of sense to me which is why I issued the change request. I'll outsource this to the actual issue in the future because that definitely makes sense. In any case, I would still like to see this changed because I'm not really happy with the proposed solution in #3279 concerning the entity name column.

alexandre-gille: It's removed, and the event title now is in the associated link caption.

MateStrysewske (changes requested):
One of the i18n tests is failing; is it possible that you manually updated the Strings.java class? If so, please revert your changes to it and use the I18nConstantGenerator instead.

Also:
* Additional filter rows are shown by default when entering the tasks view even if none of the additional filters have been filled in; dateFilterOption should be ignored when determining whether or not the additional rows are shown

alexandre-gille:
@MateStrysewske Done

MateStrysewske (approved):

PR #3493, opened 2020-11-16
#3174 consider excluded countries when generating import templates

### LINKED ISSUE #3174: Import template for cases and contacts has few fields which are specific to Switzerland. [0.5]
### Bug Description
Import template for cases and contacts has swiss related fields such as "Has covid app" and "covid code generated".

### Steps to Reproduce
1. Login to sormas as admin
2. Select Cases and click on import at the top right
3. On the new window displayed, select download import template and look for the fields "has covid app" and "covid code generated" on the downloaded csv

### Expected Behavior
The swiss specific fields should not be displayed in the import template for international server.

### Screenshots
import template_contacts
import template_cases

### System Details
* Device:
* SORMAS version: 1.50.0-snapshot
* Android version/Browser: chrome and Firefox
* Server URL: https://test.sormas.netzlink.com/
* User Role: admin

vlad-ciucescu:
closes #3174

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/importexport/ImportFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/importexport/ImportFacadeEjb.java
@@ -66,6 +66,7 @@ import javax.inject.Provider;
 import de.symeda.sormas.api.CountryHelper;
 import de.symeda.sormas.api.utils.HideForCountries;
 import de.symeda.sormas.api.utils.HideForCountriesExcept;
+import de.symeda.sormas.api.utils.fieldvisibility.checkers.CountryFieldVisibilityChecker;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/importexport/ImportFacadeEjb.java
```diff
@@ -439,7 +442,23 @@ private void appendListOfFields(List<ImportColumn> importColumns, Class<?> clazz
 				continue;
 			}
 
-			Method readMethod = null;
```
vidi42, line 442: Instead of writing a new implementation for this, maybe you could use the `CountryFieldVisibilityChecker` which is already available in the `sormas-api` project

vidi42 (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/importexport/ImportFacadeEjb.java
vlad-ciucescu: nice

vidi42 (approved):

PR #3509, opened 2020-11-18
2644 - Enable the usage of population data in campaign diagrams

danRosca:


sormas-backend/src/main/java/de/symeda/sormas/backend/campaign/data/CampaignFormDataFacadeEjb.java
```diff
@@ -304,29 +318,169 @@ public CampaignFormDataReferenceDto getReferenceByUuid(String uuid) {
 		return campaignFormDataService.getAllActiveAfter(date).stream().map(c -> convertToDto(c)).collect(Collectors.toList());
 	}
 
+	public List<CampaignDiagramDataDto> getDiagramDataByAgeGroup(
```
BarnaBartha, line 321: i think this is not ok - you are adding infrastructure and population information into the CampaignDiagramDataDTO.. this should contain the data entered by users for a campaign
the population data should be used only for % charts to know what 100 % is
@MartinWahnschaffeSymeda can you please shed some light on this ?

BarnaBartha (changes requested):

PR #3525, opened 2020-11-18
#2409fusion doublon merge

mouloud-mala:


benchaabaner:
Hi @MateStrysewske , this PR is the same as #2884 which is in a weird state. We created a new one.

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseCriteria.java
```diff
@@ -111,6 +113,7 @@
 	private FacilityTypeGroup facilityTypeGroup;
 	private FacilityType facilityType;
 	private Boolean includeCasesFromOtherJurisdictions = Boolean.FALSE;
+	private ContactReferenceDto contact;
```
MateStrysewske, line 116: Is this actually used?

sormas-api/src/main/resources/strings.properties
```diff
@@ -521,15 +524,15 @@ infoUploadDocumentTemplate=Select a ".docx"-Document Template you would like to
...
+infoMergingExplanation = This view is designed to assist you in detecting and merging duplicate entities. entity are always displayed as pairs, the entity with the earlier creation date being on top.<br/><br/>You can use the creation date filter to only view potential duplicates that were created in a specific time period. The creation date is automatically generated when a entity is entered into SORMAS and can not be changed by users. Therefore, once you have reviewed all duplicates from a specific time period in the past, you will never have to do it again as there will be no new entities created during this time frame.<br/><br/>Please note that most likely not all displayed entities are actual duplicates. There is the potential for false positives, so <b>please thoroughly review the entities before merging them!</b> In addition to the information provided in the table, you can also click on the IDs of the entities to open them in a new tab. Keep in mind that it is much worse to falsely merge unique entities together than to have some duplicates in the system.
```
MateStrysewske, line 527: Error in this string:

[...] "entities. entity are always displayed as pairs, [...]" => "entities. Entities are always displayed as pairs,"

sormas-api/src/main/resources/strings.properties
```diff
@@ -521,15 +524,15 @@ infoUploadDocumentTemplate=Select a ".docx"-Document Template you would like to
...
+infoMergingMergeDescription = Choose this option only if you are <b>sure that the two cases/contacts are the same!</b> You should click on this button next to the entity with the most recent or more complete information. Click on the IDs of the entities to open them in new tabs in order to view all their data. When the entity you chose is missing information that is present in the other entity, it will be updated. Also, contacts, samples, tasks and case management information will be transfered. Afterwards, the case you did not choose will be deleted.
```
MateStrysewske, line 528: You probably have to use separate strings for "infoMergingPickDescription" for contacts and cases because this string refers to contacts being transfered, which obviously isn't the case when contacts are merged.

sormas-api/src/main/resources/strings.properties
```diff
@@ -521,15 +524,15 @@ infoUploadDocumentTemplate=Select a ".docx"-Document Template you would like to
...
+infoMergingPickDescription = Clicking on this button will simply delete the entity that was not picked. However, no information will be copied to the entity that remains active. All contacts or cases, samples and tasks will be lost as well. Choose this option if you are sure that both entities are the same, but the other entity has not been updated with any information that needs to be transfered to the one that remains active.
```
MateStrysewske, line 529: Same problem here. Cases are never lost anyway, and contacts aren't lost when merging contacts because contacts don't have contacts.

MateStrysewske (changes requested):
* Please check all strings and captions in the merge guide again; e.g. there's still the heading "How to merge cases" which should be "How to merge cases/contacts"
* Contacts without source cases are ignored when detecting potential duplicates; I guess you need to extend the cazeFilter to also detect two contacts as duplicates that both don't have a source case

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactListCriteriaBuilder.java
```diff
@@ -297,4 +304,39 @@ public Predicate buildContactFilter(
 
 		List<Expression<?>> forProperty(SortProperty sortProperty, Root<Contact> contact, ContactJoins joins);
 	}
+
+	public List<Selection<?>> getContactIndexSelectionsAll(Root<Contact> contact, ContactJoins joins) {
```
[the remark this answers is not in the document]
mouloud-mala: Yes because it's in relation with the second constructor wich doesn't have some parameters in ContactIndexDto, getCommunity, get(Person.SYMPTOM_JOURNAL_STATUS), get(Contact.CHANGE_DATE) and get(Contact.EXTERNAL_ID)

sormas-api/src/main/resources/strings.properties
mouloud-mala: Are you speaking about "infoMergingPickDescription" or "infoMergingExplanation" ? If it is about "infoMergingPickDescription", it is only used once in MergeGuideLayout for cases and contacts. Can we call different Strings from the same Label depending on the imports.

sormas-api/src/main/resources/strings.properties
mouloud-mala: infoMergingPickDescription = Clicking on this button will simply delete the entity that was not picked. However, no information will be copied to the entity that remains active. All samples and tasks will be lost as well. Choose this option if you are sure that both entities are the same, but the other entity has not been updated with any information that needs to be transfered to the one that remains active.       Is this OK for you ?

MateStrysewske: Yep!

sormas-api/src/main/resources/strings.properties
MateStrysewske: No, there's a passage in "infoMergingPickDescription" that refers to contacts being transfered, which doesn't really make sense if you're merging contacts.

benchaabaner: infoMergingPickDescription = Clicking on this button will simply delete the entity that was not picked. However, no information will be copied to the entity that remains active. All samples and tasks will be lost as well. **Choose this option if you are sure that both entities are the same, but the other entity has not been updated with any information that needs to be transfered to the one that remains active.**    

Are you talking about this last part of the sentence ?

Do we need to have different sentences for cases and for contacts merges ?  This sentence is used once in MergeGuideLayout for cases and contacts. Can we call different Strings from the same Label depending on the imports in MergeGuideLayout ?

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactListCriteriaBuilder.java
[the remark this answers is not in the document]
MateStrysewske: Could you then maybe restructure the constructors in a way that puts the additional variables to the end of it? You then don't need two full getContactIndexSelections methods, but only a method that appends the additional variables to the selections made by the existing getContactIndexSelections method

sormas-api/src/main/resources/strings.properties
MateStrysewske: Oh, I'm sorry! I referenced the wrong property, that's my bad. Have a look at "infoMergingMergeDescription" please, that string includes the passage that I was talking about.

MateStrysewske (changes requested):

MateStrysewske:
Looks like I made some kind of mistake while answering to your questions which resulted in another change request being made. Please just ignore that. My answers are also included in the discussion above.

MateStrysewske:
@mouloud-mala Seems like something went really, really wrong during the merge with development. Basically you've reverted changes that have been made on the development branch, removed new classes etc. I can't review this pull request in this state; please fix these problems and let me know when your branch is in an acceptable state again.

MateStrysewske:
Closing this pull request because it hasn't been updated in a while.

PR #3528, opened 2020-11-18
#3463 update validation javadoc

### LINKED ISSUE #3463: [External Journal] Improve the i18n of the interface [0.5]
More precisely, it happens here (`PersonFacadeEjb`, L. 323, called whenever an existing person is saved):
```java 
	private void handleExternalJournalPerson(PersonDto existingPerson, PersonDto updatedPerson) {
		SymptomJournalStatus status = existingPerson.getSymptomJournalStatus();
		if (SymptomJournalStatus.REGISTERED.equals(status) || SymptomJournalStatus.ACCEPTED.equals(status)) {
			ExternalPersonValidation validationResult = externalJournalService.validatePatientDiaryPerson(updatedPerson);
			if (!validationResult.isValid()) {
				throw new ValidationRuntimeException(validationResult.getMessage());
			}
		}
		// 5 second delay added before notifying of update so that current transaction can complete and new data can be retrieved from DB
		final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
		Runnable notify = () -> externalJournalService.notifyExternalJournalPersonUpdate(existingPerson, updatedPerson);
		executorService.schedule(notify, 5, TimeUnit.SECONDS);
	}
```
In `ExternalJournalService.validatePatientDiaryPerson` the validation of phone numbers appears to be quite strict:
```java
		if (StringUtils.isNotEmpty(email)) {
			EmailValidator validator = EmailValidator.getInstance();
			validEmail = validator.isValid(email);
		}
		if (StringUtils.isNotEmpty(phone)) {
			validPhone = false;
			PhoneNumberUtil phoneUtil = PhoneNumberUtil.getInstance();
			try {
				Phonenumber.PhoneNumber germanNumberProto = phoneUtil.parse(phone, "DE");
				validPhone = phoneUtil.isValidNumber(germanNumberProto);
			} catch (NumberParseException e) {
				logger.warn("NumberParseException was thrown: " + e.toString());
			}
		}
		boolean validBirthdate = true;
```
especially it seems that only valid German phone numbers are accepted?

_Originally posted by @syntakker in https://github.com/hzi-braunschweig/SORMAS-Project/issues/3451#issuecomment-725451992_

vlad-ciucescu:
closes #3463 

I don't know what else I can document here

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
@@ -315,7 +315,8 @@ public class ExternalJournalService {
 
 	/**
 	 * Check whether a person has valid data in order to be registered in the patient diary
-	 * 
+	 * NOTE: since CLIMEDO is only used in Germany, only German numbers are considered valid at the moment
+	 *
 	 * @param person
 	 *            the person to validate
 	 * @return the result of the validation
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
@@ -48,6 +48,15 @@ import de.symeda.sormas.api.person.SymptomJournalStatus;
 import de.symeda.sormas.backend.common.ConfigFacadeEjb;
 import de.symeda.sormas.backend.person.PersonFacadeEjb;
 
+/**
+ * This service provides methods for communicating with external symptom journals.
+ * <br>
+ * Currently, 2 symptom journals are supported: PIA and CLIMEDO.
+ * <br>
+ * Methods containing <code>symptomJournal</code> refer to PIA.
+ * <br>
+ * Methods containing <code>patientDiary</code> refer to CLIMEDO.
+ */
 @Stateless
 @LocalBean
 public class ExternalJournalService {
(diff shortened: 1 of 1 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
```diff
@@ -315,7 +315,8 @@ public PatientDiaryRegisterResult registerPatientDiaryPerson(PersonDto person) {
 
 	/**
 	 * Check whether a person has valid data in order to be registered in the patient diary
-	 * 
+	 * NOTE: since CLIMEDO is only used in Germany, only German numbers are considered valid at the moment
+	 *
```
FredrikSchaeferVitagroup, line 319: We could for example state that a PatientDiary is climedo, and a SymptomJournal is a PIA related system. They also have sormas-sb now, but that is, as far as I know, just a variant of PIA...

FredrikSchaeferVitagroup (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
FredrikSchaeferVitagroup: We always write code for SymptomJournals and PatienDiaries, but these two actually just refer to these two systems. For people to understand you comment, it should be clear that climedo is the only PatientDiary so far.

vlad-ciucescu: I added clarifications in a javadoc on the `ExternalJournalService` class

FredrikSchaeferVitagroup (approved):

PR #3531, opened 2020-11-19
Feature 2946 new epi data form

MateStrysewske:


sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -5568,5 +5570,291 @@ ALTER TABLE country OWNER TO sormas_user;
...
+    DROP COLUMN dateofprophylaxis, DROP COLUMN visitedhealthfacility, DROP COLUMN contactwithsourcerespiratorycase, DROP COLUMN visitedanimalmarket, DROP COLUMN camels, DROP COLUMN snakes;*/
+
+UPDATE epidata SET exposuredetailsknown = 'YES' WHERE (exposuredetailsknown IS NULL OR exposuredetailsknown != 'YES') AND (SELECT COUNT(id) FROM exposures WHERE exposures.epidata_id = epidata.id LIMIT 1) > 0;
+
+UPDATE epidata SET changedate = now();
+
+INSERT INTO schema_version (version_number, comment) VALUES (278, 'Epi data migration #2949');
+
+-- 2020-10-21 Set contact with source case known for all existing cases #2946
+UPDATE epidata SET contactwithsourcecaseknown = 'YES' FROM cases WHERE cases.epidata_id = epidata.id AND (SELECT COUNT(id) FROM contact WHERE contact.resultingcase_id = cases.id > 0);
```
sergiupacurariu, line 5857: please review this statement (SELECT COUNT(id) FROM contact WHERE contact.resultingcase_id = **cases.id > 0);**
 the last bracket might not be correct: cases.id > 0 - ?

sormas-api/src/main/resources/enum.properties
```diff
@@ -509,6 +526,14 @@ FollowUpStatus.Desc.FOLLOW_UP=The follow-up process is running
 FollowUpStatus.Desc.LOST=The follow-up process could not be continued because the person was not available
 FollowUpStatus.Desc.NO_FOLLOW_UP=No contact follow-up is being done
 
+GatheringType.PARTY=Party
+GatheringType.RELIGIOUS=Religious Gathering
+GatheringType.MUSICAL=Choir/Singing Club/Orchestra
+GatheringType.OTHER=Other
+
+HabitationType.MEDICAL=Stay in a Medial Institution
```
sergiupacurariu, line 534: Medical :)

sergiupacurariu:
Related to the creation of the "Contacts with source case" in Epi Data. I saw that it is possible to create New Source Contacts with the same person as many times as you want. On saving the Source Contact if the duplicate detection appears and you choose a person from the duplicate list still a new Source Contact is created. Is this a feature because I cannot think of a situation in which you need to add the same person many times as source contact for the same case.

sergiupacurariu:
When a case is accessed by a user from a different jurisdiction the data is pseudonymized. If this user when you go Epi Data tab everything is gray out. The user can see the list of exposures but disappear entirely is this intentional?

sormas-backend/src/main/resources/sql/sormas_schema.sql
MateStrysewske: True, the bracket has to be moved in front of the '>' - thanks for catching that!

PR #3547, opened 2020-11-23
Feature 3253 send sms to cases

BarnaBartha:


sormas-api/src/main/java/de/symeda/sormas/api/user/UserReferenceDto.java
```diff
@@ -26,22 +26,21 @@
 
 	private static final long serialVersionUID = -8558187171374254398L;
 
-	public UserReferenceDto() {
+	private String firstName;
```
MateStrysewske, line 29: Missing pseudonymization

sormas-api/src/main/resources/strings.properties
```diff
@@ -921,3 +921,14 @@ SurvnetGateway.notificationEntryNotSent = Entry could not be sent
 SurvnetGateway.notificationErrorSending = Error when sending entry
 
 warningDashboardMapTooManyMarkers = There are %d places to display and it is possible that your browser will freeze while displaying them.
+
+# Messaging
+sms = SMS
```
MateStrysewske, line 926: Most of the additions here should be moved to captions because they're used as captions. enterSMS is probably the only one that should stay here.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -3079,6 +3083,48 @@ public boolean isCaseEditAllowed(String caseUuid) {
...
+						person.getUuid()));
+			}
+		});
+	}
+
+	@Override
+	public long countCasesWithMissingMessageType(List<String> caseUuids, MessageType messageType) {
+
+		return caseService.count((cb, root) -> {
+			final Join<Object, Object> personJoin = root.join(Case.PERSON, JoinType.LEFT);
```
MateStrysewske, line 3109: Use Case and Person instead of Object

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -3079,6 +3083,48 @@ public boolean isCaseEditAllowed(String caseUuid) {
...
+						"NotificationDeliveryFailedException when trying to notify person about: %s" + "Failed to send " + e.getMessageType()
+							+ " to person with UUID %s.",
+						subject,
+						person.getUuid()));
+			}
+		});
+	}
+
+	@Override
+	public long countCasesWithMissingMessageType(List<String> caseUuids, MessageType messageType) {
```
MateStrysewske, line 3106: * This method will fail if it's called for more than 32,000 cases; use IterableHelper to execute this batched
* Rename the method to "countCasesWithMissingContactInformation" and add a comment to explain what it's supposed to do

sormas-backend/src/main/java/de/symeda/sormas/backend/common/messaging/ManualMessageLogService.java
```diff
@@ -0,0 +1,58 @@
...
+
+	public static final int MANUAL_MESSAGE_LOG_LIMIT = 5;
+
+	public ManualMessageLogService() {
+		super(ManualMessageLog.class);
+	}
+
+	@Override
+	public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<?, ManualMessageLog> from) {
+		return null;
```
MateStrysewske, line 32: Throw UnsupportedOperationException instead (see e.g. TherapyService)

sormas-backend/src/main/java/de/symeda/sormas/backend/common/messaging/ManualMessageLogService.java
```diff
@@ -0,0 +1,58 @@
...
+		final CriteriaBuilder cb = em.getCriteriaBuilder();
+		final CriteriaQuery<ManualMessageLog> cq = cb.createQuery(ManualMessageLog.class);
+		final Root<ManualMessageLog> manualMessageLogRoot = cq.from(ManualMessageLog.class);
+
+		final Subquery<Person> casePersonSubQuery = cq.subquery(Person.class);
+		final Root<Case> caseRoot = casePersonSubQuery.from(Case.class);
+		casePersonSubQuery.where(cb.equal(caseRoot.get(Case.UUID), caseUuid));
+		casePersonSubQuery.select(caseRoot.get(Case.PERSON));
+
+		Predicate filter = createUserFilter(cb, cq, manualMessageLogRoot);
```
MateStrysewske, line 46: Doesn't make sense because the user filter would return null anyway

sormas-backend/src/main/java/de/symeda/sormas/backend/common/messaging/ManualMessageLogService.java
```diff
@@ -0,0 +1,58 @@
...
+		return null;
+	}
+
+	public List<ManualMessageLog> getByCaseUuid(@NotNull String caseUuid, MessageType messageType) {
+
+		final CriteriaBuilder cb = em.getCriteriaBuilder();
+		final CriteriaQuery<ManualMessageLog> cq = cb.createQuery(ManualMessageLog.class);
+		final Root<ManualMessageLog> manualMessageLogRoot = cq.from(ManualMessageLog.class);
+
+		final Subquery<Person> casePersonSubQuery = cq.subquery(Person.class);
```
MateStrysewske, line 41: This seems very unperformant to me, especially because you're selecting the whole person which loads a lot of completely unnecessary data from the database. Why not just use joins to compare the ID of the case person to the ID of the message log person?

On another note, since you've added the message log as a lazy property to the person anyway, you could just use personJoin.get(Person.MANUAL_MESSAGE_LOGS) if I'm not mistaken.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/messaging/MessagingService.java
```diff
@@ -100,20 +105,53 @@ public void sendMessage(User recipient, MessageSubject subject, String messageCo
...
+		}
+	}
+
+	private void sendMessage(
+		String subject,
+		String messageContent,
+		String emailAddress,
+		String phoneNumber,
+		String recipientUuid,
+		final String recipient,
```
MateStrysewske, line 141: recipient = recipientType

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -5899,4 +5899,22 @@ ALTER TABLE person_history ADD COLUMN armedforcesrelationtype varchar(255);
 
 INSERT INTO schema_version (version_number, comment) VALUES (283, 'Add column armedforcesrelationtype #3418');
 
--- *** Insert new sql commands BEFORE this line ***
+-- 2020-11-17 Manually send SMS #3253
+CREATE TABLE manualmessagelog
```
MateStrysewske, line 5903: primary key and sys_period are missing

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/messaging/SmsComponent.java
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		mainLayout.addComponent(new Label(I18nProperties.getString(Strings.enterSMS)));
+
+		smsTextArea = new TextArea();
+		smsTextArea.setWidth(100, Unit.PERCENTAGE);
+		smsTextArea.setRows(4);
+		mainLayout.addComponent(smsTextArea);
+
+		final HorizontalLayout charactersLayout = new HorizontalLayout();
```
MateStrysewske, line 49: Try to use the TextFieldWithMaxLengthWrapper that is used for the event description

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/messaging/SmsComponent.java
BarnaBartha: i also use this count to show the number of messages that is characters / 160 - should i extend this wrapper?

sormas-backend/src/main/resources/sql/sormas_schema.sql
BarnaBartha: ok
i'm not sure the sys_period is needed in this case, why would we need it ?

sormas-backend/src/main/java/de/symeda/sormas/backend/common/messaging/MessagingService.java
BarnaBartha: ok

sormas-backend/src/main/java/de/symeda/sormas/backend/common/messaging/ManualMessageLogService.java
BarnaBartha: rewritten it to select only the id in the subquery

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
BarnaBartha: is it a realistic scenario that the UI will send 32K + selected caseUuids ?
i can do batch counting but that feel really odd to me

MateStrysewske: It's theoretically possible because the bulk edit view is not lazy (since that doesn't work with "select all"); I agree it's not that realistic, but it's a potential bug that we should rule out nonetheless I think

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/messaging/SmsComponent.java
MateStrysewske: Either that or do that independent of this component (e.g. by just adding another line below the field that counts the number of SMS); I leave that for you to decide

sormas-backend/src/main/resources/sql/sormas_schema.sql
MateStrysewske: You're right, we don't need a history table so should not be necessary

MateStrysewske (changes requested):
* The number of characters and messages should be shown as soon as the window is open, not only after the user has started typing
* Remove spacing between number of characters and number of messages
* Remove margin from the main layout in the "Send SMS" popup
* The logic to check whether or not a person has a phone number also has to check for empty/blank entries, not only for null
* Add a confirmation notification when messages have been sent (tray notification)
* Don't display anything related to sending SMS when ConfigFacade.isSmsServiceSetUp is false

MateStrysewske (changes requested):

PR #3576, opened 2020-11-24
Add symptoms onset and sample latest collection date columns in case directory #3272

### LINKED ISSUE #3272: Add the columns "dates of symptoms" and "date of sampling" in the detailed view of cases
### Situation Description

In the context of monitoring clusters, French scientists agree that a cluster is under control when no new case is recorded within 14 days of the last positive case in the cluster. Version 1.50 will have the functionality to access, from the event, the case directory directly filtered only on cases related to the event. Adding the two columns "date of symptom onset" and "date of sample" in the detailed view will allow users to have this information.

### Feature Description
Complete the detailed view of the case directory with these two complementary variables to allow better monitoring and management of clusters.

Variables to display in column in the case directory :

- Symptom onset date = symptoms.onsetDate

- Collection date = Sample.sampleDateTime

### Additional Information

alexandre-gille:
Closes #3272

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseIndexDetailedDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseIndexDetailedDto.java
@@ -30,6 +30,8 @@ public class CaseIndexDetailedDto extends CaseIndexDto {
 	public static final String LATEST_EVENT_ID = "latestEventId";
 	public static final String LATEST_EVENT_STATUS = "latestEventStatus";
 	public static final String LATEST_EVENT_TITLE = "latestEventTitle";
+	public static final String LATEST_SAMPLE_DATE_TIME = "latestSampleDateTime";
+	public static final String SAMPLE_COUNT = "sampleCount";
 
 	@PersonalData
 	@SensitiveData
@@ -53,6 +55,8 @@ public class CaseIndexDetailedDto extends CaseIndexDto {
 	private String latestEventId;
 	private String latestEventTitle;
 	private EventStatus latestEventStatus;
+	private Date latestSampleDateTime;
+	private Long sampleCount;
 
 	private UserReferenceDto reportingUser;
 
@@ -66,7 +70,7 @@ public class CaseIndexDetailedDto extends CaseIndexDto {
 								Integer age, ApproximateAgeType ageType, Integer birthdateDD, Integer birthdateMM, Integer birthdateYYYY, Sex sex,
 								Date quarantineTo, Float completeness, FollowUpStatus followUpStatus, Date followUpUntil, Date changeDate, Long facilityId,
 								String city, String street, String houseNumber, String additionalInformation, String postalCode, String phone,
-								String reportingUserFirstName, String reportingUserLastName, int visitCount, long eventCount) {
+								String reportingUserFirstName, String reportingUserLastName, int visitCount, long eventCount, Date latestSampleDateTime, long sampleCount) {
 
 		super(id, uuid, epidNumber, externalID, personFirstName, personLastName, disease, diseaseDetails, caseClassification, investigationStatus,
 				presentCondition, reportDate, reportingUserUuid, creationDate, regionUuid, districtUuid, districtName, communityUuid,
@@ -83,6 +87,8 @@ public class CaseIndexDetailedDto extends CaseIndexDto {
 		this.phone = phone;
 		this.reportingUser = new UserReferenceDto(reportingUserUuid, reportingUserFirstName, reportingUserLastName, null);
 		this.eventCount = eventCount;
+		this.latestSampleDateTime = latestSampleDateTime;
+		this.sampleCount = sampleCount;
 	}
 
 	public String getCity() {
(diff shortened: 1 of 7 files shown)


sormas-api/src/main/resources/captions.properties
```diff
@@ -340,6 +340,7 @@ CaseData.smallpoxVaccinationReceived=Was a Smallpox vaccination received in the
 CaseData.smallpoxVaccinationScar=Is a Smallpox vaccination scar present?
 CaseData.surveillanceOfficer=Responsible surveillance officer
 CaseData.symptoms=Symptoms
+CaseData.symptomsOnsetDate=Symptoms onset date
```
MateStrysewske, line 343: I'd like you to remove this property and use the already existing `Symptoms.onsetDate` property instead for the column caption (see my comment in CaseGridDetailed.java)

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseListCriteriaBuilder.java
```diff
@@ -102,6 +106,22 @@
 					cb.isFalse(eventCountRoot.get(EventParticipant.DELETED))));
 			eventCountSq.select(cb.countDistinct(event.get(Event.ID)));
 			selectionList.add(eventCountSq);
+
+			// Latest sampleDateTime subquery
+			Subquery<Timestamp> latestSampleDateTimeSq = cq.subquery(Timestamp.class);
+			Root<Sample> sample = latestSampleDateTimeSq.from(Sample.class);
+			Path<Timestamp> sampleDateTime = sample.get(Sample.SAMPLE_DATE_TIME);
+			latestSampleDateTimeSq
+				.where(cb.equal(sample.get(Sample.UUID), joins.getSamples().get(Sample.UUID)), cb.isFalse(sample.get(Sample.DELETED)));
```
MateStrysewske, line 115: Any reason you're using the UUID here and the ID in the count subquery? ID would generally be preferable because it's shorter

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseIndexDetailedDto.java
```diff
@@ -53,6 +56,9 @@
 	private String latestEventId;
 	private String latestEventTitle;
 	private EventStatus latestEventStatus;
+	private Date latestSampleDateTime;
+	private Long sampleCount;
+	private Date symptomsOnsetDate;
```
MateStrysewske, line 61: => symptomOnsetDate

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseGridDetailed.java
```diff
@@ -1,25 +1,31 @@
...
 import de.symeda.sormas.ui.ControllerProvider;
+import de.symeda.sormas.ui.utils.DateFormatHelper;
 import de.symeda.sormas.ui.utils.ShowDetailsListener;
 import de.symeda.sormas.ui.utils.UuidRenderer;
 
 public class CaseGridDetailed extends AbstractCaseGrid<CaseIndexDetailedDto> {
 
 	private static final long serialVersionUID = 3734206041728541742L;
 
+	private static final String LATEST_SAMPLE_DATE_TIME_AND_SAMPLE_COUNT = Captions.CaseData_latestSampleDateTime;
```
MateStrysewske, line 27: Use `I18nProperties.getPrefixCaption`

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseGridDetailed.java
```diff
@@ -94,5 +118,9 @@ protected void initColumns() {
 					value.getBirthdateYYYY(),
 					I18nProperties.getUserLanguage()),
 			new TextRenderer());
+
+		((Column<CaseIndexDetailedDto, Date>) getColumn(CaseIndexDetailedDto.SYMPTOMS_ONSET_DATE))
+			.setRenderer(new DateRenderer(DateFormatHelper.getDateFormat()))
```
MateStrysewske, line 123: use setCaption here together with `I18nProperties.getPrefixCaption` to get the onsetDate caption from Symptoms

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseGridDetailed.java
alexandre-gille: I did it like this because it's been done the same in other grid classes, and it works.
I'm not reluctant to change this at all, but I kinda wanna understand why would it be better to use a random string here and add a .setCaption(I18nProperties.getPrefixCaption(blabla, blabla)) below ?
For the sake of clarity, to remove a little bit of "magic" ?

MateStrysewske: Mainly because it's more consistent. The Captions file is auto-generated, so it can easier break things if the property is renamed. The properties used for getPrefixCaption are manually added by the developers. It's true that we're using Captions.xxx in other grids, but I'd say that's probably the less ideal way.

MateStrysewske (approved):

PR #3585, opened 2020-11-25
log exception in sormas 2 sormas test

leventegal-she:


vlad-ciucescu (approved):
Tested locally, test no longer fails

vlad-ciucescu (approved):

PR #3593, opened 2020-11-26
Feature 3424 contact follow up cancel

BarnaBartha:


sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -303,6 +303,13 @@ public ContactDto saveContact(ContactDto dto, boolean handleChanges) {
...
 			updateContactVisitAssociations(existingContactDto, entity);
 
+			final boolean convertedToCase =
+				(existingContactDto == null || existingContactDto.getResultingCase() == null) && entity.getResultingCase() != null;
+			final boolean dropped = entity.getContactStatus() == ContactStatus.DROPPED;
+			if (dropped || convertedToCase) {
+				entity.setFollowUpStatus(FollowUpStatus.CANCELED);
+			}
+
 			contactService.updateFollowUpUntilAndStatus(entity);
```
MateStrysewske, line 313: Would probably make sense to put this into an else block to make sure that CANCELED isn't overridden again immediately (and because executing the logic would not make sense anymore and just slow down contact saving)

MateStrysewske (changes requested):

PR #3597, opened 2020-11-26
#3372 provide external API info at runtime

### LINKED ISSUE #3372: [External Journal] Provide API information at runtime [0.5]
### Situation Description

Symptom journals like PIA have API relevant information stored in questionnaires. This means those systems could adapt to some API changes without deploying a new software version. For this, they need information from SORMAS.

What the Conventic developers have in mind is the following scenario: a new symptom is added to visits in SORMAS. This change is released so that external symptom journals have the possibility to upload visits with the new symptom. Now users want to do exactly that, but they have to wait for a new PIA version to be released and deployed to their system.

### Feature Description

If SORMAS would provide information about the structure of the ExternalVisitDto (including all of its symptoms) at runtime, users were able to add a new symptom to a questionnaire, independent of the PIA version they use. This would facilitate and speed up the process of reporting new symptoms with external symptom journals.

- [ ] Implement a new SORMAS-external-visits-API GET endpoint to provide all information about all endpoints of the External Visits controller. This information can be extracted from the swagger.json automatically generated a build time. Methods to extract this info are already implemented in the ExternalVisitsResourceTest in #3358.

### Possible Alternatives

We could force PIA to stick to the API versions we provide, and release matching PIA versions. This would keep the above scenario more complicated and slow.

### Additional Information

We can not make SORMAS-external-visits-API and PIA versions totally independent of each other. In other scenarios, e.g. if a non backward compatible major version of the SORMAS-external-visits-API is released, PIA still has to adapt and be deployed accordingly. This feature shall just facilitate the process to adapt PIA to new minor versions of the SORMAS-external-visits-API.

vlad-ciucescu:
closes #3372

### DIFF
--- a/sormas-rest/src/main/java/de/symeda/sormas/rest/external/ExternalRestResourceConfig.java
+++ b/sormas-rest/src/main/java/de/symeda/sormas/rest/external/ExternalRestResourceConfig.java
@@ -1,0 +1,62 @@
+package de.symeda.sormas.rest.external;
+
+import javax.servlet.ServletConfig;
+import javax.ws.rs.ApplicationPath;
+import javax.ws.rs.core.Context;
+
+import io.swagger.v3.jaxrs2.Reader;
+import io.swagger.v3.jaxrs2.integration.JaxrsOpenApiContextBuilder;
+import io.swagger.v3.oas.integration.OpenApiConfigurationException;
+import org.glassfish.jersey.jackson.JacksonFeature;
+import org.glassfish.jersey.server.ResourceConfig;
+import org.glassfish.jersey.server.filter.RolesAllowedDynamicFeature;
+
+import com.google.common.collect.Sets;
+
+import de.symeda.sormas.rest.swagger.SwaggerConfig;
+import io.swagger.v3.jaxrs2.integration.resources.OpenApiResource;
+import io.swagger.v3.oas.integration.SwaggerConfiguration;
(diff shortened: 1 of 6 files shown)


JonasCir:
I love this! Always wanted to build this, but never had the time, thank you! :)

sormas-rest/src/main/java/de/symeda/sormas/rest/external/ExternalRestResourceConfig.java
```diff
@@ -0,0 +1,58 @@
...
+		register(RolesAllowedDynamicFeature.class);
+		register(JacksonFeature.class);
+
+		SwaggerConfig.init();
+
+		Info info = new Info().title("SORMAS external symptom journal API")
+			.description(
+				"The purpose of this API is to enable communication between SORMAS and other symptom journals. "
+					+ "Only users with the role ``REST_EXTERNAL_VISITS_USER`` are authorized to use the endpoints. "
+					+ "If you would like to receive access, please contact the team to have a user set up. "
```
vidi42, line 45: No sure if we want to point users towards the dev team in order to receive access, since most of the time the system administrators are responsible for giving users roles and access.
Maybe rephrase this to something like: "please contact the System Adminsitrator. For technical details use the contact bellow"

vidi42 (approved):

PR #3598, opened 2020-11-26
Potential performance improvements

HolgerReiseVSys:
#3914


HolgerReiseVSys:
I have noticed that in `PersonService::buildSimilarityCriteriaFilter` `cb.literal(name)` is set directly, not a parameter.
Same goes for the other filters. 
Having a reusable query definition could be beneficial to import performance.

MartinWahnschaffeSymeda:
@HolgerReiseVSys unit tests throw exception

> org.jboss.weld.context.ContextNotActiveException: WELD-001303: No active contexts for scope type javax.transaction.TransactionScoped

I will continue to look into this tomorrow, but if you have a quick idea on this, let me know.

PR #3604, opened 2020-11-27
#3411 add salutation to person

### LINKED ISSUE #3411: SurvNet Adaptations - Create new field “Salutation” for persons [0.5]
### Feature Description
Create new field “Salutation” for persons 

New fields (only DE)-
“Anrede”(de-de) | “Salutation” (en) [SurvNet field name: Salutation]
Place new field under the field “name”, it serves as a text block for letters

If possible, the user should be able to create a customized salutation .

Data Type: Dropdown/enum
Values (de-de) | (en)
• Sehr geehrter Herr| dear Sir
• Sehr geehrte Frau | dear Madame
• Sehr geehrte Familie | dear family
• Sehr geehrte Erziehungsberechtigte des Kindes | dear guardian of the child
• Sehr geehrte Dame, sehr geehrter Herr | dear Sir and Madame

**out of scope:**
If possible, the user should be able to create a customized salutation.

leventegal-she:
Closes #3411


vlad-ciucescu (approved):
Everything works ok, both on mobile & the web app. Only the language key needs to be changed

vlad-ciucescu:
After discussion in daily: should this be only for Germany? Should this be available in the app?

bernardsilenou:
Since it does not harm the other systems and app, it would be good to add also. If we do not add and and the app is later used in Germany, then we need to rework the issue again.

FredrikSchaeferVitagroup:
Tests fail. Please fix.

sormas-app/app/src/main/java/de/symeda/sormas/app/person/edit/PersonEditFragment.java
```diff
@@ -164,6 +166,7 @@ public static void setUpLayoutBinding(
 			false);
```
FredrikSchaeferVitagroup, line 167: We have the same problem as in [https://github.com/hzi-braunschweig/SORMAS-Project/pull/3636#discussion_r533942951](https://github.com/hzi-braunschweig/SORMAS-Project/pull/3636#discussion_r533942951). The salutation also shows in the app for non german servers.
We should either find a fix, or use the workaround as in the linked pull request.

FredrikSchaeferVitagroup (changes requested):
What do you think about adding the salutation to the CaseCreateForm and ContactCreateForm? 
In my opinion that is reasonable because the salutation is often handled together with the name of a person, which is often entered in the create forms.
One could also argue that the salutation is often used in the context of addresses, and addresses are not in the create forms, though.

sormas-app/app/src/main/res/layout/fragment_person_read_layout.xml
```diff
@@ -64,6 +64,13 @@
 
             </LinearLayout>
 
+            <de.symeda.sormas.app.component.controls.ControlTextReadField
```
FredrikSchaeferVitagroup, line 67: The fields also shows here for non german servers.

FredrikSchaeferVitagroup (changes requested):

leventegal-she:
> What do you think about adding the salutation to the CaseCreateForm and ContactCreateForm?

Talked with Markus about it, and decided to not to add on the creation forms

sormas-api/src/main/java/de/symeda/sormas/api/person/PersonDto.java
```diff
@@ -119,6 +122,13 @@
 	@PersonalData(mandatoryField = true)
 	@SensitiveData(mandatoryField = true)
 	private String lastName;
+	@HideForCountriesExcept
+	@PersonalData
+	@SensitiveData
+	private Salutation salutation;
+	@PersonalData
+	@SensitiveData
+	private String otherSalutation;
```
syntakker, line 131: What is the intention of property `otherSalutation`? If possible, rename this to clarify which context this property is intended to be used in instead of `salutation` (is it a `customSalutation`?)
If the intention is to allow salutations not covered by the enum, it might make sense to add an additional method here to resolve the salutation, something like
```java
return otherSalutation != null ? otherSalutation : salutation.toString();
```

leventegal-she: @syntakker it is a `customSalutation` used when the salutation id OTHER.

I don't see the use of the additional method you described and we don't have that kind of methods

syntakker: @leventegal-she the conditional statement from my comment is needed whenever a String is rendered from a person's salutation, so I would advocate to implement it in a central place (like it is done in `LocationDto.checkIsEmptyLocation`).

sormas-api/src/main/java/de/symeda/sormas/api/utils/EnumHelper.java
```diff
@@ -0,0 +1,27 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package de.symeda.sormas.api.utils;
+
+public class EnumHelper {
```
syntakker, line 18: :-)
However, the use of this class for salutation still requires the developer to be aware of the mechanics behind salutation.
Would it be an option to move this code to a method like `PersonDto.renderSalutation`?
Or at least wrap it in such a method, in case this class is intended to be used for other purposes, too.

leventegal-she: This method is used for export where we have specific export DTO classes.

I would not create the method `PersonDto.renderSalutation` because it's not needed for now and it doesn't make sense to create unused methods. It can be created later when needed.

syntakker: A method like `PersonDto.renderSalutation` would make document templating a lot easier. A method like this could be accessed generically.
I can resolve this in other ways, but it would be the second implementation of the above resolution of the correct salutation string.

leventegal-she: You can call EnumHelper::toString in your implementation and there will be no double implementation. But I wouldn't mix this kind of business methods into DTOs because it can affect the serialization(e.g. converting to json in the REST api or export); it's value could accidentally be added to the serialized object.

Be careful with using this field for document generation because this field is only for german systems for now.

syntakker: Yes, already ran in an issue with `LocationDto.isEmptyLocation()` in a json serialization :-)

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6084,4 +6084,15 @@ CREATE TABLE labmessage (
 CREATE TABLE labmessage_history (LIKE labmessage);
 
 INSERT INTO schema_version (version_number, comment) VALUES (291, 'Add LabMessage #3486');
+-- SurvNet Adaptations - Create new field “Salutation” for persons #3411
+ALTER TABLE person
+    ADD COLUMN salutation varchar(255),
+    ADD COLUMN othersalutation varchar(512);
```
FredrikSchaeferVitagroup, line 6090: Just confirmed with Maté, we should make this text instead of varchar(512), as text has a variable size.

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6084,4 +6084,15 @@ CREATE TABLE labmessage (
...
+
+ALTER TABLE person_history
+    ADD COLUMN salutation varchar(255),
+    ADD COLUMN othersalutation varchar(512);
```
FredrikSchaeferVitagroup, line 6094: Also here...

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1811,6 +1811,12 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				currentVersion = 253;
 				getDao(Contact.class).executeRaw("ALTER TABLE events ADD column riskLevel varchar(255);");
 
+			case 254:
+				currentVersion = 254;
+
+				getDao(Person.class).executeRaw("ALTER TABLE person ADD COLUMN salutation varchar(255)");
+				getDao(Person.class).executeRaw("ALTER TABLE person ADD COLUMN otherSalutation varchar(512)");
```
FredrikSchaeferVitagroup, line 1818: Just confirmed with Maté, we should make this text instead of varchar(512), as text has a variable size.

FredrikSchaeferVitagroup (changes requested):

sormas-backend/src/main/resources/sql/sormas_schema.sql
leventegal-she: It is wired, there was a ticket when all text columns were changed to fixed size varchar, we don't have any text column in our DB. Why would we switch now to text?

leventegal-she: ok, I talked with Mate. I will change the varchar to text

FredrikSchaeferVitagroup (approved):

PR #3619, opened 2020-11-27
Add institutional partner option to events source type #3207

### LINKED ISSUE #3207: Add "health insure agency" as category to variable “Source Type” (EVENT)
### Feature Description
In France, most of the signals come from Health Insurance but can also be sent by other institutional partners such as national education, local authorities (town halls, departmental council, etc.). The methods currently proposed in the “TYPE OF SOURCE” variable do not meet the needs of France.

### Proposed Change
We propose to complete the modalities of the “Source Type” variable with the “institutional partner” modality. The selection of this modality then opens a new variable which specifies the partner: National education, Health insurance, Health establishment, Medico-social establishment ... This new variable can be configured by each user from a repository table loaded in the "Configuration" part of SORMAS.

### Additional Information

alexandre-gille:
Closes #3207


sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -152,7 +151,7 @@
 	public static final String DATABASE_NAME = "sormas.db";
 	// any time you make changes to your database objects, you may have to increase the database version
 
-	public static final int DATABASE_VERSION = 247;
+	public static final int DATABASE_VERSION = 248;
```
MateStrysewske, line 154: There was a mistake by one of our developers in version 247; please increase this to 249 to make it work again

MateStrysewske (changes requested):
* EventReadFragment is missing in the app
* There should probably also be a srcInstitutionalPartnerTypeDetails text field in case "Other" is chosen (to allow users to specify it); we have this pattern for all other enums that allow the selection of "Other"

ftavin:
>     * There should probably also be a srcInstitutionalPartnerTypeDetails text field in case "Other" is chosen (to allow users to specify it); we have this pattern for all other enums that allow the selection of "Other"

Good point ! This is a functional change to be confirmed in the issue with bernard and carolin, I have copy/pasted the question to the issue.

alexandre-gille:
It's done :heavy_check_mark:

sormas-api/src/main/java/de/symeda/sormas/api/event/EventExportDto.java
```diff
@@ -278,6 +284,16 @@ public EventSourceType getSrcType() {
 	}
 
 	@Order(19)
+	public InstitutionalPartnerType getSrcInstitutionalPartnerType() {
```
MateStrysewske, line 287: Unify this and the details field below and build an appropriate string directly in the constructor (see CaseExportDto.educationType for an example)

sormas-app/app/src/main/res/layout/fragment_event_read_layout.xml
```diff
@@ -209,6 +210,28 @@
...
+            <LinearLayout
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:orientation="horizontal">
+
+
+                <de.symeda.sormas.app.component.controls.ControlTextReadField
+                        android:id="@+id/event_srcInstitutionalPartnerType"
+                        style="@style/ControlFirstOfTwoColumnsStyle"
+                        app:value="@={data.srcInstitutionalPartnerType}"
```
MateStrysewske, line 222: Read-only fields can't have an "=" sign after the @ sign (because this is the trigger for two-way databinding)

MateStrysewske (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/event/EventExportDto.java
```diff
@@ -128,6 +126,19 @@ public EventExportDto(
 		this.jurisdiction = new EventJurisdictionDto(reportingUserUid, surveillanceOfficerUuid, regionUuid, districtUuid, communityUuid);
 	}
 
+	private String buildSrcInstitutionalPartnerTypeString(
```
MateStrysewske, line 129: Move this to a separate EventHelper class (in the api); DTos should ideally not contain such logic. You also don't need a StringBuilder, you can directly return the respective value in the if/else if body.

MateStrysewske (changes requested):
Have you checked and tested the app? E.g. there's an import for InstitutionalPartnerType missing, so the app is unable to build.

alexandre-gille:
> Have you checked and tested the app? E.g. there's an import for InstitutionalPartnerType missing, so the app is unable to build.

Unfortunately, I've never been able to launch the app, following the `DEVELOPMENT_ENVIRONMENT.md`. Hence I cannot test anything at all.
And I don't see where this missing import should be :shrug:

MateStrysewske:
@alexandre-gille Okay, I'll take care of that this time. But please make sure to get your Android Studio up and running next week. If you need support, drop me a message at our Gitter (it's linked on the main site of this repository). It's really, really important that you test the stuff you're submitting both in the web as well as the mobile app.

MateStrysewske:
@alexandre-gille Just realised that I of course can't fix this myself because I can't push to your pull request. There's an import for InstitutionalPartnerType missing in Event.java in the mobile app. I'll approve and merge the PR after that's added.

MateStrysewske (approved):

PR #3620, opened 2020-11-27
#3520 campaign refinements

### LINKED ISSUE #3520: Campaign refinements [2]
### Proposed Change

- [x] Add captions to the filters on the campaign data view, as we are doing it on the campaign dashboard (form, province, district, ... filters)
- [x] Add caption to search field in campaings view. Call it "search"
- [x] When entering new data (click "new form" button), the form that opens should pre-select the current campaign
- [x] Add a "Campaign" label on the left side of the Campaign selection field in the campaign data view
- [x] campaign dropdown (dashboard and campaign data); instead of selecting blank to unfilter, "All campaigns" would be more informative for users
- [x] Create campaign form: (i) note is cut off, does not wrap
- [x] add an "edit" header to the column in the campaign data list table
- [x] the edit button should make the mouse cursor change to make clear that it can be clicked. Same for the campaings list
- [x] hide the uuid column in the campaigns view
- [x] make formatting of input field warning popup in create form left aligned (see screenshot below)
- [x] Add a preview for campaign phases on the dashboard

grafik

MartinWahnschaffeSymeda:
closes #3520


sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/campaigndata/CampaignDataView.java
```diff
@@ -24,17 +24,12 @@
 
 import javax.naming.NamingException;
 
+import com.vaadin.ui.*;
```
MateStrysewske, line 27: Check your import settings, * imports should only be used for 99+ imports

MateStrysewske (changes requested):

PR #3648, opened 2020-12-01
Feature 3294 inform external journal about canceled or completed follow ups

### LINKED ISSUE #3294: [External Journal] Inform External Journal about canceled or completed follow-ups [2]

FredrikSchaeferVitagroup:
closes #3294


sormas-api/src/main/java/de/symeda/sormas/api/person/JournalPersonDto.java
```diff
@@ -33,6 +35,8 @@
 	private Sex sex;
 	@SensitiveData
 	private Date latestFollowUpEndDate;
+	@SensitiveData
```
leventegal-she, line 38: Normally enum and date fields are not sensitive, is this a special case?

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -388,8 +403,82 @@ public Date getLatestFollowUpEndDateByUuid(String uuid) {
...
+				return FollowUpStatus.NO_FOLLOW_UP;
+			}
+		}
+	}
+
+	private boolean listOnlyContainsStatus(List<FollowUpStatusDto> list, FollowUpStatus prameterStatus) {
+		if (list.isEmpty()) {
+			return false;
+		}
+		if (prameterStatus == null) {
```
leventegal-she, line 472: I wouldn't throw `IllegalArgumentException` in a private function and also the exception message is too generic, it doesn't tell which parameter is wrong.

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -729,6 +818,11 @@ public static PersonDto toDto(Person source) {
 		return target;
 	}
 
+	// needed for tests
+	public void setExternalJournalService(ExternalJournalService externalJournalService) {
```
leventegal-she, line 822: The production code should not contain test code. 

You can let it if there is no other quick solution. Maybe would be nice to add a git issue for investigating mocking possibilities.

leventegal-she (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/person/JournalPersonDto.java
FredrikSchaeferVitagroup: Well, I thought if the latestFollowUpEndDate is sensitive, the followUpStatus should be, too.
Is this tag used for pseudonomization? If so, maybe it's not a bad idea to leave it like that, because the JournalPersonDto is involved in transmission of person data to external systems.
I'm not sure about this, though.

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
FredrikSchaeferVitagroup: I'm going to replace it with an assert statement.

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
FredrikSchaeferVitagroup: This change was cherry-picked from another branch. I would prefer to leave it like that for now, although I think you have a valid point.

sormas-api/src/main/java/de/symeda/sormas/api/person/JournalPersonDto.java
leventegal-she: Yes `@SensitiveData` and `@PersonalData` annotations are used for pseudonymization 
For me it seems that `latestFollowUpEndDate` is wrongly marked as sensitive data.

The requirement for *personal data* and *sensitive data* is:  data from which somebody can identify a person is personal and all personal data and all free text and things relates to the location of the case/contact/person/sample/etc is sensitive.

I don't think it makes sense to pseudonymize `latestFollowUpEndDate` and `followUpStatus`.

FredrikSchaeferVitagroup: Okay, I removed both of the tags.

leventegal-she (approved):

Issue #3417, opened 2020-11-06
Add new community facility value for Germany to ReportingType [0.5]


markusmann-vg:
### Feature Description
Refine field “type of reporting” for cases

Case > case information > type of reporting

Change values (only DE) –
- [ ] Add a new value “Gemeinschaftseinrichtung § 34“.
- [ ] New Caption for „community facility“ in Germany to „Gemeinschaftseinrichtung … § 8“ [SurvNet field name: NotificationType]

Data Type: Dropdown/enum
Values (de-de) | (en)
• nicht erhoben- | data not collected
• -nicht ermittelbar- | unknown
• Arzt/Ärztin | doctor
• Labor | Laboratory
• Eigne Ermittlungen | own determination
• Gemeinschaftseinrichtung gemäß § 8 (1) 7 IfSG | Community facility (§ 8 IfSG)
• Gemeinschaftseinrichtung gemäß § 34 IfSG | Community facility (§ 34 IfSG)
• Weiterleitung von einem anderen GA | Forwarding
• Krankenhaus | Hospital or stationary care

GabrielePoggensee:
IfSG relevant - § 9: Definition of notification data to be reported to the local health department

SurvNet: Data not transmitted to state and public health authority

JonasCir:
I'll give this a shot

JonasCir:
@markusmann-vg @GabrielePoggensee I know how to implement the fields mentioned by Markus, however, I am confused about 

>  SurvNet: Data not transmitted to state and public health authority

Does this mean, the fields need to be present in SORMAS, but the fields are never exported to SurvNet at all?

JonasCir:
Blocked on https://github.com/hzi-braunschweig/SORMAS-Project/issues/3653

kwa20:
<img width="345" alt="Screen Shot 2021-01-15 at 11 29 54 AM" src="https://user-images.githubusercontent.com/70317594/104714131-1c970d80-5725-11eb-80ba-03c8608466c6.png">

"not raised" (nicht erhoben) is basically the same as an empty entry or null value and I think should therefore be removed.

bernardsilenou:
@kwa20 You are right. 
- To be consistent with other sections, I would additionally say we rename "not determinable" to "unknown" 
- delete "not raised" option and migrate old records to unknown (ie not determinable)

kwa20:
[...]

PR #3660, opened 2020-12-02
[IMIS] #3653 locale aware DB migration

### LINKED ISSUE #3653: Introduce locale aware DB migrations
### Situation Description
See #3417 as an example. In this issue, we need to do a migration, but only for Germany. Right now, our DB setup does not support this.

### Feature Description
Introduce migrations based on the locale server property as discussed with @MateStrysewske.

JonasCir:
Closes #3653 
Currently a draft as I don't know the best way to include a similar logic in the App. Ideas would be highly appreciated  @MateStrysewske 

---

I reused mainly the existing logic. A new table `schema_version_locale` is introduced to store the version of the migration.

Changing the locale will not rollback the applied changeset as:

1. I don' think this will occur on production systems (Correct me if I'm wrong)
2. This would require significant logic. If we want to have this in general, I think a library like [liquibase](https://www.liquibase.org/) is better suited than doing this by ourselves.

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
@@ -102,7 +102,13 @@ public class StartupShutdownService {
 
 	static final String SORMAS_SCHEMA = "sql/sormas_schema.sql";
 
+	private static final String SORMAS_SCHEMA_LOCALE_FORMAT = "sql/sormas_schema_%s.sql";
+
 	static final String AUDIT_SCHEMA = "sql/sormas_audit_schema.sql";
+
+	static final  String SCHEMA_VERSION_TABLE = "schema_version";
+
+	static final  String SCHEMA_VERSION_LOCALE_TABLE = "schema_version_locale";
 
 	private static final Pattern SQL_COMMENT_PATTERN = Pattern.compile("^\\s*(--.*)?");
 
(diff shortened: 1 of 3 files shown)


MateStrysewske:
@JonasCir I don't really have the time to think about a solution in the mobile app right now unfortunately. We've found a major issue with the web backend though. Your proposed solution does only work when we assume that

a) the local changes will NOT be executed when a new server is set up and
b) every server instance for a given locale is updated whenever a new version of SORMAS is released.

In all other circumstances, there's the high probability that there will be changes in the core migration script that will break the local migration.

Example:

Server S is on SORMAS 2.52.0, database version 500 . Developers for Germany have added a local migration script for SORMAS 2.53.0 that refers to table A. 2.53.0 is released, but server S is not updated. Core developers of SORMAS are changing the name of table A for SORMAS 2.54.0. Server S is updated to 2.54.0, the table is renamed in the core script, afterwards the local script is executed which refers to the old name of table A which is no longer existing. The database migration, and thus the server deployment, fails.

MateStrysewske:
@JonasCir In addition, we also need to add local-aware DB migration to the Android app if we do this, which means we also need a concept and a solution for that.

MateStrysewske:
@JonasCir We potentially also came up with a solution for #3417 that doesn't require database migration

JonasCir:
Alright great, thanks @MartinWahnschaffeSymeda . Let's continue the discussion in  #3417 then :)

MateStrysewske:
Closing this because it's not needed right now and would need to be refined anyway.

PR #3667, opened 2020-12-03
3249-Missing error message when creating contacts

danRosca:


MateStrysewske (changes requested):
Only do this for error and warning notifications, not for any other notification types

MateStrysewske (changes requested):
Warning notifications are missing

MateStrysewske (approved):

Issue #3413, opened 2020-11-06
SurvNet Adaptations - Create new field “name of guardians” for persons [0.5]


markusmann-vg:
### Feature Description
Cases > Open Case-person
Contacts > Open Contact-person
Events > Open Event-participants

Place the field under the field “Staatsangehörigkeit” | “Nationality”. 
Data Type: string

New field –
“Namen der Erziehungsberechtigten” (de-de) | „Name of the guardians“ (en) [SurvNet field name: NamesofGuardians]

bernardsilenou:
@markusmann-vg  This is similar to #2744 It would be good to refine both. Some German users also need name and contact information of  legal guardians which can be mother, father, friends, organization, etc
The new generic design for #2744 may handle all these needs. @kwa20 would add a mockup

GabrielePoggensee:
Not IfSG-relevant, but needed for the local health departments (e.g., quarantine order)

kwa20:
@GabrielePoggensee @markusman-vg this would go hand in hand with #2744

vidi42:
- [ ] Rename the field to `namesOfGuardians`

- [ ] Update translations

AndyBakcsy-she:
Validated issue on the last version of SORMAS-Version: 1.53.0-SNAPSHOT on https://test-de1.sormas.netzlink.com/ environment.

PR #3670, opened 2020-12-03
#3413 add name of other guardians to person

leventegal-she:
Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/3413


sormas-app/app/src/main/res/layout/fragment_person_edit_layout.xml
```diff
@@ -471,13 +471,20 @@
...
                     android:id="@+id/person_fathersName"
-                    android:nextFocusForward="@+id/person_phone"
+                    android:nextFocusForward="@+id/person_nameOfOtherGuardians"
                     app:value="@={data.fathersName}"
                     app:imeOptions="actionNext"
                     style="@style/ControlSecondOfTwoColumnsStyle" />
 
             </LinearLayout>
 
+            <de.symeda.sormas.app.component.controls.ControlTextEditField
```
FredrikSchaeferVitagroup, line 481: As you probably know, this field also shows in for apps of non-german servers.

sormas-app/app/src/main/res/layout/fragment_person_read_layout.xml
```diff
@@ -365,6 +365,11 @@
 
             </LinearLayout>
 
+            <de.symeda.sormas.app.component.controls.ControlTextReadField
```
FredrikSchaeferVitagroup, line 368: As you probably know, this field also shows in for apps of non-german servers.

FredrikSchaeferVitagroup (changes requested):
Should we add this field to the CaseExport and ContactExport?

leventegal-she:
> Should we add this field to the CaseExport and ContactExport?

I don't think we need to export those fields because none of other similar fields like *Mothers's name, Father's name, etc.* are exported

sormas-app/app/src/main/res/layout/fragment_person_edit_layout.xml
leventegal-she: Will be hidden after I merge the salutation field branch.

FredrikSchaeferVitagroup:
> > Should we add this field to the CaseExport and ContactExport?
> 
> I don't think we need to export those fields because none of other similar fields like _Mothers's name, Father's name, etc._ are exported

Okay.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/person/Person.java
```diff
@@ -428,6 +430,14 @@ public void setFathersName(String fathersName) {
 		this.fathersName = fathersName;
 	}
 
+	public String getNameOfOtherGuardians() {
```
FredrikSchaeferVitagroup, line 433: Also here, it should be namesOfOtherGuardians (plural), or?

FredrikSchaeferVitagroup (changes requested):

FredrikSchaeferVitagroup (approved):

Issue #2869, opened 2020-09-11
Person address refinement - Phase 3 [5]


ChristopherRiedel:
### Situation Description
Continuation of #2746 
### Feature Description
- [x] Show mainAddress und addresses together in one list in Webapp
- [x] Find pattern for getting person (main-)address (e.g. primary address)
- [x] Rename Person.address to Person.mainAddress

ChristopherRiedel:
Yesterday @HolgerReiseVSys had a better idea for the implementation of this issue: The entity stays as it is, I will only rename the adress to mainadress and the unification of adresses and the designation as mainadress is only done in the UI and in the Dtos.
I discussed this idea with @MateStrysewske and we decided to implement it like this.

syntakker:
@ChristopherRiedel @MateStrysewske What is he motivation for this renaming of the Person's property address to mainAddress?
This change breaks backward compatibility, which affects at least csv import (relevant especially for automated generation of csv from external sources) and document templating (here a redo of existing templates is necessary).

syntakker:
After commit dd34050, the related Tests `CaseImporterTest`, `ContactImporterTest` and `QuarantineOrderFacadeEjbTest` fail, these model the affected use cases described above.

ChristopherRiedel:
> @ChristopherRiedel @MateStrysewske What is the motivation for this renaming of the Person's property address to mainAddress?
The background is that not long ago the address list was added to the person. To better distinguish the main adress from the adresses, the adress is renamed to mainadress.

ChristopherRiedel:
@syntakker 
> After commit [dd34050](https://github.com/hzi-braunschweig/SORMAS-Project/commit/dd34050ba02afa296233645e4161ec9ebacfc362), the related Tests `CaseImporterTest`, `ContactImporterTest` and `QuarantineOrderFacadeEjbTest` fail. These cover the affected use cases described above.

The other tests will be fixed by me

syntakker:
@ChristopherRiedel @MateStrysewske This change comes at the cost that features already in production stop working without notice, both for csv import and document templating. It requires to communicate clearly which actions have to be taken by to keep these functionalities working (such as updating all templates), and it will nevertheless plunge some users into debugging why things do not work as expected anymore.
Just for clarification in the code, this cost is quite high.

[...]

PR #3682, opened 2020-12-03
fix quarantine order tests (#2869)

syntakker:
If the renaming of a Person's address to mainaddress is kept (see discussion https://github.com/hzi-braunschweig/SORMAS-Project/issues/2869#issuecomment-738455536), this PR fixes the `QuarantineOrderFacadeEjbTest`.
Other tests still fail.


ChristopherRiedel:
Is included in https://github.com/hzi-braunschweig/SORMAS-Project/pull/3691

PR #3728, opened 2020-12-09
Automate formatting and check it during build

### LINKED ISSUE #3543: Automate formatting and check it in CI
Currently, formatting the code (including ensuring license headers are correctly added/updated) requires manually configuring your IDE, which itself possibly requires installing additional plugins.

There's also no process for checking that files are properly formatted, so when formatting a file you might end up actually fixing bad formatting from a previous commit, leading to changes that are unrelated to the bugfix/feature being worked on thus making code reviews (or committing, if the develop decides to ignore those unrelated reformatting changes) harder.

Automating all of this would raise the overall project quality and streamline on-boarding of new contributors.

Fortunately, this is relatively easy to do through [Spotless](https://github.com/diffplug/spotless) and their Maven plugin.

I did make those changes locally as I didn't want to install the IntelliJ IDEA plugin allowing me to configure it, and then take time to configure both that plugin and the license headers; and I'd happily contribute them back.

Is that something you'd be interested in?

tbroyer:
Fixes #3543


sormas-base/license-header
```diff
@@ -0,0 +1,14 @@
+/*
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2020 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
```
tbroyer, line 3: This will soon have to be changed to 2021 for modified files to have the correct header.

sormas-base/pom.xml
```diff
@@ -828,6 +830,31 @@
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-resources-plugin</artifactId>
 			</plugin>
+
+			<plugin>
+				<groupId>com.diffplug.spotless</groupId>
+				<artifactId>spotless-maven-plugin</artifactId>
+				<version>2.6.1</version>
+				<configuration>
+					<ratchetFrom>origin/development</ratchetFrom>
```
tbroyer, line 839: With this, Spotless will only reformat and check files that have changed compared to origin/development, it also helps keeping the license header date up-to-date.
This however means that people need to have the hzi-braunschweig/SORMAS-Project repository defined as their `origin` remote, i.e. not named as `upstream`, or that they need to keep their own `origin/development` in sync with `upstream/development`.

An alternative would be to remove that line and reformat all files with new license header dates each year. If we decide to reformat everything though, then we can use `$YEAR` in the license-header template; that way, the template itself doesn't need to change, you only have to run `mvn spotless:apply` each New Year's Day and it will update all dates in source files.

See the documentation of the plugin for more information: https://github.com/diffplug/spotless/tree/main/plugin-maven

tbroyer:
Documentation on project setup would likely need some change too; I've only made the changes that allow me to work on the project.

sormas-base/pom.xml
MartinWahnschaffeSymeda: In order to answer this, we need to decide what the process of using this should look like.

Most developers already have the automatic formatting set-up in their IDEs. I wouldn't want them to have to run spotless before committing. Here using spotless should just be an alternative to setting up the IDE.

The big benefit from my point of view is to have this as part of the CI that is run for each PR to automatically detect when something is not formatted correctly.

Does the ratchetFrom work like this for the Github Actions CI builds? If that's the case then we can probably also expect this from all developers.

On the year automatism: From my understanding it would mean that we will only update the files that have been edited, so it would probably be better to just run a full update of the headers at the start of the year.

MartinWahnschaffeSymeda (changes requested):

MateStrysewske:
Closing this PR because it hasn't been updated in a while.

PR #3735, opened 2020-12-09
codeql CI

JonasCir:
@MartinWahnschaffeSymeda 
This is an example for GitHub's own security scanner.
It copies the build phase of #3732 at the moment, but this could be changed to reuse artifacts from previous build steps I think:)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,61 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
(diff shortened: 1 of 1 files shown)


JonasCir:
Ideally, this would run as dedicated job after the build job of #3732 in one workflow

JonasCir:
Since the Actions PR landed, I will reuse the build phase to safe time, see the [docs](https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts#uploading-build-and-test-artifacts).

JonasCir:
currently stale, will revise when I have time again

Issue #3398, opened 2020-11-05
Add values to healthFacility in import template comments


kwa20:
### Feature Description
case import template
### Problem Description
the healthFacility column in the case import template only mentions the "name of facility" as an entry possibility in the comments. However, when something other than a facility from the deposited infrastructure data is desired to be imported, the values "NO_FACILITY" or "OTHER_FACILITY" would be applicable. Users are a bit confused as to why some imports fail.
### Proposed Change
Add "NO_FACILITY" and "OTHER_FACILITY" to the comments next to "name of facility"
### Additional Information
<img width="545" alt="Screen Shot 2020-10-28 at 9 21 16 PM" src="https://user-images.githubusercontent.com/70317594/98229731-90b17a00-1f5a-11eb-8027-06fa70e58c7f.png">
<img width="907" alt="Screen Shot 2020-10-28 at 9 20 08 PM" src="https://user-images.githubusercontent.com/70317594/98229744-95762e00-1f5a-11eb-969a-6732e895601e.png">

syntakker:
@kwa20  Would you opt for `NO_FACILITY`, `OTHER_FACILITY`, `<text>`? Or do you think it is better to insert the list of facilities that are configured in the current instance from which the import template is generated (this would explain what to do in depth)?
For both option we should try to make clear that it is critical to have the same combination of name/description and facilityType.

syntakker:
These three issues should be considered together:
#3398, #3755, #3739

kwa20:
@syntakker If it is possible to insert the whole list of facilities, this would need to change consecutively with every additional entry of facilities in the infrastructure data. Just using 'name of facility' might therefore be the easy solution. If we consider listing all deposited facilities, this might get a bit messy when deposited facilities accumulate over time. The advantage would be that facility names are already referenced as usable variables. In both cases, NO_FACILITY and OTHER_FACILITY should be mentioned first.

PR #3740, opened 2020-12-10
Improve documentation of facilityReference in Data Dictionary (fixes #3739, #3398)

### LINKED ISSUE #3739: Improve documentation of facility import in the Data Dictionary
### Bug Description

The documentation of type `FacilityReference` describes only the special strings NO_FACILITY and OTHER_FACILITY. The Data Dictionary does not cover the import of configured facilities.

### Expected Behavior

 The use of the name of a facility configured in SORMAS should be documented here

### System Details
* SORMAS version: 1.53.0-SNAPSHOT

syntakker:
The documentation of type `FacilityReference` in the Data Dictionary now reads:

Type | Value | Caption | Description
-- | -- | -- | --
FacilityReference | OTHER_FACILITY | Other facility | This facility should be used if the desired facility does not (yet) exist in Sormas for any reason. The corresponding facility type must be filled in, so that it is clear what kind of facility it is.
  | NO_FACILITY | Home or other place | This facility should be used if explicitly not a facility but another place (e.g. at home) is meant. In this case, the corresponding facility type must be empty.
  | \<text\> | Configured Facility | This should be used, if a facilty named \<text\> has been configured for the district and community. The corresponding facility type must be filled in.

Previously, only `OTHER_FACILITY` and `NO_FACILITY` have been documented here.


MateStrysewske (changes requested):
Maybe it's a good idea to also add to the description of OTHER_FACILITY that the user should ideally also fill in the facilityDetails field to specify the facility name?

syntakker:
@MateStrysewske Currently the description of healthFacility is "Provide the name of the facility the patient was admitted". Would it help to change this into "Provide the facility the patient was admitted"?
The reason would be to encourage the user to investigate how to provide a facility where he now is left with some imports working (if the facility is configured) and others not (if the facility is not configured - however the user did as indicated and provided the name of the facility).

MateStrysewske:
@syntakker Makes sense to me :)

syntakker:
@MateStrysewske Done.
Additionally, I've made the addition of the FacilityReference table generic. Before it was hard-coded to be displayed in Case and Person, but missing for Location. Now the Data Dictionary keeps up if FacilityReference is added somewhere else.

syntakker:
@MateStrysewske there's now also an info for Facility in the csv import template (#3398).

sormas-api/src/test/java/de/symeda/sormas/api/doc/DataDictionaryGenerator.java
```diff
@@ -184,14 +184,6 @@ private XSSFSheet createEntitySheet(XSSFWorkbook workbook, Class<? extends Entit
 			Class<?> fieldType = field.getType();
 			if (fieldType.isEnum()) {
 				// use enum type name - values are added below
-//				Object[] enumValues = fieldType.getEnumConstants();
```
MateStrysewske, line 187: Why was this removed?

syntakker: The value list produced by the commented code has been replaced by the enum tables generated by this code in line 255ff:
```
		// enums
		for (Class<Enum<?>> usedEnum : usedEnums) {
			rowNumber = createEnumTable(sheet, rowNumber + 1, usedEnum);
		}
```

My impression was that the commented code just has been forgotten there.
If the decision to implement this change was not final and the intention was indeed to possibly revive the commented code, I'll restore the removed lines.

MateStrysewske: I guess the code is really not needed anymore, so you might as well remove it completely to keep it clean :)

MateStrysewske (changes requested):

sormas-api/src/test/java/de/symeda/sormas/api/doc/DataDictionaryGenerator.java
syntakker: The comment `// use enum type name - values are added below` was introduced in the same commit as the new code. It does not refer to the commented code, but to the new solution where the enum type is inserted here instead of its values, and the value table is added below.
The change dates back to November 2018, so I guess chances for the removed code to have been intended for revival are indeed minimal :-)

kwa20:
@syntakker @MateStrysewske This is only minor because the change will already be a big help but the text should refer to "healthFacilityDetails" as required instead of "facilityDetails" because this is what is referenced for CaseData and necessary for a successful import.
facilityImport2
facilityImport
facilityImport3

syntakker:
@kwa20 Yes, I'm aware of this.
However, facilities occur in several places, e.g., there is a `person.placeOfBirthFacility`.
The documentation in the Data Dictionary refers to "the *corresponding* facilityType / facilityDetails" now, while "corresponding" is omitted for the sake of brevity in the import template.
If you think it is very important to have the exact wording, I can try and figure out a solution.

kwa20:
@syntakker  Oh I get it ok, I think we should leave it then, it's already much clearer.

PR #3741, opened 2020-12-10
Feature 3152 symptoms reordering

BarnaBartha:


FredrikSchaeferVitagroup:
This will affect the external symptom journals, right?

MateStrysewske:
@BarnaBartha Fredrik is right; under these circumstances, I'd suggest to revert the name correction for now.

Issue #3412, opened 2020-11-06
SurvNet Adaptations - Create new fields “Country of birth” and “nationality” for persons [1]


markusmann-vg:
### Feature Description
Create new fields “Country of birth” and “nationality” for case persons, contact persons and event participants

Cases > Open Case-person
Contacts > Open Contact-person
Events > Open Event-participants

New fields -
“Geburtsland” (de-de) | “Country of birth” (en) [SurvNet field name: CountryofBirthMYT]
“Staatsangehörigkeit” (de-de) |”Nationality” (en) [SurvNet field name: CitizenshipMYT]
“Geburtsname” (de–de) | “Birth name” (en) [SurvNet field name: Birthname]

Place these fields under existing field “Geschlecht” | “Sex” Data

Data Type: Dropdown/enum
Values (de-de) | (en)
Länderliste des RKI (DE) | list of countries (RKI)
WHO country list (International)

markusmann-vg:
Prerequisite: https://github.com/hzi-braunschweig/SORMAS-Project/issues/2993

ChristopherRiedel:
@markusmann-vg There is already an issue for nationality #1752

GabrielePoggensee:
IfSG-relevant, but **not**  for COVID-19

Candice-Louw:
@markusmann-vg @leventegal-she - would it be possible to make this visible on all servers, for all countries, please (requested also by CH for example)?

Discussed with @kwa20 and @bernardsilenou as international feature.  If a seperate Issue should be created for this, please let me know and I'll do it.

PR #3746, opened 2020-12-10
3412 SurvNet person country of birth

leventegal-she:
Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/3412


vlad-ciucescu (approved):
Works well both on webapp & mobile app

PR #3774, opened 2020-12-14
#3718 allow members of same family to have the same email in Climedo

### LINKED ISSUE #3718: [CLIMEDO Integration] Allow use of the same contact details when only the First Name is different [1]
### Feature Description
Persons in SORMAS can be registered in as CLIMEDO patients.

### Problem Description
CLIMEDO offers the feature to register multiple family members to the same email address, but SORMAS does not allow this, even if the persons have the same last name. The check for email duplication will fail in SORMAS.

### Proposed Change
Allow multiple persons in SORMAS to use the same email address to register in CLIMEDO if they have the same last name (belong to the same family).

### Additional Information

- Checking for email duplication already implies retrieving the person data from CLIMEDO. No additional overhead will be necessary.
- Consider the improbable case in which two persons have the same last name, same emails but are not in the same family?

vlad-ciucescu:
closes #3718

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
@@ -387,7 +387,13 @@ public class ExternalJournalService {
 			.map(PatientDiaryIdatId::getIdat)
 			.map(PatientDiaryPersonDto::getPersonUUID)
 			.anyMatch(uuid -> person.getUuid().equals(uuid));
-		return notUsed || samePerson;
+		boolean sameFamily = response.getResults()
+				.stream()
+				.map(PatientDiaryPersonData::getIdatId)
+				.map(PatientDiaryIdatId::getIdat)
+				.map(PatientDiaryPersonDto::getLastName)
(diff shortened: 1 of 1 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
```diff
@@ -387,7 +387,13 @@ private boolean isEmailAvailable(PersonDto person) {
 			.map(PatientDiaryIdatId::getIdat)
 			.map(PatientDiaryPersonDto::getPersonUUID)
 			.anyMatch(uuid -> person.getUuid().equals(uuid));
-		return notUsed || samePerson;
+		boolean sameFamily = response.getResults()
+				.stream()
+				.map(PatientDiaryPersonData::getIdatId)
+				.map(PatientDiaryIdatId::getIdat)
+				.map(PatientDiaryPersonDto::getLastName)
+				.anyMatch(lastName -> person.getLastName().equals(lastName));
```
FredrikSchaeferVitagroup, line 395: We should also check wether the firstName is different, right?

FredrikSchaeferVitagroup (changes requested):

FredrikSchaeferVitagroup (approved):

PR #3778, opened 2020-12-14
2644 - Enable the usage of population data in campaign diagrams

danRosca:


sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/campaigns/CampaignDashboardDiagramComponent.java
```diff
@@ -239,6 +241,20 @@ public void buildDiagramChart(String title, boolean isCommunityGrouping) {
 		hcjs.append("]");
 		hcjs.append("}");
 		campaignColumnChart.setHcjs(hcjs.toString());
+		if (showPercentages) {
```
MateStrysewske, line 244: This now displays the warning below every diagram that shows percentages, even if they don't use population data at all or there is no missing population data. Also, the hint stays on the screen even if the user changes from population data to total values. It should always be hidden in that case.

MateStrysewske (changes requested):
The diagrams now have a lot of unnecessary space between them because the height is set to less than 100 % by default. I don't quite understand the reason for that change. There should be no space between the diagrams if it's not required.

MateStrysewske (changes requested):
* Don't show the warning message when the chart is not supposed to use population data
* Don't show the warning message when there actually isn't any missing population data

MateStrysewske (approved):

Issue #3485, opened 2020-11-15
Translations > Missing variables


Candice-Louw:
### Bug Description
Certain strings are not translated in SORMAS, despite all values having been saved and translated in CrowdIn.  For some of these variables, there seems to be no associated/existing listing in CrowdIn to translate.

### Steps to Reproduce
Set language to de-DE or de-CH for example.  English variables show up.

1. Cases and Contacts > Open Case/Contact > Follow-up > "Symptomatic", "Temperature"
![image](https://user-images.githubusercontent.com/65529128/99197319-140c7000-2792-11eb-9cff-8175f0546975.png)

2. Cases and Contacts > Open Case/Contact > Person > Addresses > "Postalcode", "AddressType"
![image](https://user-images.githubusercontent.com/65529128/99197487-1fac6680-2793-11eb-851f-a47800fd360a.png)

3. Cases > Open Case > Therapy > "Prescription Period", "Prescription Route", "Treatment Route"
![image](https://user-images.githubusercontent.com/65529128/99197548-7fa30d00-2793-11eb-861b-b3270b4ad8a1.png)

4. Cases > Custom Export > 
![image](https://user-images.githubusercontent.com/65529128/99197734-854d2280-2794-11eb-87d3-1bbe26159ee4.png)

5. Samples >
![image](https://user-images.githubusercontent.com/65529128/99198142-f1308a80-2796-11eb-8337-087ce93c09d2.png)

6. Info > "About" + "Documents"
![image](https://user-images.githubusercontent.com/65529128/99198599-0824ac00-279a-11eb-8c6b-c32e68bf32d0.png)

7. Contacts > Follow-up View > "Report Date"
![image](https://user-images.githubusercontent.com/65529128/99290852-2d1f2a80-283f-11eb-9fe2-da7718e59b48.png)

### Expected Behavior
Translated strings.

### System Details
* Device: Web
* SORMAS version: 1.51.1
* Android version/Browser: Chrome
* Server URL: https://test1401.sormas-oegd.de and https://sb-ch.sormas.netzlink.com/
* User Role: admin, national user

### Additional Information
Possibly related: https://github.com/hzi-braunschweig/SORMAS-Project/issues/3131

@kwa20 @bernardsilenou @syntakker - Please add anything that I may have missed?

kwa20:
This one also exists on the case status map layers.

[...]

PR #3782, opened 2020-12-15
#3485 - fix translation issues

BarnaBartha:


sormas-api/src/main/resources/captions_de-DE.properties
```diff
@@ -58,6 +58,8 @@ remove=Entfernen
 reportingUser=Meldender Nutzer
 
 # About
+about=About
```
MateStrysewske, line 61: Revert changes to this file please; they contain a lot of errors and will be overwritten by Crowdin anyway. Translations aren't taken care of by us. (I know it was necessary to add them for testing purposes)

sormas-ui/src/main/java/de/symeda/sormas/ui/visit/VisitGrid.java
```diff
@@ -64,19 +65,21 @@ public VisitGrid(VisitCriteria criteria) {
 			VisitIndexDto.VISIT_STATUS,
 			VisitIndexDto.VISIT_REMARKS,
 			VisitIndexDto.DISEASE,
-			VisitIndexDto.SYMPTOMATIC,
```
MateStrysewske, line 67: Replace all references to SymptomsDto that you introduced with VisitIndexDto again; only the I18N_PREFIX should be taken from SymptomsDto (because symptomatic and temperature are still properties of VisitIndexDto here, we're only interested in changing where the caption is generated from)

MateStrysewske (changes requested):

sormas-api/src/main/resources/enum.properties
```diff
@@ -595,7 +595,7 @@ Language.FR_CH = Français (Suisse)
 # MapCaseDisplayMode
 MapCaseDisplayMode.CASE_ADDRESS = ... by home address
 MapCaseDisplayMode.FACILITY = ... by facility
-MapCaseDisplayMode.HEALTH_FACILITY_OR_CASE_ADDRESS = ... by facility or home address
+MapCaseDisplayMode.FACILITY_OR_CASE_ADDRESS = ... by facility or home address
```
MateStrysewske, line 598: Fix this for all other languages as well, please

sormas-ui/src/main/java/de/symeda/sormas/ui/visit/VisitGrid.java
```diff
@@ -70,13 +71,15 @@ public VisitGrid(VisitCriteria criteria) {
 
 		((Column<VisitIndexDto, Date>) getColumn(VisitIndexDto.VISIT_DATE_TIME))
```
MateStrysewske, line 72: Not all instances of SymptomsDto.XXX have been reverted here, please check again - it should only be used for SymptomsDto.I18N_PREFIX

MateStrysewske (changes requested):

MateStrysewske (approved):

Issue #3772, opened 2020-12-14
Evaluate sortability for generated properties


DavidBaldsiefen:
### Situation Description

Currently, generated properties such as participant count and case count on the event overview are not sortable. This is true in many other cases too.

### Feature Description

It should be evaluated wether (and how) these properties can be made sortable.

### Additional Information

See #2016 too.

Relevant for #3367

StefanKock:
Scope of this ticket needs to be better defined. Is it only about events?

StefanKock:
The problem with the sorting not in the database is likely, that it does not really sort deeply with lazy loading as a user might expect. As far as I can see it, this approach sorts only the chunk currently requested.

DavidBaldsiefen:
@StefanKock 
> The problem with the sorting not in the database is likely, that it does not really sort deeply with lazy loading as a user might expect. As far as I can see it, this approach sorts only the chunk currently requested.

Yes that is correct, however as a user i would think that the list i currently see is also the list that will be sorted if i click on one of the columns. Does it make an actual difference though? As far as i can see the returned list can (in theory) be infinetely large at the moment. I would see the bigger problem if we had pagination for example.

> Scope of this ticket needs to be better defined. Is it only about events?

It was inspired by the specific requested use case where Events couldn't be sorted by participant, case and deathcount. My original way was to find a general approach that could be used in other cases with similar problems too, but I found that the getIndexList-Methods are often too specific, so that it might make sense to use this issue just for the event properties (and keep it in mind if the problem comes up anywhere else)

MateStrysewske:
@DavidBaldsiefen As a user, I expect the whole list to be sorted, e.g. if I'm looking at the case directory with 20,000 entries, I expect the sorting to take into account all these 20,000 entries. This works fine with in-database sorting, but it won't work with sorting the result of the query because that result will only contain a subset of the 20,000 entries due to the lazy loading. As a result, only that subset will be sorted, and that will result in false sorting.

DavidBaldsiefen:
@MateStrysewske I am not sure. If I, as a user, set some filters etc. and then receive a resulting subset of the total database containing all the entries which fit my filters, I would expect that sorting only sorts that subset. Or does the `getIndexList` method only show _some_, but not _all_ entries fitting my filters? Then of course it would make a difference if I sort the table afterwards.

MateStrysewske:
[...]

PR #3795, opened 2020-12-15
[DRAFT] Make generated Properties sortable (#3772)

DavidBaldsiefen:
This PR implements #3772


Seal33:
@MateStrysewske I have implemented one possible solution to make generated properties (such as case and deathcount for events) sortable. For this I just used Collections.sort on the generated indexList before returning, which worked quite well.

It would be really nice if you could look at this solution and give me some feedback. If it's fine by you, I can then go and implement this in several other places (e.g. case grid)

MateStrysewske:
@DavidBaldsiefen I guess this doesn't really work because of lazy loading, because it will only sort the sublist of data that is currently returned, and that will result in a false sorting overall. @HolgerReiseVSys @StefanKock @MartinWahnschaffe Can you correct me if I'm wrong? You only have to look at the first commit in this PR

MartinWahnschaffe:
@MateStrysewske Your are right. It won't work that way. The sorting has to be applied in the SQL statement, otherwise it can't work with lazy loading.

DavidBaldsiefen:
> @MateStrysewske Your are right. It won't work that way. The sorting has to be applied in the SQL statement, otherwise it can't work with lazy loading.

In that case I don't see any way to make those properties sortable, as they are fetched in completely different queries and then attached to the already sorted result set of the main query. Or do you have any other idea?

Issue #3730, opened 2020-12-09
Introduce GitHub Actions for CI


Seal33:
### Feature Description

Switching from Travis CI to GitHub Actions provides an immediate solution to the very slow CI process.

This Issue and linked PR would not disable Travis CI, but enable GitHub Actions in addition, so that the results can be evaluated and compared later on.

### Problem Description

The current Travis CI process is very slow, in part due to CI processing queuing up for a lot of time.
Build minutes are also limited as announced last months (#3541) . In comparison, GitHub offers unlimited build minutes for public repositories ([source](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions))

This is of particular high risk as currently, PRs get regulary merged without the CI process finishing, posing security and stability risks (of the last 50 commits to development branch, none passed the tests, probably because most didn't even start) (see #3711 )

### Proposed Change

Enabling and integrating GitHub Actions is a fast and streamlined solution. It also opens up a whole lot of possibilities due to the full GitHub integration, like disabling merging for PRs whose tests failed.

In addition, the tests for JDK8 and JDK11 run fully parallel and finished nearly twice as fast as the current Travis CI (total 16:09min vs 27:05min)

### Additional Information

[Example Build with GitHub Actions](https://github.com/ImisDevelopers/SORMAS-Project/actions/runs/410539795)

[Travis CI Build for comparison](https://travis-ci.org/github/hzi-braunschweig/SORMAS-Project/builds/748367014)

[GitHub Actions Documentation](https://docs.github.com/en/free-pro-team@latest/actions)

MartinWahnschaffeSymeda:
@JonasCir Thanks a lot for the PR!
I think we should take this opportunity to only do builds for Java 11. We are using Java 11 since July now and should phase out the active support for Java 8 (see #2617).

From my point of view it would also make sense to remove the travis ci integration as part of this. Having two build systems active would be a waste of server times for them.

JonasCir:
Thank you Martin! I'll update the PR now.

MartinWahnschaffeSymeda:
@HolgerReiseVSys mentioned that it should be sufficient to call mvn test instead of mvn verify. This would save at least the time to build the widgetset and since we don't have any integration tests that are run here, we wouldn't loose anything.

@JonasCir What do you think?

JonasCir:
[...]

PR #3796, opened 2020-12-15
[#3730] back to verify

JonasCir:
@MartinWahnschaffeSymeda Title says it all :D we need to go back to verify b/c mvn test does not really setup serverlibs suitable for sormas-cargoserver, so an error is thrown

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   test:
-    name: mvn test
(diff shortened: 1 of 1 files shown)


MartinWahnschaffeSymeda:
I agree. Since the time difference between verify and test was just 1 minute, it's currently not worth it to solve this in a different way (e.g. exclude cargo here).

JonasCir:
Thanks :) CI is green again

Issue #3822, opened 2020-12-17
security update vulnerable dependencies


AxelNennker:
### Bug Description
Sormas uses these vulnerable dependencies
[ERROR] apache-jsp-8.0.9.M3.jar: CVE-2020-8022, CVE-2016-5388, CVE-2016-8735, CVE-2018-8014, CVE-2017-5648, CVE-2017-12617
[ERROR] apache-jsp-9.2.14.v20151106.jar: CVE-2017-7658, CVE-2017-7657
[ERROR] htmlunit-2.19.jar: CVE-2020-5529
[ERROR] htmlunit-core-js-2.17.jar: CVE-2020-5529
[ERROR] jetty-server-9.2.14.v20151106.jar: CVE-2017-7658, CVE-2017-7657
[ERROR] jetty-xml-9.2.14.v20151106.jar: CVE-2017-7658, CVE-2017-7657
[ERROR] tapestry-4.0.2.jar: CVE-2020-17531
[ERROR] websocket-client-9.2.13.v20150730.jar: CVE-2017-7658, CVE-2017-7657

### Steps to Reproduce
1.  add owasp-dependency-check to sormas-base/pom.xml
2. `mvn org.owasp:dependency-check-maven:6.0.3:check`

```
[ERROR] apache-jsp-8.0.9.M3.jar: CVE-2020-8022, CVE-2016-5388, CVE-2016-8735, CVE-2018-8014, CVE-2017-5648, CVE-2017-12617
[ERROR] apache-jsp-9.2.14.v20151106.jar: CVE-2017-7658, CVE-2017-7657
[ERROR] htmlunit-2.19.jar: CVE-2020-5529
[ERROR] htmlunit-core-js-2.17.jar: CVE-2020-5529
[ERROR] jetty-server-9.2.14.v20151106.jar: CVE-2017-7658, CVE-2017-7657
[ERROR] jetty-xml-9.2.14.v20151106.jar: CVE-2017-7658, CVE-2017-7657
[ERROR] tapestry-4.0.2.jar: CVE-2020-17531
[ERROR] websocket-client-9.2.13.v20150730.jar: CVE-2017-7658, CVE-2017-7657
```

These depencies were not updated in 
https://github.com/hzi-braunschweig/SORMAS-Project/issues/3584
Althoug https://github.com/hzi-braunschweig/SORMAS-Project/pull/3626 already listed them as vulnerable.

### Expected Behavior
[...]

Issue #3819, opened 2020-12-17
Add OWASP maven dependency check to CI


JonasCir:
### Feature Description
Include  [OWASP dependency-check](https://jeremylong.github.io/DependencyCheck/) into the CI to catch vulnerable dependencies as early as possible (CC @AxelNennker)

### Proposed Changes
I already tested the maven dependency-check as _reporting_ plugin locally. By this, we do not increase CI build times as we can run the plugin on schedule or concurrently to the build.
I currently work on integrating the [CodeQL](https://github.com/github/codeql) security scanner GitHub action into the CI. The maven dependency check could be a job of this security workflow.

### Needs discussions
- How do we triage findings?
- The security tab on GitHub has an API which can be used by the actions. It would be nice to report the findings directly there

AxelNennker:
I moved the owasp dependency-check-maven plugin to the reporting section.
Probably the following needs to removed because excluding vulnerable dependencies from the report is kind of lying, right?!
```
						<configuration>
							<failBuildOnCVSS>8</failBuildOnCVSS>
							<suppressionFiles>
								<suppressionFile>project-suppression.xml</suppressionFile>
							</suppressionFiles>
						</configuration>
```

JonasCir:
@AxelNennker Thanks, yes you are right. We don't want to miss anything :)

JonasCir:
We use trivy as part of our GitHub action build and rely on GitHub's dependalert as well (both check for vulnerable dependencies). As far as I know, the maven dependency check also runs in Jenkins, so I think we can close this?

PR #3825, opened 2020-12-17
Security updates

AxelNennker:
update versions of vulnerable dependencies found by the owasp plugin

### DIFF
--- a/sormas-base/pom.xml
+++ b/sormas-base/pom.xml
@@ -710,6 +710,46 @@
 
 			<!-- ** Vaadin END ** -->
 
+<!-- *** Security updates *** -->
+<!-- These are mostly vulnerable dependencies of vaadin8 -->
+
+<!-- https://mvnrepository.com/artifact/net.sourceforge.htmlunit/htmlunit -->
+<dependency>
+    <groupId>net.sourceforge.htmlunit</groupId>
+    <artifactId>htmlunit</artifactId>
+    <version>2.45.0</version>
+</dependency>
+
+<!-- https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-webapp -->
+<dependency>
+    <groupId>org.eclipse.jetty</groupId>
+    <artifactId>jetty-webapp</artifactId>
+    <version>11.0.0</version>
+</dependency>
+
+<!-- https://mvnrepository.com/artifact/org.eclipse.jetty/apache-jsp -->
+<dependency>
+    <groupId>org.eclipse.jetty</groupId>
+    <artifactId>apache-jsp</artifactId>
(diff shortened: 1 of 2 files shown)


JonasCir:
@AxelNennker Thank you! Could you move the plugin to the reporting section? By this we still can check for vulnerable dependencies (concurrently to the build), but do not prolong the build phase itself.

AxelNennker:
it seems the owasp-depency-check plugin does not take the suppression-file into account. 
That works on my machine locally.
Could somebody @JonasCir please have a look why this might fail on the CI machine?

MateStrysewske:

MartinWahnschaffeSymeda:
@AxelNennker Does this solve #3819 and #3822? Would be best to create separate PRs for both.

AxelNennker:
I assume @JonasCir knows about #3819 This PR tried to update dependencies to non-vulnerable versions and exclude some that have no fixes yet from the dependency-check-maven. Personally I would run dependency-check-maven on every build because the most time is spent on the vaadin8 processing anyways and the dependency-check-maven processing does not take that much time. But that is my personal preference. [dependabot](https://dependabot.com/) and integrating dependency-check-maven into a seperate CI run make sense too. On my work projects I always configure a scheduled CI run each morning which notifies me of vulnerabilities. That should be enough for most projects, I think.
#3819 is putting dependency-check-maven into the reporting section which is different.

Most currently found vulnerabilities are hard to exploit because an attacker would e.g. submit CSS or other data into the build chain, thereby taking over the build chain and infecting the release version. That is difficult.
Another personal preference of mine is to update all dependencies regularly regardless of the security risk. Most updates improve the product - here Sormas. If not then we should report back to the dependency owner and thus help providing a better environment for all developers.

I think that https://github.com/hzi-braunschweig/SORMAS-Project/issues/3822 is not fixed. See the arguments above.
Fixing requires naming a lot of transitive dependencies in Sormas' pom.xml because Sormas direct dependencies might use a latest version of something but the dependencies used by the direct dependencies might have updates which are not included by maven in the build artifact unless they are explicitly named. Even when the found vulnerabilities are in "provided" dependencies (see build system attack argument)

Updating transitive dependencies makes Sormas' pom.xml longer and harder to maintain.
But naming/listing them shuts up dependency-check-maven which is a good thing because your don't what to do a manual security check again and again. Automation is good for security. Updates are good for security.
I assume there is a 3rd party security evaluation additionally. Which is probably not public, so I don't know about that.
Maybe Sormas should write on the official web-site who did the security review. If you need recommendations for security reviewers, contact me for (biased) recommendations.

But, of course, you can have a different opinion on all that and that is totally valid.

Issue #3662, opened 2020-12-02
Show area, region and district filter fields for all users


tomabadaniel:
### Bug Description
Before the last two releases, we had Regions| Provinces drop downs filtering the data shown on the dashboard or directory. After the new release feature that introduced the cascading and readonly function of the drop-downs, Province admin unit does not show any more without first selecting a region. The users normally measure the data collected by provinces and hence the dashboard and directory should allow "all provinces" data to be shown across all regions. This was the reason we had this issue about switching between bar charts and column charts (https://github.com/hzi-braunschweig/SORMAS-Project/issues/3349).

Second point, when the National level user logs in, he/she sees Regions (still called all areas), Provinces (still called all regions) and Districts (all districts). The labels should reflect the admin unit specified. When the Provincial user logs in,  he sees only provinces and districts, and when the district level user logs in, he sees only districts instead of clusters (communities). There should be a breadcrumb read-only admin units filtered with the respective names when logged in as a Provincial or District level user. And district level user should have clusters activated for him/her
### Steps to Reproduce 
1. Log in as National user, Go to dashboard, Elements are shown by default for all regions. Click on the province filter to show elements for all provinces without selecting a region to display- It is not activated
2. Province filter is disabled. When you select a region, then province filter is activated, when you select a province, the district filter is enabled
3. Log in as Provincial user, he only sees Districts active without seeing the breadcrumbs of what Province or Region he belongs to

4. Log in as District user, he only sees Districts which is strange because he should see clusters
### System Details
* Device:
* SORMAS version:
* Android version/Browser:
* Server URL:
* User Role:

MartinWahnschaffeSymeda:
I have created #3717 for the "group by" change request.

Bugs that need to be fixed here:

- [ ] Add AF translation for "all regions", etc. to crowdin
- [ ] area, region and district filters should always be visible. When the jurisdiction level of the user is region, the area and region field should be pre-filled and read-only (same goes for other user jursidiction levels).

@danieltomaba I have changed the name of the title accordingly

MartinWahnschaffeSymeda:
@BarnaBartha I have added the translations to crowdin. Once it is merged to the pull request, #3810 will have to be merged.
I have changed this at 14:30 MEZ - so the last commit of the pull request has to be later.

PR #3834, opened 2020-12-18
#3662 - show jurisdiction fields for campaign dashboard always - auto…

BarnaBartha:
… fill & disable based on user jurisdiction


MateStrysewske (changes requested):
* Filter fields are not emptied when the "parent" filter is set to null

Issue #3766, opened 2020-12-14
Update to Keycloak 12 [3]


StefanKock:
### Problem Description

Update to Keycloak 12 when available, see https://github.com/keycloak/keycloak/releases.

### Proposed Change

1. Update Maven dependencies concerning Keycloak and review open points of #3580.
2. Push relevant infos to SORMAS-Docker for update to Keycloak 12.

JonasCir:
https://github.com/keycloak/keycloak/releases
Seems to be released

Issue #3602, opened 2020-11-26
Migrate users from SORMAS Authentication to Keycloak automatically [3]


vidi42:
### Situation Description
At the moment the SORMAS system supports enabling authentication trough Keycloak (#2745) , however this is mostly designed to be done for new systems, where there are no active users. 

An automated migration already exists for default users like admin, NatUser, etc which activates them in Keycloak at server startup with their default credentials.

However for non-default (external users) there is not automated mechanism, only a manual one:
1. An admin logs into the SORMAS Web app
2. Goes trough each user which it wants to migrate and fills their email address and saves them
3. The users will receive an email to activate their account and setup a new password

Since already running systems might have multiple users, an automated process is needed to fulfil this tasks.

### Feature Description
An automated script which can import all existing SORMAS users into Keycloak and allow them to access their account trough it.

#### Restrictions
Since the passwords in SORMAS are hashed they cannot be imported into Keycloak, so the users will have to setup new passwords.
If the users don't have an email address they will not be able to receive an email notification to setup their accounts.

### Possible Alternatives
1. Keep the manual approach

2. An automated startup script which goes trough all existing users and imports them into Keycloak with a random password.
Advantages: users which have an email address will receive an email to activate their account and setup a password
Disadvantages: users which don't have an email address still have to be manually filled by an admin with an email address

3. External script which can be run at any time by a DevOps person to activate users with a predefined password and provide users with the possibility to change their email address once logged in.
Advantages: Password can be communicated to the users and they don't need to have an email address setup upfront
Disadvantages: The passwords have to be communicated to each user otherwise individually, otherwise risk having users guessing passwords for other users

### Additional Information
Manual approach will be available for any of the alternatives.

vidi42:
In scope of this story
- [ ] Add UI field for users to change their email in settings panel (bellow language) - only when Keycloak is active as Authentication Provider
- [ ] Create scripts for migrating users into Keycloak

markusmann-vg:
@vidi42 Please provide a documentation how it's done and how it is possible to import encrypted passwords

JonasCir:
Heads up @vidi42 : There is also https://github.com/hzi-braunschweig/SORMAS-Project/pull/3837 about to land which fixes security issues

vidi42:
@JonasCir Thanks for the heads up.
We already have this in plan for the Next Sprint with #3766

AndyBakcsy-she:
[...]

PR #3837, opened 2020-12-18
CVE-2020-14389

AxelNennker:

It was found that Keycloak before version 12.0.0 would permit a user with only view-profile role to manage the resources in the new account console, allowing access and modification of data the user was not intended to have.
https://nvd.nist.gov/vuln/detail/CVE-2020-14389

Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/3766


StefanKock:
This will be addressed in #3602 and touches more than just the version in Base POM.

PR #3838, opened 2020-12-18
replace disfunctional crashlytics and update to latest firebase

AxelNennker:

The old version of crashlytics is not supported anymore.
The new version is now in com.google.firebase.crashlytics.FirebaseCrashlytics

This PR moves to the new crashlztics and updates firebase to the current version


MateStrysewske:

Issue #3814, opened 2020-12-16
Wrongful order of latest samples


kwa20:
### Bug Description
The 3 latest samples are displayed in the wrong order in the detailed export.
### Steps to Reproduce
1. create at least three samples for an entity (e.g. case) and choose different sample collection dates.
2. do a detailed export of that entity (case) and check the latest samples

### Expected Behavior
Latest samples should be displayed in the correct order. latest = most recent; 2nd latest = 2nd most recent; 3rd latest = first of those three
### Screenshots
<img width="224" alt="samplesExport" src="https://user-images.githubusercontent.com/70317594/102397812-b2ccfc00-3fde-11eb-93e3-6e6f98b56524.PNG">
<img width="452" alt="samplesExport2" src="https://user-images.githubusercontent.com/70317594/102397816-b52f5600-3fde-11eb-8701-7e6e3a34e535.PNG">

### System Details
* Device:
* SORMAS version: 1.52.1
* Android version/Browser: chrome
* Server URL: test1401.sormas-oegd.de
* User Role: admin, national user

### Additional Information
I would suggest to adjust the strings since it's probably easier then re-programming it.

raghupola-vg:
tested the export on latest version and issue is Fixed.

PR #3839, opened 2020-12-18
#3814-add descending ordering of the samples by date

danRosca:


BarnaBartha (approved):
looks good, please cover this sorting also in the export test

Issue #3583, opened 2020-11-25
Contacts > Contacts without "Responsible Region" and "Responsible District" are not visible in the Contact Directory Default View


Candice-Louw:
### Bug Description
After implementation of https://github.com/hzi-braunschweig/SORMAS-Project/issues/2137, Contacts without "Responsible Region" and "Responsible District" are not shown in the Contact Directory Default View anymore.

This results in massive amounts of Contacts not being visible to contact tracers unless they use these workarounds:
1. Adjust all user roles to include the National User role right (not ideal as additional data can then be seen).
2. As a National User, everytime the Contacts Directory is opened, the contact tracers have to add the filter "Show Contacts outside this jurisdiction" (not ideal as not all are aware of this and where to find it).

Mass editing of Contacts can't be done to adjust the region/district etc. yet.  Would each Contact have to be updated by hand to include this information?

Are there other solutions to this, please?

Thousands of contacts are 'disappearing' from the contact tracers' view and they are starting to produce double entries to 'fix' this which results in bad data quality and inefficient operation (Kt. Bern @rafaelpaffenholz ).

### Steps to Reproduce
1. Contacts > New Contact (Responsible Region and Responsible District are compulsory - expected)
![image](https://user-images.githubusercontent.com/65529128/100212876-2ca52480-2f0e-11eb-87c3-5cdb1a5f41a5.png)

2. Choose Source Case to link to Contact > Confirm (Responsible Region and Responsible District are NOT compulsory anymore - not expected?)
![image](https://user-images.githubusercontent.com/65529128/100212982-51999780-2f0e-11eb-91c9-ae7d3485c153.png)

3. Remove Case > (Responsible Region and Responsible District are compulsory - expected)
![image](https://user-images.githubusercontent.com/65529128/100213056-742bb080-2f0e-11eb-9a11-f004e3451e72.png)

4. When a Contact is saved with a source Case, the Responsible Region and Responsible District are not recorded.  The "null" values which are allowed for Contacts for these fields are not taken into consideration during the display in the list.

### Expected Behavior
+ Contact Tracers should be able to view all Contacts - even those with a null value - without user role adjustment.
+ Responsible Region and Responsible District should be made compulsory (could affect prior data negatively)
+ Make this feature configurable (?)
+ Alternatively, a bulk edit mode should be available for Contacts to assign a Responsible Region and/or District.

### System Details
* Device: Web
* SORMAS version: 1.51.1+
[...]

PR #3840, opened 2020-12-18
#3583 - Contacts > Contacts without "Responsible Region" and "Respons…

sergiupacurariu:
…ible District" are not visible in the Contact Directory Default View


BarnaBartha (approved):
looks good, but these changes should be covered by tests

sergiupacurariu:
testGetIndexList from ContactFacadeEjbTest was updated

BarnaBartha (approved):

Issue #3458, opened 2020-11-11
Add set clear to unknown to mobile app [0.5]


bernardsilenou:
### Feature Description
The button "set clear to unknown" was added to the web app but not on the mobile app 
we can use a fluid row with 3 columns just as we have for the clinical confirmation variable
### Problem Description
affected screens:

- [ ] symptoms 

### Proposed Change
- Add the set clear to unknown  button on mobile app.

AndyBakcsy-she:
Verified the existence and the functionality of the "set clear to unknown" button on the mobile app 
Tested on personal Android device with an upgrade from version 1.52.0 to 1.54.0-SNAPSHOT

raghupola-vg:
"Set clear to No" button is not fully visible on the Mobile app.
It is good to have it's own row for the buttons text to be fully visible.
![Set cleared to no button is not fully visible](https://user-images.githubusercontent.com/62763628/104911269-e35fd700-598a-11eb-86d2-7f50e05ea14f.png)

PR #3847, opened 2020-12-21
#3458 - set clear to unknown button for symptoms on mobile app

BarnaBartha:


MateStrysewske (changes requested):
The buttons aren't properly readable on devices with a low resolution. I suggest going into the accessibility options in your emulator and increasing the display size (because that's closer to the tablets they're using e.g. in Nigeria). I suggest to move the both "Set cleared to" buttons to a new line below the "Clear all" button and set the "Clear all" button to full width.

MateStrysewske (approved):

PR #3855, opened 2020-12-21
#3766 update keycloak to 12.0.0 because of CVE-2020-14389

AxelNennker:

See https://nvd.nist.gov/vuln/detail/CVE-2020-14389

keycloak 12.0.0 is available and older versions are vulnerable.


vidi42:
The CVE makes a reference to the New User Console which is a server side functionality. Updating the adapters doesn't solve this issue.
Closed similarly to https://github.com/hzi-braunschweig/SORMAS-Project/pull/3837

AxelNennker:
updating the version to 12.0.0 shuts up dependency-check-maven
I would update AND update the server.
Especially as it does not hurt to update. It is all pros no cons.

Issue #3818, opened 2020-12-17
Duration logging for EJB methods [2]


StefanKock:
### Situation Description

Currently we usually have to guess where performance degration lies when functions are used.

### Feature Description

On request enable logging of duration of EJB methods to get a hint where it takes too long.

Idea to solve:
 - https://stackoverflow.com/questions/37883022/detailed-log-of-executing-java-ee-application-in-production
 - https://www.adam-bien.com/roller/abien/entry/simplest_possible_ejb_3_12
 - Logging on TRACE level, have the logger commented out in logback.xml

### Possible Alternatives

Write logger statements manually within critical methods, but leads to boilerplate code and potentially leaves a lot of blind spots.

### Additional Information

As a PoC I already proved that this is working with annotations, but I did not find how to define it with XML for all EJBs.

StefanKock:
I finished the feature configurable on DEBUG (quick hint to duration of methods) and TRACE level (start of method calls with parameters and all durations) in `logback.xml`.

[#3818 - Example log for SamplesView.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/5710498/3818.-.Example.log.for.SamplesView.txt)

StefanKock:
> What is the performance impact of adding the interceptor?
> Is it still relevant when the logging is disabled?

According to Adam Bien, no-op interceptors are quite irrelevant for performance: https://www.adam-bien.com/roller/abien/entry/what_is_the_interceptors_overhead

StefanKock:
Updated example with new class name and not measuring _Starting_ logger statement in duration.
[#3818 - Example log for SamplesView.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/5733870/3818.-.Example.log.for.SamplesView.txt)

StefanKock:
I struggle to fix the failing unit test.

StefanKock:
I still struggle to get rid of failing unit test now tied to my change but somehow provoking it. On earlier runs where tests were repeated automatically, this happened:
```
[INFO] Running de.symeda.sormas.backend.util.IterableHelperTest
Error:  Tests run: 3, Failures: 0, Errors: 3, Skipped: 0, Time elapsed: 0.457 s <<< FAILURE! - in de.symeda.sormas.backend.util.IterableHelperTest
Error:  testExecuteBatched(de.symeda.sormas.backend.util.IterableHelperTest)  Time elapsed: 0.446 s  <<< ERROR!
[...]

PR #3857, opened 2020-12-22
#3818: Log method invocation duration

StefanKock:
I would like to have feedback on this. Example logs are here: https://github.com/hzi-braunschweig/SORMAS-Project/issues/3818#issuecomment-747519744

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
@@ -1,0 +1,69 @@
+package de.symeda.sormas.backend.util;
+
+import javax.interceptor.AroundInvoke;
+import javax.interceptor.InvocationContext;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import de.symeda.sormas.api.utils.DateHelper;
+
+/**
+ * <p>
+ * Purpose: Log duration of method invocation, usually to be to detect performance leaks in the EJB call stack.<br />
+ * You can activate logging with DEBUG level to get all durations over {@value #SIGNIFANT_DURATION} ms.
+ * To see all EJB method starts with parameters and all durations, set log level to TRACE.
+ * </p>
+ * <p>
+ * <b>Warning:</b> Activating this logging has some performance impact itself, so only active it temporarly.
+ * </p>
+ * 
+ * @see https://www.adam-bien.com/roller/abien/entry/simplest_possible_ejb_3_12
+ */
+public class MethodDurationLogger {
+
+	/**
+	 * Avoids fast but noisy and mostly internal EJB calls to be logged on DEBUG level.
+	 */
+	private static final int SIGNIFANT_DURATION = 5;
+
+	private final Logger logger = LoggerFactory.getLogger(MethodDurationLogger.class);
+
+	@AroundInvoke
+	public Object logInvokeDuration(InvocationContext context) throws Exception {
+
+		if (!logger.isDebugEnabled()) {
+			// Directly proceed if logging is not activated
+			return context.proceed();
+		}
+
+		long startTime = DateHelper.startTime();
+		boolean traceEnabled = logger.isTraceEnabled();
+		if (traceEnabled) {
+			logger.trace("Started: {} with parameters '{}'", getInvokedMethod(context), context.getParameters());
+		}
+
+		try {
+			return context.proceed();
+		} finally {
+			long ms = DateHelper.durationMillies(startTime);
+			if (traceEnabled || ms > SIGNIFANT_DURATION) {
+				logger.debug("Finished in {} ms: {}", ms, getInvokedMethod(context));
+			}
+		}
+	}
+
+	/**
+	 * @return Better readable method name (with EJB class) that was invoked.
+	 */
+	static String getInvokedMethod(InvocationContext context) {
+
+		String targetName = context.getTarget().toString();
+		int classBegin = targetName.lastIndexOf(".") + 1;
+		int innerClassBegin = targetName.indexOf("$");
+		int instanceBegin = targetName.indexOf("@");
+
+		String ejbName = targetName.substring(classBegin, innerClassBegin > 0 ? innerClassBegin : instanceBegin);
+		return String.format("%s.%s", ejbName, context.getMethod().getName());
+	}
+}
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/util/PerformanceLoggingInterceptor.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/util/PerformanceLoggingInterceptor.java
@@ -20,14 +20,14 @@ import de.symeda.sormas.api.utils.DateHelper;
  * 
  * @see https://www.adam-bien.com/roller/abien/entry/simplest_possible_ejb_3_12
  */
-public class MethodDurationLogger {
+public class PerformanceLoggingInterceptor {
 
 	/**
 	 * Avoids fast but noisy and mostly internal EJB calls to be logged on DEBUG level.
 	 */
 	private static final int SIGNIFANT_DURATION = 5;
 
-	private final Logger logger = LoggerFactory.getLogger(MethodDurationLogger.class);
+	private final Logger logger = LoggerFactory.getLogger(PerformanceLoggingInterceptor.class);
 
 	@AroundInvoke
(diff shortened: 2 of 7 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
```diff
@@ -0,0 +1,69 @@
...
+ * </p>
+ * 
+ * @see https://www.adam-bien.com/roller/abien/entry/simplest_possible_ejb_3_12
+ */
+public class MethodDurationLogger {
+
+	/**
+	 * Avoids fast but noisy and mostly internal EJB calls to be logged on DEBUG level.
+	 */
+	private static final int SIGNIFANT_DURATION = 5;
```
HolgerReiseVSys, line 28: I ❤️ signifants

sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
```diff
@@ -0,0 +1,69 @@
...
+
+	@AroundInvoke
+	public Object logInvokeDuration(InvocationContext context) throws Exception {
+
+		if (!logger.isDebugEnabled()) {
+			// Directly proceed if logging is not activated
+			return context.proceed();
+		}
+
+		long startTime = DateHelper.startTime();
```
HolgerReiseVSys, line 40: Maybe the timer should be started after the initial trace logging.

sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
```diff
@@ -0,0 +1,69 @@
...
+		boolean traceEnabled = logger.isTraceEnabled();
+		if (traceEnabled) {
+			logger.trace("Started: {} with parameters '{}'", getInvokedMethod(context), context.getParameters());
```
HolgerReiseVSys, line 43: Invoking `::toString` for the parameters may be costly.
Also, this may have side effects, e.g. trigger lazy loading.
It could make sense to print `className`@`hash` for not explcitly supported types.
Enums could be printed as `className`.`name`.

HolgerReiseVSys (commented):
What is the performance impact of adding the interceptor?
Is it still relevant when the logging is disabled?

sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
```diff
@@ -0,0 +1,69 @@
...
+	 * Avoids fast but noisy and mostly internal EJB calls to be logged on DEBUG level.
+	 */
+	private static final int SIGNIFANT_DURATION = 5;
+
+	private final Logger logger = LoggerFactory.getLogger(MethodDurationLogger.class);
+
+	@AroundInvoke
+	public Object logInvokeDuration(InvocationContext context) throws Exception {
+
+		if (!logger.isDebugEnabled()) {
```
BarnaBartha, line 35: if some1 does not define this logger in their logback.xml it'll be debug by default thus logging of durations will happen, i am not sure we want this, if so we should exted this check with || logger.level == null

StefanKock:
> 
> 
> What is the performance impact of adding the interceptor?
> Is it still relevant when the logging is disabled?

According to Adam Bien, no-op interceptors are quite irrelevant for performance: https://www.adam-bien.com/roller/abien/entry/what_is_the_interceptors_overhead

sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
```diff
@@ -0,0 +1,69 @@
...
+ * You can activate logging with DEBUG level to get all durations over {@value #SIGNIFANT_DURATION} ms.
+ * To see all EJB method starts with parameters and all durations, set log level to TRACE.
+ * </p>
+ * <p>
+ * <b>Warning:</b> Activating this logging has some performance impact itself, so only active it temporarly.
+ * </p>
+ * 
+ * @see https://www.adam-bien.com/roller/abien/entry/simplest_possible_ejb_3_12
+ */
+public class MethodDurationLogger {
```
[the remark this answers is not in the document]
StefanKock: Okay, I can rename it like this

sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
StefanKock: Yes, good idea. No need to measure Logger overhead

sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
StefanKock: Hmm, side effects would be nasty. I will think about it.

sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
StefanKock: - I think the Logger level cannot be null. If we do not define it for a specific logger, the root log level applies.
- In any case, logging needs to be configured for the usual use case. So the logback.xml needs to be changed as proposed as default behaviour. If it keeps DEBUG level, there might be a tiny performance impact, but nothing critical is printed to the logfile.
- As fas as I saw it, there is no way to disable the interceptor by configuration (except editing the ejb-jar.xml inside the artifact to deploy).

StefanKock:
A not related unit test fails (line 1735), not reproducable on my system:
```
[INFO] Running de.symeda.sormas.backend.util.IterableHelperTest
Error:  Tests run: 3, Failures: 0, Errors: 3, Skipped: 0, Time elapsed: 0.494 s <<< FAILURE! - in de.symeda.sormas.backend.util.IterableHelperTest
Error:  testExecuteBatched(de.symeda.sormas.backend.util.IterableHelperTest)  Time elapsed: 0.436 s  <<< ERROR!
java.lang.NullPointerException
	at de.symeda.sormas.backend.util.IterableHelperTest.assertExecuteCount(IterableHelperTest.java:47)
	at de.symeda.sormas.backend.util.IterableHelperTest.testExecuteBatched(IterableHelperTest.java:26)

Error:  testExecuteBatched(de.symeda.sormas.backend.util.IterableHelperTest)  Time elapsed: 0.448 s  <<< ERROR!
org.mockito.exceptions.misusing.UnfinishedVerificationException: 

Missing method call for verify(mock) here:
-> at de.symeda.sormas.backend.util.IterableHelperTest.assertExecuteCount(IterableHelperTest.java:47)

Example of correct verification:
    verify(mock).doSomething()

Also, this error might show up because you verify either of: final/private/equals()/hashCode() methods.
Those methods *cannot* be stubbed/verified.
Mocking methods declared on non-public parent classes is not supported.


Error:  testExecuteBatched(de.symeda.sormas.backend.util.IterableHelperTest)  Time elapsed: 0.448 s  <<< ERROR!
org.mockito.exceptions.misusing.InvalidUseOfMatchersException: 

Misplaced or misused argument matcher detected here:

-> at de.symeda.sormas.backend.util.IterableHelperTest.assertExecuteCount(IterableHelperTest.java:47)

You cannot use argument matchers outside of verification or stubbing.
Examples of correct usage of argument matchers:
    when(mock.get(anyInt())).thenReturn(null);
    doThrow(new RuntimeException()).when(mock).someVoidMethod(anyObject());
    verify(mock).someMethod(contains("foo"))

This message may appear after an NullPointerException if the last matcher is returning an object 
like any() but the stubbed method signature expect a primitive argument, in this case,
use primitive alternatives.
    when(mock.get(any())); // bad use, will raise NPE
    when(mock.get(anyInt())); // correct usage use

Also, this error might show up because you use argument matchers with methods that cannot be mocked.
Following methods *cannot* be stubbed/verified: final/private/equals()/hashCode().
Mocking methods declared on non-public parent classes is not supported.
```

sormas-backend/src/main/java/de/symeda/sormas/backend/util/MethodDurationLogger.java
StefanKock: After some research I found some articles that consider toString() with maliciuous side effects as bad programming style. I would not address this.

StefanKock:
`IterableHelperTest` still fails

BarnaBartha (approved):

PR #3859, opened 2020-12-22
Feature 3602 migrate users

### LINKED ISSUE #3602: Migrate users from SORMAS Authentication to Keycloak automatically [3]

vidi42:
Closes #3602


BarnaBartha (approved):
nice feature
tested on local keycloak and migration works well

Issue #3226, opened 2020-10-19
About section for campaign SORMAS (APMIS) [1]


MartinWahnschaffeSymeda:
### Proposed Change
![grafik](https://user-images.githubusercontent.com/23217632/96474515-2f3b9c80-1233-11eb-9cba-15bec3549882.png)

- [x] split the "SORMAS Version: ..." line into two lines (See screenshot)
- [x] When a custombranding.name is defined in the properties, use "... powered by SORMAS"
- [x] When automatic case classification is disabled (configured in properties) or the CASE_SURVEILANCE  feature is disabled, hide the "case classification" link
- [x] When CASE_SURVEILANCE and EVENT_SURVEILLANCE are both disabled, hide the "data dictionary" link
- [x] Allow the server admin to provide custom documents. Provide all files that are in the "custom/aboutfiles" folder as download (the "aboutfiles" subfolder does not exist yet). Use the file name and extension for the label. Put these files on top of all other document links. 

### Additional info
Technical manual is already dropped, so if none of the documents is accessible, then we should not show the "Documents" caption.

MartinWahnschaffeSymeda:
- [ ] update the related translations on crowdin and merge the crowding pull request that contains those changes afterwards.

PR #3882, opened 2020-12-28
#3226 about section for campaign sormas

alexcaruntu-vita:
#3226

Open points/questions:

1. We need translations for the new texts added.
2. Who will create the 'aboutfiles' subfolder? Do we need to update any scripts?


sormas-api/src/main/resources/captions_de-CH.properties
```diff
@@ -58,7 +58,8 @@ remove=Entfernen
 reportingUser=Meldender Nutzer
 
 # About
-aboutSormasVersion=SORMAS-Version
+aboutSormasVersion=Version
```
BarnaBartha, line 61: please revert the changed to the translation files
they will be added by crowdin, you only need to modify the captions.properties

BarnaBartha (changes requested):
@alexcaruntu-vita 
1. no, the translations will be added from crowdin
2. i suggest we do create this directory "aboutfiles" from within the setup scripts (normally whoever is responsible for the customisation of that server can also do this but it just leaves one more small step for error if they misname this directory name)

PR #3888, opened 2020-12-30
#3805: revert testDetectAnnotationChangesFast timing and update ci to run only this test

### LINKED ISSUE #3805: Auditor Test crashes during GitHub Actions
```
Error:  Failures: 
Error:    AuditorTest.testDetectAnnotationChangesFast:289 
Expected: a value less than <1000000L>
     but: <1551807L> was greater than <1000000L>
```

Suggested approach

- [x] Increase expected number to prevent the crash for new PRs
- [x] Investigate why timings are different when running on the GitHub servers

vidi42:
Closes #3805

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   test:
     name: mvn verify
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
       - name: Checkout repository
@@ -37,4 +37,4 @@ jobs:
       - name: Run mvn verify
         # FIXME(@JonasCir) see https://github.com/hzi-braunschweig/SORMAS-Project/issues/3730#issuecomment-745165678
         working-directory: ./sormas-base
-        run: mvn -T 1C verify -B -ntp
+        run: mvn verify -B -ntp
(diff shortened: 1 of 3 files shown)


vidi42:
The issue was most likely caused by slow Azure machines, because it's not reproducible anymore.

Some info about the specs of the machines against which the checks are run https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners.

Also in the next few weeks the `ubuntu-latest` will be switched to Ubuntu 20.04 which might bring some different performance number with it so it's worth to check afterwards as well. More info https://github.com/actions/virtual-environments/issues/1816

In the meantime I tried increasing parallelizing the build using approaches suggested here https://www.baeldung.com/maven-junit-parallel-tests, however the way the project is setup at the moment it's not optimized for parallel runs.

JonasCir:
Thanks for investigating, I also will keep an eye on the CI :) Could either add a FIXME with this link (https://www.baeldung.com/maven-junit-parallel-tests) to the code or a open an issue such that we do not forget about the parallel testing?

vidi42:
I've opened a new issue for this https://github.com/hzi-braunschweig/SORMAS-Project/issues/3912

PR #3889, opened 2020-12-30
#3768 - Improve handling of tag attributes and unit tests

### LINKED ISSUE #3768: HtmlHelper: Improve handling of tag attributes and unit tests [2]
### Problem Description

1. `V7UuidRenderer.encode` this is still vulnerable to HTML injection: "<a title='" + HtmlHelper.cleanHtml(value) + "'>" . Same for V7CaseUuidRenderer, ReferenceDtoHtmlProvider, CaseUuidRenderer, UuidRenderer and also maybe other classes. 
2. No unit test at all for all this stuff before and now.
3. Constant `HtmlHelper.EventActionWhitelist` violates code convention.

### Proposed Change

1. Idea: New method `HtmlHelper.cleanHtmlAttribute("title", value)`.
2. Add unit tests at least to all methods tackled by this ticket.
3. Change constant to capaitalized letters: `EVENTACTION_WHITELIST`.

danRosca:
Closes #3768

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/utils/HtmlHelper.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/utils/HtmlHelper.java
@@ -23,7 +23,7 @@ import org.jsoup.safety.Whitelist;
 // This class provides general XSS-Prevention methods using Jsoup.clean
 public class HtmlHelper {
 
-	public static final Whitelist EventActionWhitelist =
+	public static final Whitelist EVENTACTION_WHITELIST =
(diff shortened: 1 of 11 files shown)


StefanKock (changes requested):
1. There is now even more copied code, because it was not centralized to `HtmlHelper.cleanHtmlAttribute("title", value)`
2. The unit tests do not test expected good case behaviour, and also does not test was happens for example with injected `<script>`.

StefanKock:
I will start a new feature branch and

PR #3893, opened 2020-12-31
remove duplicate code, remove usage of deprecated method, NotNull

AxelNennker:

remove duplicate code by creating and using addSinceFilter
remove usage of deprecated method sampleService.createUserFilter
added some NotNull


MateStrysewske:

Issue #3717, opened 2020-12-07
Add "group by" option to campaign dashboard [2]


MartinWahnschaffeSymeda:
### Feature Description
Currently the campaign dashboard automatically groups data by the jurisidiction level below the filtering jurisdiction. E.g. when showing data of one area, it is grouped by region.

### Problem Description
Users sometimes need to be able to group by a "lower" jurisdiction level (e.g. showing data of the whole country, but grouping it by region (province).

### Proposed Change

- [x] Add a "group by" field to the campaign dashboard. Add it next to the existing filters, with some additional spacing. Values are AREA, REGION, DISTRICT, COMMUNITY. (NULL Selection is not allowed,  preselected => the sub-jurisdiction of the user)
- [x] When the area filter is set from no value to any value, automatically set the grouping to REGION. When the region filter is set from no value to any value, set the grouping to DISTRICT, etc. When any filter is set to no value, automatically set the grouping as-well.
- [x] The user should always be able to change the grouping.

PR #3901, opened 2021-01-04
#3717 - add group by jurisdiction level option to campaign dashboard

BarnaBartha:
#3717


MartinWahnschaffeSymeda:
@BarnaBartha Code looks good. Three improvements:

- [ ] Playing around with it, I feel that we should limit the grouping options based on the filter: Only allow groupings equal or below the chosen filter (e.g. don't allow to group by area, when we are filtering by a region). Only keep valid options in the drop down list.
- [ ] Only allow to group by jurisdictions two levels below the current filter (e.g. for the whole country don't allow to group by district, because there are too many. For areas don't allow to group by community).
- [ ] When I login as surveillance supervisor, the default grouping is region, but should be district. Could depend on the order the field values are set - maybe the area filter value is set last and sets the grouping to region.

Issue #3906, opened 2021-01-04
remodel AbstractAdoService to avoid UnsupportedOperationException


AxelNennker:
The method createUserFilter is abstract and the current class hierarchy leads an exception needed to be thrown when createUserFilter makes no sense for a class to implement it.

I suggest to add a new abstract class AbstractUserAdoService that has an abstract createUserFiler methods.
This class is only the a base-class when it is needed.

Avoiding UnsupportedOperationException being thrown minimizes developer surprises. With the new class hierarchy calling createUserFilter always works without unexpected exceptions.

Changes needed:
- AbstractAdoService is not abstract anymore
- a new class AbstractUserAdoService extends AbstractAdoService and createUserFilter is abstract in this class
- all classes that need createUserFilter are now based on AbstractUserAdoService

The result of this re-modelling are fewer surprises for developers

PR #3908, opened 2021-01-04
Fewer unsupported operation exception #3906

AxelNennker:
This PR minimizes the need for UnsupportedOperationException 
```
	public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<?, Visit> from) {
		// getAllUuids and getAllAfter have custom implementations
		throw new UnsupportedOperationException();
	}
```

This minimizes surprises for developers because a runtime exception is moved to a compile check.

#3906


MateStrysewske (changes requested):
The name of the new service is a bit ambiguous and kind of hints at the service being used to retrieve information from the user table. I therefore suggest to:

* Rename the new AbstractUserAdoService to AdoServiceWithUserFilter
* Rename AbstractAdoService to BaseAdoService (since it's not abstract anymore)
* Add a protected constructor to BaseAdoService to make sure that it can't be instantiated

MateStrysewske (approved):

Issue #3410, opened 2020-11-06
SurvNet Adaptations - Create new field “Reporting district” for cases and contacts [1]


markusmann-vg:
### Feature Description
Contacts > Open contact information
Cases > Open case information

New field (only DE) –
“Meldelandkreis” (de – de) |”reporting district” (en) [SurvNet field name: ReportingCounty]
(The reporting district can be different from the competent district)

Data Type: Dropdown/enum
Values (de-de) | (en)
• Namen der Landkreise (ganz Deutschland) | name of districts (prefill with the district of the reporting user)

### Additional Information

![Bildschirmfoto 2020-12-09 um 15 31 09](https://user-images.githubusercontent.com/62540178/101642788-afac9b80-3a33-11eb-9192-4cfd22217864.png)

bernardsilenou:
@markusmann-vg How does this differ from the "responsible district". When an entity is "transferred / data correction" the new responsible district is interpreted as the reporting district. My suggestion is that we map this variable to responsible jurisdiction for all entities.

kwa20:
@markusmann-vg Is the reporting district not basically the district that the Sormas instance belongs to, which sends the case to the survnet instance and eventually to the higher health institutions?

AndyBakcsy-she:
Verified issue on SORMAS Version: 1.54.0-SNAPSHOT

PR #3918, opened 2021-01-06
3410 survnet reporting district field

leventegal-she:
Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/3410

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
@@ -161,9 +161,7 @@ public class CaseDataForm extends AbstractEditForm<CaseDataDto> {
 									CaseDataDto.RABIES_TYPE))) +
 					fluidRowLocs(9, CaseDataDto.OUTCOME, 3, CaseDataDto.OUTCOME_DATE) +
 					fluidRowLocs(3, CaseDataDto.SEQUELAE, 9, CaseDataDto.SEQUELAE_DETAILS) +
-					fluidRowLocs(CaseDataDto.REPORTING_TYPE,
-							"")
-					+
+					fluidRowLocs(CaseDataDto.REPORTING_TYPE, CaseDataDto.REPORTING_DISTRICT) +
 					fluidRowLocs(CaseDataDto.CASE_ORIGIN, "") +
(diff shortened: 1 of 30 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
```diff
@@ -309,9 +307,9 @@ protected void addFields() {
 		addField(CaseDataDto.INVESTIGATION_STATUS, NullableOptionGroup.class);
 		addField(CaseDataDto.OUTCOME, NullableOptionGroup.class);
 		addField(CaseDataDto.SEQUELAE, NullableOptionGroup.class);
-		if (isConfiguredServer(CountryHelper.COUNTRY_CODE_GERMANY)) {
-			addField(CaseDataDto.REPORTING_TYPE);
-		}
+
+		addField(CaseDataDto.REPORTING_TYPE);
```
vlad-ciucescu, line 311: Is it needed to make this available for all countries?

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
```diff
@@ -309,9 +307,9 @@ protected void addFields() {
...
+		((ComboBox) addField(CaseDataDto.REPORTING_DISTRICT)).addItems(FacadeProvider.getDistrictFacade().getAllActiveAsReference());
```
vlad-ciucescu, line 312: Is this needed for all countries or only for Germany? From the story my understanding is that it should be for Germany only.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
leventegal-she: It will be hidden for other countries because of the @HideForCountrysExcept annotation in CaseDtaDto.

The idea behind this annotation was to eliminate this kind of if statements.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
leventegal-she: It will be hidden for other countries because of the @HideForCountrysExcept annotation in CaseDtaDto.

Loading all districts could cause performance issues for other countries but it's not a big query so it should not really affect the performance

vlad-ciucescu (approved):

Issue #3902, opened 2021-01-04
AggregateReportService currentUser possible null pointer exception


AxelNennker:
The code here makes no sense
https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java#L88
The same issue is in WeeklyReportService, TaskService as well in the method createUserFilter.

Either currentUser is null then there would be a null pointer exception in line https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java#L87
Or currentUser is cannot be null then the if-statement can be simplified.

Some implementations of createUserFilter test for currentUser being null before it is dereferenced.

if currentUser can be null then this code leads to a null pointer exception.

PR #3920, opened 2021-01-06
3902 condition is always true

AxelNennker:
This PR fixes https://github.com/hzi-braunschweig/SORMAS-Project/issues/3902


MateStrysewske (approved):
Just a note: There are build problems when adding @NotNull annotations to overridden methods without also adding the @NotNull annotation to the base method. We fixed this for saveEvent already, so nothing you need to do, but please keep it in mind. :)

PR #3921, opened 2021-01-06
2947 type of place details in events entities

sergiupacurariu:


sormas-backend/src/main/java/de/symeda/sormas/backend/event/Event.java
```diff
@@ -304,6 +310,32 @@ public void setMeansOfTransportDetails(String meansOfTransportDetails) {
 		this.meansOfTransportDetails = meansOfTransportDetails;
 	}
 
+	@Column(columnDefinition = "text")
```
BarnaBartha, line 313: are you expecting here string of unlimited size ? i think it would be better to define it as varchar with limited size (the current default in sormas is 512)

BarnaBartha (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/event/Event.java
sergiupacurariu: updated

BarnaBartha (approved):

PR #3929, opened 2021-01-07
#3573 - add description to discard button + navigate back to campaign…

### LINKED ISSUE #3573: Go back to campaign data list view after editing campaign data [0.5]
### Problem Description
When pressing the save or discard button in the campaign data form, the view currently stays active. For users this is a bit confusing, because in contrast to cases and contacts the campaign data has only one tab/form.

### Proposed Change

- [x] When pressing discard or save in the campaign data form, close the form and go back to the campaign data list view
- [x] Add a tooltip to the discard button saying "Discards any unsaved changes".

BarnaBartha:
Closes #3573


sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/CampaignController.java
```diff
@@ -219,7 +215,20 @@ public void discard() {
 		}
 		form.setValue(campaignFormData);
 
-		final CommitDiscardWrapperComponent<CampaignFormDataEditForm> component = new CommitDiscardWrapperComponent<>(form, form.getFieldGroup());
+		final CommitDiscardWrapperComponent<CampaignFormDataEditForm> component = new CommitDiscardWrapperComponent(form, form.getFieldGroup()) {
```
MartinWahnschaffeSymeda, line 218: @BarnaBartha Please use CommitDiscardWrapperComponent.addCommitListener instead

MartinWahnschaffeSymeda (changes requested):

MartinWahnschaffeSymeda:
@JonasCir I ran into a strange issue here. The unit test `I18nConstantsUpdatedTest.testConstantsAreUpdated` is failing. The reason is that `I18nConstantGenerator.writeI18nConstantClass` seems to receive the content of a property file from another branch when calling `I18nProperties.class.getClassLoader().getResourceAsStream(propertiesFileName)`. Seems like this is related to the caching mechanism used by the CI server.

I'll work around this by merging the development branch into this branch, but it would be good to at least understand what is going on here. Do you have an idea?

JonasCir:
@MartinWahnschaffeSymeda Not really :D One thing that I could think of is that some temporary files are written to the directories that are cached and under rare circumstances don't get cleaned up properly and therefore carry over to other branches, but I am guessing. From other projects, I know, that CI caching sometimes just behaves strange, so a simple "have you tried to turn it on and off again" helps :D I keep an eye on it

Issue #3871, opened 2020-12-23
Contacts > Contact Directory > Empty date range 'From... to' results in error


Candice-Louw:
### Bug Description
Removing the dates 'from... to' for Contacts (follow-up visits view), results in an error.

### Steps to Reproduce
1. Contacts > Contact Directory > Follow-up Visits view
2. Remove dates in 'From... To'
3. Error

### Expected Behavior
No error / warning message / other proposed solution?

### Screenshots
![image](https://user-images.githubusercontent.com/65529128/103007942-9546e780-4534-11eb-8c26-4864cbee0003.png)

### System Details
* Device: PC/Web
* SORMAS version: 1.53
* Android version/Browser: Chrome
* Server URL:  https://release-de.sormas.netzlink.com/
* User Role: admin

### Additional Information
Tested with PIA Symptomtagebuch configuration.

bernardsilenou:
I think we should show an error massage saying "Please specify a valid date range"

AndyBakcsy-she:
Validated issue on the latest version of SORMAS Version: 1.54.0-SNAPSHOT

PR #3937, opened 2021-01-08
#3871 contacts contact directory empty date range from to results in error

alexcaruntu-vita:


MateStrysewske (changes requested):
The "Please specify a valid date range" message pops up when the "to" date is removed, but not when the "from" date is removed; I think there's a short piece of code missing in the fromReferenceDate listener that is still present in the toReferenceDate listener (see lines 642 to 646 compared to the same code part in the fromReferenceDate listener)

MateStrysewske (approved):

PR #3943, opened 2021-01-08
Rename replyingUser to lastModifiedByUser and display the creator when it isn't set

### LINKED ISSUE #3719: Actions directory - the name of the person who wrote the action does not appear when creating the action
### Bug Description
In the event part, followed by actions, it is possible to describe an action carried out within the framework of the investigation. After creating the action, it goes to the actions directory. Contrary to what is expected, the person who wrote the action does not appear in the directory under the "action replying user" field. His name only appears after modifying or completing the action.

### Steps to Reproduce
1. create an action
2. save an action
3. in the directory, the name does not appear

### Expected Behavior
1. create an action
2. save an action
3. in the directory, the name should appear

### System Details
* Device:
* SORMAS version: 1.51.1
* Android version/Browser:
* Server URL:
* User Role:

### Additional Information
@ftavin @bernardsilenou

alexandre-gille:
Fixes #3719


MateStrysewske (changes requested):
The word "user" is pretty redundant for the fields that you've renamed; "replyingUser" makes sense because "replying" in itself doesn't hint at the field containing a user; "lastModifiedBy" or "actionLastModifiedBy" indicates that pretty clearly though. Could you therefore please remove "user" from the field names that you've changed (unless it's necessary to understand what the field is about)?

MateStrysewske (changes requested):
The logic that you've applied to the index list (show the last modifying user, if the action hasn't been modified yet, show the creator instead) doesn't work for the detailed export yet.

alexandre-gille:
Well, I thought this behavior was relevant only because it prevents us from having too many columns on the data-grid and therefore wasn't required on exports

MateStrysewske:
@alexandre-gille That's true, it might be fine to have it split into two columns in the export, but right now there's only one "Action last modified by" column which does NOT include the creator if the action wasn't modified yet, so it's actually missing information that is present in the grid.

alexandre-gille:
So, you wish to keep it as is ? Or I revert this and add a dedicated column for the creator ?

MateStrysewske:
@alexandre-gille You could either replicate the logic from the index list (one column used for both) or you could add another column for the creator; both is fine by me. I think I prefer the first solution because it's more in line with the index list, but in that case you need to fix the problem that the column entry is empty if there is no last modifying user (because the creator is not shown then).

alexandre-gille:
Do you imply there is currently a bug in this dev ? Or you just didn't see I already pushed the a commit to modify this export ? 
Because I tested it again and I everything seems to work fine.

MateStrysewske:
@alexandre-gille I actually missed the commit, yes. Sorry!

MateStrysewske (approved):

alexandre-gille:
No problem ;)

PR #3959, opened 2021-01-11
Send a notification to event participants related to a confirmed case

### LINKED ISSUE #3794: Notification to the investigator of an event when a person attending an event becomes positive for COVID
### Situation Description
The management and investigation of an event leads teams to understand the characteristics of an event, the actions implemented and the people participating in the event. To date, the list of people participating in an event is static and has not changed. However, it would be interesting for SORMAS to alert the person in charge of the investigation when a person becomes positive for COVID and becomes a case. The objective would be to optimize the uses of SORMAS, in particular for monitoring clusters and chains of contamination by developing a mechanism that informs the investigator as quickly as possible of the evolution of a status of a person involved in an event. 

### Feature Description
-	The creation of a case / the transformation of a contact in the event of a person participating in an event triggers an automatic task (note, this automatic task should be individualized because the activation of all automatic tasks is inconceivable) and a notification addressed to the person in charge of the investigation. For contacts of a case and / or for participants in an event (eg screened in a group)
-	Ideally and if possible, the events for which the mechanism is triggered would be configurable : only the events for which this option is activated would be subject to such monitoring. The advantage of making this mechanism optional is that it does not increase the sending of notifications and limit them only to events / clusters deemed critical.

### Additional Information
@ftavin @bernardsilenou @Olivierleuci

aponcon-atolcd:
Fixes #3794


sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
```diff
@@ -145,6 +145,44 @@ public EventService() {
...
+				cb.isNotNull(eventJoin.get(Event.SURVEILLANCE_OFFICER)),
+				cb.not(cb.equal(eventJoin.get(Event.SURVEILLANCE_OFFICER), caze.getReportingUser())));
+
+		cq.where(
+				cb.and(
+						diseaseFilter,
+						personFilter,
+						dateFilter,
+						surveillanceOfficerFilter,
+						cb.isFalse(eventJoin.get(Event.DELETED)),
```
MateStrysewske, line 178: Can you use `createActiveEventsFilter` here or isn't that possible because you're using event participant as root?

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #3963, opened 2021-01-11
#3945: add new logger to logback to disable debug logs for Apache Http client

### LINKED ISSUE #3945: Credentials logged
### Bug Description
When accessing SORMAS, secured with keycloak, via sormas-rest (BASIC Auth),
it has been noticed that the log contains user credentials.
The statement seems to be lacking a logger name and contains the term "grant type".

### System Details
* SORMAS version: 1.52.1

vidi42:
Closes #3945


StefanKock:
Since no logger statements are documented to be suppressed, I am not able to review this. Maybe there could be a logger config done on class level.

StefanKock (commented):
Since no logger statements are documented to be suppressed, I am not able to review this. Maybe there could be a logger config done on class level.

vidi42:
Update to only include the wire logs in apache http components.
The log `org.apache.http.wire` is defined at `org.apache.http.impl.conn.ManagedHttpClientConnectionFactory:66`

vidi42:
Updated to suppress all `org.apache.http` logs which contain headers and wire logs intended for debugging.
https://hc.apache.org/httpcomponents-client-4.5.x/logging.html

Issue #3524, opened 2020-11-18
SurvNet Adaptation -  Dedicated fields for technical and non-technical external IDs [2]


HolgerReiseVSys:
### Situation Description
In SurvNet all transferred entities (case, contact, person, outbreak,organisation etc.) have a technical (GUID) and a non-technical (Token; "Aktenzeichen", "Kurzname") id.
SORMAS only features one field "externalID" ("externalId", for Persons).

### Feature Description
Fields for both should be available.

Current situation:
- ExternalID is used technical and non-technical

Wanted:
- ExternalID will be renamed to ExternalToken ("Aktenzeichen" in Germany)
- new field for ExternalID and will be used for identifying the GUID in SurvNet

### Additional Information
The technical id (ExternalID) might have to be visible all the time.
It yields a specific format and should usually not be edited manually.

Production systems in Germany probably used the externalID field for the non-technical id.

markusmann-vg:
Add it to cases, contacts, events and persons

cazacmarin:
@markusmann-vg @HolgerReiseVSys 
I saw the misconfiguration for those entities related to some fields.
Cases and contacts for externalID has:
@HideForCountriesExcept(countries = {
      COUNTRY_CODE_GERMANY,
      COUNTRY_CODE_SWITZERLAND })
Person has:
@HideForCountriesExcept(countries = CountryHelper.COUNTRY_CODE_GERMANY)
Events - has non from above.
Question: Should I use the same signature from Cases and Contacts for Person and Events? Or I should leave it as is.

bernardsilenou:
@HolgerReiseVSys @markusmann-vg @carolinverset @ftavin
The French users get data set of contacts from external apps, import in SORMAS and does the management. Each data set of contacts comes with a code for all the records and users are required to map each record in SORMAS with the external datasets in the other app. 
They requested to have a second external id and I thought this ticket may server the purpose.
Thus, 
- maintain the old external id as the technical external id 
- add the new non-technical external id
- Activate both for all instances

What is you thought on this please? 
It would be nice no to add a third kind of external id

cazacmarin:
ExternalToken field for cases, contacts, events and persons was added. As well for mobile app.

kwa20:
@HolgerReiseVSys @JonasCir Do we know how we handle previous entries in the external ID field not related to SurvNet?

JonasCir:
@kwa20 i have no idea tbh.

[...]

PR #3966, opened 2021-01-11
#3524 - SurvNet Adaptation - Dedicated fields for technical and non-t…

cazacmarin:
…echnical external IDs (for cases, contacts, events and persons) - as well for mobile app.


JonasCir:
@HolgerReiseVSys FYI

Issue #3508, opened 2020-11-18
Sort areas/regions/districts alphabetically in campaign charts [0.5]


MartinWahnschaffeSymeda:
### Proposed Change
Sort the entries on the x-axis of campaign charts alphabetically (areas/regions/districts/communities).
See CampaignDashboardDiagramComponent.axisKeys.

### Additional Information
Can be done in the ui. I18n is not relevant here.

PR #3970, opened 2021-01-12
#3508 - Order axis keys alphabetically in campaign charts

alexcaruntu-vita:


MartinWahnschaffeSymeda:
@BarnaBartha Can you have a look at this, please? I suspect that simply reorderung the axis labels will not work, since the series data has to match it, but I'm not sure.

BarnaBartha:
@MartinWahnschaffeSymeda @alexcaruntu-vita Indeed sorting the x-axis categories does not solve it, the series data will be assigned to the wrong caption. What i suggest we do is to remove the axisKeys list and rename the axisCaption to xAxisInformation (or something similar) make in a LinkedHashMap and sort it via the caption values and use instead of the current axisKeys xAxisInformation.keySet()

sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/campaigns/CampaignDashboardDiagramComponent.java
```diff
@@ -225,12 +230,12 @@ private void appendData(
 		StringBuilder hcjs,
 		CampaignDiagramSeries series,
 		Map<Object, CampaignDiagramDataDto> seriesData) {
-		for (Object axisKey : axisKeys) {
-			if (seriesData.containsKey(axisKey)) {
+		for (Object axisInfo : xAxisInfo.keySet()) {
```
BarnaBartha, line 233: i think this should remain as axisKey

sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/campaigns/CampaignDashboardDiagramComponent.java
```diff
@@ -66,23 +65,29 @@ public CampaignDashboardDiagramComponent(
...
+			if (objectCampaignDiagramDataDtoMap.containsKey(groupingKey)) {
 				throw new RuntimeException("Campaign diagram data map already contains grouping");
 			}
-			objectCampaignDiagramDataDtoMap.put(diagramData.getGroupingKey(), diagramData);
+			objectCampaignDiagramDataDtoMap.put(groupingKey, diagramData);
 		}
 
+		xAxisInfo = axisInfo.entrySet()
+			.stream()
+			.sorted(Map.Entry.comparingByValue())
```
BarnaBartha, line 88: with my example data i found it would be nice to have a case insensitive sorting here (that is a should be before W)

Issue #3652, opened 2020-12-01
Allow only admins to create new campaigns for now [0.5]


tomabadaniel:
### Feature Description
RIght now, configuration of new campaigns are done at JSON definition level. When users create a new campaign, they do not configure the charts, forms etc so the dashboard structure is always affected. 
### Proposed Change
As we do not have a "replicate campaign" feature(https://github.com/hzi-braunschweig/SORMAS-Project/issues/3289), whereby the user can replicate a previous campaign and edit it for a new campaign, which preserves the structure and framework of the campaign, Let only admins be able to create new campaigns for now. Once we have that feature in place, we can open the access back to the national users and all.

MartinWahnschaffeSymeda:
- [ ] Also campaign form validation should only be available for admins.

AndyBakcsy-she:
verified issue on SORMAS Version: 1.54.0-SNAPSHOT

MartinWahnschaffeSymeda:
Looks like regional and district users are still able to edit campaigns. This should not be the case. Only national users and admins should be able.

BarnaBartha:
Refine:
- [ ] do we need separation for edit and create rights for Campaigns (create only admin, edit admin & national user)?
- [ ] what exactly should happen when not having edit right (everything disabled on detail view / only save button disabled) ?

MartinWahnschaffeSymeda:
We should implement it like this:

- [ ] When the user has only view and not edit rights, open the form as read-only (as we do for campaigns, etc.)
- [ ] In the campaigns list (see picture below) use "view" together with the vaadin "eye" icon to open the campaign form, when the user has no edit right (instead of "edit" and the "edit" icon)
- [ ] remove the "open" column that currently switches to the campaign data list. Since the campaign data list is the default view anyway and comes with a filter, this is no longer needed and would be confusing to keep

![grafik](https://user-images.githubusercontent.com/23217632/104026077-c3c7f200-51c5-11eb-823d-0ee3110de5d0.png)

@danieltomaba As originally planned, we would like to keep this simple and limit creation *AND editing* of campaigns to admins (not national users). Is there a use case that is important for the pilot that would require us to split edit and create user rights?

MartinWahnschaffeSymeda:
- [x] Also fix the campaign menu entry not being selected when opening the campaigns view

PR #3972, opened 2021-01-12
#3652 - use campaign edit right in campaign grid and campaign directory

BarnaBartha:
#3652


MartinWahnschaffeSymeda:
@BarnaBartha Looks good. I have moved the edit check to the controller, though

Issue #3873, opened 2020-12-23
Missing lastContactDate in contact import results in missing follow-up calculation


kwa20:
### Bug Description
Contacts can be imported with a missing date of last contact. This, however, is stated as a required field for the import. If this field is left empty, the calculation does not fall back to the report date and just never gets calculated until the lastContactDate is given.
### Steps to Reproduce
1. Import contacts without a lastContactDate
2. Check Follow-up section

### Expected Behavior
Follow-up should also be calculated when contacts were imported, conform to how it is done via a normal contact creation.
### Screenshots
<img width="435" alt="contactImportFollowUp" src="https://user-images.githubusercontent.com/70317594/103018640-e01d2b00-4545-11eb-8ea8-68437d493a46.PNG">
<img width="278" alt="contactImportFollowUp2" src="https://user-images.githubusercontent.com/70317594/103018643-e14e5800-4545-11eb-8b71-47d16a1f7a5f.PNG">

### System Details
* Device:
* SORMAS version: 1.52.1
* Android version/Browser: firefox
* Server URL: test1402.sormas-oegd.de
* User Role: admin, national user

bernardsilenou:
* date of last contact should not be a required field when importing contacts because it is not required when creating contacts from front end.
* Recalculate followup until date each time the contact, case or event participant entity is saved (initial creating + update). 
* The same need as in #3864, thus implement both as one

kwa20:
@bernardsilenou you're right, the problem here is not lastContactDate not being a required field but the missing initial calculation of the follow-up when the contact was created via import

raghupola-vg:
Tested by importing a contact and works fine on latest build.

Issue #3424, opened 2020-11-06
Update automatic "Cancel follow up" flow [1]


markusmann-vg:
There is a followUpStatus for Contacts and for Cases.  It's an enum with the following values:
`FOLLOW_UP,
COMPLETED,
CANCELED,
LOST,
NO_FOLLOW_UP;`
We should use this to synchronize with climedo, and make sure that the followUpStatus is updated in any reasonable situation 
- [ ] when a Contact is dropped
- [ ] when a Contact is converted to a Case in a system that has Case follow up disabled
- [ ] when a Case classification is set to 'Not a case' in a system with Case follow up enabled. Here, we need to make sure that we don't unreasonably cancel Person's follow up. Imagine a Person has a Contact. The Contact is the converted to a Case. Because of that, the Contact follow up is disabled. Now the Case classification is set to 'Not a case' which disables the person's case follow up as well. We should not have a Person who had a valid (yet converted) Contact and no follow up at all.

_Originally posted by @FredrikSchaeferVitagroup in https://github.com/hzi-braunschweig/SORMAS-Project/issues/3294#issuecomment-721538138_

markusmann-vg:
results from: https://github.com/hzi-braunschweig/SORMAS-Project/issues/3294

BarnaBartha:
After some clarification with Fredrik:
- [ ] when a contact is dropper => set follow-up status to 'Canceled'
- [ ] when a Contact is converted to a Case in a system that has Case follow up disabled => set follow-up status to 'Canceled'
- [ ] when a Case classification is set to 'Not a case' in a system with Case follow up enabled. Here, we need to make sure that we don't unreasonably cancel Person's follow up. Imagine a Person has a Contact. The Contact is the converted to a Case. Because of that, the Contact follow up is disabled. Now the Case classification is set to 'Not a case' which disables the person's case follow up as well. We should not have a Person who had a valid (yet converted) Contact and no follow up at all. => set all converted contacts of case back to follow up status ' follow up' where the followUpUntil date is not in the past and contact status is not set to dropped  (do also when Case follow up is disabled)

Candice-Louw:
@BarnaBartha - _when a Contact is converted to a Case => set follow-up status to 'Canceled'_ - would this logic be applied when importing Contacts (line listing or file) as well, please?  Should we create a seperate issue for this?

[...]

Issue #2641, opened 2020-08-12
Status is not persisted in the import of contacts 


markusmann-vg:
### Bug Description
Problem: We want to import (CONFIRMED) contacts with the status Tracking completed (COMPLETED), and the status dropped (DROPPED), because the isolation time is already over. 
With a few test entries we then noticed that the information in the csv (COMPLETED and DROPPED) is overwritten by SORMAS and is displayed in Sormas as FOLLOW_UP (in tracking) and ACTIVE (active contact). 

### Steps to Reproduce
1.see problem

### Expected Behavior
The imported values must be adopted and not overwritten by the import process of SORMAS?

### System Details
* Device:
* SORMAS version:
* Android version/Browser:

bernardsilenou:
I think this is caused by the fact that completed means there is a followup visit registered on the followup until date.
While implementing this issue, I suggest we also implement issue #1855. Contacts status is set to dropped when followup is completed. This is not right because all contacts will definitely become dropped after 14 days. Just as in cases, we should use 'dropped' to mean 'discarded'.

PR #3976, opened 2021-01-12
#3873 - execute follow up calculation logic also for import

BarnaBartha:
- the issue due to which this was ignored for import from #2641 has been fixed in the meantime with #3424


MateStrysewske:
@BarnaBartha Seems like there are errors in the code, can you check that?

MateStrysewske (approved):

PR #3981, opened 2021-01-12
Optimized contact, case, deathcount queries #3276

DavidBaldsiefen:
This PR implements #3276 and optimizes the queries implemented with PR #3772.


Seal33:
@MateStrysewske

This is the optimized solution I came up for instead of #3367 
I've reduced it to two queries (one for participantcount, deathcount, casecount and one for contact counts).

Please not this is still WIP, but I wanted to check back with you if you are fine with the queries I've implemented in `EventFacadeEjb.getIndexList(...)`. I am still trying to find ways to further optimize these queries.

The new generated queries are:
`select event2_.uuid as col_0_0_, count(eventparti0_.id) as col_1_0_, sum(cast(case when eventparti0_.resultingCase_id is not null then 1 else 0 end as int8)) as col_2_0_, sum(cast(case when case1_.outcome='DECEASED' then 1 else 0 end as int8)) as col_3_0_ from EventParticipant eventparti0_ left outer join cases case1_ on eventparti0_.resultingCase_id=case1_.id cross join events event2_ where eventparti0_.event_id=event2_.id and eventparti0_.deleted=false group by event2_.uuid`
image


`select event2_.uuid as col_0_0_, count(eventparti0_.id) as col_1_0_, sum(cast(case when exists (select eventparti3_.id from EventParticipant eventparti3_ where eventparti3_.resultingCase_id=contact1_.caze_id and eventparti3_.event_id=eventparti0_.event_id) then 1 else 0 end as int8)) as col_2_0_ from EventParticipant eventparti0_ cross join Contact contact1_ cross join events event2_ where eventparti0_.event_id=event2_.id and eventparti0_.person_id=contact1_.person_id and eventparti0_.deleted=false and contact1_.deleted=false group by event2_.uuid`
image


Using pgAdmins Query Explain Analyze function, I've measured the runtimes on my local DB as follows
Timings|Planning Time(ms) | Execution Time(ms)
-|- | -
New Queries| 0.576 | 0.491
Old Queries| 1.641 | 0.578

Obviously one would need to test this with a much larger database for more realistic timings.

MateStrysewske:
The only thing that looks a bit suspicious is the two eventparticipant invocations into the GroupAggregate in the contact query; not sure whether it's possible to get rid of those though. Apart from that, I think it looks good! It's probably a good idea to create missing indexes for eventparticipant because that tends to significantly increase performance especially on systems with a lot of data.

Seal33:
Thanks for the reply, I will!
Yes, I've been wondering about those seemingly redundant invocations too, but they appear to be caused by query optimization and I haven't yet found a way get rid of them.

By missing indexes, do you mean something like `EventParticipant.personContacts`, so that I can directly reference those instead of going EventParticipant->Person<-Contact?

Seal33:
The feature would (finally) be ready for review, so hopefully this can be part of 1.55.

I've decided that it would look best if I use a small ComboBox for selecting the contact count calculation method, as it looks much better than adding an out-of-place checkbox. In my opinion it also doesn't make much sense to add the method to the filters, as that would require reloading the table everytime it is changed (including re-executing all queries which aren't cached). I would also have to split the contact count query into two.


image

The two additional scans for the second query cannot be removed, as they are required for the `where eventparti3_.resultingCase_id=contact1_.caze_id and eventparti3_.event_id=eventparti0_.event_id`.
Adding the above mentioned indexes should also boost performance a good bit on those queries.

Apart from that, I've tested the queries for general functionality, and found that they seem to work as intended. I will solve sortability in another issue after this is merged ( #3772 ).

DavidBaldsiefen:
@MateStrysewske Could you review this?

sormas-api/src/main/java/de/symeda/sormas/api/event/EventIndexDto.java
```diff
@@ -36,6 +36,8 @@
 	public static final String PARTICIPANT_COUNT = "participantCount";
 	public static final String CASE_COUNT = "caseCount";
 	public static final String DEATH_COUNT = "deathCount";
+	public static final String CONTACT_COUNT = "contactCount"; // number of contacts whose person is involved in this event
```
MateStrysewske, line 39: Move the comments to the actual field definitions (as JavaDoc comments, not in-line comments)

sormas-api/src/main/java/de/symeda/sormas/api/event/EventIndexDto.java
```diff
@@ -36,6 +36,8 @@
...
+	public static final String CONTACT_COUNT_SOURCE_IN_EVENT = "contactCountSourceInEvent"; // number of contacts whose person is involved in this event, and whjere the source case is also part of the event
```
MateStrysewske, line 40: Typo ("whjere")

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6291,4 +6291,12 @@ ALTER TABLE labmessage ADD COLUMN messagedatetime timestamp;
 ALTER TABLE labmessage_history ADD COLUMN messagedatetime timestamp;
 
 INSERT INTO schema_version (version_number, comment) VALUES (310, 'Fix labmessage table #3486');
+
+-- 2020-01-13 Add indexes to optimize event directory performance #3276
+CREATE INDEX IF NOT EXISTS idx_eventparticpant_person_id ON eventparticipant USING hash (person_id);
```
MateStrysewske, line 6296: Typo ("eventparticpant"), also in the line below

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventGrid.java
```diff
@@ -180,6 +189,16 @@ private String buildSourceMediaText(EventIndexDto event) {
 		return (srcMediaWebsite != null ? srcMediaWebsite : "") + " " + (srcMediaName != null ? "(" + srcMediaName + ")" : "");
 	}
 
+	public void setContactCountMethod(boolean SourceInEvent) {
```
MateStrysewske, line 192: use camelcase

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventGrid.java
```diff
@@ -180,6 +189,16 @@ private String buildSourceMediaText(EventIndexDto event) {
...
+		if (SourceInEvent) {
```
MateStrysewske, line 193: Can be written shorter like this:
```
getColumn(EventIndexDto.CONTACT_COUNT_SOURCE_IN_EVENT).setHidden(!sourceInEvent);
getColumn(EventIndexDto.CONTACT_COUNT).setHidden(sourceInEvent);
```

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsView.java
```diff
@@ -92,6 +92,9 @@
 	private Label relevanceStatusInfoLabel;
 	private ComboBox relevanceStatusFilter;
 
+	// calculation method to use for contact counts
```
MateStrysewske, line 95: I don't think the comment is necessary here, the field is pretty self-explanatory; if you want to use a comment, please use a JavaDoc style comment instead (since it's about describing a class field)

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsView.java
```diff
@@ -411,6 +414,7 @@ public HorizontalLayout createStatusFilterBar() {
 				relevanceStatusFilter.setId("relevanceStatusFilter");
 				relevanceStatusFilter.setWidth(140, Unit.PERCENTAGE);
 				relevanceStatusFilter.setNullSelectionAllowed(false);
+				relevanceStatusFilter.setTextInputAllowed(false);
```
MateStrysewske, line 417: What's the reason for this change?

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsView.java
```diff
@@ -445,8 +450,27 @@ public HorizontalLayout createStatusFilterBar() {
...
 				bulkOperationsDropdown.setVisible(viewConfiguration.isInEagerMode());
+				bulkOperationsDropdown.setCaption("");
 				actionButtonsLayout.addComponent(bulkOperationsDropdown);
 			}
+
+			if (isDefaultViewType()) {
+				// Contact Count Method Dropdown
+				contactCountMethod = new ComboBox();
+				contactCountMethod.setCaption(I18nProperties.getCaption(Captions.Event_contactCountMethod));
+				contactCountMethod.addItem(0);
```
MateStrysewske, line 461: I would prefer to add a dedicated ContactCountMethod enum (can probably be in the UI since it's not needed in the API or backend) with the two values, and therefore to also move the translations to enum.properties. You can then also use this enum for the `setContactCountMethod` method in EventGrid.java instead of passing a boolean.

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsView.java
DavidBaldsiefen: I stumbled upon the fact that you could currently write text in the relevance filter, which doesn't make any sense as the typed-in-text wasn't processed anyway. So I thought I'd quickly disable it while working in the same file anyway, for a more cleaner UI.

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventContactCountMethod.java
```diff
@@ -0,0 +1,6 @@
+package de.symeda.sormas.ui.events;
+
+public enum EventContactCountMethod {
+	COUNT_ALL_CONTACTS,
```
MateStrysewske, line 4: I'd suggest to use shorter names here:

ALL,
SOURCE_CASE_IN_EVENT

since the enum name already suggests that it's about contact counts

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventGrid.java
```diff
@@ -189,14 +189,9 @@ private String buildSourceMediaText(EventIndexDto event) {
...
-	public void setContactCountMethod(boolean SourceInEvent) {
-		if (SourceInEvent) {
-			getColumn(EventIndexDto.CONTACT_COUNT_SOURCE_IN_EVENT).setHidden(false);
-			getColumn(EventIndexDto.CONTACT_COUNT).setHidden(true);
-		} else {
-			getColumn(EventIndexDto.CONTACT_COUNT).setHidden(false);
-			getColumn(EventIndexDto.CONTACT_COUNT_SOURCE_IN_EVENT).setHidden(true);
-		}
+	public void setContactCountMethod(EventContactCountMethod method) {
+		getColumn(EventIndexDto.CONTACT_COUNT_SOURCE_IN_EVENT).setHidden(!(method == EventContactCountMethod.COUNT_ALL_CONTACTS));
```
MateStrysewske, line 193: Don't these two code lines mean that both columns are displayed when the count method is "all" and none of them is displayed when the count method is "source case in event"?

MateStrysewske (changes requested):

MateStrysewske (changes requested):
Looks good! The only thing I'd like to see changed here is that the column that lists the amount of contacts with their source case in the event should have a different caption in the detailed export - right now it's "Only count contacts [...]" which is fine for the option, but not a good heading. A better heading would be "Contacts with source case in event".

DavidBaldsiefen:
> Looks good! The only thing I'd like to see changed here is that the column that lists the amount of contacts with their source case in the event should have a different caption in the detailed export - right now it's "Only count contacts [...]" which is fine for the option, but not a good heading. A better heading would be "Contacts with source case in event".

The easiest way to do that would be to change the method drop-down options to "contacts with source case in event" and "all contacts". Given that the caption above the dropdown is "Contact Count Method", it should still be self-explanatory. Or what do you think?

DavidBaldsiefen:
I have fixed the export captions. I've also added an option to show both calculation methods at once. This way, the user can also see both columns in the detailed export.

MateStrysewske (approved):

PR #3988, opened 2021-01-13
Merge branch 'development' into feat-3977_keycloak_password_policy

### LINKED ISSUE #3977: Update Keycloak Default Password Policy [0.5]
### Situation Description

At the moment there is no Password Policy defined by default for systems that choose to use Keycloak as their Authentication Provider and each System owner is responsible of defining one.

### Feature Description

Since having no Password Policy should not be the case for any system and to help any new System Owner during the setup we would setup a default set of rules which are already required by some Health Centers.

Length of minimum 12 characters
At least 1 upper case letter 
At least 1 lower case letter
At least 1 digit
At least 1 special character

vidi42:
Closes #3977


sormas-base/doc/keycloak.md
```diff
@@ -0,0 +1,75 @@
...
+Email configurations are optional and are not part of the default configuration.
+
+In case the system relies on users activating their own accounts it's required to configure these settings.
+
+#### Custom Configuration
+
+The configuration provided by default is the minimum required configuration for Keycloak to work together with SORMAS.
+
+The Keycloak configuration can be adjusted by any user with admin rights, however **beware** that any change to the default 
+configuration might rend the system unusable.
```
vlad-ciucescu, line 65: render

vlad-ciucescu (approved):

chThie:
https://github.com/hzi-braunschweig/SORMAS-Project/issues/3977#issuecomment-759441542

PR #3989, opened 2021-01-13
Feature #3768 improve handling of tag attributes

StefanKock:


chThie:
I think the method `public static String buildHyperlinkTitle(String title, String caption)` is still vulnerable for javascript injection. It is possible to break out of the title attribute of the <a> tag with a single quotation mark and this allows to add additional attributes to the tag.  
An attacker could add an href attribute and by this execute javascript when someone clicks on the link.

Example:
title: `test' href="javascript:alert('Test');" class='`
caption: `Test`
output: `<a title='test' href="javascript:alert('Test');" class=''>Test</a>`

Please correct me if this prevented somewhere else or if I got something wrong.

StefanKock:
> I think the method `public static String buildHyperlinkTitle(String title, String caption)` is still vulnerable for javascript injection. It is possible to break out of the title attribute of the tag with a single quotation mark and this allows to add additional attributes to the tag.

Thanks for contributing. I now prevent any but the intended tag attribute "title"

Issue #3835, opened 2020-12-18
Text search on directories(configuration, cases, event participants, contacts etc) loses focus


StefanKock:
### Bug Description

Using text search for example on `sormas-ui/#!configuration/countries` is not focussed anymore after a short time delay (different behaviour than on other views).

### Steps to Reproduce
1. Go to countries page.
2. Type some text into the text search field and wait some seconds.
3. Grid is reloaded and focus on textfield is lost, so keep typing does not go into the field, and trying to remove the text result in switching to last URL in browser history.

### Expected Behavior

Do not drop focus as for example UsersView text search.

### System Details
* SORMAS version: 1.53.0-SNAPSHOT
* Browser: Firefox
* Server URL: https://test-de.sormas.netzlink.com

bernardsilenou:
Here is my comment copied from the duplicate issue #3877

The goal of #2225 and #2542 was to solve this issue. However, letters are deleted when system does the lazy loading.

I do not know the best solution rather than recommending that we keep all the letters typed by the user on the screen and only activate the filter when the user press the "Enter" or apply filter icon key
This is also needed for sub directories like event participants, contacts (within a case), ...

ftavin:
@bernardsilenou I'm not sure this is exactly the same issue than #3877 : in our case the text field don't lose focus

But yes, both issues can probably be solved the same way if the cause is the same (refresh of the form)

bernardsilenou:
I would reopen #3877. The person assigned to any of the tickets should also have a look at the associated one.

StefanKock:
I closed the duplicate and added this ticket to the sprint.

carolinverset:
@StefanKock 
I have a little question : Do you think this bug will be fixed in version 1.54? many users ask us about this question. 
Thanks a lot

AndyBakcsy-she:
Validated issue on the latest version of SORMAS Version: 1.54.0-SNAPSHOT on test.sormas.netzlink.com

PR #3990, opened 2021-01-13
#3835 text search on directories loses focus

alexcaruntu-vita:


StefanKock (approved):
I reviewed the code and tested it, no objections. I think it would be good to squash the commits because I think it is not beneficial to keep the three small commits in the history.

Issue #3997, opened 2021-01-14
Error messages aren't properly logged for new entities pushed to the server


MateStrysewske:
### Bug Description
The actual error messages for new entities pushed to the server (i.e. those without a change date) aren't properly logged.

PR #4000, opened 2021-01-14
#3997 - Extended logging to deal with new entities

MateStrysewske:


JonasCir (approved):
Thanks :) :shipit:

JonasCir:
Huh but there is a test failure

PR #4001, opened 2021-01-14
#3753 Add the signal evolution date in the event file and in the event directory

### LINKED ISSUE #3753: Add the signal evolution date in the event file and in the event directory
### Situation Description

It often happens that a signal relating to an event is supplemented by information received several days / week after the initial signal. Currently SORMAS offers the date of the signal (not modifiable in the HMI), and the date of the event which can be modified.
The teams in the field tell us the need to have two date fields to trace on the one hand the date of the initial event and, on the other hand, the date of change of the signal (today the teams modify the date of the event but thus overwrite the information on the initial date of the event).

NB: This date is different from the date of modification of the signal which is updated automatically with each intervention on the event file (eg: correction of a typing error); the requested field will be entered by the investigator who is informed by a partner that the initial situation has changed and that the investigation must resume

### Feature Description

We propose to complete the event sheet with a second date block allowing teams to enter the last modification date of the signal. This information is essential and should also be included in:
- the event directory
- the actions directory
- basic and detailed export
### Additional Information
@ftavin @bernardsilenou

aponcon-atolcd:
Fixes #3753


aponcon-atolcd:
Hi
Last modifications are about a rebase on development branch.
While testing the commits, I've encountered an issue in the detailed export of events and event actions (the CSV files are empty and an error alert is showing up).
This issue is also present into the test server so this issue is not related to these commits.

MateStrysewske:
@aponcon-atolcd Yes, I've already encountered that issue as well. We're working on it!

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1849,7 +1849,12 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				getDao(Person.class).executeRaw("ALTER TABLE person ADD COLUMN externalToken text;");
 				getDao(Event.class).executeRaw("ALTER TABLE events ADD COLUMN externalToken text;");
 
-					// ATTENTION: break should only be done after last version
+			case 260:
+				currentVersion = 260;
+				getDao(Event.class).executeRaw("ALTER TABLE events ADD COLUMN signalEvolutionDate timestamp");
```
MateStrysewske, line 1854: Not sure whether it's actually needed, but please add a semicolon to the end of the SQL statement just to be safe

MateStrysewske (changes requested):
I think the comment should generally only be displayed in forms when the date has actually been filled in; you've already done that for the read fragment in the mobile app, but I think it should also be the case for edit forms. For the mobile app, you can try using dependencyParentField in combination with dependencyParentVisibility=false and dependencyParentValue=null. Should work, if not you have to fall back to a listener.

MateStrysewske (changes requested):
The comment field is still visible in the web UI when no date has been entered

aponcon-atolcd:
This time the web ui and mobile ui are both modified.
The latest push also contains modification of the labels and property naming.
And the work was also rebased as merge were no longer available.

MateStrysewske:
@aponcon-atolcd Hey, can you resolve the merge conflicts again please?

aponcon-atolcd:
@MateStrysewske Done :)

MateStrysewske (changes requested):
I like the dynamic captions of the two fields! However, "Dropped evolution date" doesn't really make sense; I would prefer to call it "Event evolution date" in this case as well (since "event" is the standard name of the entity).

aponcon-atolcd:
> I like the dynamic captions of the two fields! However, "Dropped evolution date" doesn't really make sense; I would prefer to call it "Event evolution date" in this case as well (since "event" is the standard name of the entity).

Done :)

MateStrysewske (approved):

Issue #3566, opened 2020-11-24
Show event participant creation window when linking a contact to an event [1]


MateStrysewske:
### Feature Description
#3133 added the possibility to link a contact to an event directly from within the contact by creating an event participant. This process does not involve opening an event participant creation window; it's all done automatically. Because of that, no involvement description is given, which results in an event participant without an involvement description being created even though the involvement description is a required field.

### Proposed Change

* When linking a contact to an event, the event participant creation window should be opened to allow the user to enter an involvement description
And
* Remove the requirement (optional instead of mandatory) from the involvement description field; this may be something to consider anyway because it might be a redundant field if someone is entering event participants e.g. for a concert (the involvement description will probably just be "took part" most of the time)

kwa20:
@MateStrysewske I agree that the involvement description doesn't have to be compulsory. Since we need to choose an event to link the contact to anyway, do you think it might be possible to show the involvement description in the same window?

<img width="866" alt="Screen Shot 2020-11-24 at 12 01 13 PM" src="https://user-images.githubusercontent.com/70317594/100093850-63b8fe80-2e58-11eb-925f-265ac746dc82.png">

 I assume this might be difficult since it relates to a different entity. If it's possible, though, users won't have to navigate through two windows.

Otherwise only a window with the involvement description could be shown after selecting the event since this represents the event participant entity thus far. This is what you've already described I guess.
<img width="480" alt="Screen Shot 2020-11-24 at 12 01 57 PM" src="https://user-images.githubusercontent.com/70317594/100094091-c90cef80-2e58-11eb-80e1-9068fc2b815c.png">
The separation of such windows also becomes very relevant when #3003 would be implemented.

MateStrysewske:
@kwa20 Adding the involvement description to the "Pick or create event" form would theoretically be possible (of course not with automatic databinding, but that would be fine); the problem is more that, in your mockup, it looks like the field is actually also used for filtering (because it's in the same section as the other filters). I think in terms of workflow, it's hard to fit it in there, but we'll have a look at it in the next sprint and see whether it's possible somehow.

kwa20:
@MateStrysewske You're right, that might not be logical. We could also move the field underneath the event list.

kwa20:
This actually applies to cases as well when linking them to events through the case form.

syntakker:
When linking a contact person to an event, the dialog title reads "Create new person"
[...]

PR #4005, opened 2021-01-14
#3566 - open event participant edit form to complete involvement desc…

BarnaBartha:
…ription when linking event to case or contact


MateStrysewske (changes requested):
* Involvement description also needs to be made optional in the mobile app
* The notification that pops up when linking a case to an event comes up way too early now; it should only come up after the event participant has been created
* Clicking on Discard in the second popup that allows to specify more person details will still link the event, but the user has to manually reload the page in order to see it in the list of events; clicking on Discard should have the same outcome than clicking Save

MateStrysewske (approved):

Issue #3488, opened 2020-11-16
[DEMIS2SORMAS] Enable users to convert ExternalLabMessages to assigned PathogenTestResults


FredrikSchaeferVitagroup:
### Situation Description

#3486 and #3487 enable users to import and parse lab results from [DEMIS](https://www.rki.de/DE/Content/Infekt/IfSG/DEMIS/DEMIS_node.html) via ExternalLabMessages. Now users need to be able to process these ExternalLabMessages so that their data is integrated into the SORMAS workflow.

### Feature Description

There shall be an ExternalLabMessage directory implemented that lists all ExternallabMessages. The user shall be able to select a ExternalLabMessage and process it. 

![Message_Directory](https://user-images.githubusercontent.com/67001822/99248123-45c01e00-2808-11eb-9168-98feab8c8a91.png)

Each ExternalLabMessge contains data about a Person, a Sample, a Laboratory and a PathogenTestResult. At the end of processing an ExternLabMessage, this data should be related to existing or put into new entities of a Person, Sample, Laboratory and PathogenTestResult in SORMAS.

The first step for that is to relate the ExternalLabMessage to a Person. The already implemented person duplicate detection shall be used to check if the person already exists in SORMAS. 

If so, the 'Pick or Create Person' from the contact and case creation shall be opened. If the user decides to pick an existing person, another view shall display all Contacts, Cases and Event Participants related to that person. The user can then decide which Contact, Case or EventParticipant the ExternalLabMessage's sample data to assign to. The user should also be able to create a new Case, Contact or Eventparticipant for that Person here. 

![Assign_PathogenTestResult](https://user-images.githubusercontent.com/67001822/99366940-93489380-28b9-11eb-8fb1-520b0f4ef574.png)

SORMAS should also check if a Sample similar to the one generated with data of the ExternalLabMessage is already assigned to any Contact, Case or EventParticipant related to the person. If so, SORMAS should reccomend to assign the ExternalLabMessage's PathogenTestResult to it. 
Based on the user's selection, SORMAS should assign the ExternalLabMessage's PathogenTestResult data to some Contact, Case or EventParticipant, creating a new Sample for that entity if necessary.

If there is no Person existing in SORMAS, the user shall be able to select to create either create a Case, a Contact or an EventParticipant. After that is finished, the ExternalLabMessage's Sample and PathogenTestResult shall be assinged to the created entity.

[...]

PR #4023, opened 2021-01-15
Feature 3488 lab message UI first approach

ChristopherRiedel:
#3488


FredrikSchaeferVitagroup:
There seems to be a small issue with screens of some aspcect ratios less extreme than ultrawide. When processing a lab message, while viewing the labmessage as HTML on the left, and some form on the right, there is a small overlap in the middle.

FredrikSchaeferVitagroup:
There is another small issue when creating a sample while processing: all required fields are marked red, except the RESULTS VERIFIED BY LAB SUPERVISOR. This is a little irritating.

FredrikSchaeferVitagroup:
When processing, the pathogenTestResult is always set to PENDING by default. Is that desired behaviour?

sergiupacurariu (commented):
on Creating new sample, "Date sample was collected" remains with red borders even after the dates were corrected but save works

FredrikSchaeferVitagroup (changes requested):

FredrikSchaeferVitagroup (approved):

sergiupacurariu (approved):

PR #4062, opened 2021-01-18
Change event's surveillanceOfficer to responsibleUser and allow more roles to be it #3751

### LINKED ISSUE #3751: Modify the "responsible surveillance officer" field in an event file
### Feature Description

To date, the event directory has a lot of information characterizing the event, the people involved in the event and the actions carried out.

In a context of strong viral spread, the number of events to be investigated is large. In order to facilitate the distribution of the signals to be investigated between the different investigators, we suggest displaying the name of the person in charge of the investigator (by modifying the "Responsible surveillance Officer" field)

### Problem Description

To date, only the names of people with the "responsible surveillance officer" profile appear in the drop-down list. However, in France, the persons in charge of the investigation are at the regional level and therefore have the contact supervisor profile.

### Proposed Change

- Modification of the “responsible surveillance officer” field which to date does offer the option for only “Surveillance Officer”. We suggest that this list be supplemented by the "surveillance supervisor" role in addition, in order to be able to display the name of the person in charge of investigating the event.

- In terms of the events directory:
o we modify the Surveillance Agent filter to have the label "responsible user" and it proposes the list of agents and supervisors (with the possibility of entering the first letters as in the creation of a task for example)
o add a "responsible user role" filter on the supervisor or agent role.
o If you select a role, it limits the people proposed by automatic completion in the responsible user filter but you can search for a person without having indicated a role beforehand
- Integration of this field in all exports

### Additional Information
@ftavin @bernardsilenou

alexandre-gille:
Closes #3751


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1948,8 +1948,7 @@ public void onCaseChanged(CaseDataDto existingCase, Case newCase) {
 
 	private void sendConfirmedCaseNotificationsForEvents(Case caze) {
 		Date fromDate = Date.from(Instant.now().minus(Duration.ofDays(30)));
-		Map<String, User> surveillanceOfficerByEventByEventUuid =
-			eventService.getAllEventUuidWithSurveillanceOfficerByCaseAfterDateForNotification(caze, fromDate);
+		Map<String, User> surveillanceOfficerByEventByEventUuid = eventService.getAllEventUuidWithResponsibleUserByCaseAfterDateForNotification(caze, fromDate);
```
MateStrysewske, line 1951: Variable should be renamed as well

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1872,6 +1872,10 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				getDao(Event.class).executeRaw("ALTER TABLE events ADD COLUMN connectionNumber varchar(512);");
 				getDao(Event.class).executeRaw("ALTER TABLE events ADD COLUMN travelDate timestamp;");
 
+			case 264:
+				currentVersion = 264;
+				getDao(Event.class).executeRaw("ALTER TABLE events RENAME surveillanceOfficer_id to responsibleUser_id;");
```
MateStrysewske, line 1877: I guess you haven't tested this? Renaming columns is not possible in the SQLite version used in Android. Our current approach is to rename the field in the entity class, but add the `name` attribute to reference the name in the database.

I have to emphasise again that it's vital to test the Android database migration.

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
```diff
@@ -325,10 +332,33 @@ protected void addFields() {
...
-			List<UserReferenceDto> assignableSurveillanceOfficers = FacadeProvider.getUserFacade()
-				.getUserRefsByDistrict((DistrictReferenceDto) districtField.getValue(), false, UserRole.SURVEILLANCE_OFFICER);
-			FieldHelper.updateItems(surveillanceOfficerField, assignableSurveillanceOfficers);
+			DistrictReferenceDto district = (DistrictReferenceDto) districtField.getValue();
+			if (district != null) {
+				List<UserReferenceDto> currentDistrictSurveillanceOfficers =
+					FacadeProvider.getUserFacade().getUserRefsByDistrict(district, false, UserRole.SURVEILLANCE_OFFICER);
+
+				Set<UserReferenceDto> responsibleUsers = new LinkedHashSet<>();
+				responsibleUsers.addAll(currentDistrictSurveillanceOfficers);
```
MateStrysewske, line 354: This logic doesn't really make sense; the field would either be populated by all surveillance officers from the region if the district filter is empty, or with all surveillance officers from the district if it's not empty, but not with both (because that would make the district filter completely redundant).

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsFilterForm.java
```diff
@@ -106,8 +109,15 @@ protected void addFields() {
 		addField(FieldConfiguration.pixelSized(EventCriteria.EVENT_STATUS, 140));
 		addField(FieldConfiguration.pixelSized(EventIndexDto.DISEASE, 140));
 		addField(FieldConfiguration.withCaptionAndPixelSized(EventCriteria.REPORTING_USER_ROLE, I18nProperties.getString(Strings.reportedBy), 140));
-		ComboBox officerField = addField(FieldConfiguration.pixelSized(EventCriteria.SURVEILLANCE_OFFICER, 140));
-		officerField.addItems(fetchSurveillanceOfficersByRegion(currentUserDto().getRegion()));
+		addField(FieldConfiguration.pixelSized(EventCriteria.RESPONSIBLE_USER, 140));
+
+		ComboBox responsibleUserRoleField = addField(
```
MateStrysewske, line 114: What exactly is the purpose of this filter?

MateStrysewske (changes requested):
A feature configuration that prevents selecting supervisors as the responsible users for events is missing. I suggest to call it "EXTEND_EVENT_RESPONSIBLE_USER_ROLES".

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
alexandre-gille: That's what I've been asked to do.

Quoting Bernard :
> Order the names with surveillance officers in the district of the event at top of list, other SO and then supervisors follow.

The sole purpose of this is to put the district's SO on top of the list.

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsFilterForm.java
alexandre-gille: Well, as its name suggest, it filters on the role of the event's designated responsibleUser.
For anything else beyond that, I'm not really the one you should ask about it I think :shrug:

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
alexandre-gille: I did test it. I checked that the app worked properly and directly in database to see if the migration was applied correctly. And it was.

alexandre-gille: SQLite supports alter table's column rename since 3.20 and android uses a superior version since API 28 (As the internet is telling me).
So, as a thoughtful development mastermind once said "It works on my machine".

What android's API level are we supposed to aim for ?

MateStrysewske: We have to support API 24 (Android 7.0). It's specified in the main readme, but that's really easy to miss of course :) however, with API level 24 it should therefore not work, so we need the workaround that I've suggested.

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
MateStrysewske: I see - I will ask Bernard what the purpose of this specification is because I don't think it makes sense, and get back to you once he's responded.

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsFilterForm.java
MateStrysewske: Makes sense - gonna ask Carolin :)

MateStrysewske:
> > A feature configuration that prevents selecting supervisors as the responsible users for events is missing. I suggest to call it "EXTEND_EVENT_RESPONSIBLE_USER_ROLES". Default should be off.
> 
> If I haven't misunderstood anything, the issue states that this wasn't required as long as it wasn't applied to case and contacts as well. Something changed in the meantime ?

That's true, I hadn't seen that part of the discussion. I'm not sure whether Chinedu and Bernard were on the same page though, could've been a misunderstanding, so I'm going to double-check that it's fine to have this without a feature configuration. I'll get back to you with more information!

MateStrysewske:
Okay, looks like the feature configuration is indeed not needed for now :)

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
MateStrysewske: @alexandre-gille Verified with Bernard that the specification was not correct. The only users added to the dropdown to select the responsible users should be surveillance supervisors from the event's region and surveillance officers from the district's region.

MateStrysewske:
@alexandre-gille I clarified all the points that were unclear, please have another look at my comments above

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventsFilterForm.java
alexandre-gille: I removed this filter, according to the discussion in the original issue.

alexandre-gille:
Done

MateStrysewske:
@alexandre-gille There are errors in this branch, e.g. a method referenced in EventsFilterForm that doesn't exist

MateStrysewske (approved):

PR #4076, opened 2021-01-19
4017 demis error messages

### LINKED ISSUE #4017: [DEMIS2SORMAS] Show error messages to the user [1]
### Situation Description

Currently, errors in the demis adapter are written to the log or not noted at all. There is just the standard error message shown to the user, and this not even in all cases. This has to be changed. 

### Feature Description

In this issue, the ExternalLabResultsFacade and related classes shall be enabled to receive error messages and output them to the user.

### Additional Information

The [ExternalJournalService.validatePatientDiaryPerson(...)](https://github.com/hzi-braunschweig/SORMAS-Project/blob/c4b8713faa58f1596a9eb7973dd341813cd980bc/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java#L325) might help get an idea of how to implement this.

This issue should probably be picked together with implementing the forwarding of error messages in the demis-adapter.

vlad-ciucescu:
closes #4017 
related PR for adapter:
https://git.vitasystems.dev/sormas/sormas-demis-adapter/-/merge_requests/4


FredrikSchaeferVitagroup (changes requested):
A part of the UI seems to be broken by the implemented changes. When opening the LabMessage directory, the LabMessageGrid seems to crash.

FredrikSchaeferVitagroup (changes requested):

FredrikSchaeferVitagroup (approved):

Issue #4084, opened 2021-01-20
Improve EventParticipant contact count queries


DavidBaldsiefen:
### Feature and Problem Description

The queries used for determining the amount of contacts (with and without source cases in event) for each eventparticipant are relatively slow, which could lead to problems with large events.

### Proposed Change

Improve Queries.

### Additional Information
This is a screenshot of the current query used to determine contact counts where the source case is in the event: 

![image](https://user-images.githubusercontent.com/4655486/105162127-8000b100-5b12-11eb-993c-5e19321b324e.png)

DavidBaldsiefen:
This is the suggested new query for event participants with source in event:

![image](https://user-images.githubusercontent.com/4655486/105162775-56945500-5b13-11eb-8fcf-d5b511088c30.png)

The other query only got slightly shorter, but its estimated cost (according to pgAdmin) and average runtime on my DB are also slightly better.

AndyBakcsy-she:
Validated ticket on the latest version of SORMAS Version: 1.55.0-SNAPSHOT on the test.sormas.netzlink.com environment

AndyBakcsy-she:
adding multiple contacts to the same Person would increase the contact count;
Person participant & source case is participant the contact count increases as vell

PR #4085, opened 2021-01-20
#4084 improve eventparticipant contact queries

DavidBaldsiefen:
In the eventparticipant list of an event, the contact-count is calculated (depending on filters: with or without source case in event).

This PR implements improved queries which should perform much better on large databases (see #4084 for details)


DavidBaldsiefen:
@MateStrysewske Could you review this? It's essentially the same query as used for #3276 , which proved to be much faster (see #4084 issue description)

MateStrysewske (approved):

Issue #3014, opened 2020-09-29
Prompt to save/discard unsaved changes opens up when leaving person form in read-only mode


MateStrysewske:
### Bug Description
When entering the person form in read-only mode, the prompt to save/discard unsaved changes open upon leaving the form. This probably also happens when not in read-only mode. The proposed fix should be to make sure that all code that changes the value of fields should be executed before the listeners that are responsible for the prompt to open are added.

kwa20:
@MateStrysewske I can confirm that this sometimes also happens to users that could edit the form but haven't made any changes before navigating somewhere else.

kwa20:
When navigating out of event participants, prompt is not triggered when changes have been made.

MateStrysewske:
Moved the issue with the event participants to #3042

MateStrysewske:
I can't reproduce this on my local system. We'll put this issue on Waiting until the next release. Maybe there's some kind of side effects that have resolved the issue in the meantime.

bernardsilenou:
@MateStrysewske I could reproduce it on sormas.symeda, 1,49,1.
When a pages eg case form is open, no information is edited on the form, when you try to navigate to another from, the pop up message for unsaved changes appear. I suggest we should only show this message of the user changed a filed on the form and do not save

MateStrysewske:
Can you please tell me the UUID of the case where you've encountered this?

bernardsilenou:
@MateStrysewske I am on sormas.symeda with version 1,49,1. I experienced this issue when leaving the case tab only. Other tabs are ok.
 Case is: TSJHHU-5ZAGAT-7XWVET-KTT7CMCQ

![image](https://user-images.githubusercontent.com/12996724/95244822-b429bd80-0812-11eb-8538-517849d47b84.png)

bernardsilenou:
@MateStrysewske https://test1401.sormas-oegd.de/sormas-ui/ is also on 1.49,1 but all works fine. Maybe the symeda server is pulling old files from 1,49,0?

MateStrysewske:
@bernardsilenou Which user are you using on the symeda system?

MateStrysewske:
@bernardsilenou Reminder :)

bernardsilenou:
@MateStrysewske sorry i have a pile of unread emails. I used this user name: Silenou. Sever is down so i cannot tell you more about the user account now

MateStrysewske:
[...]

PR #4097, opened 2021-01-21
#3014 - do not mark as dirty location fields changed by the converter

BarnaBartha:


MateStrysewske (changes requested):
The logic needs to be applied to other forms where GPS fields are used outside a LocationEditForm as well. This at least includes the CaseDataForm and the FacilityEditForm. Make sure to check whether the StringToAngularLocationConverter is used somewhere else for GPS coordinate fields and include those as well.

A hacky solution (i.e. adding these forms as additional special cases to the CommitDiscardWrapperComponent) is totally fine.

MateStrysewske (approved):

PR #4101, opened 2021-01-21
Add risk level column and filters to event and action directories #3951

### LINKED ISSUE #3951: Show cluster risk level on event and actions directory and add filter
### Situation Description
#3271 added a risk level to events that are clusters. However, users cannot see risk level information on the directories and will therefore have to open events individually to find out their risk level.  To have a more informative overview about the clusters, adding this information as well as a filter would be quite useful (e.g. different people might want to focus on different risk levels).
### Feature Description
I suggest to add the risk level on the event and actions directories inside the event status. Since the risk level is only assigned to clusters, a separate column is not necessary because it would have many empty rows.
### Additional Information
clusterCriticality2

clusterCriticality3

alexandre-gille:
Closes #3951


MateStrysewske:
@alexandre-gille Looks fine! Can you resolve the merge conflict, please?

alexandre-gille:
Done

MateStrysewske (approved):

PR #4114, opened 2021-01-22
Ensure that case and contact creation is only available for event with a disease #3919

### LINKED ISSUE #3919: When converting event participant into contact or case, alert the user if the disease of the event is not indicated
### Situation Description
The transformation of a person participating in an event into contact results in the opening of a window asking to specify the characteristics to be inserted in the contact file, including illness. If the disease was not specified in the description of the event, this field will be empty and cannot be modified, even though it is a mandatory field.

### Feature Description
Since the disease of the event cannot be a mandatory field, we propose to develop an alert message asking the user to specify the illness of the event.

### Additional Information

aponcon-atolcd:
Fixes #3919


MateStrysewske (changes requested):
I don't really like the solution to potentially let getCaseCreateComponent and getContactCreateComponent return null, and it forces you to unnecessarily check for null in a lot of places where events don't even play a role. I suggest to directly check whether the event in question does have a disease or not in `createFromEventParticipant` and `createFromEventParticipantDifferentDisease` (you only have to make sure that `disease` is not null there) for cases, and the `create(EventParticipantReferenceDto)` method for contacts.

MateStrysewske (approved):

Issue #839, opened 2018-10-19
Add xlsx to export options (cases and contacts directory)


MartinWahnschaffe:

AxelNennker:
Sormas is already using Apache POI but only in test and doc.
https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-api/pom.xml#L57
https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-api/src/test/java/de/symeda/sormas/api/doc/XssfHelper.java
https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-api/src/test/java/de/symeda/sormas/api/doc/DataDictionaryGenerator.java

The Apache POI version used is 3.17. Should this be updated to 4.1.2 ?

The XssfHelper and other Excel-related code in doc look useful for implementing this issue. Where should code be moved to so that is available to importers and exporters as well?

PR #4117, opened 2021-01-22
#839 XLSX Export

AxelNennker:

- Apache POI 3.17 poi-ooxml usage not only for test and doc but also for export
- renamed GridExportStreamResource to GridExportStreamResourceCSV
- created GridExportStreamResourceXLSX, it's not called yet

Missing
- ~~units tests~~ more and better unit tests 
- ~~UI to select export format (CSV, XLSX)~~ UI is now in user settings

- maybe update Apache POI to 4.1.2
- maybe add 'about' sheet like in XssfHelper.addAboutSheet


MateStrysewske:
@StefanKock Could you have a look at this when you have time?

syntakker (changes requested):
Hi @AxelNennker , sorry for the late review and the overhead caused by this delay.

This PR introduces a massive amount of duplicate code, since the new feature is implemented by copyin large portions of code all along the call stack. This needs a healthy dose of refactoring.

- I would propose to pass the export format as an extra argument in the call stack, but handle the distinction according to the format only where it really makes a difference, i.e., when creating stream resources or writing cell data to the stream. To do so, please introduce an interface for the stream writer (see comments in `ExportStreamUtils`, `GridExportStreamResource*` etc.).

- The extra column `exportFormat` has to be added in the data base.

- The selection of the `exportFormat` in user settings might be hard to find. Please consider adding a checkbox in the export dialogs in those places with popup windows for export. Still, to have this selection persisted as a user specific setting makes sense, so this selection should be persisted when changed.

- As you suggested, unit tests may be added.

- Please make sure code formatting is configured (see https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/docs/DEVELOPMENT_ENVIRONMENT.md#step-5-configure-code-formatting-and-import-settings)

Given the progress the SORMAS development has made in the meantime along with the number of change requests, starting off a new branch from current development might be easier than going through an error prone merge hell.

AxelNennker:
I'll have a look at this PR again. I guess too much changed after the csv performance issue.
Probably have to start over.

syntakker:
@AxelNennker perfect, thank you!
Let me know if I can help or in case you want to discuss details.

PR #4121, opened 2021-01-25
[Do not merge] Search for source of slow CI

DavidBaldsiefen:
**DO NOT MERGE**

This PR is just there so that I can test GitHubs CI runtimes by commiting.

Current CI runtimes are over 20 minutes, and it would be the goal to get it back down to around ~15 minutes.


DavidBaldsiefen:
Closed in favor of #4134

Issue #4028, opened 2021-01-15
facility selection fills location fields


ftavin:
### Situation Description

On the location bloc in SORMAS (can be found in event and person-related entities), you can select facilities from a list and you can indicate an address just below. We would like to add location information to facilities so that the selection of a facility automaticaly fills the address fields.

This feature will be split in two issues :
- add location information to facilities (#4027)
- facility selection fills location fields (the present issue)

### Feature Description

The selection of a facility on a location block will automaticaly copy the facility's location information to the corresponding fields of the block (street, house number, additional information, postal code, city, area type, latitude, longitude) if at least one of them is filled on the facility side. If there is already information on the fields, it is replaced by the facility's information.

Afterward, the user is still free to edit this information which will still be stored in the corresponding entity : if the address of the facility changes later, the address of the event/case/contact should not change.

### Additional Information

This issue has to be done AFTER #4027 because of the need for the new fields.

@carolinverset @bernardsilenou

bernardsilenou:
@markusmann-vg Dev France would work on this for 1.55

PR #4124, opened 2021-01-25
Facilities' address #4027 #4028 

### LINKED ISSUE #4027: Add location information to facilities
### Situation Description

On the location blocks in SORMAS (can be found in event and person-related entities), you can select facilities from a list and you can indicate an address just below. We would like to add location information to facilities so that the selection of a facility automaticaly fills the address fields.

This feature will be split in two issues :
- add location information to facilities (the present issue)
- facility selection fills location fields (#4028)

### Feature Description

We need to add to the facility entity the following informations : (all are optional)
- street
- house number
- additional information
- postal code
- area type

The new fields will not be shown in the list (configuration pane) but will be added to a new "detailled export".
The import mechanism will be modified to include the new fields (still optional)

### Additional Information

Later, phone and email address could be added as aked in #3931 but this will not be done in this issue.

@carolinverset @bernardsilenou

alexandre-gille:
Closes #4027 #4028


sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1877,6 +1877,14 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				getDao(Event.class).executeRaw("ALTER TABLE events ADD COLUMN evolutionDate timestamp;");
 				getDao(Event.class).executeRaw("ALTER TABLE events ADD COLUMN evolutionComment text;");
 
+			case 265:
+				currentVersion = 265;
+				getDao(Facility.class).executeRaw("ALTER TABLE facility ADD COLUMN street varchar(4096);");
+				getDao(Facility.class).executeRaw("ALTER TABLE facility ADD COLUMN houseNumber varchar(255);");
```
MateStrysewske, line 1883: These new fields should use the same lengths as their equivalents in the Location entity because we might run into problems otherwise; as far as I can see, this affects houseNumber and postalCode (should be 512; also needs to be changed for the definition in sormas_schema.sql)

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/facility/Facility.java
```diff
@@ -60,6 +62,16 @@
...
 	private String city;
+	@Column(length = COLUMN_LENGTH_DEFAULT)
+	private String postalCode;
+	@Column(length = COLUMN_LENGTH_BIG)
+	private String street;
+	@Column(length = COLUMN_LENGTH_DEFAULT)
+	private String houseNumber;
+	@Column(length = COLUMN_LENGTH_BIG)
+	private String additionalInformation;
+	@Column
```
MateStrysewske, line 73: Don't know why this is written this way in Location.java, but it should actually be `@Enumerated(EnumType.STRING)`

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -286,6 +286,35 @@ protected void addFields() {
...
+
+			// Fill in the address fields based on the selected facility
+			// We don't want the location form to automatically change even if the facility's address is updated later 
+			// on, so we only trigger it upon a manual change of the facility field
+			// We use isAttached() to avoid the fuss when initializing the form, it may seems a bit hacky, but it is
+			// necessary because isModified() will still return true for a short duration even if we keep the very same 
+			// value because of this field dependencies to other fields and the way updateEnumValues works
+			if (facility.isAttached()) {
+				if (facility.getValue() != null) {
+					FacilityDto facilityDto = FacadeProvider.getFacilityFacade().getByUuid(((FacilityReferenceDto) facility.getValue()).getUuid());
```
MateStrysewske, line 298: I think blindly overwriting values here is not a good idea. Instead, I propose the following solution:

For all of these fields, check whether the facility has a value for it. If not, don't overwrite whatever has been entered in the location form. If yes, check whether the location form has a value for it. If no, overwrite that empty value with the value from the facility. If yes, check whether the value is identical; If yes, do nothing; If no, show a popup window that asks the user to confirm overwriting the location details with the details from the facility (or to reject; in that case, nothing is overwritten).

I suggest the following associated text:
"The selected facility has location details that are different from the location details you are currently editing. Do you want to overwrite the details of the current location with those from the facility?"

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -286,6 +286,35 @@ protected void addFields() {
...
+					cityField.setValue(facilityDto.getCity());
+					postalCodeField.setValue(facilityDto.getPostalCode());
+					streetField.setValue(facilityDto.getStreet());
+					houseNumberField.setValue(facilityDto.getHouseNumber());
+					additionalInformationField.setValue(facilityDto.getAdditionalInformation());
+					areaType.setValue(facilityDto.getAreaType());
+					tfLatitude.setConvertedValue(facilityDto.getLatitude());
+					tfLongitude.setConvertedValue(facilityDto.getLongitude());
+				} else {
+					cityField.clear();
```
MateStrysewske, line 308: Just as above, blindly clearing these fields is not a good idea. I actually think we should never do that because even if the user removes the facility from the form, the location details might still be relevant (e.g. it's likely that the city is still the same).

MateStrysewske (changes requested):
Please also add the logic that auto-fills the location fields based on the selected facility to the mobile app to make it consistent.

alexandre-gille:
Done

MateStrysewske (changes requested):
Hey! If you can solve the merge conflicts again and fix the small issues that I've added, I'll do a functional review today and if everything is fine, it will still be merged for the upcoming release :)

MateStrysewske (changes requested):
* `!Double.valueOf(tfLatitude.getValue())` (and the longitude equivalent I guess) results in a parsing error
* Warning message is triggered and pops up multiple times immediately when opening a location form with a facility that has address details in it in the mobile app

alexandre-gille:
>     * `!Double.valueOf(tfLatitude.getValue())` (and the longitude equivalent I guess) results in a parsing error
> 
>     * Warning message is triggered and pops up multiple times immediately when opening a location form with a facility that has address details in it in the mobile app

I fixed the double issue. I was stupid enough to test it with round numbers...

I'm not able to reproduce the pop-up issue on the mobile app though. Is there anything specific on this facility ?

MateStrysewske:
@alexandre-gille I don't think so. I'll check that tomorrow again and debug it a bit to see whether I can find what the exact source of the problem is. Since everything else in this PR is fine, we shouldn't have any problems getting this into the next version :)

MateStrysewske (changes requested):
* The buttons in the dialog that asks the user whether they want to overwrite the location details with those from the facility should be "Yes" and "No" instead of "Confirm" and "Cancel"

The bug in the mobile app is reproducable on my end:
* Open a case in the web app that your mobile app user has access to; open the person view and add an address to the person address list; select a facility in this address that has location details filled in (ideally all of them, that will probably make sure you run into the problem); save
* Synchronise your mobile app with the backend
* Open the case from the person that you've added the address for, enter edit mode and navigate to the person tab
* Click on the card with the address that you've created; the address edit dialog opens, and you should have a number of prompts popping up immediately that ask the user whether they want to overwrite changes

alexandre-gille:
That's right. In the case form, the valueCange is triggered multiple times in a row and then doesn't work anymore... I have no clue why :shrug: 
It's working properly in the event form though. At least the triggering part. But the facilities retrieved by InfrastructureHelper.loadFacilities(...) don't have the newly added fields populated, so the only thing we actually replace in the city field. There's not field specified when we trace back to the FacilityDao's method which does the query, I just assumed it pulled everything by default. Do I have to change something elsewhere to make it work ?

MateStrysewske:
@alexandre-gille Does your last commit fix the problem that you were talking about?

alexandre-gille:
It doesn't.

MateStrysewske:
@alexandre-gille If I understood you correctly, the facilities in your mobile app are missing the location information, right? Since you've only added the fields that populate them from the DTO in your last commit (so probably after you did the changes to your facilities and pulled the data to your mobile app), the facilities won't get synchronised again because they already were. Can you try opening the settings screen in the app and click on "Resynchronize data"? That should solve that problem hopefully.

If I didn't understand you correctly, please let me know.

aponcon-atolcd:
The branch is rebased.

@MateStrysewske I tried to reproduce the issue you mentioned (following the steps you've described), but I wasn't able to reproduce it.
I tried from a newly created database generated from the branch and and in a second time, recreated the database from the development branch and upgraded it with the changes provided by these commits. Both of these tests didn't resulted in your issue.
Can you check if the issue is still there for you?

MateStrysewske:
@aponcon-atolcd I'll get back to this PR on Tuesday and test whether the issue still persists on my end again. :)

MateStrysewske:
@aponcon-atolcd The problem still happens for me, but only if the location details differ from the facility details.

* Add an address with a facility to a person or to an event
* Change something about the address/event location so that it differs from the facility's details (e.g. the street name) and save
* Open the address popup again - you will get four or five prompts at the same time because the value change listener is called that many times and always triggers the prompt because the details differ

aponcon-atolcd:
@MateStrysewske Thanks a lot for your comment and for your test. Now I understand better the root of the issue. It will be fixed as soon as possible ;)

aponcon-atolcd:
@MateStrysewske I reproduced the issue. The issue is now fixed. Thanks again for helping me identifying the origin of the issue.

MateStrysewske:
@aponcon-atolcd Thanks! If you resolve the merge conflict, I'll merge this PR :)

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -2020,6 +2020,15 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 						"ALTER TABLE eventParticipants ADD COLUMN sormasToSormasOriginInfo_id bigint REFERENCES sormasToSormasOriginInfo(id);");
 				getDao(EventParticipant.class).executeRaw("ALTER TABLE eventParticipants ADD COLUMN ownershipHandedOver boolean;");
 
+			case 279:
```
MateStrysewske, line 2023: Needs to be 280, and the version number at the top of the class needs to be updated to 281

MateStrysewske (changes requested):

aponcon-atolcd:
@MateStrysewske Indeed, the version numbers in Android app are fixed now

MateStrysewske (approved):

PR #4131, opened 2021-01-26
Feature/4129 Server property for default user creation

### LINKED ISSUE #4129: Server property to disable (or enable) default user creation
### Situation Description
Netzlink mentioned that they want to reduce the number of steps needed to setup SORMAS. One common step is to disable the default users (except admin).

### Feature Description
Add a property to `sormas.properties` which allows to control the creation of default users. Admin user should always be included. This  could be done by checking for the property in `StartupShutdownService.createDefaultUsers`.

### Possible Alternatives
Couple it with the SORMAS developer mode?

JonasCir:
closes #4129

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
@@ -540,7 +540,7 @@ public class ConfigFacadeEjb implements ConfigFacade {
 	    return getInt(DASHBOARD_MAP_MARKER_LIMIT, -1);
 	}
 
-	public boolean getCreateDefaultUsers(){
+	public boolean isCreateDefaultUsers(){
 	    return getBoolean(CREATE_DEFAULT_USERS,true);
 	}
 
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
@@ -537,11 +537,11 @@ public class ConfigFacadeEjb implements ConfigFacade {
 	}
 
 	public int getDashboardMapMarkerLimit() {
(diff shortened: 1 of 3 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
```diff
@@ -355,6 +355,15 @@ private void createDefaultUsers() {
 			userService.persist(admin);
 			userUpdateEvent.fire(new UserUpdateEvent(admin));
 
+
+			if (!configFacade.getCreateDefaultUsers()) {
```
MateStrysewske, line 359: Move this new code and the admin creation code to above the calls to retrieve regions, districts, communities etc. as that's unnecessary if no default users are created anyway

sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
```diff
@@ -534,7 +536,11 @@ public void validateAppUrls() {
 	}
 
 	public int getDashboardMapMarkerLimit() {
-		return getInt(DASHBOARD_MAP_MARKER_LIMIT, -1);
+	    return getInt(DASHBOARD_MAP_MARKER_LIMIT, -1);
+	}
+
+	public boolean getCreateDefaultUsers(){
```
MateStrysewske, line 542: boolean getters usually start with "is" instead of "get"

MateStrysewske (changes requested):
Please use automatic code formatting (see development environment setup instructions), ConfigFacadeEjb.java is not properly formatted

sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
JonasCir: can we settle for should?

JonasCir: fixed

MateStrysewske (approved):

PR #4143, opened 2021-01-26
Feature 3629 hide epid number

alexcaruntu-vita:


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
```diff
@@ -119,6 +120,10 @@ protected void addFields() {
 		epidField.setInvalidCommitted(true);
 		style(epidField, ERROR_COLOR_PRIMARY);
 
+		TextField externalIdField = addField(CaseDataDto.EXTERNAL_ID, TextField.class);
+		externalIdField.setInvalidCommitted(true);
```
MateStrysewske, line 124: Not necessary, the external ID doesn't have any warning associated with it that would prevent the form from being saved

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
alexcaruntu-vita: Then should I remove it from epidField as well (see above)?

MateStrysewske: No, the epid number field has this warning associated

MateStrysewske (approved):

Issue #4133, opened 2021-01-26
Examine slow CI Times related to PerformanceLoggingInterceptor


DavidBaldsiefen:
### Problem Description

The time it takes for `mvn verify` to complete on GitHub actions increased massively around the implementation of the `PerformanceLoggingInterceptor` class with #3818. It went from roughly 15 to over 20 minutes. Just removing the implemented files allowed the CI process to complete much faster again (See PR #4121).

### Proposed Change

First of all, it should be examined what part of #3818 causes this slow down. After that it should be determined if and how this problem can be solved, or if we should simply accept the slower times.

### Additional Information

Graph of CI-Timings over time:
![image](https://user-images.githubusercontent.com/4655486/105843693-547d3b00-5fd8-11eb-9950-26d9f34f61ed.png)

#3818 was implemented around position ~95

DavidBaldsiefen:
It seems like `PerformanceLoggingInterceptorTest.java` is the lone file responsible for the ~10minute increase in testing times. However, according to the logs, it only takes 2 seconds to execute. See PR #4134

StefanKock:
Running unit tests of sormas-backend on my machine with regards to `PerformanceLoggingInterceptorTest`
10:00 min = 604s as-is
9:40 min = 578s Mockito.reset(context, method)
8:00 min = 472s test method interior removed

I think I will rework the test that I do not have to mock `InvocationContext` and `Method`, that should remove the problem.

DavidBaldsiefen:
Hmm 153b132 didn't finish much faster. Would it somehow be possible to move this test invocation to the very end (or very beginning) of all tests? This way we could see if the problem is some missing cleanup, causing performance to get lost from the time on when the test is first executed.

StefanKock:
Problem solved by not mocking `InvocationContext` and `java.lang.reflect.Method`, mock reset did barely help.

PR #4145, opened 2021-01-27
#4133: avoid mocking

StefanKock:
Testing CI performance


StefanKock:
I took the rework and merged it to development, it reduced the run from 24 min to 16 min

PR #4146, opened 2021-01-27
Bugfix 3923 inconsistent show more less filter button

alexcaruntu-vita:


MateStrysewske (changes requested):
The label still displays "Show more filters" when navigating to another page and back to the original one after setting e.g. the region filter

syntakker:
The "Show more/less filters" button has a border which appears on activation and randomly disappears afterwards. It should not have a border.
Screenshot_20210223_121400

After switching to German language, the layout ist still inconsistent: There is a linebreak after "Show less filters" button, and spacing is odd.
image

Please move the whole section "show more/less filters", "reset filters", "apply filters" to an extra line of its own, which is consistently displayed below all the filter inputs.

syntakker:
> The label still displays "Show more filters" when navigating to another page and back to the original one after setting e.g. the region filter

This is fixed.

syntakker (changes requested):
See https://github.com/hzi-braunschweig/SORMAS-Project/pull/4146#issuecomment-784138579

MateStrysewske:
@syntakker @alexcaruntu-vita I have to disagree with one part of the latest change request. The border is nothing that we should change right now because it's a standard Vaadin style and it's way too minor to "fix" this.

syntakker (approved):

Issue #4125, opened 2021-01-25
CQ/CT value field does not get enabled during result entry in sample creation window [0.5]


kwa20:
### Bug Description
When entering a pathogen test result in the sample creation window, the cq/ct value field does not get enabled when PCR or cq value detection were selected as the type of test. This also affects the DEMIS interface.
### Steps to Reproduce
1. Add a new sample in any entity.
2. Add a test result in the same process.
3. Select either PCR or cq value detection as the type of test.
4. Check for subsequent cq/ct value field.

### Expected Behavior
When PCR or cq value detection were selected as the type of test, the cq/ct value field should get enabled. This should also be true when processing lab messages through the DEMIS interface.
### Screenshots
<img width="475" alt="cq ctValueMissing" src="https://user-images.githubusercontent.com/70317594/105725988-dfebc300-5f29-11eb-9dcc-d70d06068869.PNG">
<img width="329" alt="cq ctValueMissing2" src="https://user-images.githubusercontent.com/70317594/105726021-e9752b00-5f29-11eb-8de4-128a2155276b.PNG">

### System Details
* Device:
* SORMAS version: 1.54.0
* Android version/Browser: firefox
* Server URL: test1402.sormas-oegd.de
* User Role: admin, national user

bernardsilenou:
I just happen to see this issue now now, If it not to late, can we migrate cq value detection to PCT, its duplicate.
CQ value detection is not a type of test but threshold af a PCR test.
@MateStrysewske @kwa20  Any idea why this was added as a type of test?

kwa20:
@bernardsilenou judging from when it was translated on crowdin, it has been around for at least a year.

bernardsilenou:
@kwa20 OK, not bad just duplicate, in case we have to map type of test to external systems, then we just need to check both entries. If not possible to be handles in this issue, then we can create a new one.

MateStrysewske:
@bernardsilenou It's very likely that it was added as a type of test because we were requested to do so ;)

bernardsilenou:
@MateStrysewske I also think so, do we need a new ticket to merge it with PCR? I checked with other and decision is to remove it.

MateStrysewske:
@bernardsilenou Yes, that would be good

raghupola-vg:
Tested on version 1.55.0-Snapshot on test environment and an Error appears while saving the test result.
[...]

PR #4155, opened 2021-01-28
#4125 - Enable CQ/CT value field in sample creation window

alexcaruntu-vita:


sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleCreateForm.java
```diff
@@ -90,6 +95,28 @@ protected void addFields() {
 				false,
 				I18nProperties.getValidationError(Validations.afterDate, testDateField.getCaption(), sampleDateField.getCaption())));
 
+		pathogenTestResultField.addValueChangeListener(e -> {
+			PathogenTestResultType testResult = (PathogenTestResultType) e.getProperty().getValue();
```
MateStrysewske, line 99: Move this code to a separate method (setCqValueFieldVisibility with testType and testResult as parameters) and call it in both listeners to avoid duplicate code

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleCreateForm.java
alexcaruntu-vita: Done

MateStrysewske (changes requested):
The CQ value is not properly taken over when creating the sample. It's empty when I open the pathogen test afterwards.

MateStrysewske (approved):

Issue #3850, opened 2020-12-21
Automated Window scrolling in contact/events handicaps workflow


Lucas-Walter-GA-Tuebingen:
### Bug Description
Window scrolls to the top automatically in section "contacts" when doing some entries below Responsible Region. 
Similar in the event section. 

### Steps to Reproduce
1. Go to contacts
2. Fill in First Name of Contact Person, Last Name of Contact Person, Sex, Report Date, Disease of Source Case, Date of Last Contact, Responsible Region, Responsible District
3. Scroll down to type of contact
4. Select one of the Contact Reasons. 

### Expected Behavior
Window should not scroll to top as this . 

### System Details
* Device: Windows 10 64-Bit
* SORMAS version: 1.52.1
* Android version/Browser: Chrome
* Server URL:
* User Role: Admin

kwa20:
duplicate of #2447

Issue #3601, opened 2020-11-26
Case, Contact and event creation forms jump back to the top after scrolling and clicking


kwa20:
### Bug Description
When trying to enter data in the case, contact and event creation form after scrolling, the window jumps back to the top before you could make an entry. 
### Steps to Reproduce
1. Create a case, contact or an event.
2. Scroll inside of the creation form
3. Fill any available field in the form.
4. Check whether the window jumps back to the top of the creation form.

### Expected Behavior
The window should not jump back to the top when clicking fields and trying to enter data.
### System Details
* Device:
* SORMAS version: 1.51.1
* Android version/Browser: Firefox
* Server URL: test1401.sormas-oegd.de
* User Role: admin, national user

ChristopherRiedel:
I can't reproduce this.

kwa20:
@ChristopherRiedel It usually happens to me when selecting a type of contact after having selected a disease. Maybe zoom in far enough so you're forced to scroll? I think it also happened during the Review when creating a contact to show the duplicate detection for conversion to cases. Maybe the video on Skype might help.

bernardsilenou:
[NetworkVideo3.zip](https://github.com/hzi-braunschweig/SORMAS-Project/files/5607949/NetworkVideo3.zip)

bernardsilenou:
This happened on test1401.

Candice-Louw:
@ChristopherRiedel - for Contacts specifically (used daily by contact tracers) if a user scrolls, and then clicks ```Disease of source case``` or ```Type of Contact```, this jump occurs:

![contactjumpingerror](https://user-images.githubusercontent.com/65529128/100520359-b6b6ec80-319d-11eb-9d3b-dc9f2bd18ef5.gif)

kwa20:
It also just happened to me on the sample creation form.

BarnaBartha:
Duplicate/similar to #2447 -  it is a generic issue and not an easy fix with vaadin. Considering the UI is planned to be rewritten with a JS framework i suggest we don't commit to this bug right now.

bernardsilenou:
> Duplicate/similar to #2447 - it is a generic issue and not an easy fix with vaadin. Considering the UI is planned to be rewritten with a JS framework i suggest we don't commit to this bug right now.

Dear Barna,
You are absolutely right but the truth is that we do not have a definite time on our timeline on when we would drop vaadin. Since this is a bug, it would be better that we have a workaround solution for now.

PR #4158, opened 2021-01-28
#2447 Prevent unintended page scrolling on create forms

### LINKED ISSUE #2447: Type of Place and District drop down selections under New event creation are buggy on the UI
### Bug Description
Drop downs for **Type of Place** and **District** appear at other positions of the page when clicked on them and UI appears bad. User has to scroll the web page to view it correctly.

Note: This happens only when the **Source Type** is selected as Media/News and Hotline/Person as they have additional fields.

### Steps to Reproduce
1. Login to Sormas and Click on events
2. Select Create New Event and Choose screening or Cluster (Might appear on other tabs too)
3. Select the Source Type as Media/News or Hotline/Person
4. Now try selecting the Type of Place and notice where the drop down is shown.
5. Do the same with District selection.

### Expected Behavior
The drop downs must appear at the right place and the web page must not be flickered/moved. 

### Screenshots
Drop downs_New event

### System Details
* Device:
* SORMAS version: 1.44.0- snapshot

DavidBaldsiefen:
This should fix #2447 (As well as #3601 and #3850).

However this change might have unintended side effects as it changes some "top-level" css, so the UI should be tested thoroughly.

**Please not that you might need to do a full page-reload as most browsers store css in their caches!**  (Ctrl-F5 on Firefox)

### DIFF
--- a/sormas-ui/src/main/webapp/VAADIN/themes/sormas/components/window.scss
+++ b/sormas-ui/src/main/webapp/VAADIN/themes/sormas/components/window.scss
@@ -2,6 +2,8 @@
 @mixin sormas-window {
 
 	.v-window {
+
+		height: 98% !important;
 	
 	  .v-window-outerheader {
 
--- a/sormas-ui/src/main/webapp/VAADIN/themes/sormas/components/window.scss
+++ b/sormas-ui/src/main/webapp/VAADIN/themes/sormas/components/window.scss
@@ -3,38 +3,44 @@
 
 	.v-window {
 
-		height: 98% !important;
-	
-	  .v-window-outerheader {
+		.v-window-outerheader {
 
-		border-bottom: 3px solid $v-focus-color;
+			border-bottom: 3px solid $v-focus-color;
+			background-color: white;
 
-        .v-window-header {
-          padding: 16px 32px 8px 32px;
-          font-size: 24px;
-          font-weight: 600;
-          color: $s-primary-color;
-        }
-	  
-	    &::after {
-	      content: null;
-	    }
-	  }
-	
-	  .v-window-contents > .v-scrollable > .v-layout {
-	  
-	    &.v-margin-left {
-	      padding-left: 32px;
-	    }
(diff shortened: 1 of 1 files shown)


MateStrysewske:
@DavidBaldsiefen This is definitely not a valid solution because it increases the size of ALL windows to 98 % of the browser height, which looks really, really bad when there's not much content to display.

DavidBaldsiefen:
> @DavidBaldsiefen This is definitely not a valid solution because it increases the size of ALL windows to 98 % of the browser height, which looks really, really bad when there's not much content to display. We could do this on a form-by-form basis, but that doesn't really help either because it heavily relies on a) the display resolution and b) the zoom factor.

Ok, I see. I tried using max-height too, but for some reason that didn't work. The problem right now is that the forms height is automatically set by vaadin, but removed once a dropdown is opened. Thus, the overflow property breaks. If I understood correctly, this problem only exists with forms extending beyond the screen's size anyway. So a Form-By-Form basis might still be viable, as most common display resulitions and zoom factors should have the same forms extending beyond the screens size (it's mainly contact and event creation tabs).

What struck me as odd is that this issue does not seem to exist with the user creation/edit form.

I'll look for a better solution.

DavidBaldsiefen:
I've found another possible solution. Locally I couldn't find any negative sideffects on other forms.

The git diff seems to be somewhat confused, but the only changes should be that the white `background-color` for the header is now enforced and that `v-scrollable` now has a fixed max-height, which should prevent it from extending beyond the screens resolution.

MateStrysewske (changes requested):
The scrolling problem is fixed, but I now don't have any margins in popup windows anymore, which looks really bad (Chrome, Windows 10):

image

DavidBaldsiefen:
I knew something was off! But for some reason i couldn't quite put my finger on it. I guess if you stare at css for to long you get some kind off tunnel vision :D
Fix is on the way.

MateStrysewske (approved):

Issue #4095, opened 2021-01-21
Migrate dependency xdocreport to docx4j [1]


StefanKock:
### Feature Description

In `TemplateEngineService`, the xdocreport library is to be replaced by docx4j both to extract mergefield names from .docx templates and to perform the merge with mergefield values.

### Problem Description

xdocreport seems to be abandoned, that's why we want to switch to better maintained lib for filling docx templates.

StefanKock:
Waiting for decision on which minor version (1.53 or 1.54) to be fixed.

StefanKock:
Final result: docx4j is used to preprocess docx templates before filling them with xdocreport

Issue #4057, opened 2021-01-18
User feedback from document generation


syntakker:
### Feature Description

Requested by users for document generation:

- sample and pathogen tests can be chosen, data is available in templates
- current user data is available
- generation of quarantine orders for event participants
- separation of document templates for cases, contacts, event participants

Already implemented in other issues:
- improved error handling (#3222)
- custom placeholders for empty variables, variables not displayed when empty (#3991)
- printable documents for events including event actions (#2906)
- code quality issues fixed (#3480)

syntakker:
Implemented in  #4171.

AndyBakcsy-she:
Validated ticket on the latest version of sormas version 1.55.snapshot

raghupola-vg:
When a Quarantine order /event handout is created, the document name is saved only with first 5 letters of the UUID and skipping the last letter.
For ex: if an event id is VAW3OD when a quarantine template is generated, it saves as VAW3O_xxxxxxxx.xxx.
It should be saved as VAW3OD_xxxxxx.xxx

syntakker:
Finding from @kwa20: When there is only one sample or pathogen test, it should be preselected in the respective ComboBox.

StefanKock:
Findings needed to be fixed, that's why the milestone is appropriatly change to Sprint 100.

syntakker:
Feedback from Review @StefanKock: `DataHelper.getShortUuid` should not be graceful for Uuids that are too short.

AndyBakcsy-she:
Validated on the latest version of Sormas Version 1.57.0-snapshot

Issue #3991, opened 2021-01-14
Make String "./." optional for empty variables in printed documents


Marko-ilmkreis:
If you are using the function for printing documents every empty field is set as "./.". This means for every field which might be empty or only used for some persons you have to set an if clause in Word.  Which is very annoying for two reasons:

* most of the times you have to update the field manually to get it working
* sometimes for unknown reason the function failed, SORMAS shows a message that there was an error and the complete word-file is emtpy

e.g. we need to use the field $case.person.address.additionalInformation for people which are living in retirement homes to adress an c/o. As this doesn't affect all cases/contacts we have  to have to if clauses.  One for the line feed (only line feed of field is not empty or "./") and on for the field itself (If its "./" leave it blank, if not print the data).

The best would be to leave every field which is empty just empty.

Marko-ilmkreis:
I discovered that the failure of the function might be belong to the length of the field description.

e.g. this works
![grafik](https://user-images.githubusercontent.com/74791118/104552956-0e1ff780-563a-11eb-8c16-660100aa4704.png)

and the same with the field $case.person.address.additionalInformation  runs into the error. Word cuts the lentgh of the field if displaying it:

![grafik](https://user-images.githubusercontent.com/74791118/104553138-69ea8080-563a-11eb-873d-d6fb537c97ac.png)

Here is an excample-file: 
[case.docx](https://github.com/hzi-braunschweig/SORMAS-Project/files/5812692/case.docx)

syntakker:
The intention of the default placeholder was to provide a solution that fits most of the use cases, both in providing visible feedback for missing values and inserting something that produces an acceptable output. "./." seemed a reasonable choice to me.

An option would be to make the default placeholder customizeable. This requires to specify how this is done, e.g. on an per-instance basis via some config. Do you have a suggestion?

Marko-ilmkreis:
@syntakker In my opinion a placeholder is not needed. Maybe the solution could be a config for the placeholder, which also allows to have none.

syntakker:
@Marko-ilmkreis Ok. I still see arguments to have a visible indication when a mergefield is intentionally left empty (both for the creator and the receiver), but making it configurable is no problem.
I would do this via the `sormas.properties` file where all the configuration for the instance is defined. Would this do for you?

Marko-ilmkreis:
@syntakker that would be fine. We could than ask Netzlink to set it for us. Do you think it is possible to have it for the current sprint?

[...]

Issue #3480, opened 2020-11-13
Enhance code quality of quarantine document generation


StefanKock:
### Problem Description

Follow up to #2696. I noticed these points because of 1. that the units tests fail on my local system.

1. Running some tests on a local path with blank (' ') fail: 2x _TemplateEngineServiceTest_, 6x of 8 in _QuarantineOrderFacadeEjbTest_
2. _TemplateEngineService.getTempDir()_ is not used.
3. _TemplateEngineService.logger_ and _.configFacade_ are superfluous.
4. _TemplateEngineServiceTest_ does not need to run as BeanTest, avoids about 14s test runtime.
5. _TemplateEngineServiceTest_: XWPFWordExtractor not closed
6. _QuarantineOrderFacadeEjbTest_: XWPFWordExtractor not closed, rdcf2 not used
7. The mentioned classes do not follow the code formatting rule "Use one blank line after method definition (but usually not for one liners like getters/setters or delegation)." , see https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/CONTRIBUTING.md

### Proposed Change

- Avoid persisting the path as String but encode it like this
```
		URL file = testClass.getClassLoader().getResource("templates/" + templateFileName);
		return new FileInputStream(new File(file.toURI()));
```
- Clean up all other other points mentioned above.

JonasCir:
@syntakker @Seal33 FYI

syntakker:
Implemented in  #4171.

PR #4171, opened 2021-01-29
Feature 3222 docgeneration

### LINKED ISSUE #2906: “Event” printing function
### Situation Description
Sormas does not offer a print function. It would be interesting in that it allows information to be transmitted to a third party without giving them access as a user. The print function would be particularly interesting for tracking events and associated actions.

### Feature Description
Develop a function allowing users to print an HTML page from the information entered for each event:

- title of the event 

- event identifier

- event status

- event date

- description of the event

- Region, department and municipality of the event

- Community Contact Person

- For each action :

Title of the action
Description of the action
Creation date
Date of last modification
Status of the action

### Additional Information

syntakker:
Fixes #2906, #3222, #3480, #3991, #4057.
The fixes from #4095 are addressed in the refactoring.


sormas-backend/src/main/java/de/symeda/sormas/backend/docgeneration/EventDocumentFacadeEjb.java
```diff
@@ -0,0 +1,54 @@
...
+import de.symeda.sormas.api.docgeneneration.EventDocumentFacade;
+import de.symeda.sormas.api.event.EventParticipantFacade;
+import de.symeda.sormas.api.event.EventReferenceDto;
+
+public class EventDocumentFacadeEjb implements EventDocumentFacade {
+
+	private static final DocumentWorkflow DOCUMENT_WORKFLOW = DocumentWorkflow.EVENT_HANDOUT;
+
+	@EJB
+	DocumentTemplateFacade documentTemplateFacade;
```
StefanKock: private modifier missing

syntakker:
Feedback from review (@StefanKock):
- missing private modifer for EJBs, e.g., in `EventDocumentFacadeEjb`
- restrict EntityDtoHelper to `EntityDto` and `ReferenceDto` if possible
- custom path reset in `@Before` and `@After` methods in `AbstractDocGenerationTest`
- remove `try`/`catch` in `DocumentTemplatesView`

sormas-api/src/main/java/de/symeda/sormas/api/EntityDtoAccessHelper.java
```diff
@@ -15,12 +30,13 @@
...
 
-	public static Object getPropertyValue(Object entity, String propertyKey) throws InvocationTargetException, IllegalAccessException {
+	public static Object getPropertyValue(HasUuid entity, String propertyKey) throws InvocationTargetException, IllegalAccessException {
 		if (entity == null) {
 			return null;
 		}
 		Class<?> entityClass = entity.getClass();
-		while (entityClass != null) {
+		while (entityClass != null
+			&& (EntityDto.class.isAssignableFrom(entity.getClass()) || ReferenceDto.class.isAssignableFrom(entity.getClass()))) {
```
StefanKock, line 39: Why is this condition:
1. Not present on current development branch
2. Would needlessly reduce the functionality to only some implementations of `HasUuid`.

StefanKock (changes requested):
A needless condition still present in `EntityDtoAccessHelper`

sormas-api/src/main/java/de/symeda/sormas/api/EntityDtoAccessHelper.java
syntakker: 1. in the current development branch, the superclass recursion is restricted by
```java
entityClass = HasUuid.class.isAssignableFrom(superclass) ? (Class<? extends HasUuid>) superclass : null;
```
which leads subsequently to breaking the `while` loop. This is in practice equivalent to the check for type `EntityDto` or `ReferenceDto` (see 2.)
2. Checking for type `HasUuid` is slightly more general than checking for `EntityDto` or `ReferenceDto`, however, the intended use case is `EntityDto` or `ReferenceDto`. Other implementations of `HasUuid` are either irrelevant (`AbstractDomainObject` is not used in this context) or are used in tests only. Thus the more restrictive check for `EntityDto` or `ReferenceDto` is actually more accurate here.

syntakker: However, allowing for `HasUuid` does not hurt here. Removed obsolete checks, too.

PR #4179, opened 2021-01-29
Feature 3613 external id token filter

alexcaruntu-vita:


MateStrysewske (changes requested):
Sorry that this was not specified accordingly, but I would like you to integrate the two filters into the already existing text filter (that is e.g. also used to filter by epid number) instead of creating new ones because the filter forms are already extremely crowded, and having to additional text filters that both only serve one specific purpose doesn't really help with that. It's also a lot easier that way because you only need to extend the filter query in the services.

MateStrysewske (approved):

Issue #2895, opened 2020-09-16
Allow to define specific colors for campaign diagram series


MartinWahnschaffeSymeda:
### Situation Description
Currently the campaign diagram colorizes the shown series based on pre-defined colours in order of appearance.

### Feature Description
Allow the definition of a specific color for a series in the campaign diagram definition.
The color should be defined as rgb using hexadecimal color codes (e.g. #FF0000 is red, see https://www.w3schools.com/colors/colors_picker.asp).

- [ ] Extend the CampaignDiagramSeries with an optional color property
- [ ] When creating a diagram for the dashboard, check if a color is defined and if so, apply it to the series

Example:

```
[
{"fieldId":"childrenLivingInHouses","formId": "vaccinationMonitoring","color":"#ff0000"},
{"fieldId":"missedChildren","formId":"vaccinationMonitoring","color":"#ff0000"}
]
```

PR #4195, opened 2021-02-01
#2895 - allow defining specific colors for campaign diagram series

BarnaBartha:


ssferrazza (approved):
Reviewed and tested locally. Verified that adding a color overrides the default setting.

Issue #3349, opened 2020-10-30
Campaign dashboard: Allow user to switch from column to bar chart [1]


MartinWahnschaffe:
### Situation Description
The campaign dashboard charts are currently all column charts.

### Feature Description
Users should be able to switch the chart type to bar chart.

- [x] Place the option to switch in the hamburger menu of the chart (change caption depending on which chart type is currently displayed, similar to percentage/total values)
- [x] DiagramType defines what is default

MartinWahnschaffeSymeda:
For now just implement the switch and keep the height as-is for now. Automatically adjusting the height should probably be done post-pilot.

MateStrysewske:
- [ ] Make sure that the diagram type that is displayed on the Dashboard when opening it is according to the value of the "diagramType" column in the database

PR #4197, opened 2021-02-01
#3349: added ability to switch from column to bar chart, where column…

ssferrazza:
… is default as defined in DiagramType

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/campaigns/CampaignDashboardDiagramComponent.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/campaigns/CampaignDashboardDiagramComponent.java
@@ -18,6 +18,7 @@ import java.util.stream.Collectors;
 
 import com.vaadin.ui.JavaScript;
 import com.vaadin.ui.JavaScriptFunction;
+import de.symeda.sormas.api.campaign.diagram.DiagramType;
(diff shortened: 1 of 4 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/campaigns/CampaignDashboardDiagramComponent.java
```diff
@@ -63,6 +65,7 @@ public CampaignDashboardDiagramComponent(
 			totalValuesWithoutStacks = true;
 		}
 
+		showAsColumnChart = DiagramType.values().length > 0 && DiagramType.COLUMN == DiagramType.values()[0];
```
BarnaBartha, line 68: This solution is ok for switching but would fail if i want to define the default diagram type as 'bar', please extend the enum type and make sure a diagram can be by default bar type

BarnaBartha (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/campaigns/CampaignDashboardDiagramComponent.java
ssferrazza: Okay, agree

BarnaBartha (approved):

Issue #4036, opened 2021-01-15
[SurvNet Interface] Change title of role field in exposures and add new field [0.5]


kwa20:
### Feature Description
Case/Contacts > Epi Data > Exposures > Exposure Details > in the role of
### Problem Description
For the SurvNet interface, role descriptions are required. Thus far such role description do not exist in an exposure and the current "in the role of" field does not fit for this purpose considering its variables.
### Proposed Change
- Add the variable "Unknown" to type of activity (exposureType).

- Move the "in the role of" field to "care/accommodation/work in facility" in #4163 

- Add a drop-down "role" field under exposure details with the following variables:
  - (null)
  - passenger / Passagier
  - staff / Personal
  - nursing staff / Pflegepersonal
  - medical staff / Medizinisches Personal 
  - visitor / Besucher
  - guest / Gast
  - customer / Kunde
  - conservatee / Betreute
  - patient / Patient
  - educator / Erzieher
  - trainee teacher / Ausbilder
  - pupil / Schüler
  - student / Student
  - parent / Eltern
  - teacher / Lehrkraft
  - unknown / Unbekannt
  - other / Sonstige

### Additional Information
<img width="418" alt="Screen Shot 2021-01-28 at 4 01 54 PM" src="https://user-images.githubusercontent.com/70317594/106156772-3d7a4c80-6182-11eb-8433-199b3b92a525.png">

MateStrysewske:
Requires refinement in #4163

MateStrysewske:
* Add "Unknown" to `ExposureType`
* Rename `PatientExpositionRole` to `ExposureRole`
* Replace the values of `ExposureRole` with the values above
* Write an SQL migration script that maps the values as follows (only relevant for test systems):
    * NOT_COLLECTED => null
    * UNKNOWN => UNKNOWN
    * WORKING_AT => STAFF
    * ACCOMMODATED_IN => GUEST
    * CARED_FOR => PATIENT
    * OTHER => OTHER

@kwa20 Is this fine for you?

AndyBakcsy-she:
Verified ticket on the latest version of Sormas Version 1.57.0-snapshot on local environment.

raghupola-vg:
"Role" field on Mobile app has to be under the Exposure Details.

raghupola-vg:
Verified the changes on Mobile App on version 1.57.0-Snapshot

PR #4213, opened 2021-02-02
#4036 - [SurvNet Interface] Change title of role field in exposures a…

sergiupacurariu:
…nd add new field


sormas-api/src/main/resources/captions.properties
```diff
@@ -1092,7 +1092,7 @@ Exposure.prophylaxis=Post-exposure prophylaxis
 Exposure.prophylaxisDate=Date of prophylaxis
 Exposure.riskArea=Risk area as defined by public health institution
 Exposure.exposureDate=Exposure date
-Exposure.patientExpositionRole=In the role of
+Exposure.exposureRole=In the role of
```
MateStrysewske, line 1095: Change to just "Role" and remove the line from all other property files (because it's technically a new field with a new caption then and should be re-translated)

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1877,6 +1877,10 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				getDao(Event.class).executeRaw("ALTER TABLE events ADD COLUMN evolutionDate timestamp;");
 				getDao(Event.class).executeRaw("ALTER TABLE events ADD COLUMN evolutionComment text;");
 
+			case 265:
+				currentVersion = 265;
+				getDao(Exposure.class).executeRaw("ALTER TABLE exposures ADD COLUMN exposureRole varchar(512);");
```
MateStrysewske, line 1882: => varchar(255) because it's an enum

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6296,4 +6296,19 @@ ALTER TABLE exportconfiguration_history
...
+--2020-02-02
+ALTER TABLE exposures RENAME patientexpositionrole TO exposureRole;
+ALTER TABLE exposures_history RENAME patientexpositionrole TO exposureRole;
+
+UPDATE exposures SET exposureRole = NULL WHERE exposureRole = 'NOT_COLLECTED';
+UPDATE exposures SET exposureRole = 'STAFF' WHERE exposureRole = 'WORKING_AT';
+UPDATE exposures SET exposureRole = 'GUEST' WHERE exposureRole = 'ACCOMMODATED_IN';
+UPDATE exposures SET exposureRole = 'PATIENT' WHERE exposureRole = 'CARED_FOR';
+
+UPDATE exposures_history SET exposureRole = NULL WHERE exposureRole = 'NOT_COLLECTED';
```
MateStrysewske, line 6308: It's not necessary to change the values in the history tables (actually it's probably a bad idea because it doesn't allow backtracking what the original values were in case someone is unhappy with the migration)

MateStrysewske (changes requested):

tkaefer (commented):
LGTM, besides the remarks from @MateStrysewske

sormas-api/src/main/resources/captions.properties
sergiupacurariu: done

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
sergiupacurariu: updated

sormas-backend/src/main/resources/sql/sormas_schema.sql
sergiupacurariu: updated

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
MateStrysewske: Still says 512 for me

MateStrysewske (changes requested):

MateStrysewske (approved):

Issue #4178, opened 2021-01-29
Contacts | New Follow-up Visits can't be created


AndyBakcsy-she:
### Steps to Reproduce
1. Open ga1.sormas-oegd.de or on local environment with the latest version of Sormas
2. Go to contacts
3. Open a Contact and go to the Follow-up Visits tab
4. Start creating a new Visit
5. Fill in some data and try to save

### Expected Behavior
A new Visit should be created

###Actual Behavior
An error message is displayed 
### Screenshots
![image](https://user-images.githubusercontent.com/70562142/106295598-5275df00-6259-11eb-9d21-b121326735dc.png)

### System Details
* Device: Windows 10
* SORMAS version: SORMAS
Version: 1.55.0-SNAPSHOT
* Android version/Browser Chrome:
* Server URL: ga1-sormas.oegd.de or Local environment
* User Role: Na
[serverlog.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/5894816/serverlog.txt)
tional user

MateStrysewske:
@markusmann-vg Just fyi, this is only happening in our current dev version, not in 1.54.1.

AndyBakcsy-she:
@MateStrysewske  This is reproducible also on the test.sormas.netzlink.com on 1.55.snapshoot
![image](https://user-images.githubusercontent.com/70562142/106431527-11a6e180-6476-11eb-8c4e-a15748e729b8.png)

FredrikSchaeferVitagroup:
Also the view itself looks quite badly arranged. Is something wrong there, too?

MateStrysewske:
@FredrikSchaefer Pretty sure that @AndyBakcsy-she just posted a snippet of his desktop here, not the whole screen ;)

FredrikSchaeferVitagroup:
Thank you, I'm aware of that :D I mean the visitsEditForm itself. For example: When you look at NERVOUS SYSTEM, the symptom to checck comes right under it. Right underneath (SKIN), the symptom is on the right. 
There also is some discrepancy in other sections: Sometimes three symptoms are on the left side, an one is one the right. In another section, two symptoms are on the left and one ist on the right.
All in all, it looks quite messy.

MateStrysewske:
Yup, there's an issue for that already

raghupola-vg:
Tested on latest deploy on version 1.55.0-snapshot on test environment and the issue is Fixed.

PR #4217, opened 2021-02-02
#4178 - Handle the null SymptomsDto uuid when creating a visit

alexcaruntu-vita:


sormas-backend/src/main/java/de/symeda/sormas/backend/symptoms/SymptomsFacadeEjb.java
```diff
@@ -37,7 +37,9 @@ public Symptoms fromDto(SymptomsDto source, boolean checkChangeDate) {
 			return null;
 		}
 
-		Symptoms target = DtoHelper.fillOrBuildEntity(source, service.getByUuid(source.getUuid()), Symptoms::new, checkChangeDate);
+		String uuid = source.getUuid();
```
MateStrysewske, line 40: I'd prefer a more compact solution here because declaring a variable and then immediately reassigning looks a bit weird to me: 
```
String uuid = source.getUuid();
Symptoms target = DtoHelper.fillOrBuildEntity(source, uuid != null ? service.getByUuid(uuid) : null, Symptoms::new, checkChangeDate);
```

MateStrysewske (changes requested):

MateStrysewske (approved):

Issue #4224, opened 2021-02-03
Failing tests under Windows


syntakker:
### Bug Description
Under Windows, there are test failures in `EventDocumentFacadeEjbTest` and `TemplateEngineTest`, where expected and actual texts to compare differ in line separators.

### Steps to Reproduce
Execute Tests.

### System Details
* Device: Windows 10 PC
* SORMAS version: 1.55.0-SNAPSHOT
* Android version/Browser:
* Server URL:
* User Role:

syntakker:
Duplicate: bug is described in https://github.com/hzi-braunschweig/SORMAS-Project/issues/2906#issuecomment-772381649

PR #4225, opened 2021-02-03
fix text comparison in failing tests under Windows (#4224)

syntakker:
Clean line separators `\r\n` before text comparison.


syntakker:
Bugfix is moved to PR #4227, where it relates to the orginal issue #2906.

Issue #4156, opened 2021-01-28
Provide SQL statements to delete cases and contacts from the database [1]


MartinWahnschaffe:
### Problem Description
In France we have a lot of cases and contacts in the system that are no longer relevant and need to be fully deleted.

### Proposed Change
- [x] Provide an SQL statement that deletes a set of contacts, including all child data (e.g. tasks, samples)
- [x] Provide an SQL statement that deletes a set of cases, including their contacts and all child data

History tables don't need to be included here.

Important: Actual deletion of data will lead to problem when the mobile app or other data synchronization mechanisms are used on the same system. For the given use case this is not relevant and can be ignored.

**The goal of this issue is NOT to provide a fully fledged deletion mechanism for data!**

bernardsilenou:
@MartinWahnschaffe 
I think we may also need:
Provide an SQL statement that deletes a set of events, including all child data (e.g. e_tasks, e_actions, event participants, ep_samples, etc)

Issue #4137, opened 2021-01-26
Activate vaccination status for COVID-19 cases, contacts and event participant [3]


bernardsilenou:
### Situation Description
* The case form for vaccine preventable diseases has the "additional medical information" section containing information about the vaccination status of the case person. Since COVID-19 did not have a vaccine, we did not activate these vaccine related variables. 
### Feature Description
* Add/adjust these fields for cases, contacts and event participants (start with cases and push that to a feature branch, continue with contacts and then event participants); all fields EXCEPT vaccine name and manufacturer need to be visible for all diseases that vaccinationStatus is currently enabled for, plus Covid-19; vaccine name and manufacturer are only visible for Covid-19

* Add the new fields to the detailed and custom case export, the detailed contact export and the detailed event participant export

* List of variables to capture (all countries, all instances):
    - vaccination status for this diseases:  vaccinated/unvaccinated/unknown (this already exists)
	if  vaccination status == vaccinated, show the following additional variables in this order:
    - date of first vaccination	
    - date of last vaccination (exist already)
	- source of vaccination information: vaccination card/ oral communication/ no evidence, unknown (this exist but  no evidence and unknown options are to be added)
	- `VaccineName`: drop down with options: (options for covid only, other diseases has only others + free txt. 
	If possible this list should be configurable for users to upload what they want, else just hard code it for now.)
	     - Pfizer–BioNTech 
	     - Moderna 
	     - Oxford–AstraZeneca
	     - Ad26.COV2.S 
	     - NVX-CoV2373 
	     - Sanofi - GSK	    
        - Unkown             
	     - Other (free txt)
  
	 - `VaccineManufacturer`: dropdown with options; this should have a reference to a vaccine name, selecting a vaccine name automatically selects the associated manufacturer (relation will be specified in a comment by @Candice-Louw below)
	   - AstraZeneca
	   -  Johnson & Johnson
	   -  Moderna
	   -  Novavax
	   -  Pfizer & BioNTech
	   -  Sanofi & GlaxoSmithKline
	   - Unknown
	   -  others + free text
	    
	 - number of valid doses= rename to "vaccine dose"; add a validator to this field that prevents saving when anything but integers from 1 to 10 (or empty) are typed in; keep the type of the field because existing servers will have values with strings in here
	 
	 - INN (description: International non-proprietary name): free txt
	 - Batch number:  free txt	 	 
	 - UNI code (description: Unique ingredient identifier code): free txt
[...]

PR #4226, opened 2021-02-03
#4156 - provide sql scripts for contact and case deletion

BarnaBartha:


sormas-backend/src/main/resources/utils/deleteCases.sql
```diff
@@ -0,0 +1,156 @@
+DROP TABLE IF EXISTS delete_cleanup;
```
StefanKock, line 1: Add `BEGIN TRANSACTION;` instead of the `DROP TABLE` statement to encapsulate the whole delete run in one transaction.

Finish the whole script with `COMMIT; ` and an empty line after that.

StefanKock, line 1: Add a comment block at the start (draft, maybe adjust and add more info if you have it in mind):
-- Deletes all cases that are marked as deleted = TRUE.
-- All linked data will also be deleted, including contacts and event participants linked to the respective case. The person will be deleted if no other entity references it.

StefanKock (changes requested):
Some remarks concerning documentation at script begin and encapsulate it in one transaction (if something goes wrong, nothing is deleted and the data will noch be in corrupt state). We can then also test it with ROLLBACK at the and instead ot COMMIT.

I think we should join the contacts script into the cases script and rename it to deleteObsoleteData.sql .

After that we should run it against some TEST instances.

StefanKock (approved):

StefanKock (changes requested):
I ran into this problem:
```
sormas=# SELECT pg_temp.truncate_rows();
ERROR:  duplicate key value violates unique constraint "delete_cleanup_pkey"
DETAIL:  Key (table_name, id)=(event, 45245) already exists.
CONTEXT:  SQL statement "INSERT INTO delete_cleanup (SELECT 'event', event_id FROM eventparticipant WHERE id IN (SELECT id FROM delete_cleanup WHERE table_name = 'eventparticipant')
      and not exists (select 1 from eventparticipant ep where ep.event_id = event_id and ep.id <> id))"
PL/pgSQL function pg_temp_20.truncate_rows() line 36 at EXECUTE
```

1. It was not the task to delete events. Please remove this.
2. I do not now if other mechanisms might find objects several times for deletion. That would run into the same duplicate key constraint.
3. I have the impression that visits of contacts might be left behind when all linked contacts and case are deleted. This ManyToMany might run into the problem I mentioned in 2.

StefanKock:
We also have a dependency to #4137. I just discussed with @leventegal-she that he will introduce a new Entity VaccinationInfo (sub of contact and eventparticipant, and likely too for cases). If we do not consider this, our script will not work for >= 1.55, if we do it will not work for <= 1.54 .

StefanKock:
I works on test.sormas. But please remove the DROP statements, they make it too noisy.

```
-- With Drop statements

BEGIN
NOTICE:  table "delete_cleanup" does not exist, skipping
DROP TABLE
CREATE TABLE
NOTICE:  function pg_temp.trunc_register(pg_catalog.varchar,pg_catalog.varchar,pg_catalog.varchar) does not exist, skipping
DROP FUNCTION
CREATE FUNCTION
NOTICE:  function pg_temp.trunc_register(pg_catalog.varchar,pg_catalog.varchar,pg_catalog.varchar,pg_catalog.varchar) does not exist, skipping
DROP FUNCTION
CREATE FUNCTION
NOTICE:  function pg_temp.trunc_delete(pg_catalog.varchar,pg_catalog.varchar,pg_catalog.varchar) does not exist, skipping
DROP FUNCTION
CREATE FUNCTION
NOTICE:  function pg_temp.trunc_delete(pg_catalog.varchar,pg_catalog.varchar) does not exist, skipping
DROP FUNCTION
CREATE FUNCTION
NOTICE:  function pg_temp.trunc_delete(pg_catalog.varchar) does not exist, skipping
DROP FUNCTION
CREATE FUNCTION
NOTICE:  function pg_temp.truncate_rows() does not exist, skipping
DROP FUNCTION
CREATE FUNCTION
 truncate_rows
---------------

(1 row)

    table_name    | delete_rows
------------------+-------------
 cases            |         108
 clinicalcourse   |         108
 contact          |           2
 epidata          |         110
 eventparticipant |          15
 healthconditions |           2
 hospitalization  |         108
 maternalhistory  |         108
 person           |         107
 porthealthinfo   |         108
 symptoms         |         113
 therapy          |         108
 visit            |           5
(13 rows)

ROLLBACK
```

```
-- Without drop statements

~ $ psql sormas < 4156_delete_script_test.sql
BEGIN
CREATE TABLE
CREATE FUNCTION
CREATE FUNCTION
CREATE FUNCTION
CREATE FUNCTION
CREATE FUNCTION
CREATE FUNCTION
 truncate_rows
---------------

(1 row)

    table_name    | delete_rows
------------------+-------------
 cases            |         108
 clinicalcourse   |         108
 contact          |           2
 epidata          |         110
 eventparticipant |          15
 healthconditions |           2
 hospitalization  |         108
 maternalhistory  |         108
 person           |         107
 porthealthinfo   |         108
 symptoms         |         113
 therapy          |         108
 visit            |           5
(13 rows)
```

StefanKock:
@BarnaBartha Please change the end of the script to have a dry run as default:
```
-- Use rollback for a dry run to see how long the operation takes, if errors occur and what amount of data will be deleted.
ROLLBACK;

-- Use commit to really delete the data (cannot be undone).
-- COMMIT;
```

Issue #3337, opened 2020-10-29
Developer Option | Configuration | Creation of Cases and Contacts is dependent on multiple fields and this should be changed


AndyBakcsy-she:
### Bug Description
This issue details the dependencies which should be fixed in order for people to use the Developer options without workarounds
### Steps to Reproduce
Issue # 1
1. Go to the Developer Options under Configuration when logged in as Admin 
2. Try and Generate some Cases/Contacts without selecting the Disease of Case/Contact

Issue # 2
1. Try and Generate some new Contacts without selecting all the Checkboxes
2. Select only one of the checkboxes and try to Generate Contacts

### Expected Behavior
In the case of issue 1.
Creation of Cases/Contacts should not be dependent on the selection of a Disease of Case/Contact

In the case of issue 2.
Users should be allowed to generate Contacts independently of which and how many checkboxes have been selected

### System Details
* Device:
* SORMAS version:
* Android version/Browser:
* Server URL:
* User Role:

MartinWahnschaffeSymeda:
We should (taken from #3247)

Use the Case/ContactGenerationConfig and move the code that is doing the actual generation to the backend
Do the generation in batches of ~50 cases/contacts and use @transactional(REQUIRES_NEW) for each batch

AndyBakcsy-she:
Validated ticket on the latest version of sormas 1.57.0-snapshot

StefanKock:
This ticket does not improve the performance/speed of the software, I therefore removed the label.

PR #4287, opened 2021-02-08
#4185 Add event generation to devmode

### LINKED ISSUE #4185: Add an option to generate Events in DevMode
### Situation Description

Currently, the DevMode can be used to generate random cases and contacts for testing and showcase purposes. The same would be needed for events, as it can be quite a lot of work to generate events manually and then fill each with participants.

### Feature Description

Add the same for events, with an option to fill the events with participants.

### Possible Alternatives

In theory, SORMAS data-gen could be used, but that is quite elaborate.

DavidBaldsiefen:
Closes #4185 and #3337 (some other issues regarding the existing case and contact generation were solved too).

Moving the generation to the backend needs to be handled in another issue, as it's not that important at the moment.


sormas-ui/src/main/java/de/symeda/sormas/ui/configuration/DevModeView.java
```diff
@@ -323,6 +441,18 @@ private VerticalLayout createContactGeneratorLayout() {
...
+		"Party",
+		"Funeral",
+		"Concert",
+		"Fair",
+		"Rallye",
+		"Demonstration",
+		"Football Match",
+		"Tournament",
+		"Festival",
+		"Carinval" };
```
MateStrysewske, line 455: => Carnival

sormas-ui/src/main/java/de/symeda/sormas/ui/configuration/DevModeView.java
```diff
@@ -332,6 +462,12 @@ private static boolean randomPercent(int p) {
 		return random().nextInt(100) <= p;
 	}
 
+	private static int randomInt(int min, int max) {
+		if (max <= min)
```
MateStrysewske, line 466: Use brackets even for one-line statements, please

sormas-ui/src/main/java/de/symeda/sormas/ui/configuration/DevModeView.java
```diff
@@ -508,6 +647,11 @@ private void generateContacts() {
 		ContactGenerationConfig config = contactGeneratorConfigBinder.getBean();
 
 		List<Disease> diseases = config.getDisease() == null ? FacadeProvider.getDiseaseConfigurationFacade().getAllDiseases(true, true, true) : null;
+		Disease disease = config.getDisease();
```
MateStrysewske, line 650: Move this definition one line up, you can then use it as well in the definition of `List<Disease> diseases`

sormas-ui/src/main/java/de/symeda/sormas/ui/configuration/DevModeView.java
```diff
@@ -803,4 +1087,112 @@ public void setCreateWithVisits(boolean createWithVisits) {
...
+			this.maxContactsPerParticipant = maxContactsPerParticipant;
+		}
+
+		public int getPercentageOfCases() {
+			return percentageOfCases;
+		}
+
+		public void setPercentageOfCases(int percentageOfCases) {
+			this.percentageOfCases = percentageOfCases;
+			if (this.percentageOfCases >= 100)
```
MateStrysewske, line 1191: Use brackets

MateStrysewske (changes requested):
* The script currently can create cases resulting from events with a disease that differs from the event disease; cases resulting from events can not have a different disease than the event disease; events WITHOUT a disease shouldn't have any resulting cases at all

DavidBaldsiefen:
> The script currently can create cases resulting from events with a disease that differs from the event disease; cases resulting from events can not have a different disease than the event disease; events WITHOUT a disease shouldn't have any resulting cases at all

Thanks for the review! Currently, my intention was that events where no disease is selected will be created with a random disease, and thus all participants and resulting cases will also have that diesease. Would that be ok? (It will still require me to change a line I missed).

Alternatively one could create disease-less events with no cases or contacts

MateStrysewske (changes requested):
I still get cases with diseases that differ from the event diseases. For example, I just generated 10 events with Measles, and the while some of the cases resulting from these events are Measles case, I also have some with Meningitis, Polio, Dengue, Cholera, etc. It seems like the event disease is overridden when the case is actually created.

DavidBaldsiefen:
@MateStrysewske it should work now, thanks for the review!

MateStrysewske (approved):

PR #4310, opened 2021-02-09
Feature 4304 case completeness count fix

BarnaBartha:


MateStrysewske (approved):

tbroyer:
Looking at Glowroot traces from France (in preprod server), and now seeing those `exists()` calls: they need optimization too!
image
Every `exists()` call returns as many rows as there are in the queried table, due to the `FROM Contacts contacts0_`. There should be no `FROM` clause in such queries (I have no idea how you'd do that with JPA/Hibernate though; if not possible then maybe just add a `LIMIT 1`?)

StefanKock:
@tbroyer Thanks for info, I found this in `CaseFacadeEjb::calculateCompleteness` and wrote this down as a performance finding.

PR #4317, opened 2021-02-09
#4276 Add date and dose validators for vaccinations

### LINKED ISSUE #4276: Vaccination Fields should have validations
### Steps to Reproduce
1. Create a Case with Vaccination
2. Set the dates of the first and last vaccination the wrong way. Use dates from a year when Covid was not existent for example. 
3. Use a float number like 1.345 for the Vaccine dose field

### Expected Behavior
These fields should have validations for this kind of scenario.
### Screenshots
image

### System Details
* Device:Windows
* SORMAS version: Sormas version 1.55.0-snapshot
* Server URL: Localhost
* User Role: NatUser

### Additional Information
this ticket is related to #4137

DavidBaldsiefen:
Closes #4276

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
@@ -651,7 +651,6 @@ public class CaseDataForm extends AbstractEditForm<CaseDataDto> {
 				false,
 				I18nProperties
 					.getValidationError(Validations.afterDate, lastVaccinationDateField.getCaption(), firstVaccinationDateField.getCaption())));
-		System.out.println(firstVaccinationDateField.getValidators().size());
 
 		ComboBox vaccineName = addField(CaseDataDto.VACCINE_NAME);
 		ComboBox vaccineManufacturer = addField(CaseDataDto.VACCINE_MANUFACTURER);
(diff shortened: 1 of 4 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
```diff
@@ -627,14 +627,31 @@ protected void addFields() {
...
+					.getValidationError(Validations.beforeDate, firstVaccinationDateField.getCaption(), lastVaccinationDateField.getCaption())));
+		lastVaccinationDateField.addValidator(
+			new DateComparisonValidator(
+				lastVaccinationDateField,
+				firstVaccinationDateField,
+				false,
+				false,
+				I18nProperties
+					.getValidationError(Validations.afterDate, lastVaccinationDateField.getCaption(), firstVaccinationDateField.getCaption())));
+		System.out.println(firstVaccinationDateField.getValidators().size());
```
MateStrysewske, line 654: Remove this line

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/NumberValidator.java
```diff
@@ -28,9 +28,11 @@
...
 	private BigDecimal minValue;
 	private BigDecimal maxValue;
+	private boolean decimalAllowed;
 
 	public NumberValidator(String errorMessage) {
 		super(errorMessage);
+		decimalAllowed = true;
 	}
 
 	public NumberValidator(String errorMessage, Number minValue, Number maxValue) {
```
MateStrysewske, line 38: You can remove the code from this constructor and instead call the new one like this to avoid redundancies:

`this(errorMessage, minValue, maxValue, true);`

MateStrysewske (changes requested):
Seems like the label wasn't added to the issue, but this also needs to be done in the mobile app, otherwise there might be wrong data coming in from there.

MateStrysewske (approved):

Issue #4309, opened 2021-02-09
[SORMAS2SORMAS] Sharing data should work, even when the receiving SORMAS instance does not have the same facilities and sends back the case [2]


markusmann-vg:
### Feature Description
It's not likely nor doable that every health department has the same infrastructural data and a complete synced list of facilities.
However they should all be able to send and receive cases and everything around it without having the exact same data.

### Proposed Change

If the facility can not be mapped, then it should be linked  to OTHER with the name as a string

- [ ] Sending back would need a check if the name of the facility below OTHER is available in the list of facilities.

Checks:
1. byExternal ID
2. by name

**Prerequisite:**
Every SORMAS L Instance should be updated to the aligned infrastructure List, when they want to use SORMAS X
https://github.com/hzi-braunschweig/SORMAS-Project/issues/4312

PR #4335, opened 2021-02-10
#4309 [SORMAS2SORMAS] Sharing data should work, even when the receiving SORMAS instance does not have the same facilities and sends back the case

leventegal-she:
Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/4309

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/PointOfEntryService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/PointOfEntryService.java
@@ -75,15 +75,15 @@ public class PointOfEntryService extends AbstractInfrastructureAdoService<PointO
 		return em.createQuery(cq).getResultList();
 	}
 
-	public List<PointOfEntry> getByExternalId(String name, boolean includeArchivedEntities) {
+	public List<PointOfEntry> getByExternalId(String externalId, boolean includeArchivedEntities) {
 
 		CriteriaBuilder cb = em.getCriteriaBuilder();
 		CriteriaQuery<PointOfEntry> cq = cb.createQuery(getElementClass());
 		Root<PointOfEntry> from = cq.from(getElementClass());
 
 		Predicate filter = cb.or(
-			cb.equal(cb.trim(from.get(PointOfEntry.EXTERNAL_ID)), name.trim()),
-			cb.equal(cb.lower(cb.trim(from.get(PointOfEntry.EXTERNAL_ID))), name.trim().toLowerCase()));
+			cb.equal(cb.trim(from.get(PointOfEntry.EXTERNAL_ID)), externalId.trim()),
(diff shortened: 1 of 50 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/PointOfEntryService.java
```diff
@@ -66,20 +67,38 @@ public PointOfEntryService() {
...
-		if (!PointOfEntryDto.isNameOtherPointOfEntry(name.trim())) {
+		if (district != null && !PointOfEntryDto.isNameOtherPointOfEntry(name.trim())) {
 			filter = cb.and(filter, cb.equal(from.get(PointOfEntry.DISTRICT), district));
 		}
 
 		cq.where(filter);
 		return em.createQuery(cq).getResultList();
 	}
 
+	public List<PointOfEntry> getByExternalId(String name, boolean includeArchivedEntities) {
```
vidi42, line 78: Maybe rename the `name` variable to `externalId`

sormas-backend/src/main/java/de/symeda/sormas/backend/region/RegionFacadeEjb.java
```diff
@@ -318,7 +318,12 @@ public void saveRegion(RegionDto dto) throws ValidationRuntimeException {
 
 	@Override
 	public List<RegionReferenceDto> getByName(String name, boolean includeArchivedEntities) {
-		return regionService.getByName(name, includeArchivedEntities).stream().map(r -> toReferenceDto(r)).collect(Collectors.toList());
+		return regionService.getByName(name, includeArchivedEntities).stream().map(RegionFacadeEjb::toReferenceDto).collect(Collectors.toList());
+	}
+
+	@Override
+	public List<RegionReferenceDto> getByExternalId(String id, boolean includeArchivedEntities) {
```
vidi42, line 325: Maybe rename `id` to `externalId`

sormas-backend/src/main/java/de/symeda/sormas/backend/region/RegionService.java
```diff
@@ -58,6 +58,24 @@ public RegionService() {
 		return em.createQuery(cq).getResultList();
 	}
 
+	public List<Region> getByExternalId(String externalId, boolean includeArchivedEntities) {
```
vidi42, line 61: Also, wouldn't it make sense to extract this method to `AbstractInfrastructureAdoService` ?

vidi42 (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/region/RegionService.java
leventegal-she: All other getByName methods are in specific entity based services, maybe at a later point could be refactored to avoid these kind of similar methods

vidi42: Alright, let's keep it for later

vidi42 (approved):

PR #4336, opened 2021-02-10
Feature 2994 add country to location details

alexcaruntu-vita:


sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -1942,6 +1942,10 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				currentVersion = 271;
 				getDao(Exposure.class).executeRaw("ALTER TABLE cases ADD COLUMN caseIdentificationSource varchar(255);");
 
+			case 272:
+				currentVersion = 272;
+				getDao(Location.class).executeRaw("ALTER TABLE location ADD COLUMN country_id BIGINT;");
```
MateStrysewske, line 1947: Foreign key is missing (= REFERENCES keyword, check an earlier upgrade for an example)

sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/LocationDialog.java
```diff
@@ -106,8 +108,21 @@ protected void initializeContentView(ViewDataBinding rootBinding, ViewDataBindin
 
 		InfrastructureHelper.initializeHealthFacilityDetailsFieldVisibility(contentBinding.locationFacility, contentBinding.locationFacilityDetails);
 
+		if (data.getCountry() == null) {
+			for (Item countryItem : initialCountries) {
+				Country country = (Country) countryItem.getValue();
+				if (country != null && ConfigProvider.isConfiguredServer(country.getIsoCode())) {
```
MateStrysewske, line 114: I don't think this code can work, because the country ISO codes are three-digit codes while our locales are either two-digit codes in case there's no country specified (e.g. "en") or have a completely different format (e.g. "de-CH"). Also, it's theoretically possible that a server in a specific country might use a different locale, for whatever reason.

I think the best solution here is to use the `country.name` property that already exists, and compare its value with the name of the country. For this, you need to make that property accessible in InfoResource and retrieve it in the Android app, similar to how the locale is currently retrieved.

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6460,11 +6460,16 @@ ALTER TABLE cases_history ADD COLUMN bloodorganortissuedonated varchar(255);
...
 
-
 -- 2021-02-05 Case identification source #3420
 ALTER TABLE cases ADD COLUMN caseidentificationsource character varying(255);
 ALTER TABLE cases_history ADD COLUMN caseidentificationsource character varying(255);
 
 INSERT INTO schema_version (version_number, comment) VALUES (324, 'Case identification source #3420');
 
+-- 2020-02-09 Add Country to location details #2994
+ALTER TABLE location ADD COLUMN country_id bigint;
```
MateStrysewske, line 6470: Foreign key is missing here as well

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventParticipantEditForm.java
```diff
@@ -84,6 +86,11 @@ protected void addFields() {
 			new PersonEditForm(PersonContext.EVENT_PARTICIPANT, event.getDisease(), event.getDiseaseDetails(), null, isPseudonymized);
 		pef.setWidth(100, Unit.PERCENTAGE);
 		pef.setImmediate(true);
+
+		CountryReferenceDto countryDto = event.getEventLocation().getCountry();
+		pef.getAddressForm().getField(LocationDto.COUNTRY).setEnabled(false);
+		pef.getAddressForm()
```
MateStrysewske, line 92: What needs to be set to required here is NOT the person's address, but the responsible jurisdiction fields of the event participant

MateStrysewske (changes requested):

sormas-backend/src/main/resources/sql/sormas_schema.sql
alexcaruntu-vita: Good catch. Solved.

sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/LocationDialog.java
alexcaruntu-vita: Country name is marked as deprecated in sormas.properties. There are many things to consider here. Although we reserved 3 characters for ISO code in our database, I had no problem to create a country with only 2 letters for ISO code. That wouldn't be a problem, since there are two standards for ISO codes, one which uses two letters and one which uses three. If I were to choose a solution, I would vote to enforce ISO code to be two letters long, and keep the current implementation. The name of the country doesn't seem that robust to me (what if we localise them in the app and have them in English in sormas.properties?).

MateStrysewske: Yeah, that's true, because it wasn't really needed anymore until now. It's not a problem to make it "undeprecated" again though if we have another use for it. The problem with the ISO code is that we provide a default list of countries that come with SORMAS, and those use the three-digit ISO codes, so that's already the standard that we can't just change all of a sudden because it would break all running servers that have already imported our default country list (or imported their own country list and used three-digit codes).

You're right that using the country name is language-specific, but that would then be something the server administrators would need to consider. Since they're also responsible for adding the countries to their application, they could ensure that the name of the country in SORMAS and the name of the country in the properties file are identical. Localisation is not a problem in that regard because it's taken care of in a separate localisation file, so server admins can choose whether to a) import the default countries in English and then rely on the localisation file to translate these countries into the languages that their users are using or b) import/create their own countries either in English or already in their language, which would then however make the localisation file unusable. In any of these two cases, relying on the country name in this issue would not be a problem because server admins could ensure that it matches.

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/AbstractForm.java
```diff
@@ -400,6 +401,11 @@ protected boolean isConfiguredServer(String countryCode) {
 		return FacadeProvider.getConfigFacade().isConfiguredCountry(countryCode);
 	}
 
+	protected CountryReferenceDto getServerCountry() {
```
alexcaruntu-vita, line 404: @MateStrysewske which would be the best class for this method, since we have a similar sequence of code in LocationsField?

sormas-backend/src/main/java/de/symeda/sormas/backend/region/GeoShapeProviderEjb.java
```diff
@@ -167,18 +166,19 @@ public GeoLatLon getCenterOfDistrict(DistrictReferenceDto district) {
...
-			logger.warn("Shape files couldn't be loaded, because no country name is defined in sormas.properties.");
-		} else {
-			loadRegionData(countryName);
-			loadDistrictData(countryName);
-		}
-		buildCountryShape();
-	}
+	/*
+	 * @PostConstruct
+	 * private void loadData() {
```
alexcaruntu-vita, line 171: @MateStrysewske I do not think that this code is used anymore, I commented because it breaks if we configure a country name. Do you remember its original intent or can I delete it?

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
```diff
@@ -372,10 +374,21 @@ protected void addFields() {
...
 			.setVisibleWhen(getFieldGroup(), EventDto.TYPE_OF_PLACE_TEXT, EventDto.TYPE_OF_PLACE, Collections.singletonList(TypeOfPlace.OTHER), true);
 		setTypeOfPlaceTextRequirement();
-		locationForm.setFieldsRequirement(true, LocationDto.REGION, LocationDto.DISTRICT);
 		locationForm.setFacilityFieldsVisible(getField(EventDto.TYPE_OF_PLACE).getValue() == TypeOfPlace.FACILITY, true);
 		typeOfPlace.addValueChangeListener(e -> locationForm.setFacilityFieldsVisible(e.getProperty().getValue() == TypeOfPlace.FACILITY, true));
 
+		countryField.addValueChangeListener(e -> {
+			CountryReferenceDto serverCountryDto = FacadeProvider.getCountryFacade().getServerCountry();
+			CountryReferenceDto countryDto = (CountryReferenceDto) countryField.getValue();
+			if (countryDto != null && serverCountryDto != null && serverCountryDto.getIsoCode().equalsIgnoreCase(countryDto.getIsoCode())) {
```
MateStrysewske, line 383: ISO code part here needs to be changed as well; also, see above (concerning requirements when country == null)

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
```diff
@@ -372,10 +374,21 @@ protected void addFields() {
...
+				locationForm.setFieldsRequirement(true, LocationDto.REGION, LocationDto.DISTRICT);
+			} else {
+				locationForm.setFieldsRequirement(false, LocationDto.REGION, LocationDto.DISTRICT);
+				regionField.setValue(null);
+				districtField.setValue(null);
```
MateStrysewske, line 388: Community and facility need to be set to null as well

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventParticipantEditForm.java
```diff
@@ -103,6 +104,13 @@ protected void addFields() {
 
 		region.addItems(FacadeProvider.getRegionFacade().getAllActiveAsReference());
 
+		CountryReferenceDto countryDto = event.getEventLocation().getCountry();
+		CountryReferenceDto serverCountryDto = FacadeProvider.getCountryFacade().getServerCountry();
+		boolean shouldBeRequired =
+			!(countryDto != null && serverCountryDto != null && countryDto.getIsoCode().equalsIgnoreCase(serverCountryDto.getIsoCode()));
```
MateStrysewske, line 110: See above

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -202,9 +205,26 @@ protected void addFields() {
 		initializeAccessAndAllowedAccesses();
 
 		if (!isEditableAllowed(LocationDto.COMMUNITY)) {
-			setEnabled(false, LocationDto.REGION, LocationDto.DISTRICT);
+			setEnabled(false, LocationDto.COUNTRY, LocationDto.REGION, LocationDto.DISTRICT);
 		}
 
+		if (FacadeProvider.getCountryFacade().getServerCountry() == null) {
```
MateStrysewske, line 211: Remove this logic; if no server country is specified, the location forms should behave like they do right now unless a country is selected (in which case everything is disabled of course)

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -202,9 +205,26 @@ protected void addFields() {
...
+			setEnabled(false, LocationDto.REGION, LocationDto.DISTRICT, LocationDto.COMMUNITY);
+		} ;
+
+		country.addValueChangeListener(e -> {
+			CountryReferenceDto serverCountryDto = FacadeProvider.getCountryFacade().getServerCountry();
+			CountryReferenceDto countryDto = (CountryReferenceDto) e.getProperty().getValue();
+			if (countryDto != null && serverCountryDto != null && serverCountryDto.getIsoCode().equalsIgnoreCase(countryDto.getIsoCode())) {
```
MateStrysewske, line 218: See above

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -202,9 +205,26 @@ protected void addFields() {
...
+				setEnabled(true, LocationDto.REGION, LocationDto.DISTRICT, LocationDto.COMMUNITY);
+			} else {
+				setEnabled(false, LocationDto.REGION, LocationDto.DISTRICT, LocationDto.COMMUNITY);
+				region.setValue(null);
+				district.setValue(null);
+				community.setValue(null);
```
MateStrysewske, line 224: Facility needs to be set to null as well

sormas-ui/src/main/java/de/symeda/sormas/ui/person/LocationsField.java
```diff
@@ -40,6 +43,18 @@ protected void editEntry(LocationDto entry, boolean create, Consumer<LocationDto
 			entry.setUuid(DataHelper.createUuid());
 		}
 
+		if (create) {
+			List<CountryReferenceDto> countryItems = FacadeProvider.getCountryFacade().getAllActiveAsReference();
+			CountryReferenceDto serverCountryDto = FacadeProvider.getCountryFacade().getServerCountry();
+			CountryReferenceDto defaultCountry = countryItems.stream()
+				.filter(
+					countryReferenceDto -> serverCountryDto != null
+						&& countryReferenceDto.getIsoCode().equalsIgnoreCase(serverCountryDto.getIsoCode()))
```
MateStrysewske, line 52: Replace ISO code with name

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
alexcaruntu-vita: ISO part does not need to be changed. _getServerCountry_ returns a _CountryReferenceDto_ based on the country name read from config. Also, from the requirements, in the case of events `region and district are compulsory fields when the server country is selected`.

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
alexcaruntu-vita: Do we also need to disable facility?

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventDataForm.java
MateStrysewske: Oh, true, sorry! I misread that. Concerning the second point, that's indeed true, but they also need to be compulsory when no country is selected

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
MateStrysewske: Yes, basically all infrastructure fields need to be disabled when a foreign country is selected

sormas-ui/src/main/java/de/symeda/sormas/ui/person/LocationsField.java
alexcaruntu-vita: See first part of https://github.com/hzi-braunschweig/SORMAS-Project/pull/4336#discussion_r576855183

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventParticipantEditForm.java
```diff
@@ -103,6 +104,13 @@ protected void addFields() {
 
 		region.addItems(FacadeProvider.getRegionFacade().getAllActiveAsReference());
 
+		CountryReferenceDto countryDto = event.getEventLocation().getCountry();
```
MateStrysewske, line 107: I think the proper logic here is to check whether region and district have been filled in for the event - that removes the need to check for the server country etc.

MateStrysewske (changes requested):
Just realised that the code that I proposed doesn't make sense. It leads to the fields never getting disabled when no country is specified in the properties. What should actually happen in that case is that if ANY country is selected, the fields are disabled. So:

If NO country is specified in the properties:
* Make the fields editable and (for events) required when NO country is selected
* Make the fields disabled and non-required when ANY country is selected

If a country IS specified in the properties:
* Make the fields editable and (for events) required when NO country OR the country specified in the properties is selected
* Make the fields disabled and non-required when ANY OTHER country is selected

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventParticipantEditForm.java
alexcaruntu-vita: My understanding, from the story, is that those fields should be required based on the server country, so what should I do? :)

MateStrysewske: They need to be required when no region and district have been filled in for the event. While this is the case when a different country has been selected for the event, checking whether the region/district have been filled in is a) easier and b) less prone to errors than checking for the country again here.

MateStrysewske (changes requested):
* When creating a new event in the web app with a country specified in the properties, the country is not selected by default
* When a country is specified in the properties, region and district are not initially required in the web app when no country is selected (only when selecting a country and then removing it again)

MateStrysewske (approved):

Issue #4307, opened 2021-02-09
Add indexes to sormas database [0.5]


BarnaBartha:
### Situation Description
The database lacks indexes and they are only added once we see a problem.
Indexes should be added everywhere where there is potential improvement by adding them.
### Feature Description
As a rule of thumb try to keep the indexes to a minimum, but:
- add index on columns that are being used in where clauses
- add index on columns that you use to join on.

DavidBaldsiefen:
Adding too much indexes may slow down the database in other places though, especially whenever lists are modified (insert). I wouldn't suggest adding more than 1 index per table whenever it's not necessary

MartinWahnschaffe:
@BarnaBartha Statements we found in the slow query logs from Ghana that my be related to this.
Please make sure that the indices that would speed those statements up are included.

1.  CaseFacade.count for a lab officer (associated via related sample.lab)

> select count(case0_.id) as col_0_0_ from cases case0_ left outer join users user1_ on case0_.reportingUser_id=user1_.id left outer join users user2_ on case0_.surveillanceOfficer_id=user2_.id left outer join users user3_ on case0_.caseOfficer_id=user3_.id left outer join Person person4_ on case0_.person_id=person4_.id left outer join users user5_ on case0_.reportingUser_id=user5_.id left outer join Region region6_ on case0_.region_id=region6_.id left outer join District district7_ on case0_.district_id=district7_.id left outer join Community community8_ on case0_.community_id=community8_.id left outer join Facility facility9_ on case0_.healthFacility_id=facility9_.id where (case0_.id in (select sample10_.associatedCase_id from samples sample10_ where sample10_.lab_id=$1) or case0_.id in (select contact11_.caze_id from Contact contact11_ left outer join users user12_ on contact11_.reportingUser_id=user12_.id left outer join users user13_ on contact11_.contactOfficer_id=user13_.id where user12_.id=$2 or user13...

2. Pathogen tests of a sample. Very strange that this is slow - maybe the system was slowed down by a running process. Still likely that adding an index will speed this up:

[...]

PR #4333, opened 2021-02-10
#3481: Added missing indexes for queries on tables related to saving …

ssferrazza:
Added missing indexes for queries on tables related to saving and reading a case.

PR #4340, opened 2021-02-11
#4307 - add indexes

BarnaBartha:


sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6467,4 +6467,44 @@ ALTER TABLE cases_history ADD COLUMN caseidentificationsource character varying(
 
 INSERT INTO schema_version (version_number, comment) VALUES (324, 'Case identification source #3420');
 
+-- 2020-02-09 Add indexes #4307
+CREATE INDEX IF NOT EXISTS idx_cases_epid_number ON cases USING gist (epidnumber gist_trgm_ops);
+CREATE INDEX IF NOT EXISTS idx_cases_person_id ON cases (person_id);
+CREATE INDEX IF NOT EXISTS idx_cases_region_id ON cases (region_id);
+CREATE INDEX IF NOT EXISTS idx_cases_district_id ON cases (district_id);
+
+CREATE INDEX IF NOT EXISTS idx_contact_person_id ON contact (person_id);
```
DavidBaldsiefen, line 6476: Remove this line, index already exists

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6467,4 +6467,44 @@ ALTER TABLE cases_history ADD COLUMN caseidentificationsource character varying(
...
+
+CREATE INDEX IF NOT EXISTS idx_samples_contact_id ON samples (associatedcontact_id);
+CREATE INDEX IF NOT EXISTS idx_samples_case_id ON samples (associatedcase_id);
+CREATE INDEX IF NOT EXISTS idx_samples_eventparticipant_id ON samples (associatedeventparticipant_id);
+
+CREATE INDEX IF NOT EXISTS idx_task_contact_id ON task (contact_id);
+CREATE INDEX IF NOT EXISTS idx_task_case_id ON task (caze_id);
+CREATE INDEX IF NOT EXISTS idx_task_event_id ON task (event_id);
+
+CREATE INDEX IF NOT EXISTS idx_visit_case_id ON visit (caze_id);
```
DavidBaldsiefen, line 6489: Remove, already exists

DavidBaldsiefen (changes requested):
Looks good to me. We will have to see how this performs on life systems. Also, these changes might lead to quite long transition times when servers with large databases are updated, as all these indexes need to be created first.

The new indexes in the contact and case directory could actually _slow down_ the case/contact import. So I am not sure if it is wise to implement them now. Maybe @MartinWahnschaffe can take a quick look?

MartinWahnschaffe:
@DavidBaldsiefen The indices Barna added are all for relations that are at the heart of the system and are relevant for almost all queries done. I'm pretty sure this will outweigh the cost of maintaining the index.

@BarnaBartha I think we should also consider the references within the event participant. Also what do you think about adding indices to the reporting user of at least case and contact? The createUserFilter logic is executed for almost all queries and relies on this.

BarnaBartha:
@MartinWahnschaffe added the indexes to eventParticipants>
The reporting user indices have already been added by Salva
https://github.com/hzi-braunschweig/SORMAS-Project/pull/4333/files

MartinWahnschaffe:
@BarnaBartha Great. Please go ahead and merge once the checks have passed. Please provide the SQL for all indices with a short message to @StefanKock for the release notes page.

PR #4342, opened 2021-02-11
4207 prohibition to work until

cazacmarin:
I set the `prohibition to work until` field to allow up to 90 days in the future.


MateStrysewske:
@cazacmarin Please don't add a maximum date for this field at all. Prohibition to work could theoretically be years into the future.

cazacmarin:
@MateStrysewske - "PROHIBITION TO WORK UNTIL"  field, is present as well for "Contacts" in Mobile application. Should  those changes(applied to cases) to be applied to this field as well?

MateStrysewske:
@cazacmarin Yes please! Thanks for pointing that out!

Issue #4187, opened 2021-01-30
[SurvNet Interface] Add Reason for hospitalization [1]


kwa20:
### Feature Description
cases > hospitalization 
### Problem Description
Thus far it is not documented whether a case was hospitalized due to the reported disease or because of another reason. This however is needed for transmitting data to SurvNet. (cf. #3312)
### Proposed Change
Add a "reason for hospitalization" drop-down field to the hospitalization tab. Also include this in the previous hospitalization creation window and additionally add information about the intensive care unit to this section.

drop-down should include the variables:
- (null)
- reported disease
- other reason
- unknown

selecting other would enable another field to specify the other reason.

In the process, move the date of isolation field to the left to be more in line with the other fields on the form.
### Additional Information
<img width="659" alt="Screen Shot 2021-01-30 at 7 23 03 PM" src="https://user-images.githubusercontent.com/70317594/106368850-32bfe300-634d-11eb-97be-f4e149594ce3.png">
<img width="433" alt="Screen Shot 2021-01-30 at 7 19 31 PM" src="https://user-images.githubusercontent.com/70317594/106368851-381d2d80-634d-11eb-83e3-04a75b113fd8.png">

 SurvNet mapping:

- null  |  nicht erhoben
- reported disease  |  Hospitalisiert aufgrund der gemeldeten Krankheit
- other reason  |  Hospitalisiert aufgrund einer anderen Ursache als der gemeldeten Krankheit
- unknown  |  Hospitalisiert Ursache ist unbekannt, nicht ermittelbar

fill "Hospitalisiert zur Isolation" in SurvNet when isolation is set to yes in Sormas and the resaon of hospitalization is the reported disease.

MateStrysewske:
* Add the new fields to both Hospitalization as well as PreviousHospitalization
* Extend the code that turns hospitalizations into previous hospitalizations when transferring cases to also include these new fields

kwa20:
@AndyBakcsy-she I could also verify the reason for hospitalization field on test.sormas.netzlink.com but couldn't verify the "stay in ICU" fields in the previous hospitalizations yet.

PR #4344, opened 2021-02-11
#4187 add reason for hospitalization

ssferrazza:
See #4187


sormas-api/src/main/java/de/symeda/sormas/api/hospitalization/HospitalizationDto.java
```diff
@@ -60,6 +62,8 @@
 	private YesNoUnknown intensiveCareUnit;
 	private Date intensiveCareUnitStart;
 	private Date intensiveCareUnitEnd;
+	private HospitalizationReasonType reasonForHospitalization;
```
MateStrysewske, line 65: I'd prefer to call the field "hospitalizationReason" - more consistent with the enum name and also shorter. The caption can stay the way it is of course.

sormas-api/src/main/java/de/symeda/sormas/api/hospitalization/HospitalizationDto.java
```diff
@@ -60,6 +62,8 @@
...
+	private String otherReasonForHospitalization;
```
MateStrysewske, line 66: Hence, this would be "otherHospitalizationReason"

sormas-backend/src/main/java/de/symeda/sormas/backend/hospitalization/Hospitalization.java
```diff
@@ -178,4 +181,21 @@ public Date getIntensiveCareUnitEnd() {
...
+	@Enumerated(EnumType.STRING)
+	public HospitalizationReasonType getReasonForHospitalization() {
+		return reasonForHospitalization;
+	}
+
+	public void setReasonForHospitalization(HospitalizationReasonType reasonForHospitalization) {
+		this.reasonForHospitalization = reasonForHospitalization;
+	}
+
+	public String getOtherReasonForHospitalization() {
```
MateStrysewske, line 194: I don't actually know whether annotations for String fields are redundant. In other places, we still use `@Column(columnDefinition = "text")` - have you left it out on purpose because you know it's redundant?

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6496,4 +6496,12 @@ ALTER TABLE cases_history
 
 INSERT INTO schema_version (version_number, comment) VALUES (326, 'SurvNet Adaptations - Create new field “Reinfection” for cases #3831');
 
+-- 2021-02-05 Add reason hospitalization #4187
+ALTER TABLE hospitalization ADD COLUMN reasonforhospitalization varchar(255);
+ALTER TABLE hospitalization ADD COLUMN otherreasonforhospitalization varchar(255);
```
MateStrysewske, line 6501: This needs to be of type "text" because it's a free text field - we use varchar(255) for enums, have used varchar(512) for free text fields for a long time, but have decided that "text" is a way better alternative

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6496,4 +6496,12 @@ ALTER TABLE cases_history
...
+ALTER TABLE previoushospitalization ADD COLUMN reasonforhospitalization varchar(255);
+ALTER TABLE previoushospitalization ADD COLUMN otherreasonforhospitalization varchar(255);
```
MateStrysewske, line 6503: Same here of course

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6496,4 +6496,12 @@ ALTER TABLE cases_history
...
+
```
MateStrysewske, line 6504: Also add the new fields to the respective history tables, please

sormas-ui/src/main/java/de/symeda/sormas/ui/hospitalization/HospitalizationForm.java
```diff
@@ -129,6 +138,11 @@ protected void addFields() {
 		final Field isolationDateField = addField(HospitalizationDto.ISOLATION_DATE);
 		final NullableOptionGroup isolatedField = addField(HospitalizationDto.ISOLATED, NullableOptionGroup.class);
 		final NullableOptionGroup leftAgainstAdviceField = addField(HospitalizationDto.LEFT_AGAINST_ADVICE, NullableOptionGroup.class);
+
+		ComboBox reasonForHospitalization = addField(HospitalizationDto.REASON_FOR_HOSPITALIZATION);
+		otherReasonForHospitalization = addField(HospitalizationDto.OTHER_REASON_FOR_HOSPITALIZATION, TextField.class);
+		otherReasonForHospitalization.setEnabled(false);
```
MateStrysewske, line 144: Not necessary (see below)

sormas-ui/src/main/java/de/symeda/sormas/ui/hospitalization/HospitalizationForm.java
```diff
@@ -225,6 +239,26 @@ protected void addFields() {
 				I18nProperties.getValidationError(Validations.beforeDate, intensiveCareUnitEnd.getCaption(), dischargeDateField.getCaption())));
 		hospitalizedPreviouslyField.addValueChangeListener(e -> updatePrevHospHint(hospitalizedPreviouslyField, previousHospitalizationsField));
 		previousHospitalizationsField.addValueChangeListener(e -> updatePrevHospHint(hospitalizedPreviouslyField, previousHospitalizationsField));
+		reasonForHospitalization.addValueChangeListener(this::updateOtherReasonForHospitalizationField);
+	}
+
+	private void updateOtherReasonForHospitalizationField(final Property.ValueChangeEvent event) {
```
MateStrysewske, line 245: We have helper methods to replace custom logic like this. Please have a look at `FieldHelper.setVisibleWhen`, you can use this to make sure that "otherHospitalizationReason" is only visible when HospitalizationType.OTHER is selected, and that it gets cleared if another HospitalizationType is selected

sormas-ui/src/main/java/de/symeda/sormas/ui/hospitalization/HospitalizationForm.java
```diff
@@ -57,9 +60,13 @@
 import de.symeda.sormas.ui.utils.NullableOptionGroup;
 import de.symeda.sormas.ui.utils.OutbreakFieldVisibilityChecker;
 import de.symeda.sormas.ui.utils.ViewMode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class HospitalizationForm extends AbstractEditForm<HospitalizationDto> {
 
+	private static final Logger LOG = LoggerFactory.getLogger(HospitalizationForm.class);
```
MateStrysewske, line 68: Remove this declaration (not necessary and the code below will be replaced)

sormas-ui/src/main/java/de/symeda/sormas/ui/hospitalization/PreviousHospitalizationEditForm.java
```diff
@@ -41,14 +44,21 @@
 import de.symeda.sormas.ui.utils.DateComparisonValidator;
 import de.symeda.sormas.ui.utils.FieldHelper;
 import de.symeda.sormas.ui.utils.NullableOptionGroup;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class PreviousHospitalizationEditForm extends AbstractEditForm<PreviousHospitalizationDto> {
```
MateStrysewske, line 50: See comments in HospitalizationForm.java

MateStrysewske (changes requested):
The mobile app part is missing. New fields are usually added to both the web and the mobile app. This is also indicated by the labels in the issue: "vaadin-app" and "android-app" have both been added there. :)

sormas-backend/src/main/java/de/symeda/sormas/backend/hospitalization/Hospitalization.java
ssferrazza: Yes, exactly. If the column name is equal to the property name there is no need to add that annotation. But I can add this to be more consistent with the other entities.

MateStrysewske: No, if String maps to "text" automatically, that's perfectly fine. Minimalism is good. :D

sormas-api/src/main/resources/captions.properties
```diff
@@ -493,6 +493,8 @@ CaseHospitalization.previousHospitalizations=Previous hospitalizations
 CaseHospitalization.intensiveCareUnit=Stay in the intensive care unit
 CaseHospitalization.intensiveCareUnitStart=Start of the stay
 CaseHospitalization.intensiveCareUnitEnd=End of the stay
+CaseHospitalization.hospitalizationReason=Reason for hospitalization
+CaseHospitalization.otherHospitalizationReason=Specify other reason for hospitalization
```
MateStrysewske, line 497: Change to "Specify reason", otherwise the caption will get too long (especially in languages like German)

sormas-api/src/main/resources/captions.properties
```diff
@@ -510,6 +512,8 @@ CasePreviousHospitalization.healthFacility=Hospital
 CasePreviousHospitalization.healthFacilityDetails=Hospital name & description
 CasePreviousHospitalization.isolated=Isolation
 CasePreviousHospitalization.prevHospPeriod=Period of hospitalization
+CasePreviousHospitalization.hospitalizationReason=Reason for hospitalization
+CasePreviousHospitalization.otherHospitalizationReason=Specify other reason for hospitalization
```
MateStrysewske, line 516: Change to "Specify reason", otherwise the caption will get too long (especially in languages like German)

MateStrysewske (changes requested):

MateStrysewske (changes requested):
* Move the reason fields below the facilityDetails field in the previous hospitalization dialog
* Reason for hospitalization is not taken over when automatically creating a previous hospitalization. When a user edits the facility of an existing field, the system asks them whether they just did a data correction or whether the case was actually transfered to another facility. If the second option is chosen, a previous hospitalization with the hospitalization details of the current hospitalization is automatically created and the current hospitalization is updated. This logic needs to be extended to include the hospitalization reason
* In the hospitalization form, disable the hospitalization reason fields when all other fields in the hospitalization form are disabled as well (I think this happens when the case is not actually in a hospital, e.g. because place of stay has been changed to "Home"

MateStrysewske (changes requested):
The automatic previous hospitalization creation also has to be adjusted in the mobile app (see PreviousHospitalizationDao there)

MateStrysewske (approved):

PR #4345, opened 2021-02-11
Feature 3716 make user roles deactivateable

sergiupacurariu:


sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserUiHelper.java
```diff
@@ -24,13 +25,20 @@
...
-	public static Set<UserRole> getAssignableRoles() {
+	public static Set<UserRole> getAssignableRoles(Set<UserRole> assignedUserRoles) {
+
+		final Set<UserRole> assignedRoles = assignedUserRoles == null ? Collections.emptySet() : assignedUserRoles;
+
 		Set<UserRole> allRoles = UserRole.getAssignableRoles(UserProvider.getCurrent().getUserRoles());
 
 		if (!FacadeProvider.getConfigFacade().isConfiguredCountry(CountryHelper.COUNTRY_CODE_SWITZERLAND)) {
 			allRoles.remove(UserRole.BAG_USER);
 		}
```
DavidBaldsiefen, line 36: Disable the `UserRole.BAG_USER` instead

sergiupacurariu: @DavidBaldsiefen That means on all servers except Swiss ones an admin should run the SQL script otherwise they will start seeing this role. Are we good with this?

DavidBaldsiefen: Hm that doesn't sound like a nice solution either. I am just concerned that a setting is applied here that should be made by some kind of config file in my opinion. But maybe that should be another Issue then

MartinWahnschaffe: I agree. This should be handled in a separate issue later.

PR #4349, opened 2021-02-11
Merge branch 'development' into 4219_not_done_sample_final_result

### LINKED ISSUE #4219: New final laboratory result (Samples)
### Situation Description
In addition to the currently available final laboratory result for a sample (Indeterminate, Pending, Negative, Positive), a new option/selection; **NOT DONE**.

**NOT DONE** would be the representation of a sample that a pathogen test was never carried out on due to one reason or the other and not just the specimen condition.

### Feature Description

- Users would still be able to enter pathogen test result details as is implemented at the moment.
- The **NOT DONE** laboratory result would **ONLY** be available for selection when the pathogen test result is on **PENDING** or there is none recorded (NA), any other pathogen test result would work as is implemented currently.
- Ie users would not be able to set final lab result of sample  to **NOT DONE** when there is at lease one pathogen test for the sample recorded with a result  != pending.

vidi42:
Closes #4219


sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleEditForm.java
```diff
@@ -64,16 +70,35 @@ protected void addFields() {
...
 		});
 	}
 
-	public void makePathogenTestResultRequired() {
+	public void fillPathogenTestResult() {
 		ComboBox pathogenTestResultField = (ComboBox) getFieldGroup().getField(SampleDto.PATHOGEN_TEST_RESULT);
+
+		boolean hasPathogenTests = FacadeProvider.getPathogenTestFacade().hasPathogenTest(getValue().toReference());
+
+		if (!hasPathogenTests) {
```
MateStrysewske, line 82: Remove the two lines that update the enum data and set the pathogen test result value in this if block. "Not done" is meant to point out that there was no pathogen test done for the sample, and that there won't be one in the future as well (i.e. the sample has been dropped without testing it). "Pending" is still meant to be the default when a sample is created and no pathogen test exists yet.

MateStrysewske (changes requested):
The new logic also needs to be implemented in the mobile app

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleEditForm.java
vidi42: In that case the user won't be able to change the status to NOT_DONE if there are no pathogen tests done.
From my understanding of this line, the user should be able to select NOT_DONE if there are no pathogen recorded.
> The NOT DONE laboratory result would ONLY be available for selection when the pathogen test result is on PENDING or there is none recorded (NA), any other pathogen test result would work as is implemented currently.

I could remove the line which sets the default value to NOT_DONE if no pathogen test is recorded, but I would suggest to keep then the approach which updates the enum data set to let the user choose between PENDING and NOT_DONE when no pathogen test was recorded.

MateStrysewske: Maybe I misunderstood the code then. :) The way you described it makes sense to me! So "pending" should still be the default, but adding "not done" to the list is of course necessary, and also to make it editable even if no pathogen test has been recorded yet.

sormas-app/app/src/main/java/de/symeda/sormas/app/sample/edit/SampleEditFragment.java
```diff
@@ -123,6 +128,10 @@ private void setUpFieldVisibilities(final FragmentSampleEditLayoutBinding conten
 		} else {
 			contentBinding.mostRecentAdditionalTestsLayout.setVisibility(GONE);
 		}
+
+		if(record.getId() == null) {
```
MateStrysewske, line 132: Code formatting

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleEditForm.java
```diff
@@ -29,6 +29,12 @@
 import de.symeda.sormas.api.sample.PathogenTestResultType;
 import de.symeda.sormas.api.sample.SampleDto;
 import de.symeda.sormas.api.utils.fieldaccess.UiFieldAccessCheckers;
+import de.symeda.sormas.ui.utils.FieldHelper;
+
+import java.util.Arrays;
```
MateStrysewske, line 34: I think your import settings are wrong; java imports should go above the symeda imports. Have you enabled "Optimize imports on the fly"? If so, that might be the case, it caused IntelliJ to not properly regard our import settings in my case.

MateStrysewske (changes requested):
It's still necessary to allow users to select all possible laboratory results even if no pathogen tests have been documented, purely because in an outbreak situation, chances are that there's not resources to document the pathogen tests, but the final laboratory result is still extremely important.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleEditForm.java
vidi42: Seems my IntelliJ just removes this plugin with each major update 🙈

MateStrysewske (changes requested):
* "Not done" also needs to be excluded from the pathogen test result field when creating a pathogen test directly while creating a sample (by selecting the checkbox at the bottom of the form)

MateStrysewske (approved):

Issue #3419, opened 2020-11-06
SurvNet Adaptations - Create new field “Suspicious case” to cases [2]


markusmann-vg:
### Feature Description
Create new field “Suspicious case” to cases

Case > case information

New field (DE only) –
“Verdachtsfall (de – de) | “Suspicious case” (en) [SurvNet field names: SuspicionNotConf; SuspicionNotDat; SuspicionNotPhysician; SuspicionNotPathOther]

If the answer is “suspicion not confirmed” then change case classification to “no case”.

Data Type: check box, date
Values (de-de) | (en)

Verdacht nicht bestätigt am | suspicous not confirmed
o Information durch Arzt/Ärztin| information provided by physician
o Nachweis eines anderen Erregers| detection of other pathogen
Data Type: Datum |date

- [ ] Add to export

GabrielePoggensee:
IfSG relevant: § 11 (1) 1 c: Indication when a suspicion has not been confirmed

Data are transmitted to RKI

kwa20:
I would actually like to refine this issue so it makes more sense in Sormas.

Only show this checkbox if 'suspect case' is selected as the case classification. Only when 'suspicion not confirmed' is checked, show the date field and the two further check boxes. Also change the classification to 'no case' as already mentioned above.

<img width="807" alt="Screen Shot 2020-12-15 at 10 02 38 PM" src="https://user-images.githubusercontent.com/70317594/102273936-65d81f80-3f23-11eb-8ef4-7c9ae8dff698.png">

MateStrysewske:
Why do we need this checkbox in the first place? Shouldn't setting the case classification to "Not a case" be enough? This will also update the date of classification, so we'd basically only need the two additional checkboxes for German systems.

kwa20:
@MateStrysewske I'd like that solution because it's minimalistic. We could also then only show the two additional checkboxes when 'no case' is selected and use 'suspicion not confirmed' as a heading for the two.

bernardsilenou:
@kwa20 I think we need to first investigate what is the epidemiological definition of suspicious case in term of RKI. In some countries, suspicious case == suspected case.
I am more in line of thinking as @Maté, this may be a category of case classification and it is a very sensitive topic.
Please lets have a call and followup on it

MateStrysewske:
@bernardsilenou I didn't write anything like that. I don't think we need a new classification for this.

kwa20:
[...]

PR #4356, opened 2021-02-11
#3419 Create new field “Suspicious case” to cases - fixing missing co…

cazacmarin:
Create new field “Suspicious case” to cases - fixing missing column on mobile app after merge.


cazacmarin:
Closes #3419

alexcaruntu-vita (approved):

Issue #4290, opened 2021-02-08
Add a new community-level userrole with web access


DavidBaldsiefen:
### Situation Description

A lot of german users need a community-level user with access to the webapp.

### Feature Description

For now, add a new user role. I will complete this issue with the required userrights asap.

Required userrights:

- listing cases & contacts in their community
- creating and sending quarantine documents & subsequently change the related quarantine order date
- Editing the case-person and contact-person

Creating new Cases and Contacts does not seem to be a requirement.

### Possible Alternatives

Waiting for the new user rights model is not an option, due to this issues urgency.

StefanKock:
Duplicate to #1960

DavidBaldsiefen:
Reopened due to new requirements not covered with #1960

DavidBaldsiefen:
I've discussed this further with the user, and we decided that it would be best (for now) if the community officer has more rights than the community informant. This way we won't have to create a new userrole every time somebody requires a community-level role to have some additional right

bernardsilenou:
Suggestions:
- we can call this role "community surveillance officer" 
- In order not to create another community contact officer role, we can  give this role the "community contact officer" rights.
-  they can have the combination of surveillance officer and contact officer rights but limited to the community of the user
- Since the community is not a required field when creating entities,  the community user would not have access to the case with out a community if it was reported by another user in the same community. Thus community users should always add the community to entities they need to have access to.
- They should not have the right to assign or edit epid number
- They should have full access to all entities with NA communities but are linked to facilities (or home addresses)  in the community of the user

DavidBaldsiefen:
>     * we can call this role "community surveillance officer"

I would suggest just "Community officer", especially if he combines the rights of community surveillance & contact officer

>     * In order not to create another community contact officer role, we can  give this role the "community contact officer" rights.
> 
>     * they can have the combination of surveillance officer and contact officer rights but limited to the community of the user

Agreed

>     * Since the community is not a required field when creating entities,  the community user would not have access to the case with out a community if it was reported by another user in the same community. Thus community users should always add the community to entities they need to have access to.

Should we force them to fill that field? There might be situations where a user from community A needs to add a case/contact which does not belong to his communities responsibility (e.g. an event in my community, where participants from other communities took part. I should still be able to add all those participants to sormas, even though I won't be able to further process their cases or contacts)

>     * They should not have the right to assign or edit epid number

I don't know exactly what the epid number does but that can be implemented.

[...]

PR #4364, opened 2021-02-12
#1960 Give Community and Hospital Informant access to the webapp

### LINKED ISSUE #1960: Web access for community and hospital informant [2]
### Feature Description

Currently, hospital informant role has access only using mobile application. Adding web access for hospital informant would have benefits.

### Problem Description

MDM and MFA is needed for the informant role if spreading the use of the system to a wide range of private enterprises providing elderly care. This adds complexity and costs of the system. The required level of security could be achieved using web interface (+ 3rd party MFA).

### Proposed Change

Hospital informant role could access the system also using web interface. Access would be needed only to features add new case and edit own case?

DavidBaldsiefen:
Closes #1960


MateStrysewske (changes requested):
Please have a look at Martin's comment here:
https://github.com/hzi-braunschweig/SORMAS-Project/issues/1960#issuecomment-698158081

There's quite some additional stuff that needs to be done for this to work properly.

DavidBaldsiefen:
> Please have a look at Martin's comment here:
> https://github.com/hzi-braunschweig/SORMAS-Project/issues/1960#issuecomment-698158081
> 
> There's quite some additional stuff that needs to be done for this to work properly.


Filter visibility & automatically setting the community did work just fine. I was not sure about the dashboard though, as from my understanding an _Informant_ wouldn't need that. I will look into it though

DavidBaldsiefen:
I've applied the requested changes regarding dashboard rights. Sorry for missing those!

> Adjust filter visibility and default values for these user roles; community needs to be automatically filled in for community informants, health facility needs to be automatically filled in for hospital informants; all automatically filled fields should be hidden for these users

I didn't do anything about this though. When creating cases, the initial fields are filled correctly. However, both userrolles will still have to select the "lowest level". For example, the community informant will still have to select the community (and is also able to select a different one). Same for the facility selection for hospital informant.

I think this makes sense, and personally wouldn't do anything about it. Also I don't see any danger regarding access rights; If a community informant creates a case for _another_ community, it won't show up in his case view again after leaving the edit.

MateStrysewske (changes requested):
I partly agree, partly not. In the mobile app - which is the main access point to SORMAS for these user roles - they are not allowed to change the jurisdiction of new cases they create because they are asked to strictly work in their assigned jurisdiction. I'm fine with allowing community informants to alter data for their own level of jurisdiction in the web app, but definitely nothing above that, so they would be allowed to change the community, but not the region and district. For hospital informants, it doesn't make sense whatsoever to allow them to alter any of these fields, because they're a user role that is specifically designed for people who are working in a hospital. They will never create cases in other hospitals, let alone in other communities, districts or regions.

Regarding default selections, those should always go down to the user's jurisdiction level, so for community officers, their community would already be pre-selected (which is the case according to my tests), and for informants, their facility also needs to be pre-selected (which would automatically be the case due to the change requested above).

DavidBaldsiefen:
> I partly agree, partly not. In the mobile app - which is the main access point to SORMAS for these user roles - they are not allowed to change the jurisdiction of new cases they create because they are asked to strictly work in their assigned jurisdiction. I'm fine with allowing community informants to alter data for their own level of jurisdiction in the web app, but definitely nothing above that, so they would be allowed to change the community, but not the region and district. For hospital informants, it doesn't make sense whatsoever to allow them to alter any of these fields, because they're a user role that is specifically designed for people who are working in a hospital. They will never create cases in other hospitals, let alone in other communities, districts or regions.
> 
> Regarding default selections, those should always go down to the user's jurisdiction level, so for community officers, their community would already be pre-selected (which is the case according to my tests), and for informants, their facility also needs to be pre-selected (which would automatically be the case due to the change requested above).

Still, I am not completely sure about the hospital informants jurisdiction. By default, hospital informants only get assigned a Region & District, as well as a facility which in turn is assigned to a community as well. But doesn't that mean I could potentially still treat patients in my hospital, which belong to a different community? E.g. after being transferred from another hospital. I am fine with pre-selecting and enforcing the facility. But I am not so sure about the community field to be honest.

MateStrysewske:
> Still, I am not completely sure about the hospital informants jurisdiction. By default, hospital informants only get assigned a Region & District, as well as a facility which in turn is assigned to a community as well. But doesn't that mean I could potentially still treat patients in my hospital, which belong to a different community? E.g. after being transferred from another hospital. I am fine with pre-selecting and enforcing the facility. But I am not so sure about the community field to be honest.

That's not how the jurisdiction fields work though. If a hospital informant is assigned a health facility inside a specific community, that health facility will not show up in the dropdown if they select a different community. So if I'm pre-selecting and enforcing the user's facility, I also have to enforce its community (or no community at all if the facility is not assigned to a community because that's also possible).

The jurisdiction fields inside a case are a little hard to grasp in that regard because they're really about "where is the patient treated?", and in the case of a hospital informant, that's ALWAYS their hospital, without any exception.

DavidBaldsiefen:
> That's not how the jurisdiction fields work though. If a hospital informant is assigned a health facility inside a specific community, that health facility will not show up in the dropdown if they select a different community. So if I'm pre-selecting and enforcing the user's facility, I also have to enforce its community (or no community at all if the facility is not assigned to a community because that's also possible).
> 
> The jurisdiction fields inside a case are a little hard to grasp in that regard because they're really about "where is the patient treated?", and in the case of a hospital informant, that's ALWAYS their hospital, without any exception.

Now I understand (I think). So to wrap it up I will do the following:

If UserJurisdictionLevel = Community
- [ ] enforce Region and District
- [ ] Make users Community default selection
- [ ] Enforce Region, District, *and* Community in android app

If UserJurisdictionLevel = Healthfacility
- [ ] enforce Region, District and Community
- [ ] enforce Facility

And in advance for #4290:
- [ ] Do the same for contact creation

MateStrysewske:
For contacts that's a bit tricky actually. If contacts have a source case, their jurisdiction can't be enforced or pre-filled at all because it's optional. If a contact does NOT have a source case however, yes, in that case both fields should be pre-filled and disabled. That means there needs to be a listener for that as well because users can choose or remove a source case directly in the contact creation form (in the web app at least). I would actually not do that in this issue right here though since neither of these two user roles are allowed to edit contacts, so we can't properly test it. Should rather be done in #4290.

DavidBaldsiefen:
@MateStrysewske I've made the necessary changes. Sadly, that took much, _much_ longer than expected due to some (still not 100% clear) bug I had regarding the `facilityType` field. Thus I was not able to implemented the discussed restrictions for contact creation for #4290 today. However, I would still consider merging #4290 and just solving that problem afterwards - the new CO role should not break any existing workflows, as every user is free to use it "at own risk". I know this is not ideal, but eventually I think in this case the demand is high enough to justify it. Ultimately I'll leave it up to PM to decide on that though.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
```diff
@@ -203,7 +205,11 @@ protected void addFields() {
 		facilityTypeGroup.setWidth(100, Unit.PERCENTAGE);
 		facilityTypeGroup.addItems(FacilityTypeGroup.getAccomodationGroups());
 		getContent().addComponent(facilityTypeGroup, FACILITY_TYPE_GROUP_LOC);
-		facilityType = addField(CaseDataDto.FACILITY_TYPE);
+		facilityType = new ComboBox();
```
MateStrysewske, line 208: Why was this change needed? Is it related to the bug that you were writing about? Because technically this should do exactly what the line that you removed is doing

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
```diff
@@ -260,7 +266,7 @@ protected void addFields() {
 		});
 		facilityOrHome.addValueChangeListener(e -> {
 			FieldHelper.removeItems(facility);
-			if (TypeOfPlace.FACILITY.equals(facilityOrHome.getValue())) {
+			if (TypeOfPlace.FACILITY.equals(facilityOrHome.getValue()) || TypeOfPlace.FACILITY.equals(facilityOrHome.getNullableValue())) {
```
MateStrysewske, line 269: This results in an exception when the field is required and HOME is selected, probably because it checks the second condition but can't retrieve a nullable value because the field isn't nullable in the first place

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
DavidBaldsiefen: Precisely. That's what I thought too and that's also why I spend several hours searching the bug. This line definitely needs to be changed.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
DavidBaldsiefen: Solved... the whole `NullableOptionGroup` had my head hurt anyway. To be honest it does not seem very well implemented, but really that should be handled in another issue as it's used quite a lot.

MateStrysewske (approved):

PR #4370, opened 2021-02-12
#4338 - Fixed locations not being cloned on case merge

### LINKED ISSUE #4338: Merging cases with exposures duplicates their locations and sets them on both exposures
### Bug Description
While updating the productive servers from 1.52.1 to 1.54.1, a few sormas servers failed after the step:
`Updating database to version 309... (de.symeda.sormas.backend.common.StartupShutdownService.java:590)`

[sormas.log.zip](https://github.com/hzi-braunschweig/SORMAS-Project/files/5961447/sormas.log.zip)

### LINKED ISSUE #4388: Make DevMode use existing persons sometimes
### Feature Description

In the DevMode, it should be possible to set a percentage based on which new cases/contacts use existing persons to build upon.

### Problem Description

Right now, for every new case/contact a new person is created. This leads to a very large person database and is also unrealistic.

### Proposed Change

Implement a method similar to `getRandomCaseReferences()` and use it accordingly.

### Possible Alternatives

SORMAS datagen

### Additional Information
One does need to make sure no unrealistic combinations get created this way; i.e. a contact person should never be identical to the source case person of that contact.

MateStrysewske:
Fixes #4338

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/exposure/ExposureDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/exposure/ExposureDto.java
@@ -660,4 +660,11 @@ public class ExposureDto extends PseudonymizableDto {
 	public void setRiskArea(YesNoUnknown riskArea) {
 		this.riskArea = riskArea;
 	}
+
+	@Override
+	public ExposureDto clone() throws CloneNotSupportedException {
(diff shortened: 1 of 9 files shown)


StefanKock (commented):
I see no problem anymore, but I am not sure if the SQL statement to remove location ids is correct (looks correct).

PR #4391, opened 2021-02-15
#3233 add seed and default configurations to devmode

### LINKED ISSUE #3233: Make case and contact data generation deterministic (devmode)
### Feature Description
It's currently possible to generate cases and contacts using the devmode.

### Problem Description
For automatic performance tests it is necessary to have the data generated in exactly the same way each time. This will make sure that differences in the data don't affect the measured times.

### Proposed Change
Add an optional "seed" field that will be used for random number generation that is used when creating cases and contacts.

### LINKED ISSUE #4540: In Configuration -> Developer: implausible case data generated
### Bug Description
For prohibition to work, implausible data is generated by the test data generator which fails in later validations.
From the data base for one of the generated cases:
```
|prohibitiontowork|prohibitiontoworkfrom|prohibitiontoworkuntil|
|-----------------|---------------------|----------------------|
|                 |2021-01-04 00:00:00  |2021-01-02 00:00:00   |
```

`prohibitiontoworkuntil` cannot be before `prohibitiontoworkfrom`.

See also #4532.

### Steps to Reproduce
1. Go to Configuration -> Developers
2. Generate cases
3. Good luck. Since test data is random, the bug may or may not reproduce

### Expected Behavior
Enforce `prohibitiontoworkuntil` to be on the same day or after `prohibitiontoworkfrom`. If possible, make it consistent with the value of `prohibitiontowork` wether or not dates are generated here.

### System Details
* Device: Linux PC
* SORMAS version: 1.58.0-SNAPSHOT
* Android version/Browser: Firefox
* Server URL: local
* User Role:

DavidBaldsiefen:
These changes will introduce seeded randomization to devmode, as well as default configurations to streamline quick (but obviously limited) performance testing.
Closes #3233 

Edit: Also closes #4540


MateStrysewske (changes requested):
Cases can potentially be assigned ot laboratories. This a) doesn't make much sense and b) involves the problem that the facility type (which is a mandatory field) is null for these cases. Please make sure that only facilities that are valid options for cases (i.e. only facilities in the facility type groups returned by `FacilityTypeGroup:getAccomodationGroups` and with a facility type where `accomodation` is true) are used when generating them.

MateStrysewske (approved):

Issue #3716, opened 2020-12-07
Make user roles deactivateable [2]


MartinWahnschaffe:
### Feature Description
Currently all user roles in sormas are hard-coded. It's not possible to add or remove user roles. A concept for this is being prepared in #898.

### Problem Description
For #3664 it needs to be possible to deactivate certain user roles.

### Proposed Change
We already have a UserRoleConfig in the system that holds custom user rights for user roles.

- [x] extend the UserRoleConfig with an "enabled" field
- [x] When a user is edited only allow "enabled" user roles. Hide all other roles in the selection. Ideally add a method to the UserRoleConfigFacade that return alls enabled user roles. User roles that don't have a config have to be considered enabled.
- [x] Exemption: When editing a user that has a disabled user role, this user role has to be shown
- [x] Provide sample SQL on how to disable a user role (insert new user role config with the disabled setting, uuid, etc.)
- [x] Check for usages of user roles e.g. you should not be able to filter for disabled user roles

### Not in scope of this issue

Don't remove disabled user roles from existing users in the database

MartinWahnschaffe:
How to deal with technical user roles (e.g. REST_EXTERNAL_VISITS_USER, SORMAS_TO_SORMAS_CLIENT):

We were thinking about adding a separate user right to distinguish between user how manager "normal" user accounts and admin users that manage technical user accounts. For now we decided against it.
Technical users are mostly set-up when the server itself is set-up. So the process can be to setup those users first and then disable the user roles that are not supposed to be used in daily user management.
If needed a user role can be disabled any time in the database.

sergiupacurariu:
Script for updating user roles

[user role config script.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/5963897/user.role.config.script.txt)

MartinWahnschaffe:
Alternative script that disables a set of user roles:

```
DO $$
DECLARE 
	_roleEnabled boolean := false;
	_roleId bigint;
	_uuid varchar;
	_userRole varchar;
	_userRoles varchar[] := array['USER_ROLE_NAME_A', 'USER_ROLE_NAME_B'];
BEGIN

	FOREACH _userRole IN ARRAY _userRoles
	LOOP

		SELECT id FROM userrolesconfig WHERE userrole = _userRole INTO _roleId;
		
		IF(_roleId is null) 
			THEN 
				_uuid = overlay(overlay(overlay(
[...]

PR #4406, opened 2021-02-16
#3716 - Make user roles deactivateable

sergiupacurariu:


MateStrysewske (changes requested):
I think you have to move the changes to a new version number because developers and test systems already have this on their system without the default, which will probably lead to problems as soon as new entries are created in these tables.

I don't know whether it's possible to add the default to a column in retrospection, if it's not, you'll have to drop the column and add it again afterwards.

FredrikSchaeferVitagroup:
Yes. But systems that do already have an entry in the userrolesconfig will not be able to execute the first statements (without default values). This already broke https://ga1.sormas-oegd.de/. Is it likely that other systems already have this table populated?

MateStrysewske (approved):

PR #4415, opened 2021-02-16
Change/4413 config retain ext token

### LINKED ISSUE #4413: [SORMAS2SORMAS] SurvNet external tokens ("Aktenzeichen") are not globally unique
### Problem Description
Do not share external tokens across GAs as there is a (high) risk for collisions. Currently we generate default tokens of the form `SOR-CVD2021-00001` in the converter. This is also the example for the simplest collision as tokens are only unique for one GA.  When the receiver of a case sends the case to SurvNet, the local converter will pick a correct external token based on their current local counter value.

### Proposed Change
Introduce a property which allows to strip the external token from the transfer object before sending it to another SORMAS instance.

### Possible Alternatives
None, otherwise we will have to live with token collisions.

@markusmann-vg and @JaquM

JonasCir:
Closes #4413 

This is left over from my work on the converter. I feel this should be included in the next release, however, @markusmann-vg should have the last word on this

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/databuilder/CaseShareDataBuilder.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/databuilder/CaseShareDataBuilder.java
@@ -23,7 +23,9 @@ import java.util.stream.Collectors;
 import javax.ejb.EJB;
 import javax.ejb.LocalBean;
 import javax.ejb.Stateless;
+import javax.inject.Inject;
 
+import de.symeda.sormas.api.SormasToSormasConfig;
 import de.symeda.sormas.api.caze.CaseDataDto;
 import de.symeda.sormas.api.contact.ContactCriteria;
 import de.symeda.sormas.api.contact.ContactDto;
@@ -36,6 +38,7 @@ import de.symeda.sormas.api.sormastosormas.SormasToSormasOriginInfoDto;
 import de.symeda.sormas.api.utils.DataHelper;
 import de.symeda.sormas.backend.caze.Case;
 import de.symeda.sormas.backend.caze.CaseFacadeEjb;
+import de.symeda.sormas.backend.common.ConfigFacadeEjb;
 import de.symeda.sormas.backend.contact.Contact;
 import de.symeda.sormas.backend.contact.ContactService;
 import de.symeda.sormas.backend.sample.Sample;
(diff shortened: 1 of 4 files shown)


HolgerReiseVSys:
@leventegal-she Could you have a look at this?

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/databuilder/CaseShareDataBuilder.java
```diff
@@ -56,13 +59,22 @@
...
 	public CaseShareData buildShareData(Case caze, User user, SormasToSormasOptionsDto options) throws SormasToSormasException {
 		Pseudonymizer pseudonymizer = dataBuilderHelper.createPseudonymizer(options);
 
 		PersonDto personDto = dataBuilderHelper.getPersonDto(caze.getPerson(), pseudonymizer, options);
 		CaseDataDto cazeDto = getCazeDto(caze, pseudonymizer);
 
+		// external tokens ("Aktenzeichen") are not globally unique in Germany due to SurvNet, therefore, do not
+		// transmit the token to other GAs, but let them generate their own token based on their local, configurable
+		// format
+		if (!sormasToSormasConfig.getRetainExternalToken()){
```
leventegal-she, line 74: Do we need this only for cases?

if yes then the option could be renamed to `retainCaseExternalToken` otherwise  the externalToken should be cleared for `persons` and `contacts` too

sormas-base/setup/sormas.properties
```diff
@@ -82,6 +82,9 @@ app.url=
 # Password of the trust store
 #sormas2sormas.truststorePass=
 
+# Controls retention of the external token when sharing entities
+#sormas2sormas.retainExternalToken=
```
leventegal-she, line 86: would be nice to have a default value

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/databuilder/CaseShareDataBuilder.java
JonasCir: Persons do not currently have tokens in SurvNet. If need be proper logic needs to be added alter

JonasCir:
addressed comments @leventegal-she

leventegal-she (approved):

JonasCir:
Could this be merged then? :) I think it should go into 1.57

leventegal-she:
> Could this be merged then? :) I think it should go into 1.57

Yes, you can merge it

Issue #4433, opened 2021-02-17
Validate uuids


BarnaBartha:
### Feature Description
Add validation to resource and facadeEjb level for UUID's

PR #4434, opened 2021-02-17
#4433 - validate UUID's on res

BarnaBartha:


syntakker:
- In case Entities have been provided by external tool chains, the validation of Uuids by `^[0-9a-zA-Z]{6}-[0-9a-zA-Z]{6}-[0-9a-zA-Z]{6}-[0-9a-zA-Z]{8}$` might be overly strict and break workflows. A relaxed validation `^[0-9a-zA-Z-]*$` might do.
- Uuids are validated before writing to the database, while Uuids already present in the database are not affected.

MartinWahnschaffe:
I also think that this may cause problems with the UUIDs we are sometimes generating from the database side. There are currently two ways this is done:

`upper(substring(CAST(CAST(md5(CAST(random() AS text) || CAST(clock_timestamp() AS text))`

and

```
overlay(overlay(overlay(
                            substring(upper(REPLACE(CAST(CAST(md5(CAST(random() AS text) || CAST(clock_timestamp() AS text)) AS uuid) AS text), '-', '')), 0, 30)
                            placing '-' from 7) placing '-' from 14) placing '-' from 21)
```

UUIDs generated with the first one will definitely not go through the reg ex, when the related entity has to be saved again.

We need to at least get the whole UUID logic clean on our side, before we can do a strict validation. I'd suggest to go with Frank's approach or something in between.

syntakker:
Another point: the swagger API defintion changed (it will for both options if we do the validation in this place). We need to make shure that this does not interfere with any implementation SurvNet side.

JonasCir:
I'm investigating

JonasCir:
DON'T merge this. I'm seeing erros when using the REST api @BarnaBartha @syntakker

JonasCir:
ok works now, something in my setup was off

PR #4466, opened 2021-02-18
#4389 Case directory filters should depend on user-jurisdictionlevel

### LINKED ISSUE #4389: Users on district level and below can't filter the case directory by jurisdiction
### Bug Description
Users on district level or below (currently not implemented) can currently not filter cases by jurisdiction. All jurisdiction fields are disabled when only everything above and including their own jurisdiction level should be disabled (so for district-level users, region and district should be pre-selected and disabled).

### Expected Behavior
Enable filter fields below the user's jurisdiction level. E.g. surveilance officers need to be able to filter by community and health facility.

DavidBaldsiefen:
Fixes #4389 

This needs thorough testing, it would be really bad if some users end up with unwanted, unchangeable filters.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/facility/FacilityFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/facility/FacilityFacade.java
@@ -26,7 +26,6 @@ import javax.ejb.Remote;
 
 import de.symeda.sormas.api.region.CommunityReferenceDto;
 import de.symeda.sormas.api.region.DistrictReferenceDto;
-import de.symeda.sormas.api.region.RegionReferenceDto;
 import de.symeda.sormas.api.utils.SortProperty;
 import de.symeda.sormas.api.utils.ValidationRuntimeException;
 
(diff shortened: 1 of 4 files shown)


MateStrysewske (changes requested):
* In general, facility category should only be selectable once at least a district has been selected; point of entry should only be selectable once at least region has been selected
* The point of entry filter is not properly working; sometimes it's enabled initially (shouldn't be), but then gets disabled once a region is selected and doesn't get enabled again even if a district is selected, sometimes it's not properly filled before clicking on "Apply filters" (so even if you select a region and district with a POE in it, it doesn't appear in the dropdown)

Users on national level:
* Facility category is selectable if no other jurisdiction is selected; should be disabled
* Point of entry filter is not usable; it's active initially (shouldn't be) and then is disabled once a region is selected, and never gets active again until the region is removed

Users on district level and below:
* The filters are expanded by default for these users because the district filter is visible and filled in; the logic that determines whether a filter is selected to then determine whether the filter bar should be extended needs to take into account the user's jurisdiction level, so for a district-level user, the district field should not matter in this calculation

DavidBaldsiefen:
I made the following changes:

- facility category and so on are only enabled once a district is selected
  - when changing district with a selected facility type and category, those will stay
- PoE is only enabled when the correct case origin is set and at least a district is selected. I can also change this so that only a region needs to be definied, but then I'll need to add another method for that. Is that a requirement?
- If the users jursidiction level leads to specific parameters being unchangeable, the fields will now be _cleared_ and disabled (so that the filters are again collapsed by default)

MateStrysewske (approved):

Issue #4269, opened 2021-02-05
Improve code coverage with unit testing for the SurvnetGateway


vidi42:
Module: `sormas-backend`
Affected class: `de.symeda.sormas.backend.survnet.SurvnetGatewayFacadeEjb`

Suggestion:
- [x] Make sure the correct payload is sent for export
- [x] Make sure the correct payload is sent for deletion

PR #4490, opened 2021-02-22
#4269 survnetgatewayejb unit tests

ssferrazza:
See #4269

### DIFF
--- a/sormas-base/pom.xml
+++ b/sormas-base/pom.xml
@@ -802,6 +802,12 @@
 				<scope>test</scope>
 			</dependency>
 
+			<dependency>
+				<groupId>com.github.tomakehurst</groupId>
+				<artifactId>wiremock</artifactId>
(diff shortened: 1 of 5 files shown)


sormas-base/pom.xml
```diff
@@ -802,6 +802,12 @@
 				<scope>test</scope>
 			</dependency>
 
+			<dependency>
+				<groupId>com.github.tomakehurst</groupId>
+				<artifactId>wiremock</artifactId>
+				<version>1.58</version>
```
vidi42, line 808: Seems that the version you used here is from 2015. Any reason for this? There are to be some vulnerabilities for versions before 2.16.0 https://www.cvedetails.com/cve/CVE-2018-9116/

vidi42 (changes requested):

vidi42 (approved):

PR #4514, opened 2021-02-23
#4513: add import user example script

### LINKED ISSUE #4513: Provide example script to import users
### Current Situation
At the moment SORMAS doesn't support a user import functionality.
In order to support importing a list of test users this has to be done at DB level at the moment.

### Feature Description
Provide an example on how to import a list of users into the DB.

vidi42:
Closes #4513

### DIFF
--- a/sormas-backend/utils/importUsersExample.sql
+++ b/sormas-backend/utils/importUsersExample.sql
@@ -50,7 +50,7 @@ DECLARE
     region_id   bigint;
 BEGIN
 
-    inserted_id := (SELECT max(id) FROM users) + 1;
+    inserted_id := nextval('entity_seq');
     salt := gen_salt('md5');
     region_id := (SELECT id FROM region WHERE epidcode = $5);
 -- add here any other infrastructure relation query
--- a/sormas-backend/utils/importUsersExample.sql
+++ b/sormas-backend/utils/importUsersExample.sql
@@ -1,0 +1,81 @@
+/*
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2020 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


sormas-backend/utils/importUsersExample.sql
```diff
@@ -0,0 +1,81 @@
...
+    LANGUAGE plpgsql AS
+$$
+DECLARE
+    inserted_id integer;
+    salt        varchar;
+    region_id   bigint;
+BEGIN
+
+    inserted_id := (SELECT max(id) FROM users) + 1;
+    salt := gen_salt('md5');
```
StefanKock, line 54: SORMAS uses here also 'sha256', why using 'md5' here?

sormas-backend/utils/importUsersExample.sql
```diff
@@ -0,0 +1,81 @@
...
+    RETURNS integer
+    LANGUAGE plpgsql AS
+$$
+DECLARE
+    inserted_id integer;
+    salt        varchar;
+    region_id   bigint;
+BEGIN
+
+    inserted_id := (SELECT max(id) FROM users) + 1;
```
StefanKock, line 53: Use `nextval('entity_seq')` as in sormas_schema.sql

StefanKock (changes requested):
I see two changes to be made, apart from that it looks good.

sormas-backend/utils/importUsersExample.sql
vidi42: https://www.postgresql.org/docs/9.6/pgcrypto.html `gen_salt` method doesn't support the sha256 algorithm for generating the salt. The only supported ones are `bf`, `md5`, `xdes` and `des`.
`bf` would be the best option, however the string it generates exceeds the supported column length.
Since this refers only to the salt, I think the `md5` option is still a pretty good one. 

This is an example and can be tweaked by anyone who intends to use it.

StefanKock: Okay, understood

StefanKock (approved):

Issue #4332, opened 2021-02-10
Copy Follow-up status comment when converting a contact into case [1]


maxiheyner:
### Bug Description
If a confirmed contact is converted into a case all information in the FOLLOW_UP STATUS COMMENT is lost.
In version 1.52.1 the user got the information that " Follow-up was canceled automatically since this contact was converted into a case" within this field (skipping all other information). In version 1.56.0 completely no information is transferred and no message appears in that field. 
### Steps to Reproduce
1. Save some information in the FOLLOW-UP STATUS COMMENT field of a contact
2. Convert this confirmed contact into a case and pay attention for this above mentioned field

### Expected Behavior
All saved information should appear after converting a contact into a case, added by the message that "Follow-up was canceled automatically since this contact was converted into a case"
### Screenshots
![image](https://user-images.githubusercontent.com/76884029/107514732-57685600-6baa-11eb-844d-24e1e4aaac7b.png)

<img width="467" alt="image" src="https://user-images.githubusercontent.com/76884029/107514080-74505980-6ba9-11eb-8094-1f53a6bce66e.png">
<img width="746" alt="image" src="https://user-images.githubusercontent.com/76884029/107514253-b7aac800-6ba9-11eb-9138-f71b3b4e0229.png">

### System Details
* Device: desktop PC
* SORMAS version: 1.52.1 additionally tested with 1.56.0
* Android version/Browser: chrome
* Server URL:
* User Role: admin

### Additional Information

- Old cases, that were converted to a case in the past are ignored

kwa20:
@maxiheyner I think the main problem here is the overwritten entry in the contact form right?
<img width="436" alt="followUpOverwritten" src="https://user-images.githubusercontent.com/70317594/107559605-a0390280-6bdc-11eb-998a-acdee6ffa07c.PNG">

bernardsilenou:
- I think we need to merge both text
- We can also clone all followup comments from the contact to the case form.

maxiheyner:
@bernardsilenou and @kwa20 we definitely need both the original comment AND the message of contact into case conversion.

AndyBakcsy-she:
Validated ticket on the latest version of 1.58.0-snapshot on local environment

PR #4535, opened 2021-02-24
#4332: Copy followup comment when converting contact to case

ChristopherRiedel:
#4332


BarnaBartha (approved):

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -330,8 +330,11 @@ public ContactDto saveContact(ContactDto dto, boolean handleChanges, boolean han
 			if (dropped || convertedToCase) {
 				contactService.cancelFollowUp(
 					entity,
-					I18nProperties
-						.getString(convertedToCase ? Strings.messageSystemFollowUpCanceled : Strings.messageSystemFollowUpCanceledByDropping));
+					new StringBuilder(entity.getFollowUpComment()).append(" ")
```
MateStrysewske, line 333: Tests are failing. I suspect this is the reason and that StringBuilder doesn't like to be initiated with null.

MateStrysewske (changes requested):

MateStrysewske (approved):

Issue #4306, opened 2021-02-09
Add handleExternalJournals FeatureType


BarnaBartha:
- place handleExternalJournals behind a feature activated check for case save

syntakker:
@BarnaBartha there seem to be redundant calls to `PersonFacade` around the relevant code here. I'll investigate further.
If I'll find things, however, I'll open a new issue for this, so you can start to review nevertheless.

Issue #4243, opened 2021-02-04
Avoid creating ScheduledExecutorService for every single delayed task [1]


jbellmann:
### Feature Description

In current implementations of [CaseFacadeEjb](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java#L1668), [ContactFacadeEjb](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java#L351) and [PersonFacadeEjb](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java#L368) a new ScheduledExecutorService is created whenever a new Task should be scheduled for delayed execution.

### Problem Description

I'm not quite sure my knowledge about EJB environments is outdated but I remember a rule that it's not recommended to create threads programmatically. Instead use something provided by the EJB environment/container.

But with current implementation for every single invocation a new ScheduledExecutorService with new ThreadFactory is created that will also create a new Thread when the task has to be executed.

So in case of massive imports and possible invocation of this method this can lead to a massive bunch of Threads created to execute delayed tasks at once. Massive Thread management can lead to heavy CPU consumption I guess and low performance.

### Proposed Change
My proposed change would be to inject a ManagedScheduledExecutorService into these EJBs and reuse the environment provided mechanism to schedule delayed tasks and maybe avoid massive creation of new threads.

### Additional Information

It's about invocation of `ExternalJournalService.notifyExternalJournalPersonUpdate(JournalPersonDto)`

StefanKock:
I looked over that code, added appropriate labels and have some remarks:

1. There should better encapsulation, the above mentioned EJBs should not all take care this scheduling.
2. `ExternalJournalService.notifyExternalJournalPersonUpdate` contains three conditions that are often not met looking how most instances are used. And it would even be better to check before if symptom journal or patient dairy is connected before checking the data.

PR #4551, opened 2021-02-26
refactor ExternalJournal update (#4243)

syntakker:
Remarks from  https://github.com/hzi-braunschweig/SORMAS-Project/issues/4243#issuecomment-778222278:

- `ExternalJournalService.notifyExternalJournalPersonUpdate` is called from `ExternalJournalService.handleExternalJournalPersonUpdate` only. Wether an external journal is active at all is checked there first (fixed in #4306).
- `ExternalJournalService.shouldNotify` is about detecting changes form updates where initially only the previous state is known. Unfortunately even `PersonDto.isEnrolledInExternalJournal()` may have been subject to change, so fetching the person data is necessary

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
@@ -237,8 +237,7 @@ public class ExternalJournalService {
 		 * new data can be retrieved from DB
 		 */
 		JournalPersonDto existingPerson = personFacade.getPersonForJournal(person.getUuid());
-		Runnable notify = () -> notifyExternalJournalPersonUpdate(existingPerson);
-		executorService.schedule(notify, 5, TimeUnit.SECONDS);
+		executorService.schedule((Runnable) () -> notifyExternalJournalPersonUpdate(existingPerson), 5, TimeUnit.SECONDS);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
```diff
@@ -220,14 +225,29 @@ public boolean notifyExternalJournalPersonUpdate(JournalPersonDto existingJourna
...
+		}
+
+		/**
+		 * The .getPersonForJournal(...) here gets the person in the state it is (most likely) known to an external journal.
+		 * Changes of related data is assumed to be not yet persisted in the database.
+		 * 5 second delay added before notifying of update so that current transaction can complete and
+		 * new data can be retrieved from DB
+		 */
+		JournalPersonDto existingPerson = personFacade.getPersonForJournal(person.getUuid());
+		Runnable notify = () -> notifyExternalJournalPersonUpdate(existingPerson);
```
StefanKock, line 240: I would inline the Runnable notify, I see no value having it assigned.

StefanKock (changes requested):
One small change, everything else looks like a refactoring I would done equally.

sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
syntakker: Done.

PR #4579, opened 2021-03-01
Feature 4090 add optional translation to campaign diagram definition

alexcaruntu-vita:


sormas-backend/src/main/java/de/symeda/sormas/backend/campaign/diagram/CampaignDiagramDefinition.java
```diff
@@ -87,4 +100,50 @@ public boolean isPercentageDefault() {
 	public void setPercentageDefault(boolean percentageDefault) {
 		this.percentageDefault = percentageDefault;
 	}
+
+	@Lob
+	@Type(type = "org.hibernate.type.TextType")
```
BarnaBartha, line 105: is there a specific reason why we are not using type json ?

BarnaBartha (approved):

Issue #4491, opened 2021-02-22
Prompt for saving data even when no data were changed for EP


Marko-ilmkreis:
### Bug Description

In #3014 were this fixed for contacts/cases. But it still appears in event participants. Like in #3014 it only appears if gps-data are entered.

### Steps to Reproduce
1. open an event participant and try to go back to the event directory.
2. make sure that gps-data are entered
2. The windows appears even if there were no data changed.

### Screenshots

![grafik](https://user-images.githubusercontent.com/74791118/108686101-3e8a6980-74f5-11eb-8fe5-c671a5d9f25f.png)

### System Details
* Device: Win 10
* SORMAS version: 1.56.1 and 1.58-snapshot
* Android version/Browser: FF
* Server URL: own instance and https://test-de1.sormas.netzlink.com/

syntakker:
Fixed by PR #4585 for issue #4477.

raghupola-vg:
Tested on version 1.58.0-snapshot on latest deploy and issue is Fixed.

syntakker:
@raghupola-vg the fix has not even been merged :-) However, the circumstances to reproduce the the bug are quite special, see PR #4585 for a description.

syntakker:
Fixed by PR #4585.

AndyBakcsy-she:
Validated ticket on the latest version of sormas version 1.58.0-snapshot on test.sormas.netzlink.com

Issue #4477, opened 2021-02-19
No prompt for saving unsaved data if the address was changed


Marko-ilmkreis:
### Bug Description

If you change the address in the person tab of contacts/cases and leave the form there is no dialog asking if unsaved data should be saved or discard.

It works if you change the name. Then a dialog is asking what to do with unsaved data.

Maybe this comes from the solution of #3014 .

### Expected Behavior

It should be asked every time the user changes data what to do.

### System Details
* Device: Win10
* SORMAS version: 1.58.-snapshot and 1.56.1
* Android version/Browser: FF
* Server URL: https://test-de1.sormas.netzlink.com and own instance
* User Role: every user

AndyBakcsy-she:
Validated ticket on the latest version of  SORMAS
Version: 1.58.0-SNAPSHOT on test.sormas.netzlik.com

PR #4585, opened 2021-03-02
prompt for unsaved address data (#4477)

syntakker:
**For review / testing:**
Please make also sure that the bugfix from #3014 (fixed in PR #4097) is still working. The bug described in #3014 manifested under these conditions:
- `country.locale=en-US` in `sormas.properties`
- browser language (not SORMAS user language!) set to German
- in a case or its person, fill one of the latitude / longitude fields.
- reload or navigate away and return to this page. Decimal fractions should be displayed with a `,`, e.g., `3,2` now
- navigate away from the page without any prior modification. No prompt to save or discard should pop up.

Additional cleanup:
Commented code removed. The code was commented in 3/26/19, so I would not expect it to be a candidate to be revived.


syntakker:
Fixes #4491, too.

BarnaBartha:
@syntakker It looks good to me, i don't know how this method was missed in the first dirty check implementation. I do not have the time to test it now though.

BarnaBartha (approved):

PR #4586, opened 2021-03-02
#3992 Use the SORMAS instance name and localized entity names for export file names

### LINKED ISSUE #3992: Use the (custom) SORMAS instance name and localized entity names for export file names [1]
### Feature Description
The current export is manually coded to start with something like "sormas_cases_" or "sormas_regions_".

### Problem Description
* For branded systems (e.g. APMIS) the "sormas" is a bit confusing.
* For systems that are completely based on another locale the english entity names are confusing.

### Proposed Change

- [x] Use the calculated instance name (ConfigFacade.getSormasInstanceName) for the first part of the prefix. Should be lower-case and also escaped to only contain valid file characters
- [x] Use the i18n name of the exported entity based on the systems locale (not the user's!) for the second part of the prefix.  Should be lower-case and also escaped to only contain valid file characters

Do this for all exports in the system: cases, contacts, events, event participants, ... regions, districts, ... Search for usages of GridExportStreamResource and DownloadUtil to find entry points for those exports.

vlad-ciucescu:
closes #3992


leventegal-she (changes requested):
Please revert the changed xls files

leventegal-she (approved):

PR #4588, opened 2021-03-02
Feature 3352 report date for epidemiological curve

alexcaruntu-vita:


MateStrysewske (changes requested):
Great job! I like the refactorings you've done! The only change request I have is to move the new filter to the top row, to the left of the region filter, and to remove the info icon because its information is now outdated (since you introduced the date type picker)

MateStrysewske (changes requested):
Last request: Please also integrate the new field into the logic that only updates the dashboard once "Apply filters" has been clicked (since it's now part of that row); Clicking on "Reset filters" should reset the field to "Most relevant date"

MateStrysewske (approved):

PR #4616, opened 2021-03-03
Be able to attach multiple files on Events and Actions

### LINKED ISSUE #4502: Event - follow-up of actions: attach documents at the level of the action carried out
### Situation Description
To date, users can attach documents at the event level. Users express the need to have this functionality at the action level (event action tab) to facilitate the readability of information.

### Feature Description
Creation of a button in the “event action” tab to attach one or several documents to each action carried out during the investigation.

### Additional Information
@ftavin @bernardsilenou @emond

### LINKED ISSUE #4504: Event - attachment function: attach multiple documents at the same time
### Situation Description
To date, it is possible to upload one document at a time in the event part. In order to save time in managing events, users have the need to import several documents at the same time.

### Feature Description
Develop the possibility of selecting several documents in its explorer in order to import all the documents referring to an event at once.

### Additional Information
@bernardsilenou @ftavin @emond

aponcon-atolcd:
Fixes #4502 
Fixes #4504

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/action/ActionDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/action/ActionDto.java
@@ -180,6 +180,10 @@ public class ActionDto extends EntityDto {
 		this.actionMeasure = actionMeasure;
 	}
 
+	public ActionReferenceDto toReference() {
+		return new ActionReferenceDto(getUuid(), getTitle());
+	}
+
 	public ReferenceDto getContextReference() {
 
 		switch (actionContext) {
--- a/sormas-api/src/main/java/de/symeda/sormas/api/action/ActionReferenceDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/action/ActionReferenceDto.java
@@ -1,0 +1,43 @@
+/*******************************************************************************
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2018 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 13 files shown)


MateStrysewske (changes requested):
Upload is cancelled whenever the user clicks somewhere outside the upload window; the upload either needs to run in the background (and not get interrupted by the user navigating with the application in any other way) or needs to be displayed in the center of the screen as a modal, with the user not being allowed to interact with the rest of the application until the upload has completed

MateStrysewske:
I just realised that the documents are uploaded anyway (at least those up until the point where the user interrupts the upload), but they're only displayed after the page is reloaded

aponcon-atolcd:
Thank you for you feedback.

There was indeed an issue with the UI : after pushing the files to the server, the UI was only updated if we clicked away from the Upload panel and into a ComboBox input.
This issue was appearing only if we tried to upload only one file.
It was related to a Polling issue so I used the ```SormasUI.get().setPollInterval(300)``` during the upload process (more exactly when the Upload panel is open). Is there a better solution than using this?

MateStrysewske:
@aponcon-atolcd Hey! I'm fine with that solution, but it's still a problem that the user interrupts the upload whenever they click somewhere outside the upload status window in the bottom right. Like I've written above, either the upload needs to continue running (and the upload window should stay on the screen) when the user clicks or navigates somewhere else, or the upload status window should be moved to a dedicated modal that doesn't allow users to further interact with the system until the upload is completed.

aponcon-atolcd:
@MateStrysewske Hi !
Now, I understand what you were talking about. I thought you were talking about the popup window, not the progress window.
Should be ok now :)

MateStrysewske (approved):

aponcon-atolcd:
For more information about the reason why it failed, the progress window were detached from the UI when the upload popup button were made hidden (either by clicking on the upload button or by clicking anywhere in the page) causing the interruption of uploads.
Making the popup always visible during the upload process fixes the issue.

bernardsilenou:
@MateStrysewske 
What the name of the parameter to activate this feature ple?
Can you set this configuration on so that we can test this feature on https://test.sormas.netzlink.com/sormas-ui/
Merci

aponcon-atolcd:
@bernardsilenou The feature name is DOCUMENTS.

Issue #3430, opened 2020-11-07
Introduce bulk edit option for tasks


kwa20:
### Situation Description
Bulk actions tasks
### Feature Description
The option to bulk edit tasks (reassign, priority and status change) might be useful for users when many changes for tasks have to be made. This feature should serve as an addition (second step implementation) to the newly to be added filters that are planned for the task management (#3278 ).

- Add options to archive and edit tasks alongside being able to delete them.
- Editing should allow changing of assignee and priority as well as the task status.

### Additional Information
<img width="631" alt="Screen Shot 2021-02-23 at 8 36 25 AM" src="https://user-images.githubusercontent.com/70317594/108813981-583db680-75b2-11eb-8887-d9b887fba487.png">

bernardsilenou:
@carolinverset @ftavin any comment

kwa20:
Add archiving as an option (cf. 4060)

ftavin:
@carolinverset @bernardsilenou same need in France (including archiving)

@kwa20 : can you edit your original description to include archiving (text + mockup) ?

kwa20:
@ftavin @carolinverset I've update the original description. Does that look fine for you?

carolinverset:
@kwa20 @ftavin 
Perfect ! Thank you very much

bernardsilenou:
@MateStrysewske @MartinWahnschaffe  Question related to archiving of tasks please: Should we allow the archiving of tasks linked to parent entities (case, contact, ep) that are active? I though we only archive child entities through their parents but bot the other way round.

MateStrysewske:
I don't see a reason to not do it. One possible use case would be to get rid of "old" tasks for a case that has been active for a while and has a lot of done tasks in order to not flood the task list with tasks that have already been performed.

ftavin:
@bernardsilenou @carolinverset @MateStrysewske 
Our team will do the dev on this one !

Marko-ilmkreis:
@ftavin will ths be implemented in the next sprint?

ftavin:
@Marko-ilmkreis
as you can see a pull request is in progress so it is likely to be integrated in the next version (1.59)

Marko-ilmkreis:
@ftavin Ok, Thanks. As its not linked to an sprint I thought there might be a delay.

maxiheyner:
Isse Tracker Comment

MateStrysewske:
[...]

Issue #4582, opened 2021-03-02
Allow all user roles to assign tasks to users with a higher jurisdiction


DavidBaldsiefen:
### Feature Description

As implemented with #4048 for regional->national users, allow community-level users to also assign tasks to regional-level users.

### Additional Information

- Use the FeatureType `ASSIGN_TASKS_TO_HIGHER_LEVEL`
- Check only for users Jurisdiction Level, **not** for specific roles. Because of the upcoming user role rework, this feature should be as generic as possible

MateStrysewske:
@DavidBaldsiefen Is it necessary that community-level users can directly assign tasks to region-level users, or would it be fine if they could assign them to district-level users (who then in turn would be allowed to assign tasks to region-level users)? Because if we do it that way, we can implement it generically as in that every user is allowed to assign tasks to users of the jurisdiction level above them.

DavidBaldsiefen:
> @DavidBaldsiefen Is it necessary that community-level users can directly assign tasks to region-level users, or would it be fine if they could assign them to district-level users (who then in turn would be allowed to assign tasks to region-level users)? Because if we do it that way, we can implement it generically as in that every user is allowed to assign tasks to users of the jurisdiction level above them.

I think that would be fine and makes sense. But I van check back with the user who requested this @stephan-2

DavidBaldsiefen:
@MateStrysewske Do you think it makes sense to limit the users of the next-higher jurisdiction level to those whose jurisdiction also "contains" my jurisdiction? E.g. only district-users when their district also contains my community.

MateStrysewske:
@DavidBaldsiefen That sounds like a very good idea, yes!

maxiheyner:
Issue tracker comment

AndyBakcsy-she:
Validated ticket on the latest version of 1.58.0-snapshot on test.sormas.netzlink.com

PR #4617, opened 2021-03-03
#3430 Add bulk edit/archive option on task directory

mathieu-petit-atol:
Closes #34430


sormas-backend/src/main/java/de/symeda/sormas/backend/task/TaskService.java
```diff
@@ -398,4 +408,19 @@ public User getTaskAssignee(Contact contact) throws TaskCreationException {
...
+	public void updateArchived(List<String> taskUuids, boolean archived) {
+
+		CriteriaBuilder cb = em.getCriteriaBuilder();
+		CriteriaUpdate<Task> cu = cb.createCriteriaUpdate(Task.class);
+		Root<Task> root = cu.from(Task.class);
+
+		cu.set(Task.CHANGE_DATE, Timestamp.from(Instant.now()));
+		cu.set(root.get(Task.ARCHIVED), archived);
+
+		cu.where(root.get(Case.UUID).in(taskUuids));
```
MateStrysewske, line 422: => Task.UUID

sormas-ui/src/main/java/de/symeda/sormas/ui/task/BulkTaskDataForm.java
```diff
@@ -0,0 +1,161 @@
...
+		reassignedToCheckbox.addValueChangeListener(e -> {
+			reassignedTo.setEnabled((boolean) e.getProperty().getValue());
+		});
+	}
+
+	private List<UserReferenceDto> getUsers() {
+		final List<UserReferenceDto> users = new ArrayList<>();
+		UserDto userDto = UserProvider.getCurrent().getUser();
+
+		users.addAll(FacadeProvider.getUserFacade().getAllUserRefs(false));
```
MateStrysewske, line 131: This will allow users to assign tasks to users that normally wouldn't be allowed to be assigned, which is not okay. This logic either needs to replicate the logic from TaskEditForm or it can't be possible to bulk-assign tasks to new users. If that's a necessity, I guess assigning new users to tasks can only be possible if e.g. only tasks associated with cases from the same district are selected, etc.

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskBulkEditData.java
```diff
@@ -0,0 +1,44 @@
...
+public class TaskBulkEditData extends EntityDto {
+
+	private static final long serialVersionUID = -7234609753914205675L;
+
+	public static final String TASK_REASSIGNED_TO = "taskReassignedTo";
+	public static final String TASK_PRIORITY = "taskPriority";
+	public static final String TASK_STATUS = "taskStatus";
+
+
+	private UserReferenceDto taskReassignedTo;
```
MateStrysewske, line 17: Change this to `assignee` and update all similarly named fields and properties (caption should be "Change task assignee")

MateStrysewske (changes requested):
Please have another look at how the archived property of e.g. cases affects the system and replicate necessary functionality for tasks. One example would be that it's necessary to add a `pullArchivedUuidsSince` method to `TaskResource` that is called in the mobile app to remove archived tasks from the device. There might be other pieces of code that also need to be considered.

sormas-ui/src/main/java/de/symeda/sormas/ui/task/BulkTaskDataForm.java
```diff
@@ -0,0 +1,169 @@
...
+	private List<UserReferenceDto> getUsers() {
+		final List<UserReferenceDto> users = new ArrayList<>();
+		UserDto userDto = UserProvider.getCurrent().getUser();
+
+		if (district != null) {
+			users.addAll(FacadeProvider.getUserFacade().getUserRefsByDistrict(district, true));
+		}
+
+		// Allow regional users to assign the task to national ones
+		if (FacadeProvider.getFeatureConfigurationFacade().isFeatureEnabled(FeatureType.ASSIGN_TASKS_TO_HIGHER_LEVEL)
```
MateStrysewske, line 142: #4582 has just introduced an extended logic that allows users of every user role to assign tasks to users of the next higher jurisdiction level. Please have a look at the changed code and make sure that the code here works identically

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskController.java
```diff
@@ -177,4 +184,137 @@ public void run() {
...
+				I18nProperties.getString(Strings.headingNoTasksSelected),
+				I18nProperties.getString(Strings.messageNoTasksSelected),
+				Type.WARNING_MESSAGE,
+				false).show(Page.getCurrent());
+			return;
+		}
+
+		// Check if tasks with multiple districts have been selected
+		DistrictReferenceDto district = null;
+		for (TaskIndexDto taskIndexDto : selectedTasks) {
```
MateStrysewske, line 220: If I'm not mistaken, this logic will make bulk-reassigning general tasks (that aren't associated with any entity) impossible. I think reassigning should also be possible if ONLY tasks without districts have been selected

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/task/BulkTaskDataForm.java
MateStrysewske: This part of the code now has to be removed (lines 154 to 161); the generic code that you added makes it redundant

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskController.java
MateStrysewske, line 220: Just realised that this will perform poorly and put a lot of stress onto the database when many tasks are selected since it requires potentially one additional query per selected task; please re-write `getTaskIndexDistrict` to instead return whether or not all tasks lie in the same OR no district for a list of given tasks (instead of one single task); that code should then also be moved to the backend

MateStrysewske (changes requested):

aponcon-atolcd:
Hi
Here are the fix you asked for.
Thanks for the review.

It seems that the tests stopped working with my PRs (a token issue apparently). Seems related to some configuration between this repository and the forked one. 🤔

MateStrysewske (changes requested):
The performance should be fine now - however, one thing that wouldn't work now if I'm not mistaken is that it wouldn't be possible to bulk-edit tasks that aren't associated with cases, contacts, events etc. (because for those, the new backend method would also return "null" since there's no district in the result list at all). I think that should be possible though. Do you actually need the district in the UI? If not, an easy solution would be to rename `getSharedDistrictByTaskUuids` to `getNumberOfAssociatedDistricts` and let it return an int. If the result is 0 or 1, bulk-editing would be allowed, 2 or more would prevent it. The batch execution could stop when the count is already 2 or higher. If that's not possible, a possible alternative would be to let the method throw an exception, but I don't know if that's good practice.

MateStrysewske:
@aponcon-atolcd By the way the problem with the GitHub CI should be fixed, so when you're making your next commit, the tests should run through again :)

aponcon-atolcd:
@MateStrysewske Thanks for your code review and for fixing the Github CI issue :) 
We need to keep the shared related district to filter the assignee list in the edit form so returning the district count is not enough. Instead I've worked to return a list containg at most 2 districts. I've also fixed a NPE issue while querying for related districts. And last fix is about not allowing edition of several tasks related to different districts.

MateStrysewske (approved):

PR #4618, opened 2021-03-03
Make unaccented search working on accented data in web app

### LINKED ISSUE #4168: Adapt the text search filter so that it is less case sensitive
### Problem Description
In France, many names or cities-names have accents. SORMAS 'text search filters are very sensitive to case and stress. Users find it difficult to filter different directories on names that include accents.

### Proposed Change
We are proposing an adaptation of SORMAS to make filters less sensitive to emphasis and box.

### Additional Information
@bernardsilenou   @ftavin can you complete with technical details

aponcon-atolcd:
Fixes #4168


MateStrysewske (changes requested):
Hey! Before I do an in-depth review of this pull request, I'd like to ask you to remove the feature type that you've added for it. As I've stated in the linked issue, I think this change is very useful for every possible SORMAS server in the world because there are people with accents (or other special characters) in their names living everywhere :)

aponcon-atolcd:
Hi Mate.
The reason why I added a Feature Type here is because this feature needs to enable a PostgreSQL extension (```CREATE EXTENSION unaccent;```). I can't put this in the ```sormas_schema.sql``` file because it needs specific rights to be able to add the extension, right that may not be enabled.
The feature type is disabled by default to ensure the software will start as expected and to be sure that listing pages will work too.

By the way, I'll need to fix the merge conflicts too and apply this work on recent works too.

MateStrysewske:
@aponcon-atolcd I see. I'm wondering whether there's another way around it because I'd really, really like to have this enabled all the time and by default. I'll talk to a colleague of mine and get back to you!

MateStrysewske:
@aponcon-atolcd Talked to @MartinWahnschaffe who has a similar issue with another feature right now. The approach will be to 

* Add the extension to the server setup script to make sure that it's enabled on new servers
* Add a hint to the release notes that server admins HAVE to install the extension before updating to the new version

Like I said, having this feature is important for all countries, so we should enforce it. Therefore my request is still valid, please remove the feature type again :)

Because the approach we decided on will require changes to the Docker version of SORMAS as well (which is not maintained by us) to make sure that development systems are still working properly, I unfortunately don't think we can merge it in time for the upcoming release next week, but I'll make sure that it's going to be part of the next one.

aponcon-atolcd:
The last push :
- removed the commit that added the FeatureType.
- rebased the commit and fixed conflicts
- added the management of the PostgreSQL extension into the file ```server-setup.sh```

MateStrysewske (approved):
Thank you! This is great! I've approved the PR, but I'll have to wait a bit until I merge it into the dev branch to make sure that our test systems are configured accordingly.

And yes, I'll add a hint to the release notes for 1.59.0 that instructs server administrators to enable this extension, and I'll probably do a tweet as soon as it's merged into development that instructs developers to do the same.

PR #4622, opened 2021-03-04
#4316 - Identify performance problems in detailed export.

### LINKED ISSUE #4316: Identify performance problems in detailed export
### Problem Description
The detailed case and contact exports can be very slow on some systems, with export of a few thousand entries taking more than 10 minutes which results in a timeout on those servers. We need to investigate this, identify performance problems and fix them.

cazacmarin:
Closes #4316

### DIFF
--- a/sormas-base/setup/sormas.properties
+++ b/sormas-base/setup/sormas.properties
@@ -210,3 +210,6 @@ sms.auth.secret=
 
 # Turn off Attribute logging in the Auditlog on instances where this is not needed
 # auditor.attribute.logging=true
+
+# Bulk operation for CSV Cases/Contacts/Events/Samples Detailed Export. Optimal value is suggested to be set to 5000.
+# stepSizeForCsvExport=5000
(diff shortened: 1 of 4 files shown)


leventegal-she (changes requested):
please add a a commented line in the sormas.properties file with a default value for `stepSizeForCsvExport` and write a few words about setting the best value for performance

sormas-base/setup/sormas.properties
```diff
@@ -210,3 +210,6 @@ sms.auth.secret=
 
 # Turn off Attribute logging in the Auditlog on instances where this is not needed
 # auditor.attribute.logging=true
+
+# Bulk operation for CSV Cases/Contacts/Events/Samples Detailed Export. Optimal value is suggested to be set to 5000.
```
leventegal-she, line 214: Please change `Bulk operation` to `Batch size`

leventegal-she (changes requested):

leventegal-she (approved):

Issue #4555, opened 2021-02-26
Add a "sampling reason" field in the sample part [3]


carolinverset:
Implementation specifications: https://github.com/hzi-braunschweig/SORMAS-Project/issues/4555#issuecomment-789585894

### Situation Description
There are many reasons for people to take tests (in anticipation of travel or hospitalization for an operation, professional reasons, personal choice, etc.). Not all people who take a test are symptomatic. It would be good if we could define in SORMAS the reasons that led people to carry out a test

### Feature Description
For the cases, add a field in "sample" allowing to define the reasons which led the cases to carry out the test.
The modalities would be as follows:

- Contact of a confirmed case
- Evocative clinical signs
- Before hospitalization
- Trip
- Professional reason
- Voluntary process
- Moving/Returning into Retirement Home
- Not specified

refined description here : https://github.com/hzi-braunschweig/SORMAS-Project/issues/4555#issuecomment-789585894

### Additional Information
@ftavin @bernardsilenou @emond-fuseau

Marko-ilmkreis:
I would ask to add
* Moving/Returning into Retirement Home

carolinverset:
Yes, good idea. I'm agree with you :-)

bernardsilenou:
@carolinverset @Marko-ilmkreis
1. Which entity do you think users would  create for these people: case, contact; ep ?
2. If case, then how do we classify them while the test result is not out: not yet classified ?

carolinverset:
@bernardsilenou @Marko-ilmkreis @emond-fuseau
1. yes French users want this field for cases only.
2. this field could appear only for cases for which a test result is available?
What do you think of ?

bernardsilenou:
@Candice-Louw 
- Since the test result is available at the time of creating the case, then we can easily determine the case classification from the test result: Negative == not a case, positive == confirmed case.
-  If result is not available, then we can still create a case with classification  == not yer classified

bernardsilenou:
Can we call it "Reason for sample collection" or "Reason for testing"? not to get confused with the epidemiological meaning of sampling.

ftavin:
@carolinverset @bernardsilenou 
I understand this field should only be visible when the sample is linked to a case, am I right ?

should this new data also appear in the samples directory as a new column ? as a filter ? in the exports ?
on the case screen, should this information be displayed on the samples bloc ?

bernardsilenou:
[...]

PR #4630, opened 2021-03-04
4555 sampling reason

leventegal-she:
Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/4555


sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6786,4 +6786,43 @@ UPDATE labmessage SET testresultverified = true;
...
+                    END;
+                UPDATE samples set samplingreason = _samplingreason, samplingreasondetails = rec._covidtestreasondetails where id = latest_sample._id;
+            END LOOP;
+    END;
+$$ LANGUAGE plpgsql;
+
+ALTER TABLE cases
+    DROP COLUMN covidtestreason,
+    DROP COLUMN covidtestreasondetails;
+ALTER TABLE cases_history
```
MateStrysewske, line 6822: Don't remove the fields from cases_history - in case something went wrong during the migration, we can use these fields to restore the data

MateStrysewske (changes requested):
Looks good! Only two small change requests :)

MateStrysewske (approved):

PR #4640, opened 2021-03-04
detect charset for import (#4050)

### LINKED ISSUE #4050: umlauts and country variables produce errors during imports
### Bug Description
When using umlauts in free text fields in the import templates a critical error appears upon importing the data. This is quite problematic since the reason of the error is not easily identifiable and users rely more and more on the import function with rising case numbers. Furthermore, importing the newly added person.birthCountry and person.citizenship variables results in a partially failed import with only a generic error message also.

Applicable across all import options.
### Steps to Reproduce 
1. import data with either umlauts being entered in text type fields and country variables in the birthCountry and citizenship fields.
2. check errors upon importing the data.

### Expected Behavior
umlauts in text type fields should be ignored during the import and importing birthCountry and citizenship fields should be able to do with country infrastructure data as variables.
### Screenshots
Sorry for the german error messages. They didn't change when switching languages.
criticalErrorFreeTextUmlauts

umlauts:
"Import failed! Import failed because of a critical error. Please contact your admin..."

importFailBirthcountryCitizenship

birthCountry, citizenship:
"Import partially successful! The import was partially successful but some of the data could not be imported due to erroneous data."

### System Details
* Device:
* SORMAS version: 1.53.0

### Additional Information

syntakker:
fixes #4050 

~~Charset detection for non-UTF-8 import files. Charset is detected using icu4j:~~
~~Import of ISO-8859-1 files is correct in unit tests. Caution: testing with larger real world examples pending.~~

Icu4j does not reliably detect the charset of larger files, so the approach is dropped. Instead an approach to deterministically cover the actual problematic cases is chosen.

The charset to read csv input files is determined according to the following steps:

1. if the file can be read in using charset UTF-8 without `MalformedInputException`, it is assumed to be a valid UTF-8 file (a ISO-8859-1 file containing special characters leads to a `MalformedInputException`, whereas an ISO-8859-1 without special characters as well as ASCII-files can be read as UTF-8 without problems, i.e. no exception, no misinterpretation of the content) 
2. otherwise, the file is read as ISO-8859-1 file. This always succeeds, as ISO-8859-1 encoding is a single byte per character. The result is checked for possibly misread UTF-8 byte sequences (see https://www.i18nqa.com/debug/utf8-debug.html resp. file `utf8sequences.txt`. If none of these sequences is encountered, the file is assumed to be a valid ISO-8859-1 file.
3. If one of the usually misinterpreted sequences is found, the charset detection defaults to UTF-8. Processing the file will very likely result in a `MalformedInputException`, which however is reasonable: the file is not a valid UTF-8 file, but contains UTF-8 byte sequences, so it is most likely indeed malformed with respect to the two charsets in consideration.

Tests are present for the cases UTF-8, UTF-8 with BOM, ISO-8859-1 with large files.


sormas-ui/src/test/java/de/symeda/sormas/ui/importer/CountryImporterTest.java
```diff
@@ -30,29 +30,34 @@
...
-		getCountryFacade().getByDefaultName("Country with ä", false).get(0);
+		CountryReferenceDto countryReference = getCountryFacade().getByDefaultName("Country with ä", false).get(0);
+		assertEquals("XYZ", countryReference.getIsoCode());
 	}
 
-
-	@Test(expected = MalformedInputException.class)
-	public void testUmlautsInCountryImportNonUTF8() throws IOException, InvalidColumnException, InterruptedException, CsvValidationException, URISyntaxException {
+	@Test
+	public void testUmlautsInCountryImportNonUTF8()
```
vlad-ciucescu, line 46: I don't think this test is needed anymore, since you created tests for other specific encodings. "non UTF8" is a bit vague anyway.

syntakker: Yes. I will clean this up and complete the tests with import files in different encodings and also with larger files to make sure we don't fail for imports that already worked.

sormas-api/src/main/java/de/symeda/sormas/api/utils/CharsetHelper.java
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	public static boolean isCharsetUTF8(File inputFile) {
+		try (InputStream inputStream = Files.newInputStream(inputFile.toPath())) {
+			BOMInputStream bomInputStream = new BOMInputStream(inputStream);
+			if (bomInputStream.hasBOM()) {
+				return true;
+			}
+			Reader reader = new InputStreamReader(bomInputStream, StandardCharsets.UTF_8.newDecoder());
+			BufferedReader bufferedReader = new BufferedReader(reader);
```
vlad-ciucescu, line 64: You can add the `BOMInputStream`, the `Reader` and the `BufferedReader` to the try resources. Otherwise you have to close them manually.

vlad-ciucescu (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/utils/CharsetHelper.java
```diff
@@ -0,0 +1,95 @@
...
+			while (bufferedReader.readLine() != null);
```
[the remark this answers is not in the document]
syntakker: I measured the time for the charset detection in unit tests (1000 lines, file size ~26kb):

- `testLargeFileUtf8WithBOM`: 0ms (no further tests if BOM is present)
- `testLargeFileUtf8`: 1ms
- `testLargeFileISO8859_1`: 3ms

Line length is moderate, tested on my local machine.
However, given the frequency of data imports and the size of pure text files, I would estimate performance should not become an issue here. Or rather: charset detection will not be the bottleneck :-)

sormas-api/src/main/java/de/symeda/sormas/api/utils/CharsetHelper.java
```diff
@@ -0,0 +1,95 @@
...
+		}
+		if (isCharsetIso8859_1(inputFile)) {
+			return StandardCharsets.ISO_8859_1;
+		}
+		return StandardCharsets.UTF_8;
+	}
+
+	public static CharsetDecoder getDecoder(File inputFile) {
+		Charset charset = detectCharset(inputFile);
+		return charset != null ? charset.newDecoder() : StandardCharsets.UTF_8.newDecoder();
```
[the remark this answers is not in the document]
syntakker: Obsolete null check removed.

sormas-api/src/main/java/de/symeda/sormas/api/utils/CharsetHelper.java
```diff
@@ -0,0 +1,95 @@
...
+		} catch (IOException e) {
+			return false;
+		}
+	}
+
+	public static boolean isCharsetIso8859_1(File inputFile) {
+		try (InputStream inputStream = Files.newInputStream(inputFile.toPath())) {
+			Reader reader = new InputStreamReader(inputStream, StandardCharsets.ISO_8859_1.newDecoder());
+			BufferedReader bufferedReader = new BufferedReader(reader);
+			while (true) {
```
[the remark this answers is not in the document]
syntakker: Done.

sormas-api/src/main/java/de/symeda/sormas/api/utils/CharsetHelper.java
syntakker: Done.

vlad-ciucescu (approved):

PR #4655, opened 2021-03-05
validate event start date before end date (#4362)

### LINKED ISSUE #4362: Multiday event dates are not correctly validated in android app
### Bug Description

The validators for multiday events are faulty in the android app

### Steps to Reproduce
1. Log into the android app
2. Open Event directory
3. Create a new event
4. Select "Multiday Event" checkbox
5.1 Select dates in the future
5.2 select endDate<startDate

### Expected Behavior

Step 5.1 should not throw an error. Step 5.2 should throw an error

### System Details
* Device: Emulated Pixel2
* SORMAS version: 1.57.0-Snapshot
* Android version/Browser: Android 11.0 API 30
* Server URL:

### Additional Information

See reference for implementation here: https://github.com/hzi-braunschweig/SORMAS-Project/pull/4317

syntakker:
fixes #4362


sormas-app/app/src/main/java/de/symeda/sormas/app/event/edit/EventValidator.java
```diff
@@ -0,0 +1,14 @@
...
+import de.symeda.sormas.app.databinding.FragmentEventEditLayoutBinding;
+
+class EventValidator {
+
+	public static void initializeEventStartDateValidation(final Event event, final FragmentEventEditLayoutBinding contentBinding) {
+		if (event != null) {
+			ValidationHelper.initDateIntervalValidator(contentBinding.eventStartDate, contentBinding.eventEndDate);
+		}
+	}
+}
```
DavidBaldsiefen, line 14: Why did you create this new class, instead of just doing the nullcheck & adding the validator inline in `EventFragment.java`?

DavidBaldsiefen (changes requested):
Multiday Events still cannot be set into the future. It should be possible to add events which are entirely in the future

sormas-app/app/src/main/java/de/symeda/sormas/app/event/edit/EventValidator.java
syntakker: Inlined.

DavidBaldsiefen (changes requested):
I've checked again, and it turns out the `startDate` is _not_ allowed to be in the future, only the end date. My mistake, sorry! Once that line is removed it can be merged

PR #4678, opened 2021-03-08
#4519 update follow-up date in external journal when converting contact to case

### LINKED ISSUE #4519: [CLIMEDO Integration] Converting a contact to a case does not update the follow-up duration in Climedo
### Bug Description
After converting a confirmed contact to a case, the Climedo patient is not updated with the latest follow-up duration
### Steps to Reproduce
1. Register a person from a contact as a patient in Climedo
2. Make the contact confirmed.
3. Convert the contact to a case.

### Expected Behavior
The follow-up duration in Climedo matches that of the case

### Actual Behavior
The follow-up duration stays the same in Climedo

### Additional Information
Although it may seem to be a problem on Climedo, this can be a problem on the SORMAS side, since we notify Climedo of the update to the follow-up date

vlad-ciucescu:
closes #4519

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
@@ -29,6 +29,7 @@ import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
+import de.symeda.sormas.api.caze.CaseDataDto;
 import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.validator.routines.EmailValidator;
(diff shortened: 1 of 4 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java
```diff
@@ -185,20 +186,20 @@ private String getPatientDiaryAuthTokenInternal() {
...
-	 * @param contact
-	 *            a contact assigned to a person already available in the external journal
+	 * @param caze
+	 *            a case assigned to a person already available in the external journal
 	 * @param previousFollowUpUntilDate
 	 *            the follow-up end date before the update
 	 */
-	public void notifyExternalJournalFollowUpUntilUpdate(ContactDto contact, Date previousFollowUpUntilDate) {
-		PersonDto person = personFacade.getPersonByUuid(contact.getPerson().getUuid());
+	public void notifyExternalJournalFollowUpUntilUpdate(CaseDataDto caze, Date previousFollowUpUntilDate) {
```
leventegal-she, line 194: It does not really make too much sense to require the whole case dto for getting it's person uuid and followup date.

would be cleaner and more performant to have only the used values as function arguments: personUuid, followupUntilDate and prevFollowupUntilDate

leventegal-she (changes requested):

leventegal-she (approved):

Issue #4666, opened 2021-03-07
Sorting responsible and reporting user columns on event directory result in error


kwa20:
### Bug Description
Trying to sort the responsible and reporting user columns on the event directory results in error
### Steps to Reproduce
1. Go to the event directory
2. Sort the responsible and reporting user columns

### Expected Behavior
sorting should work fine.
### Screenshots
<img width="843" alt="userColumnEventsError" src="https://user-images.githubusercontent.com/70317594/110254248-b2fde780-7f8e-11eb-8e25-fbb3f3fa1930.PNG">

### System Details
* Device:
* SORMAS version: 1.57.1
* Android version/Browser: firefox
* Server URL: test1401.sormas-oegd.de
* User Role: admin, national user

raghupola-vg:
Verified on version 1.58.0-Snapshot and issue is Fixed.

PR #4685, opened 2021-03-08
#4666 - Fixed sorting by responsible and reporting user

alexcaruntu-vita:


sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventFacadeEjb.java
```diff
@@ -355,6 +356,12 @@ public long count(EventCriteria eventCriteria) {
 					order.add(sortProperty.ascending ? cb.asc(expression) : cb.desc(expression));
 					expression = location.get(Location.ADDITIONAL_INFORMATION);
 					break;
+				case EventIndexDto.REPORTING_USER:
```
MateStrysewske, line 359: Please also sort by last name, similar to how it's done in `ContactFacadeEjb:getContactFollowUpList` (currently lines 877ff on the dev branch) for both reporting user and responsible user

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #4695, opened 2021-03-09
[#4507] tweak task reassignment on case change

### LINKED ISSUE #4507: Cases > Case > Changing the "Facility" results in reassignment of an existing Task to a random other user
Updated definition: https://github.com/hzi-braunschweig/SORMAS-Project/issues/4507#issuecomment-793539839

### Bug Description
Changing `Facility` for Cases assigns a random new `Assigned to:` value for Tasks associated to the Case.

### Steps to Reproduce
1. Cases > Open/Create Case1 > 
2. Create Task for Case1 and assign it to "User A".
3. Case1 > Set ` Place of Stay` to be "Facility", `Facility Category` as "Medical Facility", `Facility Type` as "Hospital", and select a value from `Facility` (from the list defined in the Configuration/Infrastructure data).
4. Save [Works fine]
5. Case1 > Change value from `Facility` (from the list defined in the Configuration/Infrastructure data) to another value > Prompt appears to change infrastructure data > Agree
6. Task associated with Case1 is now randomly assigned to "User X" (and not "User A" as it is supposed to stay).

### Expected Behavior
Task assignment should not change.

### Screenshots
image
image
image

### System Details
* Device: Web
* SORMAS version: 1.56.1
* Server URL: https://test1401.sormas-oegd.de/ (Majatovic)
* User Role: admin

### Additional Information

JonasCir:
Closes #4507

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -1787,13 +1787,19 @@ public class CaseFacadeEjb implements CaseFacade {
 			boolean regionChanged = !existingCase.getRegion().getUuid().equals(newCase.getRegion().getUuid());
 			boolean districtChanged = !existingCase.getDistrict().getUuid().equals(newCase.getDistrict().getUuid());
 
-			boolean communityChanged = existingCase.getCommunity() != null
-				&& newCase.getCommunity() != null
-				&& !newCase.getCommunity().getUuid().equals(existingCase.getCommunity().getUuid());
+			// check if infrastructure was changed, added, or removed from the case
 
-			boolean facilityChanged = existingCase.getHealthFacility() != null
+			boolean communityChanged = (existingCase.getCommunity() != null
+				&& newCase.getCommunity() != null
+				&& !newCase.getCommunity().getUuid().equals(existingCase.getCommunity().getUuid()))
+				|| (existingCase.getCommunity() == null && newCase.getCommunity() != null)
+				|| (existingCase.getCommunity() != null && newCase.getCommunity() == null);
+
+			boolean facilityChanged = (existingCase.getHealthFacility() != null
 				&& newCase.getHealthFacility() != null
-				&& !existingCase.getHealthFacility().getUuid().equals(newCase.getHealthFacility().getUuid());
+				&& !existingCase.getHealthFacility().getUuid().equals(newCase.getHealthFacility().getUuid()))
+				|| (existingCase.getHealthFacility() == null && newCase.getHealthFacility() != null)
+				|| (existingCase.getHealthFacility() != null && newCase.getHealthFacility() == null);
 
 			if (regionChanged || districtChanged || communityChanged || facilityChanged) {
 				reassignTasksOfCase(newCase, false);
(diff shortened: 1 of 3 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1774,6 +1774,34 @@ public void archiveOrDearchiveCase(String caseUuid, boolean archive) {
...
+	 */
+	private void updateTasksOnCaseChanged(Case newCase, CaseDataDto existingCase) {
+		// In case the health facility of a case changes
+		// check if a new surveillance officer for this facility was also assigned.
+		// If yes re-assign the tasks associated with this case to this SurvOff
+		if (existingCase != null
+			&& newCase.getHealthFacility() != null
+			&& existingCase.getHealthFacility() != null
+			&& !newCase.getHealthFacility().getUuid().equals(existingCase.getHealthFacility().getUuid())
+			&& newCase.getSurveillanceOfficer() != null
```
JonasCir, line 1789: hm I guess this can be removed?

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1774,6 +1774,34 @@ public void archiveOrDearchiveCase(String caseUuid, boolean archive) {
...
+	 * a case has been created/saved
+	 */
+	private void updateTasksOnCaseChanged(Case newCase, CaseDataDto existingCase) {
+		// In case the health facility of a case changes
+		// check if a new surveillance officer for this facility was also assigned.
+		// If yes re-assign the tasks associated with this case to this SurvOff
+		if (existingCase != null
+			&& newCase.getHealthFacility() != null
+			&& existingCase.getHealthFacility() != null
+			&& !newCase.getHealthFacility().getUuid().equals(existingCase.getHealthFacility().getUuid())
```
JonasCir, line 1788: alternatively I could just drop this check implies the tasks to be reassigned when a new officer working at the facility of the case is assigned? I think this might be even better?

MateStrysewske: Not sure whether I understand you correctly, and I'm also not sure whether I properly understand the logic you've implemented. 

I think what I would do here is to keep the current logic that triggers `reassignTasks` to be called, and extend `reassignTasks` to only actually reassign tasks when the case's jurisdiction doesn't match the jurisdiction of the currently assigned user anymore. Since `reassignTasks` is also used to reassign tasks from users for which the user roles have changed, it would also be necessary to add a `forceReassignment` paramter to that method that is `false` if it's called from `onCaseChanged` and `true` if called from `UserFacadeEjb:removeUserAsSurveillanceAndContactOfficer`.

What do you think about that proposal?

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2037,16 +2058,53 @@ public void setResponsibleSurveillanceOfficer(Case caze) {
...
+			boolean mismatch;
+
+			if (taskAsignee == null) {
+				// no one is assigned so we skip detailed checks and go directly reassignment.
+				mismatch = true;
+			} else {
+				boolean regionMismatch = !taskAsignee.getRegion().getUuid().equals(caze.getRegion().getUuid());
+				boolean districtMismatch = !taskAsignee.getDistrict().getUuid().equals(caze.getDistrict().getUuid());
+
+				boolean communityMismatch = taskAsignee.getCommunity() != null
```
MateStrysewske, line 2088: For community and health facility, the situation where 
* the user has a community/facility
* the case does not have a community/facility

also has to count as a mismatch (because it means the case is no longer associated with their jurisdiction)

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2037,16 +2058,53 @@ public void setResponsibleSurveillanceOfficer(Case caze) {
...
 			}
 
-			assignOfficerOrSupervisorToTask(caze, task);
+			User taskAsignee = task.getAssigneeUser();
+			boolean mismatch;
+
+			if (taskAsignee == null) {
+				// no one is assigned so we skip detailed checks and go directly reassignment.
+				mismatch = true;
+			} else {
```
MateStrysewske, line 2084: Use `else if (!forceReassignment)` because these calculations don't matter if that's set to true

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1774,6 +1774,41 @@ public void archiveOrDearchiveCase(String caseUuid, boolean archive) {
...
+	private void updateTasksOnCaseChanged(Case newCase, CaseDataDto existingCase) {
+		// In case that *any* jurisdiction of the case has been changed, we need to see if we need to reassign related tasks.
+		// Tasks can be assigned to various user roles and users, therefore it is crucial to make sure
+		// that no tasks related to cases are assigned to officers lacking jurisdiction on the case.
+
+		if (existingCase != null) {
+			boolean regionChanged = !existingCase.getRegion().getUuid().equals(newCase.getRegion().getUuid());
+			boolean districtChanged = !existingCase.getDistrict().getUuid().equals(newCase.getDistrict().getUuid());
+
+			boolean communityChanged = existingCase.getCommunity() != null
```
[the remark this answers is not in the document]
JonasCir: good catch!

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1774,6 +1774,41 @@ public void archiveOrDearchiveCase(String caseUuid, boolean archive) {
...
+				&& newCase.getCommunity() != null
+				&& !newCase.getCommunity().getUuid().equals(existingCase.getCommunity().getUuid());
+
+			boolean facilityChanged = existingCase.getHealthFacility() != null
```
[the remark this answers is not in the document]
JonasCir: done

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2037,16 +2058,53 @@ public void setResponsibleSurveillanceOfficer(Case caze) {
...
+	public void reassignTasksOfCase(Case caze, boolean forceReassignment) {
+		// for each task that is related to the case, the task assignee must match the jurisdiction of the case
+		// otherwise we will reassign the task
 		for (Task task : caze.getTasks()) {
 			if (task.getTaskStatus() != TaskStatus.PENDING) {
 				continue;
 			}
 
-			assignOfficerOrSupervisorToTask(caze, task);
+			User taskAsignee = task.getAssigneeUser();
```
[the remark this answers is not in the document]
JonasCir: fixed typo

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
JonasCir: added :)

MateStrysewske (approved):

Issue #4489, opened 2021-02-22
[SurvNet Interface] The autogenerated external Token should be unique [2]


kwa20:
### Feature Description
External Token auto generation during transmission to SurvNet
### Problem Description
Thus far, the auto generated Token is not being compared to other tokens in the system to prevent duplicate Tokens. Usually, the external Token should be unique since it's used as the main reference for communication with the higher health institutions.
### Change

- If the user leaves the field of the duplicate token, the red frame, exclamation point and mouse over should persist.
- Add validation for unique token to externalToken
- Validation is not blocking the save functionality

=> Re-Use EPID-Number validation

- If the user leaves the field of the duplicate token, the red frame, exclamation point and mouse over should persist.
- Add validation for unique token to externalToken
- Validation is not blocking the save functionality

In Scope:

- [ ] Persons 
- [ ] cases
- [ ] contacts
- [ ] events
- [ ] event participants

Out of scope:
Add the functionality to Android-App

<img width="853" alt="Screen Shot 2021-02-22 at 9 23 31 AM" src="https://user-images.githubusercontent.com/70317594/108682108-354ace00-74f0-11eb-8258-2885bb6c2676.png">

### Additional Information

- See `CaseDataForm.setEpidNumberError` how the component error is displayed (it's not `com.vaadin.ui.Validator`)

JonasCir:
I guess this was caused by updating the converter, but leaving the rest of the system as it was? If you bump the converter to a new version, the counter likely gets reset. I think this is not a SORMAS issue in particular. I would suggest to check this in the converter or make sure that the update process does not break assumptions

JonasCir:
Even if you flash the message, there is nothing the user could do about it as the correct solution would be to connect to the converter DB and set the counter to the appropriate value

kwa20:
@JonasCir That might be true for the autogeneration, but it's at least attracting attention so the user is aware that a duplicate exists and may edit it manually if desired.

JonasCir:
I get your point :) The problem is that even if he notice it, he cannot do anything about it. A manual fallback won't work as the user then need to coordinate how to chose the tokens. The only way to handle this properly is to make sure that upgrades are properly executed and the DB entry of the counter is not compromised

markusmann-vg:
As the token cannot be unique over different systems, how do we proceed when a case was send via SORMAS 2 SORMAS and the Token is already taken.

JonasCir:
[...]

PR #4711, opened 2021-03-10
#4489 [SurvNet Interface] The autogenerated external Token should be unique

leventegal-she:
Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/4489


sormas-ui/src/main/java/de/symeda/sormas/ui/utils/ValidationUtils.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import com.vaadin.server.UserError;
+import com.vaadin.ui.Label;
+import com.vaadin.v7.ui.TextField;
+
+import de.symeda.sormas.api.i18n.I18nProperties;
+
+public class ValidationUtils {
+
```
ssferrazza, line 29: Please add a private constructor to classes containing only static utility methods.

ssferrazza, line 29: Please add a private constructor to classes containing only static utility methods.

ssferrazza (changes requested):

ssferrazza (approved):

PR #4752, opened 2021-03-12
#4582 Allow users to assign tasks to users of the next higher jurisdictionlevel

### LINKED ISSUE #4582: Allow all user roles to assign tasks to users with a higher jurisdiction

DavidBaldsiefen:
Closes #4582

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserFacadeEjb.java
@@ -46,6 +46,7 @@ import org.apache.commons.beanutils.BeanUtils;
 
 import de.symeda.sormas.api.region.DistrictReferenceDto;
 import de.symeda.sormas.api.region.RegionReferenceDto;
+import de.symeda.sormas.api.user.JurisdictionLevel;
 import de.symeda.sormas.api.user.UserCriteria;
 import de.symeda.sormas.api.user.UserDto;
 import de.symeda.sormas.api.user.UserFacade;
@@ -82,6 +83,7 @@ import de.symeda.sormas.backend.user.event.PasswordResetEvent;
 import de.symeda.sormas.backend.user.event.UserCreateEvent;
 import de.symeda.sormas.backend.user.event.UserUpdateEvent;
 import de.symeda.sormas.backend.util.DtoHelper;
+import de.symeda.sormas.backend.util.JurisdictionHelper;
 import de.symeda.sormas.backend.util.ModelConstants;
 
 @Stateless(name = "UserFacade")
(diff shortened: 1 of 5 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/util/JurisdictionHelper.java
```diff
@@ -262,4 +263,23 @@ public static EventParticipantJurisdictionDto createEventParticipantJurisdiction
...
+		case NATION:
+			return JurisdictionLevel.NONE;
+		case REGION:
+			return JurisdictionLevel.NATION;
+		case DISTRICT:
+			return JurisdictionLevel.REGION;
+		case COMMUNITY:
+		case POINT_OF_ENTRY:
+		case HEALTH_FACILITY:
+		case LABORATORY:
```
MateStrysewske, line 278: Region and district aren't mandatory for laboratories, so I'd rather give them jurisdiction level NONE; external laboratories are supposed to be outside the country, so they should also receive NONE

sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserFacadeEjb.java
```diff
@@ -129,6 +131,50 @@
...
+
+			District district = null;
+			List<UserRole> superordinateRoles = UserRole.getWithJurisdictionLevels(superordinateJurisdiction);
+			if (user.getDistrict() != null) {
+				district = districtService.getByReferenceDto(user.getDistrict());
+			} else if (user.getHealthFacility() != null) {
+				district = facilityService.getByReferenceDto(user.getHealthFacility()).getDistrict();
+				superordinateRoles.addAll(UserRole.getWithJurisdictionLevels(JurisdictionLevel.COMMUNITY));
+			} else if (user.getLaboratory() != null) {
+				district = facilityService.getByReferenceDto(user.getLaboratory()).getDistrict();
```
MateStrysewske, line 164: Laboratories don't necessarily have a district; corresponding to the change request below, I think they should just be excluded from this logic for now

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserFacadeEjb.java
```diff
@@ -129,6 +131,47 @@
...
+	 */
+	public List<UserReferenceDto> getUsersWithSuperiorJurisdiction(UserDto user) {
+		JurisdictionLevel superordinateJurisdiction =
+			JurisdictionHelper.getSuperordinateJurisdiction(UserRole.getJurisdictionLevel(user.getUserRoles()));
+
+		List<User> superiorUsersList;
+
+		switch (superordinateJurisdiction) {
+		case NATION:
+			return null;
```
MateStrysewske, line 147: The list actually has to contain all national-level users in this case - otherwise this code breaks the change concerning allowing region-level users to assign tasks to nation-level users

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserFacadeEjb.java
DavidBaldsiefen: Can we assume there is always only one nation per SORMAS instance? So We do not need a method named `getAllByNationAndUserRoles`

MateStrysewske: Right now that's the case, yes

MateStrysewske (approved):

PR #4753, opened 2021-03-12
Add a button to unlink cases from events #2984

### LINKED ISSUE #2984: Add an icon to unlink cases from events
### Feature Description
When a case is mistakenly linked to the event, it is impossible to break this link without deleting the case. There are users who have retrospectively and linked cases to wrong events and are blocked from correcting this link. They can not also delete the case because it has been for long in the system and has many contacts
### Proposed Change
- Add a delete icon on the case form to break the association between the case and event 
- Add a description of what  this icon means when you hover the mouse on it : "Unlink this case as resulting from this event" (this is optional. We can also use an information box saying "Unlink event", by this way we need no description. This second option may not look good design wise since there is a pencil icon by the side)
- When users click on the unlink icon, show a dialog box saying "Are you sure you want to unlink this case from this event" : yes, no, cancel. 
- If yes, then break the link but maintain the case and event participant in their respective tables. This is a similar method that is called when the case associated to an event is deleted, but in this case, we do not delete the case that was initially set at resulting from event. 
- Please also hard delete the resulting case id on the event participant table or the other way round.
- Show a message saying "Case was was successfully unlinked from event"
### Possible Alternatives
- Will add more points soon, need to run

### Additional Information

image

mathieu-petit-atol:
Closes #2984


sormas-api/src/main/resources/strings.properties
```diff
@@ -709,6 +709,7 @@ messageEventLinkedAsSuperordinate = The selected event was successfully linked t
 messageEventLinkedAsSubordinate = The selected event was successfully linked to this event as a subordinate event
 messageEventSuperordinateEventUnlinked = The link between this event and its superordinate event was successfully removed
 messageEventSubordinateEventUnlinked = The link between this event and its subordinate event was successfully removed
+messageEventParticipationUnlinked = The link between this cases and the event was successfully removed
```
MateStrysewske, line 712: this cases => this case

sormas-api/src/main/resources/strings.properties
```diff
@@ -995,6 +996,9 @@ DiseaseNetworkDiagram.lowRisk = Low risk
 DiseaseNetworkDiagram.selectByClassification = Select by Classification
 DiseaseNetworkDiagram.subheading = The arrows indicate the direction of transmission
 
+# Unlink events
+unlinkPopupCaption = Are you sure you want to unlink this case from this event ?
```
MateStrysewske, line 1000: Change the property name to `confirmationUnlinkCaseFromEvent`

MateStrysewske (changes requested):
I feel like the PR is incomplete; at least I can't find any code that adds the icon that can be clicked to unlink cases from events. Is there something I'm missing?

mathieu-petit-atol:
> I feel like the PR is incomplete; at least I can't find any code that adds the icon that can be clicked to unlink cases from events. Is there something I'm missing?

It's done with the `addUnlinkEventListener` variable ; when you set this variable it's automatically show the unlink button with the function `drawDisplayedEntries` in the `EventList` class.

MateStrysewske:
> > I feel like the PR is incomplete; at least I can't find any code that adds the icon that can be clicked to unlink cases from events. Is there something I'm missing?
> 
> It's done with the `addUnlinkEventListener` variable ; when you set this variable it's automatically show the unlink button with the function `drawDisplayedEntries` in the `EventList` class.

Ah, sorry! Stupid me! :D Thanks for the clarification.

sormas-api/src/main/resources/strings.properties
```diff
@@ -995,6 +996,9 @@ DiseaseNetworkDiagram.lowRisk = Low risk
 DiseaseNetworkDiagram.selectByClassification = Select by Classification
 DiseaseNetworkDiagram.subheading = The arrows indicate the direction of transmission
 
+# Unlink events
+confirmationUnlinkCaseFromEvent = Are you sure you want to unlink this case from this event ?
```
MateStrysewske, line 1000: Move the property to the section in this file that contains the other confirmations and remove the whitespace before the question mark

MateStrysewske (changes requested):
The popup currently has the same heading and caption; heading should be shorter, e.g. "Unlink case from event"

sormas-api/src/main/resources/strings.properties
```diff
@@ -181,6 +181,8 @@ confirmationRemoveGridRowCancel=No
 confirmationSuperordinateEventDiscardUnsavedChanges = Linking or unlinking a superordinate event will discard all unsaved changes you have made to this event. Are you sure you want to continue?
 confirmationLocationFacilityAddressOverride = The selected facility has location details that are different from the ones you are currently editing. Do you want to overwrite them with those from the facility?
 confirmationCancelExternalFollowUpPopup=Are you sure you want to cancel external follow-ups in the eDiary?
+confirmationUnlinkCaseFromEvent = Are you sure you want to unlink this case from this event?
+confirmationUnlinkCaseFromEventHeader = Unlink case from event
```
MateStrysewske, line 185: Call this headingUnlinkCaseFromEvent and move it to the heading section, please

MateStrysewske (changes requested):

MateStrysewske:
@mathieu-petit-atol Thanks! Sorry for the inconvenience, but there's another merge conflict (that probably came in today) - can you fix that again? Apart from that, the PR is ready to be merged.

mathieu-petit-atol:
Done

MateStrysewske (approved):

Issue #4668, opened 2021-03-08
[SurvNet Interface] Events > Add new field "Internal ID" [0.5]


Candice-Louw:
### Situation Description
The concept of generating an automatic "Herd Identification number" for outbreaks based on international standards is required.

### Feature Description
Events > Open Event > Introduce "Internal ID" (en-) / "Name im Gesundheitsamt" (de-) field [DE only]:
![image](https://user-images.githubusercontent.com/65529128/110285721-f6367580-7fe3-11eb-9714-739c2c7c1783.png)

Initially - this field will be a free-text and filled in by hand.

### Additional Information
SurvNet Mapping (Field Tab):

IdField | FieldName | Code | Schema2 | Schema3 | Schema4 | IdParent | DataTypeName | FieldTypeName | Sort | XmlName | GuiText | GuiTextShort
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
202 | NameGA | Outbreak | 1 | 1 | 1 | 0 | String255 | Simple | 2 | NameGA | Name im Gesundheitsamt | NameGA

![image](https://user-images.githubusercontent.com/65529128/110293084-6649f900-7fee-11eb-93bd-e13d357e921d.png)

AndyBakcsy-she:
Verifid ticket on the latest version of sormas 1.58.0-snapshot on web

PR #4765, opened 2021-03-15
#4668 [SurvNet Interface] Events > Add new field "Internal ID"

leventegal-she:
Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/4668


sormas-backend/src/main/java/de/symeda/sormas/backend/event/Event.java
```diff
@@ -629,4 +631,13 @@ public void setSormasToSormasOriginInfo(SormasToSormasOriginInfo originInfo) {
 	public void setSormasToSormasShares(List<SormasToSormasShareInfo> sormasToSormasShares) {
 		this.sormasToSormasShares = sormasToSormasShares;
 	}
+
+	@Column(columnDefinition = "text")
```
vidi42, line 635: The ID in SurvNet (`NameGA`) seems to be limited to a 255 characters string. Having the internalId as a "text" data type in SORMAS might end up in issues when sending events to SurvNet.

vidi42 (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/event/Event.java
leventegal-she: I'm not sure this needs to be handled in SORMAS, I would say it's the responsibility of the converter to handle this kind of issues.

In SORMAS we have varchar(255) only for enums, for free text we have varchar(512) or text.

@MateStrysewske what do you think, should the size of this field be limited to 255 in SORMAS?

MateStrysewske: @leventegal-she No, I don't think so. I guess the danger of someone entering a value here that is too long is pretty low anyway. Moreover, if this limitation changes at any point in the future, we'll have troubles with removing it in the mobile app, so I'd rather have it as a text right from the start.

vidi42 (approved):

PR #4780, opened 2021-03-15
[#4218] add properties and make sure they are not null

### LINKED ISSUE #4218: Introduce disease properties to switch between basic and extended classification [3]
### Situation Description
The NCDC Data Transition team in accordance with the **WHO** standards has requested that we add the following case classifications under **Measles** Disease 

- Clinically confirmed
- Epidemiologically linked.

### Feature Description
In addition to the case classifications currently available under **Measles** (NOT YET CLASSIFIED, SUSPECT CASE, PROBABLE CASE, CONFIRMED CASE, NOT A CASE), a request has been made to add two more; **Clinically confirmed** and **Epidemiologically linked**.

The automatic case classification by the system would be put on hold for the disease (Measles) as a clear algorithm for the selection is yet to be achieved instead users having the right user privilege should be able to manually select case classifications.

### Additional Information

JonasCir:
closes #4218

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseEditFragment.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseEditFragment.java
@@ -71,7 +71,6 @@ import de.symeda.sormas.app.backend.classification.DiseaseClassificationAppHelpe
 import de.symeda.sormas.app.backend.classification.DiseaseClassificationCriteria;
 import de.symeda.sormas.app.backend.common.DatabaseHelper;
 import de.symeda.sormas.app.backend.config.ConfigProvider;
-import de.symeda.sormas.app.backend.disease.DiseaseConfiguration;
 import de.symeda.sormas.app.backend.disease.DiseaseVariant;
 import de.symeda.sormas.app.component.Item;
 import de.symeda.sormas.app.component.controls.ControlPropertyField;
@@ -224,11 +223,11 @@ public class CaseEditFragment extends BaseEditFragment<FragmentCaseEditLayoutBin
 	private void updateCaseConfirmationVisibility(FragmentCaseEditLayoutBinding contentBinding) {
 
 		if (record.getCaseClassification() == CaseClassification.CONFIRMED) {
-			Disease disease = record.getDisease():
-			boolean extendedCalssification = DiseaseConfigurationCache.usesExtendedClassification(disease);
+			Disease disease = record.getDisease();
+			boolean extendedCalssification = DiseaseConfigurationCache.getInstance().usesExtendedClassification(disease);
 
 			if (extendedCalssification) {
-				boolean extendedCalssificationMulti = DiseaseConfigurationCache.usesExtendedCalssificationMulti(disease);
+				boolean extendedCalssificationMulti = DiseaseConfigurationCache.getInstance().usesExtendedClassificationMulti(disease);
 				if (extendedCalssificationMulti) {
 					contentBinding.caseDataClinicalConfirmation.setVisibility(VISIBLE);
 					contentBinding.caseDataEpidemiologicalConfirmation.setVisibility(VISIBLE);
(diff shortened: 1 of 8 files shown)


sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseEditFragment.java
```diff
@@ -224,9 +223,12 @@ private void setUpFieldVisibilities(final FragmentCaseEditLayoutBinding contentB
 	private void updateCaseConfirmationVisibility(FragmentCaseEditLayoutBinding contentBinding) {
 
 		if (record.getCaseClassification() == CaseClassification.CONFIRMED) {
-			DiseaseConfiguration diseaseConfiguration = DatabaseHelper.getDiseaseConfigurationDao().getDiseaseConfiguration(record.getDisease());
-			if (diseaseConfiguration.getExtendedClassification()) {
-				if (diseaseConfiguration.getExtendedClassificationMulti()) {
+			Disease disease = record.getDisease();
+			boolean extendedCalssification = DiseaseConfigurationCache.getInstance().usesExtendedClassification(disease);
```
MateStrysewske, line 227: Typo (below as well)

sormas-app/app/src/main/java/de/symeda/sormas/app/util/DiseaseConfigurationCache.java
```diff
@@ -43,6 +43,10 @@
 	private List<Disease> caseBasedDiseases = new ArrayList<>();
 	private List<Disease> aggregateDiseases = new ArrayList<>();
 	private List<Disease> followUpEnabledDiseases = new ArrayList<>();
+
+	private Map<Disease, Boolean> extendedClassificationDiseases = new HashMap<>();
```
MateStrysewske, line 47: I don't think we need maps here, lists that contains all diseases using the extended classifications should be enough and can then be used with `contains`

sormas-backend/src/main/java/de/symeda/sormas/backend/disease/DiseaseConfiguration.java
```diff
@@ -32,7 +32,12 @@
 
 	public static DiseaseConfiguration build(Disease disease) {
 		DiseaseConfiguration configuration = new DiseaseConfiguration();
+
 		configuration.setDisease(disease);
+		configuration.setExtendedClassification(disease.isDefaultExtendedClassification());
```
MateStrysewske, line 37: This should be removed - the system should be able to deal with null values because we have the defaults

sormas-backend/src/main/java/de/symeda/sormas/backend/disease/DiseaseConfigurationFacadeEjb.java
```diff
@@ -44,6 +44,10 @@
 	private List<Disease> caseBasedDiseases = new ArrayList<>();
 	private List<Disease> aggregateDiseases = new ArrayList<>();
 	private List<Disease> followUpEnabledDiseases = new ArrayList<>();
+
+	private Map<Disease, Boolean> extendedClassificationDiseases = new EnumMap<>(Disease.class);
```
MateStrysewske, line 48: See above - use lists instead of maps

MateStrysewske (changes requested):
CaseReadFragment is missing (still uses getDiseaseConfiguration and fails if the configuration is null)

sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseEditFragment.java
JonasCir: fixed

sormas-app/app/src/main/java/de/symeda/sormas/app/util/DiseaseConfigurationCache.java
JonasCir: changed to list with contains

JonasCir: actually we stick to maps, thank you @MateStrysewske :)

sormas-backend/src/main/java/de/symeda/sormas/backend/disease/DiseaseConfiguration.java
JonasCir: you are correct, in earlier versions this was mandatory but now with properties it works :)

MateStrysewske (approved):

PR #4794, opened 2021-03-16
#4742 display raw xml when lab message cannot be converted

### LINKED ISSUE #4742: [DEMIS Adapter] Return XML representation of a labMessage when conversion to HTML fails
### Feature Description

The DEMIS-Adapter provides a method to convert labMessages to viewable format. This method parses the original demis notification (represented as XML) and returns a html representation of it.

### Problem Description

When something goesg wrong in this process, the adapter returns an error. An error message is then showed to the user in the SORMAS-UI. Like that, when something goes wrong, the user can just not open a labMessage to find out more about it, which blocks the processing.

### Proposed Change

Some users dealt with XML-representation of demis notifications before. If the conversion of a DEMIS notification fails, it would be better to return its XML-representation instead of just returning an error message. Like that, users could at least see some (ugly formatted) information about the labMessage instead of not being able to see any details of a labMessage at all.

vlad-ciucescu:
closes #4742


FredrikSchaeferVitagroup:
When viewing an unconverted labMessage, the vertical scrollbar is hidden when one has not scrolled down completely. That's not so nice. Maybe setting the panels height to 100% (e.g. panel.setHeightFull()) might help in this case.

FredrikSchaeferVitagroup:
Also, when procesing a labMesage, the popupWindows shows up behind the currently opened window. So you process a message with unconverted labMessageDetails (and no error message) and after that you see the popupWindow statig that the labMessage could not be converted. That should not be so.

vlad-ciucescu:
The problem is that the vertical scroll bar belonged to the layout and the horizontal scroll bar belonged to the panel inside the layout. The only solution which I could find was to make the panel fixed height, so both scrollbars are displayed in the panel.

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageEditForm.java
```diff
@@ -52,7 +52,7 @@ protected void addFields() {
 		addFields(LabMessageDto.UUID, LabMessageDto.MESSAGE_DATE_TIME);
 		labMessageDetails = new Label();
 		Panel detailsPanel = new Panel(labMessageDetails);
-		detailsPanel.setWidth(550, Unit.PIXELS);
+		detailsPanel.setHeight(500, Unit.PIXELS);
```
FredrikSchaeferVitagroup, line 55: I would propose to leave the fixed width.
How did you get to setting the height to 500 pixels?  I would prefer if the heigt was set to 100% (which will relate to available screen height)...

FredrikSchaeferVitagroup (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageEditForm.java
vlad-ciucescu: If I set the height to 100% then the panel displays the entire messages, without a vertical scroll bar, and the scroll bar of the layout will be used. In this case, we have to scroll to the bottom to see the horizontal scroll bar of the panel.

FredrikSchaeferVitagroup: Does every element in the hierarchy above the panel have the height set to 100%. That is necessary for the panel's 100% to work...

FredrikSchaeferVitagroup (approved):

PR #4798, opened 2021-03-17
#4050 Fix FileReader compilation issue for java versions earlier than 11

### LINKED ISSUE #4050: umlauts and country variables produce errors during imports

vidi42:
Fixes #4050

### DIFF
--- a/sormas-api/src/test/java/de/symeda/sormas/api/utils/CharsetHelperTest.java
+++ b/sormas-api/src/test/java/de/symeda/sormas/api/utils/CharsetHelperTest.java
@@ -22,7 +22,6 @@ import static org.junit.Assert.assertTrue;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.net.URISyntaxException;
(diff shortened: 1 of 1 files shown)


StefanKock:
SORMAS does not support JDK lower than 11 see [here](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/SERVER_SETUP.md), why is this an issue?

vidi42:
In that case we need to adjust the Nightly builds which seem to be using JDK 8

StefanKock:
Maybe I was too quick: Since we need to support Java 7 on Android and this is part of sormas-api, the change is likely be needed / consistent.

StefanKock:
To which ticket relates this (missing ticket Id in commit)

syntakker:
It relates to issue #4050.

syntakker (approved):

Issue #4771, opened 2021-03-15
Show "sent to SurvNet" including the last Date of sending bellow the Send to SurvNet Button [2]


vidi42:
### Situation Description
Extracted from #3947 

### Feature Description
Add history of SurvNet transfers on cases and events.

### Additional Information

![Screenshot 2021-03-17 at 11 01 08](https://user-images.githubusercontent.com/62599627/111441672-35657480-8710-11eb-9261-7d6b1cb587b1.png)

Entities sharable with SurvNet at the moment:
* Cases
* Events

New data model:
#### `ExternalShareInfo`
naming is generic as described in #3947
> @markusmann-vg : In UI "SurvNet" should be replaced with "Reporting-SW"

| Column       | Type        |
|--------------|-------------|
| id           | bigint      |
| uuid         | varchar(36) |
| creationdate | timestamp   |
| changedate   | timestamp   |
| caze_id      | bigint      |
| event_id     | bigint      |
| sender_id    | bigint      |
| status         | varchar(255)    |
* each share will create a new entry in the `externalshareinfo` table which can then be displayed in a history view:
  * Row content: ```User X transferred at <date and time>```
  * Ordered by `creationDate` descending - the last one on the top
* the structure is similar to the `sormastosormasshareinfo` for easier refactoring in the future

MartinWahnschaffe:
@kwa20 @JaquM We have adjusted the mockup a bit, to make it more generic and re-usable for other purposes. Please have a quick look at it on the top of the page to confirm whether this is ok.

kwa20:
@MartinWahnschaffe Looks fine for me. I'm just wondering what it will look like if a case has been sent e.g. 10 times. Will there be a page break at some point?

leventegal-she:
@kwa20 It is a paginated list, so it will display the last 5 transfers and the pagination buttons like on the samples list.

vidi42:
Found 2 issues:

1. If a user which doesn't have jurisdiction to view the user's data (it's pseudonymized) access a case the server crashes with a NPE and the `Share to Reporting Tool` box is hidden.
This doesn't seem to be a big issue, because the SurvNet converter doesn't process anyway pseudonymized cases however we should prevent the NPE.
```
java.lang.NullPointerException
	at de.symeda.sormas.ui.externalsurveillanceservice.ExternalShareInfoList$ExternalShareInfoListEntry.<init>(ExternalShareInfoList.java:94)
[...]

PR #4800, opened 2021-03-17
#4771 Show "sent to SurvNet" including the last Date of sending bellow the Send to SurvNet Button

leventegal-she:
Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/4771


sormas-api/src/main/resources/captions.properties
```diff
@@ -2233,11 +2233,11 @@ sormasToSormasSendLabMessage=Send to another organization
 BAGExport=BAG Export
 
 # Survnet Gateway
-SurvnetGateway.title=SurvNet
-SurvnetGateway.send=send to own SurvNet instance
-SurvnetGateway.unableToSend=Unable to send
-SurvnetGateway.confirmSend=Confirm sending
-SurvnetGateway.sendShort=Send to SurvNet
+ExternalSurveillanceToolGateway.title=Reporting Software
```
vidi42, line 2236: I think we should rename it to either `Reporting System` or `Reporting Tool` because this wording it's too long for the UI.
image
Ideally would be to adjust the UI for any wording length, but I would be ok with just the renaming also.

vidi42 (changes requested):

vidi42 (approved):

PR #4802, opened 2021-03-17
Feature 4564 add checkbox probable infection environment to exposures

### LINKED ISSUE #4564: [SurvNet Interface] Add checkbox "probable infection environment" to exposures [2]
### Situation Description
After consultation with the RKI, the InfectionEnvironmentSetting in SurvNet and the exposures in Sormas differ conceptually from each other by either asking the user to document ALL relevant exposures during the incubation period or only the probable ones, where the user is very sure, that this infection setting led to the infection.
### Feature Description

- Add a checkbox called "probable infection environment" to individual exposures in epiData.
- Add info "mouse overs" to elaborate on the field on the main form as well as the individual checkboxes.
- Make this field and the info mouse over DE only.
- Show the mouse over info next to exposure details known only when the SurvNet interface is configured.
- Show the last sentence in the info box in exposures only when the SurvNet interface is configured.

### Mapping

**Export:**
- When the checkbox is checked, transmit the exposure as an infectionEnvironmentSetting entry to SurvNet.

**Import:**
- During the initial import, all transmitted data from SurvNet need to have the box checked in order to allow resending it via the export.
- if exposures get merged in any way, checking this box should always be favoured over a non-checked box.

### Additional Information
Screen Shot 2021-02-27 at 7 27 41 PM
Screen Shot 2021-03-09 at 1 22 25 PM

FredrikSchaeferVitagroup:
closes #4564


sormas-api/src/main/resources/strings.properties
```diff
@@ -541,6 +541,7 @@ infoCanceledBy = Canceled by %s using bulk action
...
 infoCaseIncidence = "Case incidence proportion" means the number of cases per 100,000 inhabitants. You can check the map key to see the thresholds that define how the districts are colorized.
 infoCaseMap = If cases are shown by home address and there are no GPS coordinates available for it, the coordinates of the location where the case has been reported are used instead.
+infoCheckPropableInfoectionEnvironment = This ckeckbox should be checked if you are sure that this exposure was the most probable infection environment for this case. Only one exposure can be marked as the probable infection environment at the same time, and that exposure will be transmitted to SurvNet.
```
MateStrysewske, line 544: => infoCheckProbableInfectionEnvironment
This ckeckbox => This checkbox

sormas-api/src/main/resources/validations.properties
```diff
@@ -103,6 +103,7 @@ visitStatus = You have to specify the status for a visit.
 softApproximateAgeTooHigh = Please make sure that this approximate age value is correct; If applicable, change it to a value lower than 150.
 validPointOfEntry = You have to specify a valid point of entry
 exportNoNameSpecified = Please type in a name for your export configuration
+epiDataMultipleInfectionEnvironments = This epidemiological data already contains an exposure that has been marked as the probable infection environment. Setting this exposure as the probable infection environment will remove that mark from the other exposure. Do you want to set this exposure as the probable infection environment of this epidemiological data?
```
MateStrysewske, line 106: Use case instead of epidemiological data, please. The checkbox actually has to be hidden for contacts anyway as I just realised (since contacts only have one exposure), but you've already implemented it that way.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -2072,6 +2072,10 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				currentVersion = 287;
 				getDao(Event.class).executeRaw("ALTER TABLE events ADD COLUMN internalId text;");
 
+			case 288:
```
MateStrysewske, line 2075: DATABASE_VERSION increase is missing

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/exposure/Exposure.java
```diff
@@ -60,6 +60,8 @@
 	private EpiData epiData;
 	@DatabaseField(foreign = true, foreignAutoRefresh = true)
 	private User reportingUser;
+	@DatabaseField(dataType = DataType.BOOLEAN)
```
MateStrysewske, line 63: Remove the dataType part (for consistency reasons, we're not really using it in other places where we define booleans in the app)

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposureForm.java
```diff
@@ -69,8 +72,8 @@
 	private static final String LOC_BURIAL_DETAILS_HEADING = "locBurialDetailsHeading";
 
 	//@formatter:off
-	private static final String HTML_LAYOUT = 
-			fluidRowLocs(ExposureDto.UUID, ExposureDto.REPORTING_USER) +
+	private static final String UUID_REPORTING_USER = fluidRowLocs(ExposureDto.UUID, ExposureDto.REPORTING_USER);
```
MateStrysewske, line 75: I don't really understand the need for this layout split. I'd prefer to just add a LOC to the layout and initialise/hide that based on the configuration criteria that you've used below.

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
```diff
@@ -123,6 +147,14 @@ private void addGeneratedColumns(Table table) {
 			return exposureTypeLabel;
 		});
 
+		table.addGeneratedColumn(COLUMN_PROBABLE_INFECTION_ENVIRONMENT, (Table.ColumnGenerator) (source, itemId, columnId) -> {
```
MateStrysewske, line 150: In line with the implementation details in the issue, I'd prefer to have an additional icon in the exposure type column instead of a new column (because we have serious issues with space in the exposures table already)

sormas-api/src/main/resources/validations.properties
FredrikSchaeferVitagroup: I have partially implemented that. The checkbox does not show in the ExposuresField for contacts, because I had the impression that it does not make sense there. It does show in the exposureForm for contacts though, as I thought there might be situations where this is helpful. Do you want me to completely hide it for contacts?
Another question: When converting a contact to a case, if that contact has an exposure, should this exposure automatically become the probable infection environment?

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposureForm.java
FredrikSchaeferVitagroup: It's because of those conditions:
- show field only in for german instances
- show the survNet warning only for german systems and when the survNetGateway ist enabled.
I thought this looks a little better when those conditions are nested. And nested conditions in a string definition look ugly.

It's not really necessary to split the layout here, though. I will remove the split.

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
FredrikSchaeferVitagroup: Ah, sorry. I misunderstood this. So an exposure in a high risk area that is the probable infection environment shall have two icons? I guess those two attributes are likely to come together.

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposureForm.java
FredrikSchaeferVitagroup: I now remember: When the static strings are generated, the attribute epidataParentClass is not available. The epidataParentClass is necessary to determine how the HTML-Layout is supposed to look like, though (as the SurvNet-warning is only supposed to show for cases). The epiDataParentClass is usable in the non-static method createHtmlLayout().
As the layout is modified neither at the beginning, nor at the end, I decided to split the HTML_Layout to stuff before the modification part, and after. Taking this into consideration, I still don't see a better way to do that.

sormas-api/src/main/resources/validations.properties
MateStrysewske: Yes, it can be completely hidden from contacts (because since there can only be one exposure per contact, that's automatically the probable infection environment if it's transmitted to SurvNet)

For the second question, yes, I think that would be pretty helpful!

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
FredrikSchaeferVitagroup: I just implemented the two icon appraoch. I think it looks okay.

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposureForm.java
MateStrysewske: Okay, fine by me then!

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
MateStrysewske: Exactly

MateStrysewske (changes requested):
Is there also a checkbox icon that has a black background and a white checkmark in it? (to be consistent with the icon for risk area)

FredrikSchaeferVitagroup:
> Is there also a checkbox icon that has a black background and a white checkmark in it? (to be consistent with the icon for risk area)

There is. I chose the inversed version because I thought it would make the icons more easily distinguishable. I've now changed it to the black background white checkmark version, so you can have a look and decide.

MateStrysewske (approved):

Issue #4774, opened 2021-03-15
Age is not sorted properly in list of EP


Marko-ilmkreis:
### Bug Description

If you click on the header of the age in the list of the EP the numbers are wrongly sorted as in these screenshot:

![grafik](https://user-images.githubusercontent.com/74791118/111158647-936f4c00-8598-11eb-8d0e-479f7af057a3.png)

As its working fine in the list of contacts and cases, it could be a bug or a missing configuration.

AndyBakcsy-she:
Validated ticket on the latest version of sormas 1.58.0-snapshot on test.sormas.netzlink.com

PR #4804, opened 2021-03-17
#4774 - Use number to express age instead of string

alexcaruntu-vita:


DavidBaldsiefen (changes requested):
Looks good to me, but you can now remove all usages of `ApproximateAgeType` in `EventParticipantIndexDto`

DavidBaldsiefen:
Can you find out why the tests failed and solve the problem? Seems like the variable was still used in some tests

alexcaruntu-vita:
I investigated and it seems that import and export of event participants are affected if we remove ApproximateAgeType. I will revert the last changes and keep just the original one.

DavidBaldsiefen (approved):

PR #4810, opened 2021-03-18
#4805 Provide SQL function to generate a UUIDv4 encoded as base32

### LINKED ISSUE #4805: Provide SQL function to generate a UUIDv4 encoded as base32
### Proposed Change
The uuids we are currently creating in SQL don't match the UUID format we are using when creating a UUID in the api (UUIDv4 encoded as base32 with a block length of 6).

We need a proper SQL function to do this. 

E.g. this is needed for #4728.

MartinWahnschaffe:
Closes #4805

@StefanKock Could you have a look at this? The UUID generation has already been reviewed by @HolgerReiseVSys, so we just need to make sure I didn't forget anything concerning the server setup part.


StefanKock (commented):
SORMAS-Docker needs to have pgcrypto installed before this is merged into development.

PR #4811, opened 2021-03-18
Import script for facilities #4728

### LINKED ISSUE #4728: Import script for facilities
### Situation Description

For scenarios where multiple SORMAS intances are used per country (e.g. one per district in Germany), there is a need to have a shared facility list. For now we will address this by importing the same list of facilities into all instances.

Problem: Importing facilities via the UI currently does not allow to update data of existing facilities, since there is no clear identifier.
Furthermore the authorities have a need to make sure that all SORMAS instances do have the facilities and thus prefer not to rely on every single SORMAS admin having to import the data.

### Feature Description

- [x] Provide an SQL script that allows to import facilities into SORMAS that are provided as CSV, following the import schema used in the UI import functionality.
- [x] The script needs to check for existing facilities based on the external ID. If the facility already exists it is supposed to update/override the facility with the values from the import file
- [x] Import lines that do not have an external ID should be ignored and ideally logged
- [x] persist script under `sormas-backend/utils`

*Important*: The data is supposed to be imported as soon as possible (ideally already into 1.57), so the new fields added in #4027 may not be there yet, but should be included when re-importing the data for 1.58.
This should also be updated when #3931 and #4066 are implemented.

### Possible Alternatives
* Provide a centralized architecture to distribute facilities to SORMAS instances. Before further considering this, it would need to be defined which country-wide data sources exists for hospitals, laboratories, doctors offices, schools and other relevant facility types. In addition to this a country wide unique identifier to be used as external ID would be very helpful when exchanging data with external systems.
* Extend the UI import functionality to allow automatic merging based on external ID and manual merging based on similar names.

### Additional Information

CSV Template that also include the fields introduced with #4755 and #3931.
[sormas_import_facility_template.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/6161770/sormas_import_facility_template.txt)

MartinWahnschaffe:
Closes #4728

### DIFF
--- a/sormas-backend/utils/importFacilities.sql
+++ b/sormas-backend/utils/importFacilities.sql
@@ -1,0 +1,70 @@
+
+/* 1. Load CSV data into temp table */
+CREATE TABLE tmp_facility (name text, regionname text, districtname text, communityname text, city text, postalcode text, street text, housenumber text, phonenumber text, emailaddress text, 
+						   	contactFirstName text, contactLastName text, contactPhoneNumber text, contactEmailAddress text, additionalinformation text, 
+							areatype varchar(255), latitude double precision, longitude double precision, type varchar(255), publicownership boolean, archived boolean, externalid text);
+COPY tmp_facility
+FROM 'C:\Users\Public\Documents\facilities.csv'
+DELIMITER ';'
+CSV HEADER;
+
+/* 2. Clean up and map data */
+/* trim whitespaces - especially important for region, district and enum values */
+UPDATE tmp_facility 
+	SET (name, regionname, districtname, communityname, city, postalcode, street, housenumber, phonenumber, emailaddress, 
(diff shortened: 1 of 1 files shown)


sormas-backend/utils/importFacilities.sql
```diff
@@ -0,0 +1,70 @@
+
+/* 1. Load CSV data into temp table */
+CREATE TABLE tmp_facility (name text, regionname text, districtname text, communityname text, city text, postalcode text, street text, housenumber text, phonenumber text, emailaddress text, 
+						   	contactFirstName text, contactLastName text, contactPhoneNumber text, contactEmailAddress text, additionalinformation text, 
+							areatype varchar(255), latitude double precision, longitude double precision, type varchar(255), publicownership boolean, archived boolean, externalid text);
+COPY tmp_facility
+FROM 'C:\Users\Public\Documents\facilities.csv'
```
StefanKock, line 7: Highlight/comment the path where the CSV file needs to be configured

StefanKock (changes requested):

Issue #4784, opened 2021-03-15
Add a country field to regions [5]


MateStrysewske:
### Situation Description
Concept issue: #4658 

For exposures (and later maybe for other use cases), users need to be able to not only specify a country, but also a continent or subcontinent in case they don't know the exact name of the country they have traveled to or have traveled to multiple countries.

### Feature Description
Beyond the definitions above, it should also be possible to create regions in other countries and allow users to select those when specifying locations in other countries.

- [x] Add a new `country` field to the `Region` entity
- [x] Extend the region edit form with the new field; read-only when editing an existing region UNLESS the country is empty
- [x] Extend the region grid with a column containing the region's country
- [x] Remove the code that makes region, district, community and health facility read-only when selecting a different country than the server country; this impacts the location and event forms; however, these fields still need to be optional in that case (this impacts the event form)
- [x] Change all code occurrences where region dropdowns are filled to the following pattern:
    * If NO country is selectable (e.g. case or contact jurisdiction), fill the region dropdown with all regions that are either in the server country or have no country
    * If a country is selectable (e.g. locations, events), use the logic above if no country or the server country is selected, otherwise fill the dropdown with the regions from the selected country
- [x] Add a version-specific database upgrade script to `StartupShutdownService.upgrade` that assigns all existing regions to the server country (if specified)

maxiheyner:
Issue tracker comment

PR #4826, opened 2021-03-18
4784 add country to region

leventegal-she:
Closes https://github.com/hzi-braunschweig/SORMAS-Project/issues/4784


MateStrysewske (changes requested):

MateStrysewske (changes requested):
The following region dropdowns aren't yet properly adjusted:
* Filter in task directory
* Filter in sample directory
* Filter in campaigns directory
* Campaign form creation and editing
* Filter in person directory (would probably make sense to be able to filter by foreign regions here, but that would require a country dropdown, and since we don't have that right now and it's definitely not part of this issue to add one, the foreign regions need to be excluded)
* Case, contact and sample dropdowns in the dev mode screen

Problems that I encountered:
* It's possible to import cases into foreign infrastructure data; didn't think about that when writing the issue; imported cases have to be rejected when the specified region is associated to a different country than the server country. However, for e.g. the person's address, it should ideally be possible (if this is too complicated we can leave it out for now and just refuse to import data when it's associated with foreign countries)
* This is probably also the case for the contact and campaign form import (haven't tested it)
* Foreign regions are shown on the Outbreaks configuration screen; should not be shown there
* Foreign regions and districts are shown when configuring line listing; should not be shown there
* When creating/editing a region, "area" needs to be disabled and set to null when selecting a different country than the server country

Additions that we need:
* The region, district, community, facility and point of entry directories need a country filter that is automatically set to the server country; should be next to the search bar for the region directory and to the left side of the region filter for the other directories

MateStrysewske (approved):

PR #4833, opened 2021-03-19
Feature 4830 inform climedo of changes when cases and contacts get deleted

### LINKED ISSUE #4830: [CLIMEDO Integration] Inform Climedo of changes when Cases and Contacts get deleted [0.5]
### Situation Description

If configured, SORMAS informs climedo of relevant changes in person data when cases or contacts of a registered person are modified. It is also necessary to inform about relevant changes when cases or contacts get deleted.

### Feature Description

Just like for changes in contacts, cases of persons, the [handleExternalJournalPersonUpdate(...)](https://github.com/hzi-braunschweig/SORMAS-Project/blob/3f4076d4499d3206ad97b395e545dde5f31f2669/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java#L229) shall be called when contacts and cases get deleted.

FredrikSchaeferVitagroup:
closes #4830

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -2271,7 +2271,11 @@ public class CaseFacadeEjb implements CaseFacade {
 			throw new UnsupportedOperationException("User " + userService.getCurrentUser().getUuid() + " is not allowed to delete cases.");
 		}
 
-		caseService.delete(caseService.getByUuid(caseUuid));
+		Case caze = caseService.getByUuid(caseUuid);
+
+		externalJournalService.handleExternalJournalPersonUpdate(caze.getPerson().toReference());
+
+		caseService.delete(caze);
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2271,7 +2271,11 @@ public void deleteCase(String caseUuid) {
 			throw new UnsupportedOperationException("User " + userService.getCurrentUser().getUuid() + " is not allowed to delete cases.");
 		}
 
-		caseService.delete(caseService.getByUuid(caseUuid));
+		Case caze = caseService.getByUuid(caseUuid);
+
+		externalJournalService.handleExternalJournalPersonUpdate(caze.getPerson().toReference());
+
+		caseService.delete(caze);
```
vlad-ciucescu, line 2278: Shouldn't the notify take place after the delete? It might not matter technically because the transaction is not over yet, but conceptually it makes sense.

vlad-ciucescu (approved):

PR #4840, opened 2021-03-20
Change 4761 activity as case ifsg

### LINKED ISSUE #4761: [SurvNet Interface] Make activityAsCase compatible for IfSG reporting (DE only) [2]
See here for implementation details: https://github.com/hzi-braunschweig/SORMAS-Project/issues/4761#issuecomment-800337829

### Feature Description
Case > epiData > activityAsCase
### Problem Description
The newly added element activityAsCase turns out to only partially fulfill the requirements of needed variables to allow a proper and clear mapping to the SurvNet interface. The mapping would be quite complicated and not very intuitive for users to fill when being used to a different structure and set of values.
### Proposed Change

- Hide "type of activity" only on german servers. Instead, show the following field:
- Add a DE only enum "Activity" with the following values:
    - Working at
    - Accommodated in
    - Cared for in
    - Unknown

- Hide "Role" only on german servers.
- Hide "type of place" only on german servers. Instead, show the following field:
- Add a DE only enum "Facility (IfSG)" in the location section with the following values:
    - Facility (§ 23 IfSG)
    - Community facility (§ 33 IfSG)
    - Facility (§ 36 IfSG)
    - Unknown
- Always show facility fields for this element on german servers.

- Show "Activity" and "Facility (IfSG)" on the overview table instead of "type of activity" and "type of place".
- Don't make any field compulsory. If this is not possible, make "Activity" the single compulsory field.

### Additional Information
Updated:
Screen Shot 2021-03-18 at 9 44 18 AM

Screen Shot 2021-03-14 at 6 46 06 PM

syntakker:
Fixes #4761

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/CountryHelper.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/CountryHelper.java
@@ -27,16 +27,9 @@ public final class CountryHelper {
 	public static boolean isCountry(String countryLocale, String country) {
 		// If the country locale is complete (e.g. de-DE), check the last (country) part; 
 		// otherwise check the first (language) part
-		if (Pattern.matches(I18nProperties.FULL_COUNTRY_LOCALE_PATTERN, countryLocale)) {
-			if (countryLocale.toLowerCase().endsWith(country.toLowerCase())) {
-				return true;
-			}
-		} else {
-			if (countryLocale.toLowerCase().startsWith(country.toLowerCase())) {
-				return true;
-			}
-		}
-		return false;
+		return Pattern.matches(I18nProperties.FULL_COUNTRY_LOCALE_PATTERN, countryLocale)
+			? countryLocale.toLowerCase().endsWith(country.toLowerCase())
+			: countryLocale.toLowerCase().startsWith(country.toLowerCase());
 	}
 
 	public static boolean isInCountries(String countryLocale, String... countries) {
--- a/sormas-api/src/main/java/de/symeda/sormas/api/CountryHelper.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/CountryHelper.java
@@ -15,9 +15,36 @@
 
 package de.symeda.sormas.api;
 
+import java.util.regex.Pattern;
+
+import de.symeda.sormas.api.i18n.I18nProperties;
+
 public final class CountryHelper {
 
 	public static final String COUNTRY_CODE_GERMANY = "de";
 	public static final String COUNTRY_CODE_SWITZERLAND = "ch";
 
+	public static boolean isCountry(String countryLocale, String country) {
+		// If the country locale is complete (e.g. de-DE), check the last (country) part; 
+		// otherwise check the first (language) part
+		if (Pattern.matches(I18nProperties.FULL_COUNTRY_LOCALE_PATTERN, countryLocale)) {
+			if (countryLocale.toLowerCase().endsWith(country.toLowerCase())) {
+				return true;
+			}
+		} else {
+			if (countryLocale.toLowerCase().startsWith(country.toLowerCase())) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	public static boolean isInCountries(String countryLocale, String... countries) {
+		for (String country : countries) {
+			if (isCountry(countryLocale, country)) {
+				return true;
+			}
+		}
+		return false;
+	}
 }
--- a/sormas-api/src/main/java/de/symeda/sormas/api/event/TypeOfPlace.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/event/TypeOfPlace.java
@@ -17,45 +17,33 @@
  *******************************************************************************/
 package de.symeda.sormas.api.event;
 
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import de.symeda.sormas.api.i18n.I18nProperties;
+import de.symeda.sormas.api.utils.HideForCountriesExcept;
 
 public enum TypeOfPlace {
 
-	FACILITY(true),
-	FESTIVITIES(false),
-	HOME(true),
-	MEANS_OF_TRANSPORT(false),
-	PUBLIC_PLACE(false),
-	SCATTERED(false),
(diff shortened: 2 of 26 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/event/TypeOfPlace.java
```diff
@@ -17,47 +17,39 @@
...
-			for (TypeOfPlace typeOfPlace : values()) {
-				if (typeOfPlace.isUsableForCases()) {
-					typesUsableForCases.add(typeOfPlace);
-				}
-			}
-		}
-		return typesUsableForCases;
-	}
+	FACILITY,
+	@HideForCountriesExcept(countries = {})
```
MateStrysewske, line 29: Parameters can be removed from these annotations

sormas-api/src/main/resources/captions.properties
```diff
@@ -145,6 +145,7 @@ ActivityAsCase.activityAsCaseType=Type of Activity
 ActivityAsCase.activityAsCaseTypeDetails=Type of Activity details
 ActivityAsCase.location=Location
 ActivityAsCase.typeOfPlace=Type of place
+ActivityAsCase.typeOfPlaceIfSG=Facility (IfSG)
```
MateStrysewske, line 148: Not necessary; caption should be changed in Crowdin for the existing typeOfPlace property

sormas-api/src/main/resources/captions_de-DE.properties
```diff
@@ -145,6 +145,7 @@ ActivityAsCase.activityAsCaseType=Art der Aktivität
 ActivityAsCase.activityAsCaseTypeDetails=Art der Aktivität Details
 ActivityAsCase.location=Ort
 ActivityAsCase.typeOfPlace=Art der Einrichtung
+ActivityAsCase.typeOfPlaceIfSG=Einrichtung (IfSG)
```
MateStrysewske, line 148: See above; also, adding German translations is not necessary in general because those will be overridden by Crowdin with the English default anyway (until there's a translation in Crowdin)

MateStrysewske (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/CountryHelper.java
```diff
@@ -15,9 +15,36 @@
...
+
 public final class CountryHelper {
 
 	public static final String COUNTRY_CODE_GERMANY = "de";
 	public static final String COUNTRY_CODE_SWITZERLAND = "ch";
 
+	public static boolean isCountry(String countryLocale, String country) {
+		// If the country locale is complete (e.g. de-DE), check the last (country) part; 
+		// otherwise check the first (language) part
+		if (Pattern.matches(I18nProperties.FULL_COUNTRY_LOCALE_PATTERN, countryLocale)) {
```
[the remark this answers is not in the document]
syntakker: Counter example:
```
countryLocale = "de-CH"
country = "de"
```

sormas-api/src/main/java/de/symeda/sormas/api/event/TypeOfPlace.java
syntakker: The parameter has to be the empty list her, since these values are not intended to be shown at all except for `ActicityAsCase` for instances in Germany.
Without the empty list, the parameter defaults to "de", so these values would be shown by default wherever `TypeOfPlace` is used on instances in Germany.

sormas-api/src/main/resources/captions.properties
syntakker: The distinction is necessary for the same reason as above.

sormas-api/src/main/resources/captions_de-DE.properties
syntakker: See above.

sormas-api/src/main/java/de/symeda/sormas/api/CountryHelper.java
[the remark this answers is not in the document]
MateStrysewske: My code should do exactly the same that your code does, just in a more compact form :) if it doesn't, that's because I forgot something

sormas-api/src/main/java/de/symeda/sormas/api/event/TypeOfPlace.java
MateStrysewske: Ah, I see! Makes sense, thanks!

sormas-api/src/main/resources/captions.properties
syntakker: The string to be used is depending on the instance's country, not the language. So a instance in Germany with user language German should still show "Facility (IfSG)", whereas an instance in another country with user language German should show "Type of place".

MateStrysewske: Absolutely true!

sormas-api/src/main/resources/captions_de-DE.properties
syntakker: See above.

sormas-api/src/main/java/de/symeda/sormas/api/CountryHelper.java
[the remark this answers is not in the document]
syntakker: `&&` => `?`, `||` => `:`
:-)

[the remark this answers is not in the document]
syntakker: Done.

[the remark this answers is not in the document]
syntakker: For some reason, github does not show the new commit on this branch to me here:
https://github.com/hzi-braunschweig/SORMAS-Project/commit/5e0b8743c2445278280446c7443de3f52e59407c

MateStrysewske (changes requested):
Looks good! The only thing I found is that the "role" field is still visible in the read dialog for activities as case in the mobile app.

syntakker:
Done. Additionally, caption for "Type of place" was not set correctly.

MateStrysewske (approved):

PR #4847, opened 2021-03-22
Feature 4777 continent to location

BarnaBartha:


sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/LocationDialog.java
```diff
@@ -288,6 +296,16 @@ public void setFacilityFieldsVisible(boolean visible, boolean clearOnHidden) {
 		}
 	}
 
+	public void setContinentFieldsVisible( boolean clearOnHidden) {
```
MateStrysewske, line 299: Code formatting

Also, the method name together with the parameter doesn't make clear what the method does. A method called setXXXVisible with a boolean parameter is usually expected to set XXX visible if the boolean parameter is true and invisible when it's false. I'd suggest to rename the method to `updateContinentFieldsVisibility` and remove the parameter because as far as I can see, it's never called with `false` anyway

Another question would be whether it's necessary to call this method from outside; can't it be automatically executed directly in this class?

sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/LocationDialog.java
```diff
@@ -288,6 +296,16 @@ public void setFacilityFieldsVisible(boolean visible, boolean clearOnHidden) {
...
+		contentBinding.locationContinent.setVisibility(InfrastructureDaoHelper.loadContinents().isEmpty() ? GONE : VISIBLE);
```
MateStrysewske, line 300: If I'm not mistaken, `loadContinents` and `loadSubcontinents` should always return false because you always add an empty item to the list, i.e. it should always contain at least one element. If that's the case, I'd rather directly call a backend method that counts the number of continents/subcontinents in the database here (also gets rid of the unnecessary logic that turns them into a list of items because those items aren't used here in the first place)

sormas-backend/src/main/java/de/symeda/sormas/backend/location/Location.java
```diff
@@ -134,6 +140,24 @@ public void setCountry(Country country) {
 		this.country = country;
 	}
 
+	@ManyToOne(cascade = {})
```
MateStrysewske, line 143: Parameter is redundant if no cascade type is chosen

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -6999,4 +6999,14 @@ ALTER TABLE country ADD CONSTRAINT fk_country_subcontinent_id FOREIGN KEY (subco
 
 INSERT INTO schema_version (version_number, comment) VALUES (352, '2020-03-17 Create continent and subcontinent #4775');
 
+-- 2020-03-19 Add continent and subcontinent to location #4777
+
+ALTER TABLE location ADD COLUMN continent_id BIGINT;
```
MateStrysewske, line 7004: Columns also need to be added to the history tables (foreign key isn't necessary there though)

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -83,7 +88,7 @@
 		//XXX #1620 are the divs needed?
 		divs(
 			fluidRowLocs(LocationDto.ADDRESS_TYPE, LocationDto.ADDRESS_TYPE_DETAILS, ""),
-			fluidRowLocs(LocationDto.COUNTRY, "", ""),
+			fluidRowLocs(LocationDto.COUNTRY, LocationDto.SUB_CONTINENT, LocationDto.CONTINENT),
```
MateStrysewske, line 91: Order should be reversed (Continent -> Subcontinent -> Country); since that would look weird when the continent and subcontinent fields are hidden, I'd suggest to move them to a separate line above the country field

MateStrysewske (changes requested):

MateStrysewske (changes requested):
- [x] ContinentFacadeEjb:getByCountry fails with a NullPointerException for countries without subcontinents
- [x] Mobile app crashes upon opening a location form for the same reason
- [x] Country list needs to be updated based on the chosen continent/subcontinent
- [ ] Subcontinent list needs to be updated based on the chosen continent

MateStrysewske (changes requested):
* Selecting a country without a subcontinent/continent needs to clear the respective field
* Countries and subcontinents need to be selectable in the mobile app when no continent/subcontinent respectively has been selected (just like in the web app)

MateStrysewske (approved):

PR #4865, opened 2021-03-24
Feature 2744 climedo logic adaptations

### LINKED ISSUE #2744: Add multiple phone numbers to person entity (case, contact, event ) [5]
### Proposed Change
- Add 2 more fields to capture other phone numbers
- Show these fields only after selecting "multiple phone numbers "
- Migration: use phone number 1 as formal phone number variable. If more than one numbers were entered in this field, then users have to manually transfer them to phone number 2
- First on web and mobile is optional
image

FredrikSchaeferVitagroup:
closes #2744


FredrikSchaeferVitagroup:
Merged after external confirmation from @BarnaBartha

PR #4866, opened 2021-03-24
#4479 - Extended parsing and added validation

### LINKED ISSUE #4479: Entering the ct value of a samples causes error
### Bug Description

If you entering a sample result in Sormas 1.56.1 with a , as seperator, you get this error:
grafik

This looks like fixed in 1.58-snapshot. But in this version a comma would be accepted and brings a wrong value at the end:

Entered value:

grafik

Value after leaving the field

grafik

### Expected Behavior

Both should be accepted a "," and a "." and hold the entered value

MateStrysewske:
Fixes #4479


JonasCir (approved):
lgtm thanks :)

Issue #4565, opened 2021-03-01
Map in statistics is not filled with data


Marko-ilmkreis:
### Bug Description

In statistics you have the option so get the data also shown on the map. Unfortunately now data were shown on the map.

### Steps to Reproduce
1. Go to statistics
2. Select the needed data and as visualization the map
3. the map stays empty

### Expected Behavior

All datas are shown on the map or the option for a map is hidden if its not possible to show the selected data.

### Screenshots

![grafik](https://user-images.githubusercontent.com/74791118/109466339-dc3be680-7a69-11eb-87ab-fc91d6954fff.png)

### System Details
* Device: Win10
* SORMAS version: 1.56.1
* Android version/Browser: FF

Marc-FS:
Same here. Maybe insufficient User rights?
1.56.1

Marko-ilmkreis:
No, its under my role. Admin, National user, National clinicain.

JonasCir:
Hi @Marko-ilmkreis, did this issue came up after the 1.56.1 update?

Marko-ilmkreis:
@JonasCir No, as far as I remember its in Sormas as long we are using it. In december the hotline told us that the developer are aware of it and going to fix it. An issue is not opened for that so far.

JonasCir:
Hm at some point I managed to work around this, let's if I can remember it :D

JonasCir:
@Marko-ilmkreis 
Hi, if you login as contact supervisor and navigate to the "Kontaktübersicht" (not statistics)

1. do you see anything there on the map? 
2. If you continue to navigate to the transmission chains ("Bitte vergrößern um die Infektionsketten zu sehen") do you see anything there?

Marko-ilmkreis:
@JonasCir I tried it and it shows the data as expected for both (1&2).

JonasCir:
Thanks, I think I identified the root [cause](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-ui/src/main/java/de/symeda/sormas/ui/statistics/StatisticsView.java#L941):

The problem is that, the code tries to resolve shape files for the districts/regions given, however, for Germany there are no such files provided by default as it seems.

Marko-ilmkreis:
Okay, that seems to be right. The map type can be selected, but this makes no sense for Germany anyway.
[...]

PR #4867, opened 2021-03-24
#4864 correctly display districts in statistics map

### LINKED ISSUE #4864: Improve Statistics Map, Districts overlap each other
### Bug Description

Currently, some districts are not correctly mapped in the statistics map section. The main reason for that is that in germany, some districts have the same name but are either a Landkreis (LK) or Stadtkreis (SK). 

### Steps to Reproduce
1. Create Cases in a LK and SK which are "inside" each other. Example: LK & SK Würzburg
2. Go to the statistics section, and generate the district map. the LK will overlap the SK.

### Expected Behavior

It should look like this: 

image

### Additional Information

This is a followup of #4565

DavidBaldsiefen:
Fixes #4864 , followup of #4565


sormas-backend/src/main/java/de/symeda/sormas/backend/region/GeoShapeHelper.java
```diff
@@ -134,21 +138,21 @@ public static MultiPolygon getPolygon(SimpleFeature feature, MathTransform trans
 	 * @return
 	 *         The name of the feature, null if no name could be found.
 	 */
-	public static String sniffShapeName(SimpleFeature feature, List<String> list) {
-		String shapeName = null;
+	public static String sniffShapeAttribute(SimpleFeature feature, List<String> list) {
```
MartinWahnschaffe, line 141: Rename list to attributeNames.
I think the old comment also needs to be adjusted. It's returning the value of the found feature attribute, not the attribute name.

sormas-ui/src/main/java/de/symeda/sormas/ui/statistics/StatisticsView.java
```diff
@@ -973,11 +969,50 @@ public void generateMap() {
...
+		// sort polygon array, so that polygons which are completely contained by another appear on top
+		List<Integer[]> indexesToSwap = new ArrayList<>();
+		int poly1index = 0, poly2index = 0;
+		for (LeafletPolygon poly1 : resultPolygons) {
+			for (LeafletPolygon poly2 : resultPolygons) {
+				if (poly1index == poly2index) {
+					continue;
+				}
+				// determine maximum latitude and longitude of each polygon
+				List<double[]> poly1LatLons = Arrays.asList(poly1.getLatLons());
```
MartinWahnschaffe, line 988: I suggest to give the LeaftletPolygon two fields minLatLon and maxLatLon and a getter method. Whenever the getter is called the first time, calculate the values. Reset them to null, when setLatLons is called.

Thisway the min/max does not have to be calculated for every combination of polygons, but just for every single polygon.

MartinWahnschaffe (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/region/GeoShapeProviderEjb.java
```diff
@@ -266,23 +269,53 @@ private void loadDistrictData(String countryName) {
...
+					String shapeDistrictId = GeoShapeHelper.sniffShapeAttribute(feature, Arrays.asList("ARS"));
+					if (shapeDistrictId == null) {
+						continue;
+					}
+					districtResult = districts.stream().filter(r -> {
+						String districtExtID = r.getExternalId();
+						if (districtExtID == null) {
+							return false;
+						}
+						return districtExtID.contains(shapeDistrictId)
```
MartinWahnschaffe, line 285: I'm not sure about using similarity here. From my understanding we are looking at IDs that should be an exact match. It would not be acceptable to use the shape for a district, when the ID is different. E.g. 142421323 vs. 142421324

MartinWahnschaffe (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/region/GeoShapeProviderEjb.java
DavidBaldsiefen: Hmm yes you're right. The problem is that (of course :P) the shapefiles we use use a slightly different approach. I think the shapefiles provide 5-digit-ID's and the external IDs in SORMAS are usually 7 digits. (something like the first two digits always being 11). I can look into this

DavidBaldsiefen: I've removed the similarity check, and added a custom check for the external ID being "110<shapefile ID>", e.g. `11012345` for the district with shapefile-ID `12345`, as that seems to be common for germany.

sormas-ui/src/main/java/de/symeda/sormas/ui/statistics/StatisticsView.java
```diff
@@ -726,22 +731,12 @@ public void generateChart() {
 		}
 		hcjs.append("],");
 
-		hcjs.append("exporting: {\n" +
```
MartinWahnschaffe, line 729: @DavidBaldsiefen Didn't notice this before. Please add a tag to keep the original formatting, because it is much easier to read.

MartinWahnschaffe (changes requested):
@DavidBaldsiefen Just one small fix. Afterwards you can merge

PR #4888, opened 2021-03-25
Feature 3545 refactoring batching

### LINKED ISSUE #3545: Option to set missing geo-coordinates based by one click for all cases/contacts [2]
## Situation Description

We are on the way to implement Sormas to our district and are ging to import thousands of contacts and cases. They have mainly all an address included, but no geo-coordinates as the data come from an excel-file ....

### Feature Description

It should be possible to create geo-cordinats for the cases and contacts where the are missing.

### Possible Alternatives

The coordinates must be set manually in every case.

ChristopherRiedel:
closes #3545


StefanKock (approved):

StefanKock:
Done in pair programming and reviewed together.

PR #4939, opened 2021-03-29
Feature 2937 campaign data aggregation view

alexcaruntu-vita:


sormas-backend/src/main/java/de/symeda/sormas/backend/campaign/statistics/CampaignStatisticsService.java
```diff
@@ -0,0 +1,403 @@
...
+				results.put(campaignStatisticsGroupingDto, campaignStatisticsDto);
+			}
+			int length = result.length;
+			CampaignFormDataEntry campaignFormDataEntry = new CampaignFormDataEntry((String) result[length - 2], result[length - 1]);
+			results.get(campaignStatisticsGroupingDto).addStatisticsData(campaignFormDataEntry);
+		});
+		return results.values().stream().collect(Collectors.toList());
+	}
+
+	public long count(CampaignStatisticsCriteria criteria) {
```
MateStrysewske, line 68: I guess the count query is useless right now because we don't have lazy loading, so please comment it out for now

sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/campaignstatistics/CampaignStatisticsGrid.java
```diff
@@ -0,0 +1,99 @@
...
+			column.setCaption(I18nProperties.getPrefixCaption(CampaignStatisticsDto.I18N_PREFIX, column.getId(), column.getCaption()));
+		}
+	}
+
+	public void reload() {
+		getDataProvider().refreshAll();
+	}
+
+	public void setDataProvider() {
+		DataProvider<CampaignStatisticsDto, CampaignStatisticsCriteria> dataProvider = DataProvider.fromFilteringCallbacks(
```
MateStrysewske, line 49: I think we should use an eager data provider here instead for now since lazy loading isn't implemented in the backend anyway; see any `setEagerDataProvider` method in one of the existing grids for reference

MateStrysewske (changes requested):
* Please make sure that there are no columns shown for fields with a CampaignFormElementType other than NUMBER and YES_NO
* Region dropdown may only contain regions from the server country (please check other region dropdowns for how this is done, was recently added)

alexcaruntu-vita:
Regarding

> * Region dropdown may only contain regions from the server country (please check other region dropdowns for how this is done, was recently added)

In order to do this I need to map region to country directly since there is no way to map area to country. This will erase the mapping between region and area and the grouping at the area level won't work.

MateStrysewske:
> In order to do this I need to map region to country directly since there is no way to map area to country. This will erase the mapping between region and area and the grouping at the area level won't work.

Is this still valid or did you find a solution for it? Asking because I see a commit that references this issue. If it's not easily solvable, I think it's fine to leave it like this for now because Afghanistan probably won't add any regions from other countries to their system anyway.

alexcaruntu-vita:
> > In order to do this I need to map region to country directly since there is no way to map area to country. This will erase the mapping between region and area and the grouping at the area level won't work.
> 
> Is this still valid or did you find a solution for it? Asking because I see a commit that references this issue. If it's not easily solvable, I think it's fine to leave it like this for now because Afghanistan probably won't add any regions from other countries to their system anyway.

It is still valid. The only way to solve it is manually, in the db. So, if I got you right, should I revert the commit that uses the server country to detect regions?

sormas-ui/src/main/java/de/symeda/sormas/ui/campaign/campaignstatistics/CampaignStatisticsView.java
```diff
@@ -185,7 +188,7 @@ public CampaignStatisticsFilterForm createFilterForm() {
 				for (CampaignFormElement element : campaignFormElements) {
 					if (element.isImportant() || !onlyImportantFormElements) {
 						String type = element.getType();
-						if (type.equals("number") || type.equals("yes-no")) {
+						if (type.equals(ELEMENT_TYPE_NUMBER) || type.equals(ELEMENT_TYPE_YES_NO)) {
```
MateStrysewske, line 190: Please use `CampaignFormElementType.fromString` for the comparison because it's case-insensitive, so:

```
if (CampaignFormElementType.fromString(type) == CampaignFormElementType.NUMBER ...
```

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #4961, opened 2021-03-30
[#4959] switch from @ Required to @ NotNull

### LINKED ISSUE #4937: ContactDto.contactClassification is not required in the API
### Bug Description
It is possible to post a contact to the API which has no `contactClassification` field. However, this triggers NPEs down the pipeline. 

### Expected Behavior
The API requires and **validates** that this field is present. Just adding an `@Required` annotation establishes a contract, but a client is still free to ignore this. 

In the long term, we need to make sure that all fields that are annotated as required are also validated in the API and if they are missing, the request needs to return an error @vidi42

### System Details
* SORMAS version: latest develop

### LINKED ISSUE #4959: Enforce DTO fields marked as @Required to be present in the request
### Bug Description
Currently we have the possibility to annotate DTO fields as required. However, as of now this is purely cosmetic and used by openAPI to create a contract which fields must be included in a correct request. Any (malicious) client is of course free to send whatever he likes and therefore is able to trigger exceptions and store malformed data which cannot be retrieved properly.

### Steps to Reproduce
See https://github.com/hzi-braunschweig/SORMAS-Project/issues/4937 as example. There are more severe issues where you can store cases/contacts via the API, but cannot open the details page or do not show up at all in the list views because of NPEs.

### Expected Behavior
The API checks the request body properly for missing fields and returns HTTP 400 (Bad Request) in case a required filed is not provided.

JonasCir:
Closes #4959 
Closes #4937

So I tested it with the fix for #4937 (which I included here in this PR as well). The API returns 400 bad request now :tada:


JonasCir:
@MateStrysewske removed the special casing for Required from the swagger generation.
Ready for review :)

MateStrysewske:
@JonasCir Looks fine to me, however I'm a bit concerned about this potentially breaking a) the mobile app synchronisation and b) external systems. Have you tried creating cases, contacts, samples, events, etc. in the mobile app and syncing them to the backend? If not, please do so.

@vidi42 @FredrikSchaeferVitagroup @HolgerReiseVSys @leventegal-she Do you see any potential problems with this in terms of external systems?

JonasCir:
@MateStrysewske I will take another testing pass, I absolutely understand that this change is critical. Also we are not in a rush to merge this therefore we should take our time and wait for feedback :)

sormas-api/src/main/java/de/symeda/sormas/api/sample/PathogenTestDto.java
```diff
@@ -57,31 +57,31 @@
...
 	private String labDetails;
-	@Required
+	@NotNull
 	@SensitiveData
 	private UserReferenceDto labUser;
-	@Required
+	@NotNull
 	private PathogenTestResultType testResult;
-	@Required
+	@NotNull
```
vidi42, line 81: `testResultText` seems to be marked here as NotNull, however on the mobile side it's not marked or enforced to be required, which results in a `400` when synchronizing Casese or Samples

FredrikSchaeferVitagroup:
I just tried to create a new case via processing a labMessage. When clicking of Save, I get an exception:
```Constraint violations: 
 (1) Kind: PROPERTY
 message: darf nicht null sein
 root bean: de.symeda.sormas.backend.contact.ContactFacadeEjb@59e65e07
 property path: saveContact.arg0.lastContactDate

FredrikSchaeferVitagroup:
As also VisitDto was modified, and I don't now what values the external journals consider not nullable, this has the potential to break the communication with external journals. We should maybe communicate this change and test quite extensively before rolling this out to production?

JonasCir:
Alright, the thing here is that the annotation where just cosmetic before, therefore, we need to figure out which values are really "required". In the past I came along fields which were required to be not null by the logic, but weren't annotated as such. In the same way, there are values which are not required but were erroneously marked as such.

sormas-api/src/main/java/de/symeda/sormas/api/sample/PathogenTestDto.java
JonasCir: fixed

JonasCir:
@MateStrysewske @MateStrysewske @StefanKock I rebased this branch. I really think we should get rid of `@Required` as it is effectively useless, a source of confusion, and a steady source of bugs over the years.

JonasCir:
Most of the current test failures are likely caused by the fact that things were marked as required, but never truly were.

MateStrysewske:
@JonasCir We can't merge this PR as long as there are test failures though

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=4961

MartinWahnschaffe:
@MateStrysewske All probems are fixed now. Please re-review (or look into the last two commits). I have removed some of the NotNull annotations based on #7648.

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=4961

MateStrysewske (changes requested):
Error description columns in error report files for imports are buggy now. E.g. for contacts, the column is just empty while for cases it's all over the place with class paths and basically not human-readable.

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=4961

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=4961

MateStrysewske (approved):

PR #4993, opened 2021-04-06
#4929 custom export captioning (i18n)

### LINKED ISSUE #4929: Event ID shows up too many times in custom export for event participants
### Bug Description
When creating a new configuration for the event participant export, the event ID shows up too many times.
### Steps to Reproduce
1. Open an event and go to the event participant tab
2. Create a new custom export configuration
3. Check selection options

### Expected Behavior
Event ID should only appear once and the other three should be event participant Id, case Id or person Id.
### Screenshots
eventParticipantExport

### System Details
* Device:
* SORMAS version: 1.58.0

DavidBaldsiefen:
Fixes #4929 by determining the correct i18n prefix using reflection.


MateStrysewske (changes requested):
I like that solution! The only thing I've captured that it's not yet working for is the event participant UUID (which still says "event ID" instead of "event participant ID") - can you check that one again please?

DavidBaldsiefen:
> I like that solution! The only thing I've captured that it's not yet working for is the event participant UUID (which still says "event ID" instead of "event participant ID") - can you check that one again please?

I fixed it by choosing a more logical priority-order of the i18n prefixes to chose when no prefix is specified.

MateStrysewske (approved):

PR #5000, opened 2021-04-06
#4928 date validation

### LINKED ISSUE #4928: Date validations don't prevent saving erroneous entries and are sometimes misleading
### Bug Description
Date validations don't prevent the user from saving erroneous data and are sometimes misleading. Examples are validations for vaccination fields and stay in ICU fields with the addition of admission date. Dates that are in the future can be saved regardless of the error message. When adjusting the date period, the error display remains, when one field was edited.
### Steps to Reproduce
1. Go to hospitalizations 
2. Enter a stay in ICU start date and then an end of stay date that lies in the future.
3. Save the hospitalization
4. Enter dates where the end date is before the start date
5. Change one of the dates so the entry is logical
6. Check the other entry
7. Do the same with the admission date and first/last vaccination date

### Expected Behavior
If the date entry is erroneous, users should be able to save it. If the date period gets corrected to a logical entry, error messages should no longer be displayed.
### Screenshots
**Saving erroneous entry:**
(current date 3/28/2021)
Screen Shot 2021-03-28 at 7 37 26 PM

**Stay in ICU**
Initial entry:

IcuDateValidation

Change:

IcuDateValidation2

**admission and ICU**
Initial entry:

IcuDateValidation3

Change:

IcuDateValidation4

**Vaccination dates**
Initial entry:

vaccinationDateValidation

Change:

vaccinationDateValidation2

### System Details
* Device:
* SORMAS version: 1.58.0

DavidBaldsiefen:
Fixes #4928 

Dates should now be validated correctly and prevent the user from saving when it makes sense. Also, changes to vaccination and hospitalization dates will now always re-evaluate both the start and end date. Validations which should be ignored upon saving were changed to purely visual indicators.


MateStrysewske (changes requested):
* Admission date receives the soft validation error ("should be after onset") even if the symptom onset date is null
* Can you please add the logic that updates both date fields when one of them is changed to the intensive care unit fields as well?

MateStrysewske (approved):

PR #5028, opened 2021-04-09
#4138: Add date and responsible user of last follow-up status change

### LINKED ISSUE #4138: Show date and responsible user of last follow-up status change [2]
Dear Sormas-Team,

most of the time the Follow-up date is to long and can not be adjusted. Actual we monitor people 14 days after there tests or extend till 24h after recovery. The Problem, the follow-up date is calculated automaticly 14 Days after the Date of Report or when the first symptoms started. This does not belong to our procedure. 

Please let us change the Follow-up date on our own.

Thank you very much and stay healty

Best regards

Julian

* Add two new fields to cases and contacts:
    * `followUpStatusChangeDate` (caption: "Date of follow-up status change")
    * `followUpStatusChangeUser` (caption: "Responsible user")
* When a user manually changes the follow-up status of a case or contact, store the date and the user into these new fields
* Vaadin app: Show the two fields below the "Follow-up status" field only if they are filled
* Vaadin app: Add a read-only text field to the follow-up form in between follow-up until and the checkbox to override the follow-up until date that contains the expected end date of follow-up (based on the calculations in the `updateFollowUpUntilAndStatus` methods in CaseFacadeEjb and ContactFacadeEjb; very likely makes sense to move the calculation of the until date to a helper method in the API to be able to access it from the UI); caption: "Expected follow-up until"

ChristopherRiedel:
closes #4138


sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactDataForm.java
```diff
@@ -960,4 +981,19 @@ private void initContactDateValidation(DateField startDate, DateField endDate, C
 			}
 		});
 	}
+
+	@Override
+	public void setValue(ContactDto newFieldValue) throws ReadOnlyException, Converter.ConversionException {
+		super.setValue(newFieldValue);
+
+		expectedUntilDate.setValue(
```
sergiupacurariu, line 989: this returns nullPointerException. If the Case follow-up feature is not activated. You are trying to set a value on a field that does not exist. Please update.

sergiupacurariu (commented):
there are 2 conflicts which need to be solved

sergiupacurariu:
when the follow-up until is overwritten the user who changed that is not saved on the case

ChristopherRiedel:
> when the follow-up until is overwritten the user who changed that is not saved on cases and contacts

Thats not part of the feature ;)

PR #5087, opened 2021-04-15
Merge branch 'development' into 1771_upload_documents_for_case_and_contact

### LINKED ISSUE #1771: Option to attach document like pdf, word, jpeg to cases and contacts (Web-App UI Part) [3]
### Situation Description
Often, cases usually have associated documents that are not registered in SORMAS. users need to be able to upload associated documents to cases and to contacts.
### Additional Information
See #3283

Phase 1: Web-App only (!) 
Android App out of scope

### Development Specifications:
- [ ] Add the ability to upload documents to cases and contacts (same logic as for events, see #3283)
  - At least add the following Metadata to each file: ID, Filename, Display Name, Type, Size, Author, Upload Date
  - Allowed Files: jpg, png, doc, docx, pdf, xls, xlsx, dot, dotx 10MB (Maybe add more filetypes, but some files like .exe should not be allowed
- [x] Make sure documents are hidden if the user does not have direct access to the case/contact (based on the jurisdiction and/or whether the user is the reporting user; probably matches whether or not the case/contact is pseudonymized)
- [x] Add a list of documents to the right side of cases/contacts (similar to how it's done for events)

vidi42:
Closes #1771


sormas-backend/src/main/java/de/symeda/sormas/backend/document/DocumentStorageService.java
```diff
@@ -77,4 +83,17 @@ private Path computeRelativePath(Document document) {
 			Integer.toString(document.getCreationDate().getHours()),
 			document.getUuid());
 	}
+
+	private void setDocumentAttributes(Document document, Path filePath) throws IOException {
+		setAttribute(filePath, "Document UUID", document.getUuid());
+		setAttribute(filePath, "Author", userService.getCurrentUser().getUserName());
+		setAttribute(filePath, "Display Name", document.getName());
+		setAttribute(filePath, "Display Name", document.getName());
```
MateStrysewske, line 91: Duplicate attribute

MateStrysewske (changes requested):
The text that's displayed if there are no documents is not fully internationalised, the words "CASE" and "CONTACT" aren't translated

PR #5109, opened 2021-04-16
Feature 3383 fix actions

### LINKED ISSUE #3383: [External Journal] Setup proper External Journal API versioning [1]
### Situation Description

External journals are dependent on our API defined in the ExternalVisitsResource. Their developers need to be informed about APi changes. This process and the necessary documents, structures and procedures are supposed to be setup with this issue. Some preliminary work has been done (there are unit test for the API in #3358 , there is some documentation in the ExternalVisitsResource and generated in #3180).

### Feature Description

What is left to do:
~~- [ ] define a sensible version number for the API in the ExternalVisitsResource~~
~~- [ ] add a comment to that version number that explains how to increase it. it is supposed to be as follows: There are major and minor versions of the API (just concerning the external visits).
Within a major version, the latest minor version is compatible with all minor versions before. This is only the case when a new optional field is added.
Otherwise, the major version changes. I think we can not provide several major API versions with dedicated paths at the same time, because it would be too complex (we would need different versions of a DTO for example).~~
~~- [ ] setup and document a standardized way to inform external developers about API changes. Maybe we could use one of our gitter channels for that. It probably makes sense to have a document with descriptions and explanations (as provided as example in #3358), that we provide an updated version of for each (or at least each major) API change. We should also define a time period we give external development teams before we release major API changes.~~
See: https://github.com/hzi-braunschweig/SORMAS-Project/issues/3383#issuecomment-748080102 bellow

FredrikSchaeferVitagroup:
closes #3383

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,10 +41,14 @@ jobs:
 
       - name: Commit external visits API spec to development
         if: github.event_name == 'push' && github.ref == 'refs/heads/development' && ${{ hashFiles(sormas-rest/target/external_visits_API.yaml) }} != ${{ hashFiles(openapi/external_visits_API.yaml) }}
-        with: ${{ secrets.UPDATER_TOKEN }}
         run: |
-            git config user.name github-API-updater
-            git config user.email github-API-updater@github.com
+            token=${{ secrets.SORMAS_VITAGROUP_TOKEN }}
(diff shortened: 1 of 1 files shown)


JonasCir (changes requested):
minor changes, just accept them right away :)

JonasCir (approved):
lgtm thanks

PR #5127, opened 2021-04-19
#4938 increase bulk-coordinate-setter performance

### LINKED ISSUE #4938: Improve performance of bulk geo-coordinate setting [2]
### Feature Description
#3545 has introduced a functionality to automatically set the geo-coordinates for the home addresses of all persons.

### Problem Description 
The way this has been implemented will lead to massive performance problems on servers with many persons in the database, and thus the feature can't be used properly. The problem is that there is a call to getIndexList without any restrictions, which might try to load hundreds of thousands of entries from the database in one query, which will very likely lead to a server timeout.

### Proposed Change
The whole process needs to be done in a batched way (not only the updateGeoLocation call as it is right now).

DavidBaldsiefen:
This PR is supposed to improve the performance of the bulk-geocoordinate-setter (or more accurately: reduce the chance for timeouts / cache overflows). At it's core, the main change is that instead of using `getIndexList` a custom query using batched calls using `OFFSET` and `LIMIT` is used. It also only fetches each persons UUID instead of fetching the entire person.

It might make sense to add a cancel-button to the popup too, that is not yet done with this change.

Closes #4938


sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonsView.java
```diff
@@ -55,14 +56,14 @@ public PersonsView() {
 		gridLayout.setStyleName("crud-main-layout");
 
 		// TODO disabled because of potential performance problems, see GitHub issue #4938
```
StefanKock, line 58: Remove comment

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -783,16 +783,16 @@ public boolean doesExternalTokenExist(String externalToken, String personUuid) {
 
 	@Override
 	public long setMissingGeoCoordinates(boolean overwriteExistingCoordinates) {
+		// The uuid-list is filtered by the users jurisdiction and retrieved in batches to avoid timeouts
```
StefanKock, line 786: Add empty line above

StefanKock (changes requested):
Added two remarks, quickly fixable.

The approach looks better than before, but I'm uncertain if fetching the person UUIDs in 500 batch size is quicker than all at once. It the complex WHERE conditions are doing performance harm, the batching makes it worse.

DavidBaldsiefen:
> Added two remarks, quickly fixable.
> 
> The approach looks better than before, but I'm uncertain if fetching the person UUIDs in 500 batch size is quicker than all at once. It the complex WHERE conditions are doing performance harm, the batching makes it worse.

I was also wondering what a good batch size would be. Another problem is that the user filter has some quite complex underlying cod (lots of joins etc.), which would also be re-executed each time.

MateStrysewske (approved):

StefanKock (approved):

PR #5133, opened 2021-04-19
Feature 5029 pcr specs and disease variant for pathogens

BarnaBartha:


sormas-api/src/main/resources/captions.properties
```diff
@@ -281,6 +281,7 @@ CampaignFormData.edit=Edit
 caseCasesList=Cases list
 caseInfrastructureDataChanged=Infrastructure data has changed
 caseCloneCaseWithNewDisease=Generate new case for 
+caseUpdateCaseWithNewDiseaseVariant=Update case disease variant
```
MateStrysewske, line 284: Move this from captions to strings and use `headingUpdateCaseWithNewDiseaseVariant` as property

sormas-app/app/src/main/res/layout/fragment_pathogen_test_edit_layout.xml
```diff
@@ -80,6 +85,13 @@
                     app:value="@={data.testedDisease}"
                     style="@style/ControlSingleColumnStyle" />
 
+                <de.symeda.sormas.app.component.controls.ControlSpinnerField
+                    android:id="@+id/pathogenTest_testedDiseaseVariant"
+                    app:dependencyParentField="@{pathogenTestTestedDisease}"
```
MateStrysewske, line 90: Don't limit the visibility of this field to only CORONAVIRUS; should be visible for every disease with disease variants

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/PathogenTest.java
```diff
@@ -113,6 +119,16 @@ public void setTestedDiseaseDetails(String testedDiseaseDetails) {
 		this.testedDiseaseDetails = testedDiseaseDetails;
 	}
 
+	@ManyToOne(cascade = {})
+	@JoinColumn(nullable = true)
```
MateStrysewske, line 123: If I'm not mistaken you can ommit the parameters for both annotations because they're the default

MateStrysewske (changes requested):

sormas-app/app/src/main/java/de/symeda/sormas/app/pathogentest/edit/PathogenTestEditActivity.java
```diff
@@ -84,7 +92,42 @@ public void saveData() {
...
+		final Case associatedCase = pathogenTestToSave.getSample().getAssociatedCase();
+
+		if (associatedCase != null) {
+			DiseaseVariant caseDiseaseVariant = DatabaseHelper.getDiseaseVariantDao().queryForId(associatedCase.getDiseaseVariant().getId());
+			DiseaseVariant newDiseaseVariant = pathogenTestToSave.getTestedDiseaseVariant();
+			if (pathogenTestToSave.getTestResult() == PathogenTestResultType.POSITIVE
+					&& pathogenTestToSave.getTestResultVerified() == true
+					&& !newDiseaseVariant.equals(caseDiseaseVariant)) {
+
+				String heading = I18nProperties.getCaption(Captions.caseUpdateCaseWithNewDiseaseVariant);
```
MateStrysewske, line 104: Wrong captions referenced, app doesn't build because of that (please make sure that the app is building correctly)

MateStrysewske (changes requested):
* Don't show the prompt to update the disease variant when no disease variant has been selected in the pathogen test
* (Web) Saving a pathogen test can potentially bring up a lot of prompts at the same time; right now, if you're saving a positive pathogen test with a different disease variant than the case for a case that is not confirmed, you will first get the prompt to update the disease variant; confirming that brings up another popup asking you to confirm the case; however, choosing "Yes" there as well brings up the warning message that the data could not be saved because it has been changed in the meantime. The solution probably is to load the case from the database again in each of these prompts

MateStrysewske (changes requested):
* PCR test specification field in mobile app doesn't have a proper caption

MateStrysewske (approved):

PR #5135, opened 2021-04-19
no geobutton highlighting with all location fields empty (#5051)

### LINKED ISSUE #5051: Close stream in TemplateEngine, refactor Geocode button highlighting
### Code Quality
In `TemplateEngine`, l.80:
- stream is not closed

In `LocationEditForm`, l.504ff:
- move hardcoded style to the proper location
- refactor `ValueChangeListener` for Location fields

### System Details
* SORMAS version: 1.59.0-SNAPSHOT

syntakker:
fixes #5051

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
@@ -419,7 +419,8 @@ public class LocationEditForm extends AbstractEditForm<LocationDto> {
 						|| facilityDto.getAreaType() != null
 						|| facilityDto.getLatitude() != null
 						|| facilityDto.getLongitude() != null
-					||(StringUtils.isNotEmpty(facilityDto.getContactPersonFirstName()) && StringUtils.isNotEmpty(facilityDto.getContactPersonLastName()))) {
+						|| (StringUtils.isNotEmpty(facilityDto.getContactPersonFirstName())
+							&& StringUtils.isNotEmpty(facilityDto.getContactPersonLastName()))) {
 
 						// Show a confirmation popup if the location's address is already set and different from the facility one
 						if ((StringUtils.isNotEmpty(cityField.getValue()) && !cityField.getValue().equals(facilityDto.getCity()))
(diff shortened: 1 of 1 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -524,19 +525,18 @@ private HorizontalLayout createGeoButton() {
...
-			Field<?> field = getField(fieldName);
-			field.addValueChangeListener(e -> {
-				if (field.isModified()) {
-					geocodeButton.addStyleName(CssStyles.GEOCODE_BUTTON_HIGHLIGHT);
-				}
-			});
-		} ;
+		Stream.of(locationGeoFields).forEach(field -> field.addValueChangeListener(e -> {
+			if (isAllFieldsEmpty(locationGeoFields)) {
+				geocodeButton.removeStyleName(CssStyles.GEOCODE_BUTTON_HIGHLIGHT);
```
StefanKock, line 535: When you empty all address fields, the geocoding button will not be highlighted. But will the coordinates already present be removed or stay there?

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
```diff
@@ -524,19 +525,18 @@ private HorizontalLayout createGeoButton() {
...
+			} else if (field.isModified()) {
```
StefanKock, line 536: Still not changed as I commented in Crowd testing: `if (field.isModified())` changes behaviour. Intended was to signal the user to click on geocoding button when the value changed. Now when you change a value (`modified = true`), click on geo button, and change than the value back as it is in the DTO (`modified = false`), the highlighting does not reappear. I would say: Just remove the if clause: If the value in any related text field changes, highlight the geocoding button as it was before.

StefanKock: Tested with @syntakker: Is okay, isModified does not change to false if the DTO value is typed in again. So my objection is not relevant.

sormas-ui/src/main/java/de/symeda/sormas/ui/location/LocationEditForm.java
StefanKock: Reviewed with @syntakker : Is okay like it is now, no functionality resets the coordinates yet if there are no useful coordinates.

PR #5141, opened 2021-04-20
#4976 Added missing pathogen test mappings

### LINKED ISSUE #4976: [DEMIS Interface] The DEMIS mapping does not apply when processing messages and only creating new pathogen test results
### Bug Description
When a user selects an existing sample and only adds a pathogen test result, the mapping of fields does not apply.
### Steps to Reproduce
1. Process a lab message
2. Choose an existing sample and only create a new pathogen test result

### Expected Behavior
The mapping should apply when only creating a new pathogen test result too
### Screenshots
**Sample + pathogen test result**

DemisMappingPathOnly2

**Pathogen test result**

DemisMappingPathOnly

### System Details
* Device:
* SORMAS version: 1.58.1
* DEMIS Adapter: 1.3.0

Alexandru-Pirvu:
Closes #4976

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
@@ -685,10 +685,13 @@ public class LabMessageController {
 
 	private PathogenTestDto buildPathogenTest(SampleDto sampleDto, LabMessageDto labMessageDto) {
 		PathogenTestDto pathogenTestDto = PathogenTestDto.build(sampleDto, UserProvider.getCurrent().getUser());
+		pathogenTestDto.setTestResult(labMessageDto.getTestResult());
 		pathogenTestDto.setTestType(labMessageDto.getTestType());
 		pathogenTestDto.setTestedDisease(labMessageDto.getTestedDisease());
+		pathogenTestDto.setTestResultVerified(labMessageDto.isTestResultVerified());
 		pathogenTestDto.setTestDateTime(labMessageDto.getTestDateTime());
 		pathogenTestDto.setTestResultText(labMessageDto.getTestResultText());
+		pathogenTestDto.setReportDate(labMessageDto.getMessageDateTime());
 		return pathogenTestDto;
 	}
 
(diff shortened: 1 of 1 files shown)


FredrikSchaeferVitagroup:
This field needs to be checked, too.
image

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
```diff
@@ -675,6 +675,10 @@ private void createPathogenTest(SampleDto sampleDto, LabMessageDto labMessageDto
 		Window window = VaadinUiUtil.createPopupWindow();
 		CommitDiscardWrapperComponent<PathogenTestForm> pathogenTestCreateComponent =
 			getPathogenTestCreateComponent(sampleDto, labMessageDto, pathogenTestDto, window);
+		CheckBox viaLimsCheckbox = pathogenTestCreateComponent.getWrappedComponent().getField(PathogenTestDto.VIA_LIMS);
+		viaLimsCheckbox.setValue(Boolean.TRUE);
```
FredrikSchaeferVitagroup, line 679: [Here](https://github.com/hzi-braunschweig/SORMAS-Project/blob/9ea12de565dace01f2544b396be2acacf69c0535/sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java#L652) is a null check involved. Is there a reason why you ommited this check in your implementation?

Alexandru-Pirvu: The PathogenTestForm contains that field. I saw no reason to check for null.

FredrikSchaeferVitagroup: Does it always exist, also for non german systems?

FredrikSchaeferVitagroup: No, it does not.

FredrikSchaeferVitagroup: After a deeper look it seems that it always exists. I am wondering what the null check is there for...

FredrikSchaeferVitagroup (approved):

PR #5152, opened 2021-04-20
#4936 fix lat/long order and add EPSG4326_WKT String to sormas.properties

### LINKED ISSUE #4936: longitude and latitude get mixed up for maps
### Bug Description
When the geo coordinates for the maps in Sormas are configured, the system mixes up longitude and latitude.
### Steps to Reproduce
1. Set the coordinates for the maps in Sormas
2. Check the case status map

### Expected Behavior
Sormas should use longitude and latitude in the correct way
### Screenshots
Screen Shot 2021-03-29 at 11 19 09 AM

**This map display actually conforms to latitude: 10.53 and longitude: 52.21**

Screen Shot 2021-03-29 at 11 12 04 AM

### System Details
* Device:
* SORMAS version: 1.58.0
* Android version/Browser: firefox, chrome

DavidBaldsiefen:
Fixes #4936 

The whole epsg4326_wkt string is now also configurable, to give users more flexibility when using custom shapefile sources

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/region/GeoShapeProviderEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/region/GeoShapeProviderEjb.java
@@ -134,7 +134,7 @@ public class GeoShapeProviderEjb implements GeoShapeProvider {
 		}
 
 		Point polygonCenter = regionMultiPolygons.get(region).getCentroid();
-		return new GeoLatLon(polygonCenter.getX(), polygonCenter.getY());
+		return new GeoLatLon(polygonCenter.getY(), polygonCenter.getX());
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/region/GeoShapeProviderEjb.java
```diff
@@ -167,22 +164,23 @@ public GeoLatLon getCenterOfDistrict(DistrictReferenceDto district) {
...
-			loadRegionData(countryName);
-			loadDistrictData(countryName);
+			loadRegionData(countryName, WKT_String);
+			loadDistrictData(countryName, WKT_String);
 		}
 		buildCountryShape();
 	}
 
-	private void loadRegionData(String countryName) {
+	private void loadRegionData(String countryName, String WKT_String) {
```
vidi42, line 183: I think it's enough naming this parameter as `wkt` since the `_String` suffix seems redundant and it matches the naming of the parameter used by `CRS.parseWKT`

sormas-base/setup/sormas.properties
```diff
@@ -146,6 +146,10 @@ sms.auth.secret=
 #geocodingLongitudeJsonPath=$.features[0].geometry.coordinates[0]
 #geocodingLatitudeJsonPath=$.features[0].geometry.coordinates[1]
 
+# EPSG4326 WKT String used to interpret GEO-Coordinate data (e.g. Axis order)
+# default: GEOGCS[\"WGS 84\",DATUM[\"WGS_1984\",SPHEROID[\"WGS 84\",6378137,298.257223563,AUTHORITY[\"EPSG\",\"7030\"]],AUTHORITY[\"EPSG\",\"6326\"]],PRIMEM[\"Greenwich\",0,AUTHORITY[\"EPSG\",\"8901\"]],UNIT[\"degree\",0.01745329251994328,AUTHORITY[\"EPSG\",\"9122\"]],AXIS[\"Long\",EAST],AXIS[\"Lat\",NORTH],AUTHORITY[\"EPSG\",\"4326\"]]
+# geocodingEPSG326_WKTstring=
```
vidi42, line 151: missing a `4`, Should be `geocodingEPSG4326_WKTstring` but I also suggest keeping the name for this property simple as `geocodingEPSG4326_WKT`

vidi42 (changes requested):

vidi42 (approved):

PR #5167, opened 2021-04-21
#5099 prevent case bulkactions from entering infinite loop

### LINKED ISSUE #5099: Setting caseOutcome to DECEASED continuously produces very large logs
### Bug Description
It occurred that setting caseOutcome to DECEASED produces very large log files (gb/min) and results in the server having to be restarted.

Thus far, we could only reproduce this by bulk editing cases and setting them to DECEASED, however, we suspect that this also can be reproduced with individual cases since it was reported by users on productive systems.

### Steps to Reproduce

1. Enter the bulk mode and select many cases
2. Edit all of them and set the caseOutcome to DECEASED
3. Set it for individual cases (exact reproduction steps still have to be identified)

### Expected Behavior
Error should result in an error message. Logs should only be written once when an error occurs and don't loop.
### System Details

* Device:
* SORMAS version: 1.58

### Additional Information
This error message seems to occur very often:

[stackoverflowerror.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/6317948/stackoverflowerror.txt)

@nliakm

DavidBaldsiefen:
Fixes #5099 

See Issue Comment for a more detailed description of the fix

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -1918,8 +1918,7 @@ public class CaseFacadeEjb implements CaseFacade {
 					return;
 				}
 				// calculate classification
-				List<PathogenTestDto> pathogenTestDtos =
-						pathogenTests.stream().map(PathogenTestFacadeEjbLocal::toDto).collect(Collectors.toList());
+				List<PathogenTestDto> pathogenTestDtos = pathogenTests.stream().map(PathogenTestFacadeEjbLocal::toDto).collect(Collectors.toList());
 				CaseDataDto newCaseDto = toDto(associatedCase);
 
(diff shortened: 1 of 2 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2334,8 +2333,7 @@ private void updatePersonAndCaseByOutcome(CaseDataDto existingCase, Case newCase
 			}
 
 			if (newCase.getOutcome() == CaseOutcome.DECEASED) {
-				if (newCase.getPerson().getPresentCondition() != PresentCondition.DEAD
-					&& newCase.getPerson().getPresentCondition() != PresentCondition.BURIED) {
+				if (!newCase.getPerson().getPresentCondition().isDeceased()) {
```
DavidBaldsiefen, line 2336: Please don't merge yet, I am currently investigating why this causes a potential NullPointerException

MateStrysewske (approved):

PR #5178, opened 2021-04-21
ignore (sub)continent value changes if fields are invisible (#4981)

### LINKED ISSUE #4981: Country lists only display countries in one subcontinent
### Bug Description
When selecting a country in the event section, only the countries of a single subcontinent is displayed. This seems to be the subcontinent that either the server locale is set to or that is deposited as country.name in the properties.
### Steps to Reproduce
1. Import all default infrastructure data
2. Open or create an event
3. Check countries in the drop-down list

### Expected Behavior
The full country list should be displayed in events since we don't show continent and subcontinent fields there.
### Screenshots
countryListEvents

### System Details
* Device:
* SORMAS version: 1.58.1

syntakker:
Fixes #4981

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/util/InfrastructureDaoHelper.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/util/InfrastructureDaoHelper.java
@@ -15,6 +15,10 @@
 
 package de.symeda.sormas.app.util;
 
+import static android.view.View.GONE;
+import static android.view.View.VISIBLE;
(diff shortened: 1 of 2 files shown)


syntakker:
@MateStrysewske this is the second best solution, so we have a fix for the release.
My gut feeling is that the cascade of ValueChangeListeners that are added and removed to invisible fields can be substantially simplified.
Given the central role of the location form, however, I'll avoid a rushed refactoring here. I'll further look into what can be done.

MateStrysewske (approved):

PR #5189, opened 2021-04-22
#4589 extend exposure table

### LINKED ISSUE #4589: Improve exposure table display [1]
### Feature Description
(cases/contacts) > epiData > exposures
### Problem Description
When exposures are entered in the table, the display of entries on the form do not provide detailed information. Therefore, users usually have to open the individual exposures to really understand the content.

**Current display:**
Screen Shot 2021-03-02 at 2 38 53 PM
The description shows what has been entered in the respective exposure.
### Proposed Change

- Display the lower hierarchical entries in the overview table if available. This should include:
    - "Type of activity" (gatheringDetails > gatheringType > exposureType, habitationDetails > habitationType > exposureType, exposureTypeDetails > exposureType). 
    - "type of place" 
        -  Facility (facility, facilityDetails > facilityType > typeOfPlace). When facility is available also keep the facilityType displayed.
        - Means of transport (Values, MeansOfTransportDetails > "Means of transport")
        - (typeOfPlaceDetails > "other")
- In the case of animal contacts, put the type of animal in parentheses after animal contact since the type of animal by itself does not make sense as a type of activity.
- Add "Role" (exposureRole) to the overview table.

**Improved display:**
Screen Shot 2021-03-02 at 3 13 34 PM
The description shows what has been entered in the respective exposure.

DavidBaldsiefen:
Closes #4589

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
@@ -126,25 +126,20 @@ public class ExposuresField extends AbstractTableField<ExposureDto> {
 	private void addGeneratedColumns(Table table) {
 		table.addGeneratedColumn(COLUMN_EXPOSURE_TYPE, (Table.ColumnGenerator) (source, itemId, columnId) -> {
 			ExposureDto exposure = (ExposureDto) itemId;
-			String exposureString = "";
-			exposureString =
+			String exposureString =
 				ExposureType.OTHER != exposure.getExposureType() ? exposure.getExposureType().toString() : exposure.getExposureTypeDetails();
 
 			// if possible, always display "lowest-level" activity type (e.g. show type of gathering instead of just "gathering")
 			if (exposure.getExposureType() == ExposureType.GATHERING && exposure.getGatheringType() != null) {
 				exposureString += " - " + (exposure.getGatheringType() != GatheringType.OTHER
 					? exposure.getGatheringType().toString()
-					: ((exposure.getGatheringDetails() != null && !exposure.getGatheringDetails().isEmpty())
-						? exposure.getGatheringDetails()
-						: GatheringType.OTHER.toString()));
+					: (StringUtils.isNotEmpty(exposure.getGatheringDetails()) ? exposure.getGatheringDetails() : GatheringType.OTHER.toString()));
 			}
 
 			if (exposure.getExposureType() == ExposureType.HABITATION && exposure.getHabitationType() != null) {
 				exposureString += " - " + (exposure.getHabitationType() != HabitationType.OTHER
 					? exposure.getHabitationType().toString()
-					: (exposure.getHabitationDetails() != null && !exposure.getHabitationDetails().isEmpty()
-						? exposure.getHabitationDetails()
-						: HabitationType.OTHER.toString()));
+					: (StringUtils.isNotEmpty(exposure.getHabitationDetails()) ? exposure.getHabitationDetails() : HabitationType.OTHER.toString()));
 			}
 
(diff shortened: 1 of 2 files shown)


FredrikSchaeferVitagroup:
I noticed that in my browser, the TYPE OF ACTIVITY now defaults to quite a long widht. 

image

Can this be changed? I have heard that some users have a low screen resolution and the exposures table already is a little troublesome in this regard.

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
```diff
@@ -117,9 +126,33 @@ protected void updateColumns() {
 	private void addGeneratedColumns(Table table) {
 		table.addGeneratedColumn(COLUMN_EXPOSURE_TYPE, (Table.ColumnGenerator) (source, itemId, columnId) -> {
 			ExposureDto exposure = (ExposureDto) itemId;
-			String exposureString =
+			String exposureString = "";
```
FredrikSchaeferVitagroup, line 129:
```suggestion
			String exposureString;
```

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
```diff
@@ -94,13 +95,21 @@ protected void updateColumns() {
 			table.setVisibleColumns(
 				EDIT_COLUMN_ID,
 				COLUMN_EXPOSURE_TYPE,
+				COLUMN_EXPOSURE_ROLE,
```
MateStrysewske, line 98: Exposure role doesn't need a generated column because all you're doing here is to make a toString. You can remove `COLUMN_EXPOSURE_ROLE` (and the code to add the generated column) and just replace it with `ExposureDto.EXPOSURE_ROLE`

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
MateStrysewske, line 129: Unnecessary statement because the string is immediately overwritten again

sormas-ui/src/main/java/de/symeda/sormas/ui/exposure/ExposuresField.java
```diff
@@ -145,11 +178,42 @@ private void addGeneratedColumns(Table table) {
...
+			ExposureDto exposure = (ExposureDto) itemId;
+			return exposure.getExposureRole() != null ? exposure.getExposureRole().toString() : "";
+		});
+
 		table.addGeneratedColumn(COLUMN_TYPE_OF_PLACE, (Table.ColumnGenerator) (source, itemId, columnId) -> {
 			ExposureDto exposure = (ExposureDto) itemId;
-			return exposure.getTypeOfPlace() != null
-				? TypeOfPlace.OTHER != exposure.getTypeOfPlace() ? exposure.getTypeOfPlace().toString() : exposure.getTypeOfPlaceDetails()
-				: "";
+			String typeOfPlaceString = "";
```
MateStrysewske, line 188: Unnecessary to initialise the String with "" here because it's always overwritten; `String typeOfPlaceString;` is sufficient

MateStrysewske (changes requested):
I'd suggest to use StringUtils.isNotEmpty in all place where you check for both != null and !isEmpty (easier to read and more comfortable to write)

MateStrysewske:
> I noticed that in my browser, the TYPE OF ACTIVITY now defaults to quite a long widht.
> 
> image
> 
> Can this be changed? I have heard that some users have a low screen resolution and the exposures table already is a little troublesome in this regard.

The column will automatically shrink if the content of the other columns requires more space

MateStrysewske (approved):

PR #5213, opened 2021-04-22
#5188 Add indicator on locationform regarding facilities and districts

### LINKED ISSUE #5188: Make facility selection on LocationForm more intuitive
### Situation Description
Currently, if a user selects a correct `facilityTypeGroup` and `FacilityType` on the LocationForm, a list of available Facilities will only appear if a district is specified:

image

Also, it is not possible to define a custom facility with these presets.

### Feature Description

- [ ] Add a visual indicator to the facility field if no district is selected, that informs the about the reason
- [ ] ~~Always enable the selection of "Other Facility"~~

DavidBaldsiefen:
Closes #5188


DavidBaldsiefen:
Converted to draft as i18n is still missing. Whoops!

MateStrysewske (approved):

PR #5215, opened 2021-04-23
5154 concurrent lab message fetch

### LINKED ISSUE #5154: [Lab Messages] Users can simultaneously fetch new messages
### Bug Description

Two different users with according rights can fetch lab messages at the same time. If both fetch processes run through, the same messages are stored twice in the sormas-db.

### Steps to Reproduce

1. Log in with two different users having right to fetch lab messages
2. With both users, click the `fetch lab messages` button in the lab message directory. Both clicks should happen in the same few seconds.
3. When new messages can be downloaded, they are all downloaded and stored twice.

### Expected Behavior

This should be avoided.

Proposal: When another user is already fetching lab messages, a popup shall show informing: `Another user seems to be fetching lab messages at the moment. If you choose to proceed, this may result in duplicate lab messages.` The user can then either click a `proceed` button that will lead to continuation of the lab message fetch, or click a `cancel` button or cancel via a window control button, which will both cancel the lab message fetch and close the popup.

### System Details

* Device: PC
* SORMAS version: 1.58
* Server URL: Local system
* User Role: National User

### Additional Information

Hint for delopers: a started and not yet finished lab message fetch should already store a systemEvent of type `FETCH_LAB_MESSAGES` with status `STARTED` in the sormas-db. This can be used to determine if another user is fetching lab messages at the moment.

Alexandru-Pirvu:
Closes #5154


sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessagesView.java
```diff
@@ -81,7 +84,7 @@ public LabMessagesView() {
 		addHeaderComponent(samplesViewSwitcher);
 
 		addHeaderComponent(ButtonHelper.createIconButton(Captions.labMessageFetch, VaadinIcons.REFRESH, e -> {
-			fetchLabmessages();
+			checkForConcurrentFetch();
```
FredrikSchaeferVitagroup, line 87: The method name seems a little misleading to me. checkForConcurrentFetch(...) does not just check, it also initiates the fetch. That's why I propose to either add some comment clarifying that or remame thie method to something like initiateLabMessageFetch(...)

sormas-api/pom.xml
```diff
@@ -2,7 +2,7 @@
 	<parent>
 		<groupId>de.symeda.sormas</groupId>
 		<artifactId>sormas-base</artifactId>
-		<version>1.59.0-SNAPSHOT</version>
+		<version>1.60.0-SNAPSHOT</version>
```
FredrikSchaeferVitagroup, line 5: Why did you manually change that? Rebasing or merging should be the way to go here.

FredrikSchaeferVitagroup (changes requested):

sormas-api/pom.xml
FredrikSchaeferVitagroup: This applies to all version changes.

sormas-api/src/main/resources/strings_en-AF.properties
```diff
@@ -1148,4 +1148,4 @@ DefaultPassword.ownUserNewPasswordSetHints=Your account has been successfully up
 DefaultPassword.otherUsersNewPasswordSetHints=Your accounts have been successfully updated. Make sure to copy the new passwords and to notify the affected users about the password change. The new passwords are displayed only once.
 DefaultPassword.newPassword=New Password
 DefaultPassword.newPasswordPlaceholder=New Password\: %s
-DefaultPassword.unchanged=- unchanged -
\ No newline at end of file
+DefaultPassword.unchanged=- unchanged -
```
FredrikSchaeferVitagroup, line 1151: As discussed earlier, this should also be removed from all but the strings.properties. Please always check the diff view (the files changed) yourself, before requesting a review!

FredrikSchaeferVitagroup (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjb.java
```diff
@@ -314,14 +313,15 @@ public long count(LabMessageCriteria criteria) {
 	}
 
 	/**
-	 * The creation of the currentSystemEvent is in this method. All the rest is outsourced to another method,
-	 * because it shall be done in one transaction. In case of uncaught exceptions, this leaves the systemEvent with status STARTED
-	 * and falls back to standard exception handling.
+	 * This method marks the previously started system events as UNCLEAR and creates a new event with status STARTED.
```
FredrikSchaeferVitagroup, line 316:
```suggestion
	 * This method marks the previously started system events as UNCLEAR (if any exist) and creates a new event with status STARTED.
```

FredrikSchaeferVitagroup (changes requested):

FredrikSchaeferVitagroup (approved):

PR #5244, opened 2021-04-27
#4482 add vaccination column and filter

### LINKED ISSUE #4482: Add vaccination status filters and columns to case and contact directories and event participants view [1]
### Situation Description
The GAs urgently need a filter and sorting option on the vaccination status of cases, contacts and event participants. 
### Feature Description
Add a column and a filter field at the detailed view of case and contact directory for the vaccination status. Add an additional filter field and column for the vaccination status at the event participants view for events. 
Might look like this for cases:

Might look like this for contacts:
image
Might look like this for event participants in events:
image

* Add a "Vaccination status" filter to the filter forms in the case, contact and event participant directories (not in the first row for cases and contacts)
* Add a "Vaccination status" column to the case and contact directories (to the right of symptom journal status)  and the event participant list (rightmost column)

DavidBaldsiefen:
- Add a "Vaccination"-Filter to Case, Contact and EventParticipant directories
- Add a "Vaccination"-Column to each directory

The captioning may still be up for discussion. I would propose the short version of this PR (Just "Vaccination" instead of "Vaccination status for this disease", as it is more readable and compact). However, `VaccinationInfo.vaccinationShort` may not be the ideal place for this caption, it might be better to just add a general `Vaccination` or `VaccinationStatus` entry to `Captions.java`

Closes #4482


MateStrysewske (changes requested):
Extend the width of the vaccination status filters to 140 (to be in line with the other filters and give the caption a bit more space)

MateStrysewske (approved):

PR #5248, opened 2021-04-27
#4437 Added disease variant filter and captions to contact title and associated cases view

### LINKED ISSUE #4437: Add disease variant filter to the contact directory and variant to source case box [1]
### Situation Description
Users cannot filter for disease variants on the contact directory. This, however, might be desired when trying to filter for contacts that took place with source cases with disease variants due to higher infectiousness etc.
### Feature Description

- Add the disease variant filter to the contact directory to allow users to further filter by not only disease but also disease variant.
- Add the disease variant of the source case in parentheses after the disease in the case box on the contact information.
- The disease variant filter is only enabled, when a disease with variants is selected

### Additional Information
Screen Shot 2021-02-17 at 1 54 00 PM
Screen Shot 2021-03-24 at 12 27 57 PM

Alexandru-Pirvu:
Fixes #4437


sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactsFilterForm.java
```diff
@@ -347,6 +353,18 @@ protected void applyDependenciesOnFieldChange(String propertyId, Property.ValueC
 			populateSurveillanceOfficersForDistrict(district);
 			break;
 		}
+		case ContactIndexDto.DISEASE: {
+			Disease disease = (Disease) event.getProperty().getValue();
+			ComboBox diseaseVariantField = getField(ContactCriteria.DISEASE_VARIANT);
+			if (disease != null) {
+				List<DiseaseVariantReferenceDto> diseaseVariants = FacadeProvider.getDiseaseVariantFacade().getAllByDisease(disease);
+				FieldHelper.updateItems(diseaseVariantField, diseaseVariants);
+				diseaseVariantField.setEnabled(true);
```
vidi42, line 362: The `diseaseVariantField` should be set to `enabled` only if the `diseaseVariants` is not empty.

vidi42 (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactsFilterForm.java
Alexandru-Pirvu: The other dependent fields in the form are enabled even though there are no "child" entries. Also, enabling results in the field description to be shown. Otherwise, it is just a blank, disabled field, which does not give any feedback to the user.

vidi42: I understand, however this behavior would be more consistent with the requirement of the ticket:
> The disease variant filter is only enabled, when a disease with variants is selected
and with the way Disease Variant filter is working for cases. see `de.symeda.sormas.ui.caze.CaseFilterForm:555`

vidi42 (approved):

PR #5281, opened 2021-04-29
3254 decouple case jurisdiction

### LINKED ISSUE #3254: Decouple Place of stay from the responsible jurisdiction from cases [13]
### Situation Description
In German health departments - but potentially also in other countries - it's possible that the place of stay of a case (e.g. the hospital they haven been admitted to) does not match its responsible jurisdiction (e.g. because the person has been submitted to a hospital in a different district than their home district).

### Feature Description
* Add new fields to cases:
    * `responsibleRegion` (caption: Responsible region)
    * `responsibleDistrict` (caption: Responsible district)
    * `responsibleCommunity` (caption: Responsible community)
* Change the caption of the existing fields by removing the "Responsible" keyword from region, district and community
* In both the case creation and case edit forms (app and web), move the "Place of stay" field to above the region, district and community

#### Case edit form/fragment
* Add a heading above the "Place of stay" field that is also called "Place of stay" and remove the caption from the "Place of stay" field
* Add a heading below the facility fields that is called "Responsible jurisdiction"
* Add a checkbox below the heading that is called "Responsible jurisdiction of this case differs from its place of stay"
    * Checked by default when a responsible region has been entered, unchecked if not
    * When unchecked, the three new fields are disabled; when checked, they can be edited
    * When checked, responsible region and district are mandatory
* Add the three new fields below this heading

#### Backend
* Every piece of code that currently checks the region, district or community of a person in terms of visibility, edit rights or anything related now needs to check whether the responsible region, district or community have been filled in, and if so, use this information instead of the region, district and community associated with the place of stay

leventegal-she:
Fixes #3254


sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseEditFragment.java
```diff
@@ -119,16 +121,24 @@
 
 	private List<Item> caseConfirmationBasisList;
 
+	private boolean isDifferentJurisdiction;
```
MateStrysewske, line 124: Rename to "differentJurisdiction" (I suggest to not name boolean fields starting with "is" - the boolean type already suggests that, and it makes naming getters and setters pretty problematic; we had issues with that in the past)

sormas-app/app/src/main/java/de/symeda/sormas/app/caze/read/CaseReadFragment.java
```diff
@@ -53,15 +53,22 @@
 
 	private Case record;
 	private CaseConfirmationBasis caseConfirmationBasis;
+	private boolean isDifferentJurisdiction;
```
MateStrysewske, line 56: See above

sormas-app/app/src/main/java/de/symeda/sormas/app/util/JurisdictionHelper.java
```diff
@@ -205,4 +212,27 @@ public static SampleJurisdictionDto createSampleJurisdictionDto(Sample sample) {
 
 		return jurisdiction;
 	}
+
+	private static ResponsibleJurisdictionDto createResponsibleJurisdiction(Case caze){
```
MateStrysewske, line 216: Code formatting is missing in this class

sormas-app/app/src/main/res/layout/fragment_case_read_layout.xml
```diff
@@ -412,6 +412,40 @@
 
                 </LinearLayout>
 
+                <TextView
```
MateStrysewske, line 415: Instead of always showing these fields (including the heading), only show them if the responsible jurisdiction is actually filled in; this also allows you to get rid of the differentJurisdiction ControlTextReadField. Easiest way is probably to wrap all of this information in a layout and use the goneIfEmpty parameter with responsibleRegion (since there should be no way to have a responsible district and community if the region is empty)

sormas-app/app/src/main/res/values/strings.xml
```diff
@@ -386,6 +386,8 @@
     <string name="heading_campaign_form_data_read">Read Campaign Data</string>
     <string name="heading_campaign_form_data_summary">Campaign Data Summary</string>
     <string name="heading_campaign_form_meta_select">Select Campaign Form</string>
+    <string name="heading_case_place_of_stay">Place of stay</string>
```
MateStrysewske, line 389: Use the headings from I18nProperties instead

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseJurisdictionPredicateValidator.java
```diff
@@ -0,0 +1,90 @@
...
+			cb.equal(joins.getRegion().get(Region.ID), currentUser.getRegion().getId()),
+			cb.equal(joins.getResponsibleRegion().get(Region.ID), currentUser.getRegion().getId()));
+	}
+
+	@Override
+	protected Predicate whenDistrictLevel() {
+		return CriteriaBuilderHelper.or(
+			cb,
+			cb.equal(joins.getDistrict().get(District.ID), currentUser.getDistrict().getId()),
+			cb.equal(joins.getResponsibleDistrict().get(District.ID), currentUser.getDistrict().getId()));
```
MateStrysewske, line 70: I wonder whether this will return true if user district == null and responsible district of the case == null? Because that would give the user access to the case even though they shouldn't have it; I don't know how exactly this works internally - can you confirm that this will still return false? Needs to be tested in any case

sormas-backend/src/test/java/de/symeda/sormas/backend/caze/CaseFacadeEditRightsTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+		CaseDataDto editableCase = createCase(rdcf2, districtUser1);
+		CaseDataDto notEditableCase = createCase(rdcf1, districtUser1);
+
+		assertThat(getCaseFacade().isCaseEditAllowed(editableCase.getUuid()), is(true));
+		assertThat(getCaseFacade().isCaseEditAllowed(notEditableCase.getUuid()), is(false));
+	}
+
+	@Test
+	public void testCanEditResponsibleJurisdiction() {
```
MateStrysewske, line 111: => testCanEditByResponsibleJurisdiction

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskEditForm.java
```diff
@@ -241,6 +244,40 @@ protected void addFields() {
 		});
 	}
 
+	private List<DistrictReferenceDto> getCaseDistricts(CaseDataDto caseDto) {
+
+		if (caseDto.getResponsibleDistrict() != null && caseDto.getDistrict() == null) {
```
MateStrysewske, line 249: This shouldn't be possible under any circumstances - is there a specific reason why you added this check?

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskEditForm.java
```diff
@@ -241,6 +244,40 @@ protected void addFields() {
...
+		}
+
+		districts.add(caseDto.getDistrict());
+
+		return districts;
+	}
+
+	private List<RegionReferenceDto> getCaseRegions(CaseDataDto caseDto) {
+
+		if (caseDto.getResponsibleRegion() != null && caseDto.getRegion() == null) {
```
MateStrysewske, line 266: See above

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseJurisdictionPredicateValidator.java
leventegal-she: Normally user district should not be null because this method used for users with role having DISTRICT jurisdiction level. 
And if the user is wrongly setup this code would crash when getting the district id by `currentUser.getDistrict().getId()`.

So in terms of data protection we are safe :)

MateStrysewske (changes requested):
Played around with the new fields, tested it, and didn't find any problem so far. Looks really solid! Only two things I've noticed:

* Caption for place of stay field is missing in the case edit form in the web app; this looks weird because it still has the required indicator. Please re-add the caption (even though it's basically a duplicate of the heading)
* Responsible jurisdiction needs to be added to at least the detailed and custom case export; ideally also to the detailed index view

MateStrysewske (approved):

PR #5284, opened 2021-04-29
#4175: Make event report date editable

### LINKED ISSUE #4175: Make event report date editable [0.5]
### Feature Description
Events have both a report date as well as an event date. Unlike the event date, the report date is not editable.

### Problem Description
Since the event date is not mandatory, we have to fall back to the report date when we want to filter events by date, e.g. for the dashboard. This is problematic for things like retrospective data.

### Proposed Change
- [x] Make the event report date editable. It needs to be a mandatory field to make sure that it can't be empty.
- [x] Add a `validate` method to `EventFacadeEjb` that is called during `saveEvent` that ensures that the report date is not empty (similar process as for cases)
- [x] Add date validation: 
    - Date of report <= Current date
    - Event start date <= Date of report

ChristopherRiedel:
Fixes #4175


alexcaruntu-vita (changes requested):
Shouldn't all these be implemented for mobile app as well?

ChristopherRiedel:
> Shouldn't all these be implemented for mobile app as well?

true, I overlooked that

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
```diff
@@ -388,13 +388,6 @@ protected void addFields() {
 		quarantineFrom = addField(CaseDataDto.QUARANTINE_FROM, DateField.class);
 		quarantineTo = addDateField(CaseDataDto.QUARANTINE_TO, DateField.class, -1);
 
-		quarantineFrom.addValidator(
-			new DateComparisonValidator(
-				quarantineFrom,
-				quarantineTo,
-				true,
-				false,
-				I18nProperties.getValidationError(Validations.beforeDate, quarantineFrom.getCaption(), quarantineTo.getCaption())));
```
DavidBaldsiefen, line 397: Why is this done here?

alexcaruntu-vita: Maybe by mistake. @ChristopherRiedel?

alexcaruntu-vita (approved):
I'm ok with all the changes. The PR can be merged after we clarify the question raised by @DavidBaldsiefen.

PR #5299, opened 2021-05-03
Add a link to the associated entity in Task notifications #4734

### LINKED ISSUE #4734: Task Notification: include the identifier of the event, case or contact in the form of a hypertext link
### Situation Description
When a task is assigned to a SORMAS user, a notification can be sent to him. To date, this notification specifies the nature of the task and the identifier of the case / contact / event associated with the task.
In the context of an epidemic recovery, the number of notifications sent can quickly become large and users waste a lot of time going from messaging to SORMAS to identify the right event to investigate.

### Feature Description
In order to facilitate access to the events to be investigated (but also to the case and contact), it is proposed to replace the identifier with a hypertext link allowing direct access to the event sheet concerned by the task.

If :
- the user is connected to SORMAS, he will be directed directly to the file concerned by the task
- the user is not connected to SORMAS, he will first be oriented towards the authentication windows and then, after having entered the login and password, on the form concerned by the task.

### Additional Information
@ftavin @bernardsilenou @emond-fuseau

aponcon-atolcd:
Fixes #4734


sormas-backend/src/main/java/de/symeda/sormas/backend/task/TaskFacadeEjb.java
```diff
@@ -700,7 +703,12 @@ public void sendNewAndDueTaskMessages() {
 						: String.format(
 							I18nProperties.getString(MessagingService.CONTENT_TASK_START_SPECIFIC),
 							task.getTaskType().toString(),
-							context.toString() + " " + DataHelper.getShortUuid(associatedEntity.getUuid()));
+							context.toString() + (associatedEntity != null ? (" " + DataHelper.getShortUuid(associatedEntity.getUuid())) : ""));
+
+					String associatedEntityLinkMessagePart = getAssociatedEntityLinkMessagePart(context, associatedEntity);
```
BarnaBartha, line 708: extract these two additions (this and on line 744) that are the same to a method, for eg. appendAssociatedEntityLink(...)

sormas-backend/src/main/java/de/symeda/sormas/backend/task/TaskFacadeEjb.java
```diff
@@ -763,6 +776,37 @@ private void validate(TaskDto task) throws ValidationRuntimeException {
...
+		String uiUrl = configFacade.getUiUrl();
+		if (uiUrl == null) {
+			return null;
+		}
+
+		StringBuilder uiUrlBuilder = new StringBuilder(uiUrl);
+		if (!uiUrl.endsWith("/")) {
+			uiUrlBuilder.append("/");
+		}
+		return uiUrlBuilder.append("#!").append(taskContext.getUrlPattern()).append("/data/").append(uuid).toString();
```
BarnaBartha, line 807: this (being in the backend module) in combination with the TaskContext (api module) enum extension contain a non-typesafe dependency to the UI, and thus changes in the UI (or the complete change of the UI with the angular UI) can break this
Indeed having a solution for this is out of scope for this ticket but a comment would be good to at least increase the chances of not having an issue in the future

BarnaBartha (changes requested):

MateStrysewske:
@aponcon-atolcd Just a quick note, clicking on the small sync icon next to the reviewer's name in the top right requests a review from that person, which makes it easier for them to realise that the PR has been changed :) I suggest to have another look at the updated Contribution guidelines, there's a new section about pull requests in there that covers this topic among others

PR #5300, opened 2021-05-03
Add the possibility to disable Case and Contacts sub tabs from the configuration file #4733

### LINKED ISSUE #4733: Make configurable the tabs of the cases, contacts or events sections
### Situation Description
The CNIL asks the Ministry to limit free text fields which increase the risk of entering personal and sensitive data. During its inspection, it issued a recommendation (which it would be good to be able to implement) consisting in hiding the tabs of the cases / contacts and event parts that are not used (eg: hospitalization, clinical course, etc.) in order to control this risk.

### Feature Description

We suggest making the tabs that each user want to display or  to hide, configurable in a configuration file.

We could imagine that a certain number of tabs are always displayed and not configurable (ex : case or person case) but that the display of the other tabs are configurable (like what exists for actions of the event for example).

Below is the list of tabs that we propose to make configurable:

- CAS: Hospitalization, symptoms, epidemiological data, therapy, clinical course, follow up

- CONTACT: epidemiological data, follow up visits

- EVENT: event actions

### Additional Information
@ftavin @bernardsilenou @emond-fuseau @olivierleuci

aponcon-atolcd:
Fixes #4733


MateStrysewske (changes requested):
@aponcon-atolcd Hey! Sorry that I came up with this so late and that it causes additional work, but I think these configurations would be better suited for the featureconfiguration table in the database (given that they're basically true/false properties) - could you do that change? So basically add them as FeatureTypes instead, and then use the FeatureConfigurationFacade methods instead of the ConfigFacade (will also mean less code overall)

sormas-api/src/main/java/de/symeda/sormas/api/feature/FeatureType.java
```diff
@@ -63,6 +63,16 @@
 		new FeatureType[] {
 			CASE_SURVEILANCE }),
 
+	// SHOW/HIDE VIEW TAB FEATURES
+	VIEW_TAB_CASES_HOSPITALIZATION(false, true, null),
```
MateStrysewske, line 67: For all of the new feature types, set `serverFeature` to `true` and `dependentFeatures` to `new FeatureType[] { CASE_SURVEILANCE }` or `new FeatureType[] { CONTACT_TRACING }` respectively

sormas-base/setup/sormas.properties
```diff
@@ -305,4 +305,16 @@ app.url=
 
 # JNDI name of the DEMIS adapter that is connected to this SORMAS instance. Specifies where additionally deployed modules are stored.
 # Example: java:global/sormas-demis-adapter/DemisLabResultsAdapter
-#interface.demis.jndiName=
\ No newline at end of file
+#interface.demis.jndiName=
+
+# Turn off theses properties to disable a tab in a Case view
```
MateStrysewske, line 310: These properties are no longer needed since they were moved to the FeatureType enum

MateStrysewske (changes requested):

MateStrysewske (changes requested):
Web part looks good now, can you also adjust the Android app to no longer display these tabs if the feature configurations are turned off?

aponcon-atolcd:
@MateStrysewske Sorry, I totally forgot to manage the Android app. Should be good now!

MateStrysewske (approved):

PR #5321, opened 2021-05-05
#4176 change which entities are rendered on dashboard map

### LINKED ISSUE #4176: Number of points on the dashboard map should be rendered by the time filter [2]
### Bug Description
After activating the setting to not display the map when it exceeds a defined number of points, the message indicates a very high number of points to display.
After discussion with Bernard, it seems that it is theoretically expected that the number of points to be displayed by default in the map are the cases of the last month and not all the points of the base?
The screenshot below indicates in fact 640 thousands points to display and the case directory of this test environment contains just this number of cases.

### Steps to Reproduce
1.go to the dashboard 
2.compare the number of points indicated in the message on the map with the number of cases: the total of cases is shown on the map when you click on "display" Only the points of cases recorded in a recent period (1 month?) should be displayed

### Expected Behavior
If I understood correctly, only points cases from a recent period (1 month?) should be displayed. 

### Screenshots
image

### System Details
* Device:
* SORMAS version: 1.52.1
* Android version/Browser:
* Server URL:
* User Role: admin

### Additional Information
@bernardsilenou @olivierleuci @emond @ftavin

### Developer Specifications

- Apply the Dashboard Date & DateType Selection to the Map. Add a checkbox to the map to filter the entities based on relevance instead (=> `CaseService.createCaseRelevanceFilter`)
  - Remove the Map layer date filter, as that might cause unnecessary confusion as to which dates are actually used
- Map Contacts should always be filtered based on the contacts `reportDate`, nothing else
- Use the `dashboardMapMarkerLimit` property for the maximum amount of entities to be shown on the dashboard map. **Make sure the marker counter uses the same algorithm that is also used to pull the entities** (this is what causes the bug described above)

DavidBaldsiefen:
Fixes #4176 

See developer specifications for a list of changes


MateStrysewske (changes requested):
Seems to work fine! I just realised that changing the date filters now has a direct influence on the map, but doesn't update it after clicking on "Apply filters" - could you make sure that the map is also reloaded when the dashboard filters are applied?

DavidBaldsiefen:
> Seems to work fine! I just realised that changing the date filters now has a direct influence on the map, but doesn't update it after clicking on "Apply filters" - could you make sure that the map is also reloaded when the dashboard filters are applied?

Hmm that's exactly how it should work. I tested it very quick and I am not completely sure what you mean... when applying a custom time filter, it is still necessary to click on "Apply Filters" afterwards for both the dashboard and the map. Is that what you mean?

MateStrysewske (approved):

PR #5329, opened 2021-05-05
#5251 - add import scripts for region, district & community

### LINKED ISSUE #5251: Provide SQL script for semi-automatic import of regions, districts and communities
### Problem Description
Currently this data can only be imported via csv in the frontend.
We have already provided a similar script for the import of facilities: #4728 

### Proposed Change

- [ ] Provide a script for regions
- [ ] Provide a script for districts
- [ ] Provide a script for communities
- [ ] Decide how to merge existing entries (e.g. name, external id)

### LINKED ISSUE #5317: Weekly report notification still pops up in mobile app, although feature is disabled
### Bug Description
"The mobile app now notified me to submit a SORMAS disease report, which I tapped then I was able to successfully submit a case of Anthrax to APMIS it appears."

Make sure that task notifications are also not showing when the tasks feature is disabled.

### Steps to Reproduce

1. Deactivate the weekly reporting (in the error scenario surveillance was also deactivated)
2. Login as a surveillance officer.
3. At the start of the week the app shows a notification to enter a weekly report

### System Details
* SORMAS version: 1.60.0-SNAPSHOT

BarnaBartha:
Fixes #5251

### DIFF
--- a/sormas-backend/utils/importCommunities.sql
+++ b/sormas-backend/utils/importCommunities.sql
@@ -1,0 +1,104 @@
+/*
+ *  SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ *  Copyright © 2016-2021 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program. If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+/* 1. Load CSV data into temp table */
+CREATE TEMP TABLE tmp_community
+(
+    name       varchar(255),
+    growthrate real,
+    regionname varchar(255),
+    districtname varchar(255),
+    archived   boolean,
+    externalid varchar(255)
+);
+
+/** TODO
+	Adjust the path below to the one that provides the CSV file.
+	Note that the user running the postgres services needs to have access rights.
+**/
+COPY tmp_community
+    FROM 'C:\Users\Public\Documents\communities.csv'
+    DELIMITER ';'
+    CSV HEADER;
+
+/* 2. Clean up and map data */
+/* trim whitespaces - especially important for district, district and enum values */
+UPDATE tmp_community
+SET (name, externalid, regionname, districtname) = (trim(name), trim(externalid), trim(regionname), trim(districtname));
+
+/* fill region_id and district_id */
+ALTER TABLE tmp_community ADD COLUMN region_id integer;
+ALTER TABLE tmp_community ADD COLUMN district_id integer;
+UPDATE tmp_community SET region_id = region.id FROM region WHERE region.name = regionname;
+UPDATE tmp_community SET district_id = district.id FROM district WHERE district.name = districtname AND district.region_id = tmp_facility.region_id;
+
+UPDATE tmp_community SET archived = 'FALSE' WHERE archived IS NULL;
+
+/* 3. Validation */
+DO
+$$
+    DECLARE
+        errordetails text;
+    BEGIN
+
+        /* make sure all regions and districts were found */
+        IF (SELECT count(*) FROM tmp_community WHERE region_id IS NULL OR district_id IS NULL) > 0 THEN
+            errordetails = (SELECT string_agg(externalid, ', ') FROM tmp_community WHERE region_id IS NULL OR district_id IS NULL);
+            RAISE WARNING 'Ignoring communities without region or district: %', errordetails;
+        END IF;
+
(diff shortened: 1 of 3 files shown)


sormas-backend/utils/importDistricts.sql
```diff
@@ -0,0 +1,111 @@
...
+        errordetails = (SELECT string_agg(externalid, ', ') FROM tmp_district WHERE externalidcount > 1);
+        RAISE WARNING 'Ignoring districts that are using the same externalid: %', errordetails;
+    END IF;
+
+    DELETE FROM tmp_district WHERE externalidcount > 1;
+
+END;
+$$ LANGUAGE plpgsql;
+
+/* 4. Update existing regions */
```
DavidBaldsiefen, line 88: `regions` -> `districts`

sormas-backend/utils/importDistricts.sql
```diff
@@ -0,0 +1,111 @@
...
+UPDATE district
+SET (name, epidcode, growthrate, region_id, archived)
+        = (d.name, d.epidcode, d.growthrate, d.region_id, d.archived)
+FROM tmp_district AS d
+WHERE district.externalid IS NOT NULL AND district.externalid = d.externalid;
+
+/* 5. Insert new regions */
```
DavidBaldsiefen, line 95: `regions` -> `districts`

sormas-backend/utils/importDistricts.sql
```diff
@@ -0,0 +1,111 @@
...
+    FROM (SELECT externalid, COUNT(externalid) as cnt FROM tmp_district GROUP BY externalid) calc
+    WHERE calc.externalid = tmp_district.externalid;
+
+    IF (SELECT MAX(externalidcount) FROM tmp_district) > 1 THEN
+        errordetails = (SELECT string_agg(externalid, ', ') FROM tmp_district WHERE externalidcount > 1);
+        RAISE WARNING 'Ignoring districts that are using the same externalid: %', errordetails;
+    END IF;
+
+    DELETE FROM tmp_district WHERE externalidcount > 1;
+
```
DavidBaldsiefen, line 84: I think the `epidcode` should also be required / `NOT NULL` (even though I already get an error when not providing one)

sormas-backend/utils/importRegions.sql
```diff
@@ -0,0 +1,115 @@
...
+    FROM (SELECT externalID, COUNT(externalID) as cnt FROM tmp_region GROUP BY externalID) calc
+    WHERE calc.externalID = tmp_region.externalID;
+
+    IF (SELECT MAX(externalidcount) FROM tmp_region) > 1 THEN
+        errordetails = (SELECT string_agg(externalID, ', ') FROM tmp_region WHERE externalidcount > 1);
+        RAISE WARNING 'Ignoring regions that are using the same externalid: %', errordetails;
+    END IF;
+
+    DELETE FROM tmp_region WHERE externalidcount > 1;
+
```
DavidBaldsiefen, line 87: As in `importDistricts.sql`, maybe make sure epid-code is not null

sormas-backend/utils/importRegions.sql
```diff
@@ -0,0 +1,115 @@
...
+$$
+DECLARE
+	errordetails text;
+BEGIN
+
+    /* make sure all areas and countries were found (considering that not all systems will import both) */
+    IF (SELECT count(*) FROM tmp_region WHERE tmp_region.area_id IS NULL AND tmp_region.country_id IS NULL) > 0 THEN
+        errordetails = (SELECT string_agg(externalID, ', ') FROM tmp_region WHERE area_id IS NULL AND country_id IS NULL);
+        RAISE WARNING 'Ignoring regions without area and country: %', errordetails;
+    END IF;
```
DavidBaldsiefen, line 62: This means the script won't work on systems without areas and countries. Maybe add 

```sql
IF (SELECT COUNT(*) FROM country) > 0 OR (SELECT COUNT(*) FROM area) > 0
```

around it, so that the script will still work on systems where neither property is used. If this returns false, all entries containing area/country information should be cleared, or a warning printed that those entries cannot be imported

DavidBaldsiefen (changes requested):
Looks good overall! I've added some smaller remarks. One thing I noticed is that you do not check if a district/community with the same name already exists. When adding those manually, that would throw an error in the application. Maybe it should be checked here too?

sormas-backend/utils/importCommunities.sql
```diff
@@ -0,0 +1,102 @@
...
+            RAISE WARNING 'Ignoring communities without district: %', errordetails;
+        END IF;
+
+        DELETE FROM tmp_community WHERE district_id IS NULL;
+
+/* make sure all entries have an external id */
+        IF (SELECT count(*) FROM tmp_community WHERE externalid IS NULL) > 0 THEN
+            errordetails = (SELECT string_agg(name, ', ') FROM tmp_community WHERE externalid IS NULL);
+            RAISE WARNING 'Ignoring communities without externalid: %', errordetails;
+        END IF;
```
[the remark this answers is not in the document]
BarnaBartha: these imports were done in a similar way as the facility import and
the decision to update existing entries was made to be using the externalId - thus we need the externalId with this import
this goes hand in hand with the name check - do you think it would be better to use the name instead of the externalId for merging the entries?

sormas-backend/utils/importRegions.sql
BarnaBartha: good point, ok

DavidBaldsiefen:
Looks good to me. However I was wondering, if we should add a (general) README to the `utils` folder on how to use the importscripts. Maybe that should be a different issue though. What do you think?

DavidBaldsiefen (approved):

PR #5336, opened 2021-05-06
Feature 3004 surveillance dashboard indicators update

### LINKED ISSUE #3004: Surveillance and contact dashboard indicators update [3]
### Bug Description
The indicators on the dashboard are difficult to easily understand. Some case indicators appear on contact dashboard, Some filters do not work and the time frame used to calculate  the indicators are not the same.
We need to refine their calculations and captions to give users the correct numbers.
### Steps to Reproduce
1.
2.

### Expected Behavior
- Surveillance dashboard should use the total number of cases  reported within the specified time filter provided by the user as denominator while total number of contacts for contact dashboard.
- All  indicators should be based on this set of cases
- sum of individual case classification categories == total New cases,  (this is an old issue, maybe it was not just corrected?) 
- custom filter is not respinsive on contact dashboard for example
More on screen shot
### Screenshots
image

image

### System Details
* Device: web
* SORMAS version: 1,49
* Android version/Browser:
* Server URL:
* User Role:

### Additional Information
Related: https://github.com/hzi-braunschweig/SORMAS-Project/issues/2815

**Surveillance Dashboard**
* Add the missing event states to the "New events" section; "Outbreak" should be red, change "Event" to orange and "Signal" to yellow, use `$s-neutral-color` for "Screening"
* Adjust the date filters for the events that are counted on the Dashboard if needed. The date of the event needs to be included in the selected time filter:
    * If only the start or end date of the event are specified: filter start<= event start date <= filter end
    * If both the start and end date of the event are specified: filter start <= event end date && event start date <= filter end
    * If neither the start nor the end date of the event are specified: filter start <= event report date <= filter end
* Add the missing case classifications `CONFIRMED_NO_SYMPTOMS` and `CONFIRMED_UNKNOWN_SYMPTOMS` to the "New cases" section and the epidemiological curve; these should only be visible in Germany
* Make sure that cases classified as one of these classifications are showing up as confirmed cases on the map (seems like right now they're shown as not yet classified)
* Make sure that "Cases in quarantine" only counts the cases in quarantine among the "New cases" and not based on its own query; the logic is similar to the event date filters, i.e. filter start <= quarantine start/end <= filter end if only one date is filled in or filter start <= quarantine end && quarantine start <= filter end if both dates are filled in
* Rename "Contacts converted to case" to "Cases resulting from contacts"; the number should show the amount of cases among the cases returned for "New cases" that are set as the resulting case of a contact; make sure that this is a distinct count because one case can be the resulting case of multiple contacts, but should only be counted once anyway
* Check why the "last reported district" in the disease carousel layout and the "last report" in the disease overview cards sometimes don't match; the logic should be the same for both, i.e. the district of the case with the latest report date among the cases displayed on the dashboard (see #5185)
* Exclude cases with the classification "Not a case" from all figures on the dashboard except the classification overview in the disease carousel layout (i.e. still display the "not a case" label there)
* Add "Unknown" to the epidemiological curve mode "Alive or Dead"; color grey, include all cases where person present condition is empty

**Contact Dashboard**
* Contacts should not be counted based on the new cases, but purely based on their report dates; this will make sure that contacts without source cases are also included
* Quarantine calculation needs to adhere to the same logic as for cases (see above)
* Make sure that the "Contacts per case" counts are correct; it's possible that no changes are required here, but it would be good to play around with it a bit and ensure that everything works as expected

alexcaruntu-vita:
Fixes #3004


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1144,12 +1144,38 @@ public long count(CaseCriteria caseCriteria) {
 		return Collections.emptyList();
 	}
 
-	private Predicate buildQuarantineDateFilter(CriteriaBuilder cb, Root<Case> caze, Date from, Date to) {
+	private Predicate buildQuarantineDateFilter(CriteriaBuilder cb, Root<Case> caze, Date fromDate, Date toDate) {
+		Predicate filter = null;
+		if (fromDate != null && toDate != null) {
+			filter = CriteriaBuilderHelper.and(
```
MateStrysewske, line 1150: I think these filters don't exactly match what's specified in the issue because they don't account for the situation where QUARANTINE_TO is filled in, but QUARANTINE_FROM isn't. I think we need an extra condition like this:

```
cb.or(
  cb.and(
    cb.isNull(caze.get(Case.QUARANTINE_TO)),
    cb.between(caze.get(Case.QUARANTINE_FROM), fromDate, toDate)),
  cb.and(
    cb.isNull(caze.get(Case.QUARANTINE_FROM)),
    cb.between(caze.get(Case.QUARANTINE_TO), fromDate, toDate)),
  cb.and(
    cb.greaterThanOrEqualTo(caze.get(Case.QUARANTINE_TO), fromDate),
    cb.lessThanOrEqualTo(caze.get(Case.QUARANTINE_FROM), toDate)));
```

This means that:
* If only QUARANTINE_FROM or QUARANTINE_TO are filled in, one of those need to be between start and end date
* If both are filled in, QUARANTINE_FROM needs to be before the end date and QUARANTINE_TO needs to be after the start date to make sure that at least one of the dates is between start and end date

If I'm not mistaken, that's semantically different from the current code, right? Or did I misunderstand something?

This probably needs to be adjusted for all three if cases then (even though the subsequent if cases are less complex because there's only one filter date)

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1347,9 +1369,14 @@ public Long countCasesForMap(RegionReferenceDto regionRef, DistrictReferenceDto
 		}
 
 		Expression<Number> maxReportDate = cb.max(caze.get(Case.REPORT_DATE));
+		Expression<Number> maxCreationDate = cb.max(caze.get(Case.CREATION_DATE));
 		cq.multiselect(caze.get(Case.DISEASE), districtJoin, maxReportDate);
```
MateStrysewske, line 1373: Don't you also need to select the creation date in order to be able to sort by it?

sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/DashboardDataProvider.java
```diff
@@ -150,18 +148,21 @@ private void refreshDataForSelectedDisease() {
...
-			caseCriteria.newCaseDateBetween(previousFromDate, previousToDate, newCaseDateType);
+			caseCriteria.newCaseDateBetween(previousFromDate, previousToDate);
 			setPreviousCases(FacadeProvider.getCaseFacade().getCasesForDashboard(caseCriteria));
 
 			if (getDashboardType() != DashboardType.CONTACTS) {
 				if (getCases().size() > 0) {
-					setTestResultCountByResultType(
-						FacadeProvider.getSampleFacade()
-							.getNewTestResultCountByResultType(getCases().stream().map(c -> c.getId()).collect(Collectors.toList())));
+					List<Long> list = new ArrayList<>();
```
MateStrysewske, line 160: What's the reason for turning this into a for loop?

sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/surveillance/components/disease/DiseaseOverviewComponent.java
```diff
@@ -0,0 +1,83 @@
...
+
+import de.symeda.sormas.api.disease.DiseaseBurdenDto;
+import de.symeda.sormas.ui.dashboard.surveillance.components.disease.burden.DiseaseBurdenComponent;
+import de.symeda.sormas.ui.dashboard.surveillance.components.disease.tile.DiseaseTileViewLayout;
+import de.symeda.sormas.ui.utils.ButtonHelper;
+import de.symeda.sormas.ui.utils.CssStyles;
+
+public class DiseaseOverviewComponent extends HorizontalLayout {
+
+	private static final int SHONW_DISEASES = 6;
```
MateStrysewske, line 21: Typo; maybe rename to NUMBER_OF_DISEASES_COLLAPSED

MateStrysewske (changes requested):
Things that are missing/incorrect (see development specifications in the issue for more details):

* Events need to fall back to their report date if no event date is specified
* Cases with the classification "Not a case" should not be counted e.g. in the disease layouts at the top left - they should be excluded from view AND calculations/counts EVERYWHERE except for the classification overview in the disease carousel layout; this e.g. also involves the quarantine counts, last reported district, fatality rate, etc.
* "Alive or dead" epi curve hasn't been adjusted to show cases where person present condition == empty as "Unknown"

sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/DashboardDataProvider.java
alexcaruntu-vita: It was a refactoring suggested by IDE: Lambda can be replaced with method reference. Should I revert it?

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
alexcaruntu-vita: Don't need to select it if I am using just for sorting. In fact, we could remove `maxReportDate` from `multiselect `as well.

MateStrysewske: Ah, interesting, didn't know about that! Yes, I think you could just remove maxReportDate as well then :)

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
```diff
@@ -736,33 +732,25 @@ private Predicate createEventDateFilter(CriteriaQuery<?> cq, CriteriaBuilder cb,
 		Date eventDateTo = eventCriteria.getEventDateTo();
 
 		if (eventDateType == null || eventDateType == EventCriteriaDateType.EVENT_DATE) {
+			Predicate eventDateFilter = null;
```
MateStrysewske, line 735: I think the event filters aren't correct yet; they basically need to do the same as the quarantine filters (because both event start and end date are optional, such as the quarantine start and end date); also, report date must only be used as a fallback in case NEITHER event start or end date are specified; if I'm not mistaken, the current logic will always fall back to the report date even if the event date is specified.

MateStrysewske (changes requested):

MateStrysewske (changes requested):
According to the issue description:

"Make sure that "Cases in quarantine" only counts the cases in quarantine among the "New cases" and not based on its own query" -> I just realised this still uses its own query, which is not necessary. You'll probably have to move the code that produces the counts to the dashboard data provider; the `getQuarantineDataForDashBoard` methods (both for cases and contacts) are likely not needed anymore

Additionally:
* "Alive or dead" Epi curve should not include "Not a case" cases
* Last reported district in the disease carousel layout should not include "Not a case" cases

sormas-backend/src/main/java/de/symeda/sormas/backend/dashboard/DashboardService.java
```diff
@@ -0,0 +1,371 @@
...
+				.getResultStream()
+				.collect(Collectors.toMap(tuple -> (CaseClassification) tuple[0], tuple -> ((Number) tuple[1]).intValue()));
+		} else {
+			result = Collections.emptyMap();
+		}
+
+		return result;
+	}
+
+	public List<DashboardQuarantineDataDto> getQuarantineData(DashboardCriteria dashboardCriteria) {
```
MateStrysewske, line 136: Please remove this method, add the quarantine from and to dates to the DashboardCaseDto and DashboardContactDto instead, and then calculate the quarantine data based on that directly in the DashboardDataProvider; the dataset used for the quarantine data needs to be identical to the dataset that is used in the rest of the dashboard, so there's no additional query needed.

sormas-backend/src/main/java/de/symeda/sormas/backend/dashboard/DashboardService.java
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	public long countCasesConvertedFromContacts(DashboardCriteria dashboardCriteria) {
+
+		CriteriaBuilder cb = em.getCriteriaBuilder();
+		CriteriaQuery<Long> cq = cb.createQuery(Long.class);
+		Root<Case> caze = cq.from(Case.class);
+		final CaseQueryContext caseQueryContext = new CaseQueryContext(cb, cq, caze);
+
+		Predicate filter = caseService.createUserFilter(cb, cq, caze, new CaseUserFilterCriteria().excludeCasesFromContacts(false));
```
MateStrysewske, line 203: use `true` here as well

sormas-api/src/main/java/de/symeda/sormas/api/dashboard/DashboardCriteria.java
```diff
@@ -0,0 +1,81 @@
...
+public class DashboardCriteria extends BaseCriteria implements Serializable {
+
+	private RegionReferenceDto region;
+	private DistrictReferenceDto district;
+	private Disease disease;
+	private CriteriaDateType newCaseDateType;
+	private Date newCaseDateFrom;
+	private Date newCaseDateTo;
+
+	private boolean includeNoCases;
```
MateStrysewske, line 21: Rename to `includeNotACaseClassification`

MateStrysewske (changes requested):

MateStrysewske (changes requested):
* The disease carousel layout now always includes cases with the classification "Not a case", but shouldn't do so; this affects fatalities, fatality rate, last reported district, cases in quarantine, cases placed in quarantine, but NOT cases resulting from contacts; it also affects the fatality count in the disease overview layout in the top right

MateStrysewske (approved):

PR #5375, opened 2021-05-10
Send a notification when creating an EventParticipant related to another event #4817

### LINKED ISSUE #4817: Notification to the responsible user when a participating person is part of other events
### Situation Description

Development in connection with # 4570 # 4571 # 4572 # 4573 consisting of grouping events together.

### Feature Description

Development consists of sending a notification to all event responsible user when a person participating in their event is attached to other events.
The objective is to be able to speed up the identification of chains of contamination and to take measures to limit the spread.

Eg:
Report date: 01/01/2021 => responsible user a: Event A -> Participant person α
Report date: 02/01/2021 => responsible user b: Event B -> Participant person α.

-> On 02/01/2021, the responsible users for events A and B will receive a notification informing that the person is involved in events A and B and that an exchange between user a and b should make it possible to rule on the need whether or not to link events A and B.

Content of the notification:
- person ID
- List of events to which the person is linked
- List of users responsible for the various events, recipients of the notification.

This notification will only be sent if:
- The disease is the same for the different events
- The events are reported in a space-time of 30 days

### Additional Information

aponcon-atolcd:
Fixes #4817


sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
```diff
@@ -194,6 +194,31 @@ public EventService() {
...
+		Join<Event, User> responsibleUserJoin = eventRoot.join(Event.RESPONSIBLE_USER, JoinType.INNER);
+
+		Timestamp timestamp = DateHelper.toTimestampUpper(date);
+		Predicate filter = cb.and(
+			cb.equal(personJoin.get(Person.UUID), personUuid),
+			cb.not(cb.equal(eventRoot.get(Event.UUID), excludedEventUuid)),
+			cb.equal(eventRoot.get(Event.DISEASE), disease),
+			createActiveEventsFilter(cb, eventRoot),
+			cb.or(
+				CriteriaBuilderHelper.greaterThanAndNotNull(cb, eventRoot.get(Event.START_DATE), timestamp),
```
BarnaBartha, line 212: are you certain this is what is required here?
from the ticket: 'reported in a space time of 30 days' - to me sounds like report date was in the last 30 days

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
```diff
@@ -227,9 +242,70 @@ public EventParticipantDto saveEventParticipant(EventParticipantDto dto, boolean
 		EventParticipant entity = fromDto(dto, checkChangeDate);
 		eventParticipantService.ensurePersisted(entity);
 
+		if (existingParticipant == null) {
+			// The Event Participant is newly created, let's check if the related person is related to other events
+			// In that case, let's notify corresponding responsible Users of this relation
+			notifyEventResponsibleUsersOfCommonEventParticipant(entity, event);
```
BarnaBartha, line 248: will take the opportunity here to mention that it is rather questionable the way sending notifications based on the feature type has been designed.. and should be addressed at some point
right now when creating a new event on a system where EVENT_PARTICIPANT_RELATED_TO_OTHER_EVENTS_NOTIFICATIONS feature is not enabled there is still a query, then some code that for each result of the query does some preparation, and then decides not to do anything with it

BarnaBartha (changes requested):

PR #5378, opened 2021-05-10
#5374 Set facility-contactperson visiblity-dependency

### LINKED ISSUE #5374: Contact person details are shown for location forms even with no facility type selected
### Bug Description

We have recently introduced contact person fields to facilities and locations. It makes no sense to display these fields for locations if no facility has been selected in the first place, however - this can in fact be pretty confusing. 

### Steps to Reproduce

1. Open any form containing a location, e.g. person or event form
2. See that the four new contact person fields are displayed

### Expected Behavior

Only display the new fields when at least a facility type has been selected (or if there's a "Type of place" selector, like e.g. for events, if "Facility" has been selected there). "Community contact person" does not belong to these new fields and is fine as it is.

### System Details

* Device:
* SORMAS version: 1.60.0-SNAPSHOT
* Server URL:
* User Role:

DavidBaldsiefen:
Fixes #5374

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/LocationDialog.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/LocationDialog.java
@@ -27,11 +27,9 @@ import org.apache.commons.lang3.StringUtils;
 import android.content.Context;
 import android.util.Log;
 import android.view.View;
-
 import androidx.databinding.ViewDataBinding;
 import androidx.databinding.library.baseAdapters.BR;
 import androidx.fragment.app.FragmentActivity;
-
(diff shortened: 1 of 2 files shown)


MateStrysewske (changes requested):
The logic isn't quite consistent for events; If I select "Facility" as type of place there, the contact person fields are displayed even when facilityType = null, and if I then select and de-select a facility type, they're hidden.

Ideally, for events, the fields should always be displayed when type of place = "Facility" (even if no facility type is selected), but that might be a bit too complicated because the location form doesn't know where it's embedded; so I guess it's enough if you make sure that the contact person fields aren't displayed initially when selecting the type of place and facility type is empty

MateStrysewske (approved):

Issue #5338, opened 2021-05-06
[Lab Messages] Improve the display of errors when viewing lab messages


FredrikSchaeferVitagroup:
### Feature Description

When a user tries to view a lab message, sormas currently tries to reach the adapter to retreive an HTML representation fo the lab message. When this is not possible, an xml representation of the lab message is displayed instead, sometimes along with an error message, sometimes not. This is currently confusing users.

### Problem Description

The current approach is that, when the adapter can be reached but to conversion to html fails, an error message is displayed to the user in a popup. When the adapter can not be reached, an error message is written to the logs instead. The implementation is [here](https://github.com/hzi-braunschweig/SORMAS-Project/blob/1ecc28d62ce7204a1480fb4a5b7b4f4058d23935/sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageForm.java#L53). Both approaches are not very good: 

The first approach shows a popup window every time a lab message shall be displayed as HTML. This means that, at every step of processing a lab message, the popup shows again. Users have to click away the popup several times before finishing.

The second approach leaves the user confused, not knowing whether the lab message is wrongly formatted, the system is broken or misconfigured. No indication of an error is visible.

### Proposed Change

When the conversion tp HTML fails, keep displaying the xml representation, but modify it so that on top there shows a message: 
`The lab message could not be converted to a graphical representation and therefore is shown in a raw format below.`
This message should show in red be followed by a short error description, e.g. the `The lab message adapter could not be found.`, or the `ExternalMessageResult.error` which holds a message from the adapter (if that error is not null or an empty string).

### Possible Alternatives

Another option would be to always show a popup when the conversion fails or the adapter can not be found. One would need to make sure then that the popup does not show again and again when processing a lab message. This alternative would probably be more effort. One could, for example, just trigger the HTML-conversion only once at the beginning of processing a lab message and show the popup then. For the further processing-steps, on would then need to hand over the graphical representation.

### Additional Information

There already is a related error message stored as [a validation string](https://github.com/hzi-braunschweig/SORMAS-Project/blob/1ecc28d62ce7204a1480fb4a5b7b4f4058d23935/sormas-api/src/main/java/de/symeda/sormas/api/i18n/Validations.java#L60).

PR #5386, opened 2021-05-10
4703 download lab messages

### LINKED ISSUE #4703: [DEMIS Interface] Introduce option to download lab messages [2]
### Situation Description
Users that process lab messages through the DEMIS interface would also like to have the opportunity to download these messages.
### Feature Description
Introduce the option to download the lab messages. For that, add a button to the directory for that purpose. This should independent of the status of the lab messages (processed, unprocessed). The downloaded format should ideally be PDF. Messages should always be downloadable independent of their status.

### Additional Information
Screen Shot 2021-03-09 at 8 39 01 PM

- [ ] Add errorhandling if the adapter cannot be accessed nor a proper html file was generated

vidi42:
Fixes #4703


FredrikSchaeferVitagroup:
Would it make sense to also add the download button to the [LabMessageButtonsPanel](https://github.com/hzi-braunschweig/SORMAS-Project/blob/0d1bbd4ab0634dba08803dcd91da0b9d14d44283/sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java#L826)? I know, that is not specified in the issue, but while trying out the feature I noticed that this probably makes sense. 
As a user I may not always be able to decide from the few details in the grid of the labMessagesView whether I want to download the message or not. I probably sometimes need to look into the detailed view (labMessageForm). From the detailed view, it is not a nice workflow to close the popup window, search for the lab message I just looked at again in the grid of the labMessagesView to find the appropriate button for download. It would instead be helpful to also have the button directly in the detailed view. 
After having sent the downloaded message via fax or any other well beloved technology from the middle ages, I could directly mark the message as forwarded with the button next to the download button in the detailed view.
What do you think about it?

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
```diff
@@ -898,4 +923,32 @@ private HorizontalLayout getLabMessageButtonsPanel(LabMessageDto labMessage, Run
...
+						I18nProperties.getString(Strings.headingLabMessageDownload),
+						I18nProperties.getString(Strings.messageLabMessageDownloadConvertFailed),
+						Notification.Type.ERROR_MESSAGE,
+						false).show(Page.getCurrent());
+			}
+
+		} catch (NamingException e) {
+			new Notification(
+					I18nProperties.getString(Strings.headingLabMessageDownload),
+					I18nProperties.getString(Strings.messageLabMessageDownloadConvertFailed),
```
FredrikSchaeferVitagroup, line 947: Mostly, the error messages are directly transmitted from the adapter. Why did you decide to store this message in SORMAS?

sormas-api/src/main/resources/strings.properties
```diff
@@ -969,6 +970,8 @@ messageUserSyncPartiallySuccessful = <b>Sync partially successful!</b><br/>The s
 messageUserSyncSuccessful = <b>Sync successful!</b><br/>All users have been synced. You can now close this window.
 messageUserRightsExportFailed = Please contact an admin and notify them about this problem
 messageNoLabMessagesSelected = You have not selected any lab messages
+messageLabMessageDownloadConvertFailed = Could not convert lab message to PDF
+messageLabMessageDownloadFailed = Something went wrong when trying to download the lab message
```
FredrikSchaeferVitagroup, line 974: I don't see this used anywhere?

sormas-backend/src/main/java/de/symeda/sormas/backend/user/KeycloakService.java
```diff
@@ -180,7 +180,9 @@ public void handleUserUpdateEvent(@Observes UserUpdateEvent userUpdateEvent) {
                 createUser(keycloak.get(), newUser);
             }
         } catch (Exception e) {
-            userUpdateEvent.getExceptionCallback().accept(e.getMessage());
+            if (userUpdateEvent.getExceptionCallback() != null) {
```
FredrikSchaeferVitagroup, line 183: Does this have to do anything with this issue?

sormas-api/src/main/java/de/symeda/sormas/api/i18n/Strings.java
```diff
@@ -850,6 +851,8 @@
 	String messageImportSuccessful = "messageImportSuccessful";
 	String messageIncompleteGpsCoordinates = "messageIncompleteGpsCoordinates";
 	String messageInvalidDatesLineListing = "messageInvalidDatesLineListing";
+	String messageLabMessageDownloadConvertFailed = "messageLabMessageDownloadConvertFailed";
+	String messageLabMessageDownloadFailed = "messageLabMessageDownloadFailed";
```
FredrikSchaeferVitagroup, line 855: I don't see it used anywhere?

FredrikSchaeferVitagroup (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/user/KeycloakService.java
FredrikSchaeferVitagroup: Okay then. I just wanted to know what it is there for.

sormas-api/src/main/resources/strings.properties
```diff
@@ -969,6 +970,7 @@ messageUserSyncPartiallySuccessful = <b>Sync partially successful!</b><br/>The s
 messageUserSyncSuccessful = <b>Sync successful!</b><br/>All users have been synced. You can now close this window.
 messageUserRightsExportFailed = Please contact an admin and notify them about this problem
 messageNoLabMessagesSelected = You have not selected any lab messages
+messageLabMessageDownloadFailed = Something went wrong when trying to download the lab message
```
FredrikSchaeferVitagroup, line 973: Could you maybe change this to:
```suggestion
messageExternalLabResultsAdapterNotFound = The external lab results adapter could not be found. This probably means you system is not configured correctly. Please contact you system administrator.
```
This would be a little more informative and a good foundation for #5338.

FredrikSchaeferVitagroup (changes requested):

FredrikSchaeferVitagroup:
> > Would it make sense to also add the download button to the [LabMessageButtonsPanel](https://github.com/hzi-braunschweig/SORMAS-Project/blob/0d1bbd4ab0634dba08803dcd91da0b9d14d44283/sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java#L826)? I know, that is not specified in the issue, but while trying out the feature I noticed that this probably makes sense.
> > As a user I may not always be able to decide from the few details in the grid of the labMessagesView whether I want to download the message or not. I probably sometimes need to look into the detailed view (labMessageForm). From the detailed view, it is not a nice workflow to close the popup window, search for the lab message I just looked at again in the grid of the labMessagesView to find the appropriate button for download. It would instead be helpful to also have the button directly in the detailed view.
> > After having sent the downloaded message via fax or any other well beloved technology from the middle ages, I could directly mark the message as forwarded with the button next to the download button in the detailed view.
> > What do you think about it?
> 
> @FredrikSchaeferVitagroup we discussed this when estimating the story and decided this will be out of scope for this ticket. One of the reason also being that the Lab message panel is a bit crowded with buttons at the bottom at the moment.
> This can be improved anyway and also the `Download` option can be added then, but has to be part of a different issue.

@JaquM  I have moved this to a dedicated ticket (#5411).

PR #5396, opened 2021-05-11
#5377 correctly handle pregnancy fields on gender change

### LINKED ISSUE #5377: Additional medical info trimester field for male patient
### Bug Description

At additional medical information if pregnancy selected trimester will also appear, in case sex is changed pregnancy and postpartum fields disappear but trimester stays.
### Steps to Reproduce

1. Create a new case with sex of the patient as female
2. At additional information select pregnancy as yes
3. Go to case person change the sex of the patient to male, other or unknown
4. At case additional medical information pregnancy and postpartum disappear but trimester field is still there

### Expected Behavior

Trimester field should also disappear just as pregnancy and postpartum do if the sex of patient is not female
### Screenshots

image

### System Details

* Device:Windows
* SORMAS version:  1.60.0-snapshot
* Server URL: https://test.sormas.netzlink.com/sormas-ui
* User Role: NatUser

### Additional Information

I think even if patient is female there should be an age limit since now you can have a 2 year old female patient and have the option of selecting pregnancy options in additional medical info !

DavidBaldsiefen:
Fixes #5377

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseEditFragment.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseEditFragment.java
@@ -175,6 +175,7 @@ public class CaseEditFragment extends BaseEditFragment<FragmentCaseEditLayoutBin
 		if (record.getPerson().getSex() != Sex.FEMALE) {
 			contentBinding.caseDataPregnant.setVisibility(GONE);
 			contentBinding.caseDataPostpartum.setVisibility(GONE);
+			contentBinding.caseDataTrimester.setVisiblity(GONE);
 		}
 
 		// Smallpox vaccination scar image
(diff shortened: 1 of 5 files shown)


FredrikSchaeferVitagroup:
I noticed that UI and app differ in behaviour. When the gender of a pregnant person is set to male in the UI, the values for the trimester gets set to null. Changing back the gender to female shows the field again, but without anything selsected.

Doing the same in the app shows the value previously filled in before the gender was set to male. This also applies for the pregnancy YesNo. Is this desired behaviour?

DavidBaldsiefen:
@FredrikSchaeferVitagroup I think this happens because the App only updates the underlying data if you synchronize it inbetween (specifically, `PersonFacadeEjb` is only called when data is synchronized). So I would say this is "desired" behaviour. To get the same behaviour as in the webapp, you would need to change the gender -> synchronize -> change gender again

FredrikSchaeferVitagroup (approved):

PR #5415, opened 2021-05-12
Add recommended Android Studio JDK Version to README

### LINKED ISSUE #6005: Increase documentation for android studio Java version
### Proposed Solution
Add a mention of the required JDK Version for Android Studio (JDK 8) to `DEVELOPMENT_ENVIRONMENT.MD`

DavidBaldsiefen:
Closes #6005


StefanKock (changes requested):
1. In Step 3 there is not mentioned to also provide JDK 8. I think this should be added.
2. No ticket id in the commit?

DavidBaldsiefen:
>     1. In Step 3 there is not mentioned to also provide JDK 8. I think this should be added.

You mean adding info on where to get the JDK? This would depend on the user OS after all

>     2. No ticket id in the commit?

No ticket created, should I?

MateStrysewske:
@DavidBaldsiefen I think what Stefan means is that Step 3 only instructs the user to download Java 11, not Java 8, which would be necessary to properly set the JDK in Android Studio. The guide should be extended with that information as well.

And yes, please create a task issue for this and ammend the last commit you did to include the issue number.

syntakker (approved):

Issue #5433, opened 2021-05-14
Integration the RestAPI testing in the automation framework


BojteTamas:
### Proposed Solution
After this change we can run REST API tests using the RESTASSURED library.

PR #5437, opened 2021-05-14
#5433-Integration the RestAPI testing in the automation framework

BojteTamas:
### Improvements

1. Added the ability to create REST API test using the automation framework

2. The REST API steps are implemented using Gherkin
 
3. The developer can analize the report and get the steps: 
Screenshot 2021-05-14 at 19 53 31

4.The developer can see the request and copy the curl:
Screenshot 2021-05-14 at 19 54 13

5. `./gradlew clean startTests -Dcucumber.tags="@API"` to run the REST API feature


Mihai-Bota:
LGTM!

Mihai-Bota (approved):

sormas-e2e-tests/src/main/resources/configuration/properties/environment/test-performance.properties
```diff
@@ -16,4 +16,6 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
-ENVIRONMENT_URL=https://test-performance.sormas.netzlink.com/sormas-ui/login#!cases/?includeCasesFromOtherJurisdictions=false&dateFilterOption=DATE
\ No newline at end of file
+ENVIRONMENT_URL=https://test-performance.sormas.netzlink.com
+REST_USER=SurvOff
+REST_PASSWORD=SurvOff38118
```
tamasvlad, line 21: I would not expose any credentials to a public repo

sormas-e2e-tests/src/test/resources/features/sanity/api/Person.feature
```diff
@@ -0,0 +1,6 @@
+@Sanity @Person
+Feature: Check Person RestApi endpoints
+
+  Scenario: Create a new event
+    Given API: I receive all person ids
+    Given API: I receive the person
```
tamasvlad, line 6: The test doesn't have any assert. Don't you need an assert to check that you actually received the person? To verify something specific for that person from the response(first name for eg)?

tamasvlad (changes requested):
console

There is no need to start the chromedriver when api tests are running. Also, there is an issue with the allure configuration for api tests

rdutu-vg (approved):

sormas-e2e-tests/src/main/resources/configuration/properties/environment/test-performance.properties
BojteTamas: Good point, I ask for this and it is ok, because the repo as well is open source. This will be tackled when we will user the jenkins/GitHub actions

sormas-e2e-tests/src/test/resources/features/sanity/api/Person.feature
BojteTamas: This is only to introduce the concept of API testing. You are right there is no assert at this moment

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/api/PersonSteps.java
```diff
@@ -0,0 +1,42 @@
...
+public class PersonSteps implements En {
+
+  @Inject
+  public PersonSteps(PersonsHelper personsHelper, ApiState apiState) {
+
+    When(
+        "API: I receive the person",
+        () -> {
+          List<String> personUuids = apiState.getResponse().jsonPath().get();
+          personsHelper.getPersonByUuid(personUuids.get(4717));
```
richardbartha, line 36: What is this number are there so many persons in the list what happens if this list changes ?

sormas-e2e-tests/src/test/java/org/sormas/e2etests/helpers/api/PersonsHelper.java
```diff
@@ -0,0 +1,45 @@
...
+public class PersonsHelper {
+
+  private final RestAssuredClient restAssuredClient;
+
+  @Inject
+  public PersonsHelper(RestAssuredClient restAssuredClient) {
+    this.restAssuredClient = restAssuredClient;
+  }
+
+  public void getPersonByUuid(String Uuid) {
```
richardbartha, line 36: Maybe we should only use one get method here and avoid duplications as much as possible. 
`private void get(String path) {
    restAssuredClient.sendRequest(
            Request.builder().method(Method.GET).path(path).build());
  }`

richardbartha (approved):
Looks ok, added two minor comments :)

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/api/PersonSteps.java
BojteTamas: This was only for integration of API, it is not a real case.

sormas-e2e-tests/src/test/java/org/sormas/e2etests/helpers/api/PersonsHelper.java
BojteTamas: Feel free to make a PR, and we can discuss more there

alinapopaiustina:
My concerns:
1. mainteinability of a API and UI automation tests in the same place: already mentioned by Vlad Tamas
2. to fire an API call is not enough argument for tight coupling of 2 otherwise independent projects. (lots of workarounds will be introduced just to keep them apart but together)
3. besides, to handle test data setup using API calls does not require rests test to be in the same place: why use an API test instead of an API call to do the job; API tests bring much more code than necessary.
4. Overall: I see no test data handling, no logging in place 

I understand decision is taken, just wanted to mention my concerns.

BojteTamas:
> My concerns:
> 
> 1. mainteinability of a API and UI automation tests in the same place: already mentioned by Vlad Tamas
> 2. to fire an API call is not enough argument for tight coupling of 2 otherwise independent projects. (lots of workarounds will be introduced just to keep them apart but together)
> 3. besides, to handle test data setup using API calls does not require rests test to be in the same place: why use an API test instead of an API call to do the job; API tests bring much more code than necessary.
> 4. Overall: I see no test data handling, no logging in place
> 
> I understand decision is taken, just wanted to mention my concerns.

Thank you for the information

Issue #5399, opened 2021-05-11
Create a new Task Automation flow


BojteTamas:
### Feature Description
We will cover the UI automation of the Create Task functionality.

PR #5457, opened 2021-05-18
#5399-Create a new Task Automation flow

BojteTamas:
Screenshot 2021-05-18 at 11 07 46


sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/application/tasks/CreateNewTaskSteps.java
```diff
@@ -0,0 +1,169 @@
...
+  public static final DateTimeFormatter TIME_FORMATTER = DateTimeFormatter.ofPattern("HH:mm");
+  protected static Task task;
+  private final WebDriverHelpers webDriverHelpers;
+
+  @Inject
+  public CreateNewTaskSteps(WebDriverHelpers webDriverHelpers, TaskService taskService) {
+    this.webDriverHelpers = webDriverHelpers;
+
+    When(
+        "^I create a new task with specific data$",
```
Mihai-Bota, line 44: I usually parse data via tables in steps 
https://www.baeldung.com/cucumber-data-tables
but I'm assuming that this will be for tomorrow automation presenation and it's using a more general approach than cucumber.
so it's just a note. everything looks fine.

Mihai-Bota (approved):
LGTM!

left a comment with a note.

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/application/tasks/CreateNewTaskSteps.java
BojteTamas: Regarding this, the data is generate in the task service, we want to be sure the creation of the task with all fields is possible

PR #5495, opened 2021-05-20
#2918 disable selection of pseudonymized entries in bulkaction mode

### LINKED ISSUE #2918: Bulk actions on all directories should respect pseudonimization [2]
### Bug Description
A user that has no right to sensitive data of a case, contact or event participant is able edit these entities through bulk actions. I rested only on event participant but I think this holds  cases and contacts also.
### Steps to Reproduce
1. log is as admin
2. go to an event that you are not allowed to edit participants
3. Use bulk action mode to delete  an ep
### System Details
* Device:
* SORMAS version: 1,48
* Android version/Browser:

* When the user tries to perform any bulk action (including opening the "Edit" window, i.e. do this before the window opens), check how many of the selected entries are pseudonymized; if at least one is pseudonymized, don't perform the bulk action, but instead present an error message to the user: "X of the cases/contacts/event participants/... you've selected is pseudonymized. Please deselect all pseudonymized cases/contacts/event participants/...." Do this for case, contact, event, eventparticipant directories
* Allow Admins to always Archive and/or Delete entities

DavidBaldsiefen:
Closes #2918


sormas-ui/src/main/java/de/symeda/sormas/ui/utils/FilteredGrid.java
```diff
@@ -66,6 +68,22 @@ public void setDataProvider(FetchItemsCallback<T> fetchItems, SerializableSuppli
...
 	}
 
+	@Override
+	public GridSelectionModel<T> setSelectionMode(Grid.SelectionMode selectionMode) {
+		GridSelectionModel<T> model = super.setSelectionMode(selectionMode);
+		if (selectionMode == SelectionMode.MULTI && PseudonymizableIndexDto.class.isAssignableFrom(getBeanType())) {
+			addSelectionListener(event -> {
+				event.getAllSelectedItems().forEach(item -> {
+					if (((PseudonymizableIndexDto) item).isPseudonymized()) {
+						deselect(item);
```
BarnaBartha, line 78: this may become quite annoying.. would be better to make the row disabled
i've checked and indeed with vaadin 8 this would be quite complicated to do so going with the idea that thee UI will be replaced it's ok.. @MateStrysewske what do you think?

BarnaBartha (approved):

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/FilteredGrid.java
DavidBaldsiefen: Yes I thought so too, but I also didn't find a cleaner solution

MateStrysewske: @DavidBaldsiefen @BarnaBartha If this works like I think it will, it could be pretty confusing for the users ("WHY CAN'T I SELECT THIS ROW?") - I think implementing the warning message should be enough; alternatively, you could display the warning message in a popup and provide a button "Deselect pseudonymized entries", which could be a nice comfort function

sergiupacurariu: In the event participant section using bulk action you can delete all event participants no matter they are pseudonymized or not.

DavidBaldsiefen: > In the event participant section using bulk action you can delete all event participants no matter they are pseudonymized or not. Cases and contacts work fine.

Were you logged in as admin? Admins are supposed to always be allowed to bulk delete/archive

sergiupacurariu: I was logged as admin indeed, sorry for that I saw that edit action is not possible to be selected. But there is another action that can be available if you are a surveillance supervisor and admin which is: "Create quarantine order documents". This action can be triggered as well on all event participants even the pseudonymized ones. Should this be possible? I think through this flow an admin could get the names of the pseudonymized persons because looking down the flow there is no pseudonymization check at all. Unfortunately, I could not check if you see the actual names in the generated documents. Please look over my assumptions if it is the case we might need to check with @MateStrysewske about this situation.

sergiupacurariu: I checked the generated documents and those generated from pseudonymized event participants do not contain person data. This situation needs to be further discussed outside of this ticket.

sergiupacurariu (approved):

Issue #5509, opened 2021-05-20
Add ContactId to the edit contact implementation


Mihai-Bota:
### Problem Description
For a better tracking in contact editing the ContactId is needed in the user Contact POJO

### Proposed Solution
add the missing ContactId to the POJO

PR #5520, opened 2021-05-24
#5509/ Added ContactId for Contacts

Mihai-Bota:
Added ContactId for Contacts pojo in order to provide history search for Contacts

image

### DIFF
--- a/sormas-e2e-tests/src/main/java/org/sormas/e2etests/pojo/Contact.java
+++ b/sormas-e2e-tests/src/main/java/org/sormas/e2etests/pojo/Contact.java
@@ -51,4 +51,5 @@ public class Contact {
   String contactCategory;
   String relationshipWithCase;
   String descriptionOfHowContactTookPlace;
+  String contactId;
 }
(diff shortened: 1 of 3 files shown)


sormas-e2e-tests/src/main/java/org/sormas/e2etests/pojo/Contact.java
```diff
@@ -51,4 +51,5 @@
   String contactCategory;
   String relationshipWithCase;
   String descriptionOfHowContactTookPlace;
+  String contactId;
```
BojteTamas, line 54: Please rename it to `uuid`, because you are already in `Contact` class

Mihai-Bota: will do. but it's not an UUID. i've tried to make it an UUID object but it's failing the conversion.

Mihai-Bota: @BojteTamas -> updated

BojteTamas: it's UUID, but it is created in other way than `UUID.random().toString()`

Issue #5530, opened 2021-05-25
Create the creation flow of person, contact, case, event, sample and task using the REST API


BojteTamas:
### Proposed Solution
Cover the basic creation of person, contact, case, event, sample and task using using the '/push' endpoint for each

PR #5532, opened 2021-05-25
#5530 create the creation flow of person contact case event sample and task using the rest api

BojteTamas:
### Done

- [X] Implemented the create of  person contact case event sample and task using the rest api with the basic payload
- [X] Update the Lombok version to 1.18.16
- [X] Split up the pojos into WEB and API
- [X] By using '@API' annotation the only api tests will not start anymore the Chrome browser


sormas-e2e-tests/src/main/java/org/sormas/e2etests/services/api/SampleApiService.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Date;
+import java.util.UUID;
+import org.sormas.e2etests.pojo.api.*;
+
+public class SampleApiService {
+
+  @Inject
+  public SampleApiService() {}
+
+  public Sample buildGeneratedEvent(Case caze) {
```
richardbartha, line 31: I think here you wanted to have buildGeneratedSample right ?

richardbartha (approved):

sormas-e2e-tests/src/main/java/org/sormas/e2etests/services/api/SampleApiService.java
BojteTamas: yes it is a typo, thank you

VGitVlad (approved):

PR #5533, opened 2021-05-25
5333 sharing external journal status

### LINKED ISSUE #5333: [SORMAS 2 SORMAS] Symptom journal registrations are shared
### Bug Description
When a case or contact is shared via Sormas 2 Sormas, symptom journal registrations are shared as well and the person seemingly remains registered in the other instance, disregarding which external journal is used. This should not happen because the follow-up is done locally and could cause potential complications when shared between instances.

### Steps to Reproduce

1. Open or create a case/contact
2. Register the person in the external symptom journal (either PIA or Climedo)
3. Share the entity with another Sormas instance (either without or with handing over ownership)

### Expected Behavior
The registration in an external symptom journal should be ignored when sharing data via Sormas 2 Sormas.

### Screenshots
diaryAndShare
diaryAndShare2

### System Details

* Device:
* SORMAS version: 1.59.1
* Server URL: test1402.sormas-oegd.de, release-de.sormas.netzlink.com

### Additional Information
@Candice-Louw @FredrikSchaeferVitagroup @markusmann-vg 
We should specify how external Journals should be handled with S2S. Things to consider:
- Should a case/contact be able to be shared with or without handing over the ownership while under external follow-up?
- If not, how do we prompt the user?
- Should external follow-up be somehow continued in the other system when handing over ownership? (likely very complicated)

Alexandru-Pirvu:
Fixes #5333


sormas-ui/src/main/java/de/symeda/sormas/ui/utils/ExternalJournalUtil.java
```diff
@@ -47,11 +48,27 @@
 	 * If the person is not registered in the external journal, a create account/register button is returned
 	 * If the person is registered, a button is returned which opens a popup with further options.
 	 *
+	 * The button is:
+	 * 	- enabled if the associated sormasToSormasEntity(case or contact) is not shared or it is shared and the current SORMAS instance has ownership
```
FredrikSchaeferVitagroup, line 52: What happens when there are multiple contacts/cases for a person, and one is shared with ownership handed over while the others are not? Is that currently possible via S2S?

Alexandru-Pirvu: Indeed, there are multiple scenarios in which there can be one case/contact without ownership and another with ownership for the same person. This indeed creates an inconsistent state for the person as both systems can alter the person data, including the external journal status.

I can think of two ways to reach that state:
* have two cases/contacts for the same person and share one with ownership and the other without
* share a case/contact without ownership and create another case/contact for the same person on the receiving instance

Those situations seem unlikely to happen though. Is this something we should treat? If so, maybe we should take care of the handling for the whole person data form.

FredrikSchaeferVitagroup: What about having two contacts for one person, and only sharing one?

Alexandru-Pirvu: Indeed, that is also possible and will lead to the same situation. However, in case the person is already registered in the external journal, sharing any associated contact will not be possible.

FredrikSchaeferVitagroup: Good point. So this problem is out of scope for this PR.

FredrikSchaeferVitagroup: Related ticket created: #5650

FredrikSchaeferVitagroup (approved):

Issue #5291, opened 2021-04-30
Don't ask for GPS access when no surveillance feature is active [1]


MartinWahnschaffe:
### Feature Description
Currently the Android app validates access to GPS when the login activity is opened.

### Problem Description
When the app is **not** used for surveillance purposes (e.g. for APMIS), it is not necessary to have access to the GPS module.
Making unnecessary use of GPS is not good from a privacy perspective. In some countries it is even problematic from a security perspective (unstable regions).

### Proposed Change
- [x] Postpone the check to after the login and initial sync is done
- [x] Only ask for GPS access when at least one surveillance related feature is enabled
- [ ] Check in the mobile app, where GPS Data is used and access is required

af-pwo:
Samsung Galaxy S9
Android 10
1.61.0 SNAPSHOT
test.sormas.netzlink.com

QA Done

App doesnt ask for GPS when no surveillance feature is activated

syntakker:
The SORMAS app still asks for GPS access with no surveillance feature enabled.

To reproduce:
- uncheck *all* features in the database's tabel `featureconfiguration` (this is overly cautious not to miss one of the features in question)
- start SORMAS server
- freshly install and start SORMAS app
- configure settings and login
- the screen to choose a PIN appears, where allowance for GPS access is asked for

![gps_allow](https://user-images.githubusercontent.com/5616564/121862347-048ae080-ccfb-11eb-96f1-3203de047328.png)

BarnaBartha:
Was rechecked together with @syntakker and is working as expected.

PR #5538, opened 2021-05-26
#5291 - don't ask for gps access when no surveillance feature is active

BarnaBartha:
Fixes #

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/util/LocationService.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/util/LocationService.java
@@ -15,12 +15,6 @@
 
 package de.symeda.sormas.app.util;
 
-import java.io.IOException;
-import java.util.List;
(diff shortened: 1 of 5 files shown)


sormas-app/app/src/main/java/de/symeda/sormas/app/util/LocationService.java
```diff
@@ -157,7 +158,7 @@ public boolean isRequestingActiveLocationUpdates() {
 
 	public void requestFreshLocation(Activity callingActivity) {
 
-		if (!validateGpsAccessAndEnabled(callingActivity)) {
+		if (ConfigProvider.getUser() == null || !validateGpsAccessAndEnabled(callingActivity)) {
```
alexcaruntu-vita, line 161: I would include the test for user inside the `validateGpsAccessAndEnabled` method snce they are always used together.

alexcaruntu-vita (approved):

PR #5588, opened 2021-05-31
Fix follow up calculation for contacts (#5548)

### LINKED ISSUE #5548: Contact follow-up calculation contradicts the expected until date
### Bug Description
When contacts are created and the follow-up duration calculation is less than 14 days (set in the database), the follow-up until date shows an error which contradicts the expected until date This results in contacts not being able to be saved properly.

cf. https://github.com/hzi-braunschweig/SORMAS-Project/issues/3243#issuecomment-843321677
### Steps to Reproduce

1. Set the follow-up duration to less than 14 days in the database
2. Create a contact
3. Check the follow-up until date

### Expected Behavior
Follow-up until date should not contradict the expected until date and prevent saving
### Screenshots
Screen Shot 2021-05-26 at 11 05 35 PM

### System Details

* Device:
* SORMAS version: 1.60

syntakker:
Fixes #5548


StefanKock (approved):
Fixed as discussed with @ChristopherRiedel

PR #5590, opened 2021-05-31
[#4626] S2S basic central keycloak deployment

### LINKED ISSUE #4626: Set up SORMAS X (eXchange) keycloak server [CONCEPT]
### Situation Description
When SORMAS instances are sending data to each other they need to be able to authenticate. Currently this is done by having a special user in each instance and sharing the login information for that user with all other instances, which is a lot of work to maintain.

### Feature Description

We need to replace this with a solution where the authentication is handled by a central keycloak. @vidi42 has created a draft of how this could work.

- [x] Set this up in a local or test environment
- [ ] Create setup guidelines
- [x] Define what may have to be changed in SORMAS or the per SORMAS keycloak instances

JonasCir:
Intends to fix #4626 

### Open points

- [x] Get feedback from @MartinWahnschaffe and @vidi42 
- [ ] Address test failures
- [ ] Write comments, docs, and tests
- [x] There is one open `todo` in the code: `validateExternalUrls()` does not work for the provides Urls
- ~[ ] Update bash scripts (I have working versions in the testbed repo)~ We will do this once when all parts are on one branch

### How to run this

Check out instructions given [here](https://git.vitasystems.dev/sormas/s2s-testbed). Once the deployment is running, login at `localhost:6080/sormas-ui/login`, create a case, and send it to the second instance.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/SormasToSormasConfig.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/SormasToSormasConfig.java
@@ -17,6 +17,10 @@ public class SormasToSormasConfig implements Serializable, Cloneable {
 	private String truststoreName;
 	private String truststorePass;
 	private boolean retainCaseExternalToken;
+	private String oidcServer;
+	private String oidcRealm;
+	private String oidcClientId;
+	private String oidcClientSecret;
 
 	public String getPath() {
 		return path;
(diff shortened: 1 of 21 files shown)


JonasCir:
Test failures result from the fact that `requestAccessTokenS2S` of course does not work. Can this be mocked somehow?

JonasCir:
See #5866 for a continuation of this PR

PR #5595, opened 2021-06-01
#5578 Wrong translation for contact confirmed (de)

### LINKED ISSUE #5578: Wrong translation for contact confirmed (de)
### Bug Description

A translation was accidently changed to a wrong value, whether a contact is confirmed.

Received via email: 
> there was a translation error I made in the 1.60 which might cause a lot of confusion. On the Contact Dashboard, the string for "Confirmed" was translated into "C. Klinisch-labordiagnostisch bestätigt" because I confused "Confirmed" with the case classification in crowdin. I think that the corresponding variable for the Contact dashboard is dashboardConfirmed. Instead of the current translation, this would simply have to be "Bestätigt" for de-DE. Do you think we could adjust this in the code for 1.60.1?

### Steps to Reproduce

1. Check contacts dashboard with german language. 

### Expected Behavior
The english translation value is "Confirmed". Replace it back to the former value "Bestätigt".

### Screenshots

grafik

### System Details

* SORMAS version: 1.60.0

af-fwa:
Fixes #5578


syntakker:
The naming convention for branches is [bugfix|feature|change]_<issue number>_<descripion>.

syntakker (approved):

PR #5624, opened 2021-06-02
4423 s2s accept reject

### LINKED ISSUE #4423: [SORMAS2SORMAS] As a recipient I want to be able to accept or reject a shared case from another SORMAS Instance [5]
### Situation Description
SORMAS 2 SORMAS interface pushes the information to the recipient's database without 
a.) letting know
b.) giving the opportunity to decline or accept the new case/contact

### Feature Description
There will be a limited set of information, that the recipient should be able see.
With this limited set of information the recipient hat to be able to either accept or reject the receiving of the case/contact
When accepting, the case/contact information will be available in the recipient's SORMAS instance
When rejecting, nothing of the information will be persisted in the recipient's instance.

### Additional Information
@HZI: The workflow has to be specified. 
1. How will the recipient know, that there is a new case/contact coming from another SORMAS instance? 
2. How will the sender know, when the attempt to send case/contact was a.) accepted or b.) rejected?
3. What information does the sender see so that he can decide wether to accept or reject?
4. Is there a time limit to reject / Accept?
5. Who is able to accept / reject?
6. Is withdrawing possible, this means: When a case was sent and not yet accepted or rejected, can the sender cancel the transmission request.
@Candice-Louw 
@kwa20 
@JaquM 
 
Brainstorming:
Add a new view in cases and contacts for "incoming/outgoing" with cases/contacts from other instances and  cases/contacts which were rejected or accepted

Answers and mockups will be provided until 24th of feb

leventegal-she:
Fixes #4423


sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/ShareRequestGridComponent.java
```diff
@@ -0,0 +1,159 @@
...
+
+		setSizeFull();
+		setMargin(false);
+
+		grid = new ShareRequestGrid(viewConfiguration.isInEagerMode(), criteria);
+		grid.setSizeFull();
+		grid.getDataProvider().addDataProviderListener(e -> updateStatusButtons());
+
+		VerticalLayout gridLayout = new VerticalLayout();
+		/* gridLayout.addComponent( */createFilterBar(filterChangeHandler, filterResetHandler)/* ) */;
```
vidi42, line 62: Maybe also add a comment here that the filter bar will only be added once more filters are introduced.

vidi42 (changes requested):

vidi42:
The ownership handover column is missing from the Share Requests grid. Is there a reason for leaving that out?

vidi42 (approved):

PR #5627, opened 2021-06-02
#5453 API and service for querying persons by external ids

### LINKED ISSUE #5453: New REST API endpoint to query person data by external ID [2]
### Situation Description
When data is transferred from SORMAS to SurvNet or other IfSG applications via the SurvNet Webservice only pseudonymized information is transferred. However, health departments need the personal information of the data to be available in the IfSG application.

To give the manufacturers of IfSG applications an option to retrieve regarding personal information from SORMAS based on a shared unique ID, the REST API should provide the data by the externalId which is the shared identifier in both systems.

### Feature Description
A new REST API endpoint for PERSONS should be provided that can be queried by the externalId of a person.

* Add a `getByExternalIds` endpoint to PersonResource that takes a list of Strings
* Add a `getByExternalIds` method to PersonFacade that works similar to the `getByUuids` method, but filters by external Ids instead

Alexandru-Pirvu:
Fixes #5453


sormas-api/src/main/java/de/symeda/sormas/api/person/PersonFacade.java
```diff
@@ -86,4 +86,6 @@
 	long setMissingGeoCoordinates(boolean overwriteExistingCoordinates);
 
 	boolean isSharedWithoutOwnership(String uuid);
+
+	List<PersonDto> getByExternalIds(List<String> uuids);
```
vidi42, line 90: You could rename the parameter to `externalIds` to match the method's name.

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
```diff
@@ -604,4 +604,17 @@ public boolean updateGeoLocation(Person person, boolean overwriteExistingCoordin
 		}
 		return geoLocationUpdated;
 	}
+
+	public List<Person> getByExternalIds(List<String> externalIds) {
```
vidi42, line 608: As `externalIds` can be of any size, a good approach is to use the `de.symeda.sormas.backend.util.IterableHelper` when using SQL `IN` statements.

See `de.symeda.sormas.backend.caze.CaseService#getByPersonUuids`

vidi42 (changes requested):

vidi42 (approved):

PR #5628, opened 2021-06-02
#4956 patch

### LINKED ISSUE #4956: [Angular][Backend] Case Save "Patching" method [8]
### Situation Description
Based on the discussion had on: https://github.com/hzi-braunschweig/SORMAS-Project/issues/4615 together with the team, it was concluded that we should implement a patching system for when saving a case (save only the updated items and not all of the case fields).

### Feature Description
- Enable "patching" for when saving a case.
- We should only save the updated fields
- Supply an end-point for the patching which should be available for the FE (angular).

### Possible Alternatives
- If there is no development time, or developer availability we can leave the "save" functionality as is.

### Additional Information
N/A

**ESTIMATION: 8 POINTS**

marius2301:
Fixes #4956


sormas-rest/src/main/java/de/symeda/sormas/rest/CaseResource.java
```diff
@@ -129,4 +130,10 @@ public CaseDataDto getByUuid(@PathParam("uuid") String uuid) {
 		return FacadeProvider.getCaseFacade().getByUuid(uuid);
 	}
 
+	@POST
+	@Path("/patchJson/{uuid}")
+	public CaseDataDto patchJson(@PathParam("uuid") String uuid, JSONObject jsonObject) throws Exception {
```
MartinWahnschaffe, line 135: Suggestion: postUpdate
Also add a comment that explains how/when to use this and what the JSON object should look like

sormas-backend/src/main/java/de/symeda/sormas/backend/common/PatchHelper.java
```diff
@@ -0,0 +1,54 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import net.minidev.json.JSONObject;
+
+public class PatchHelper {
+
+	public static <T> T patchJson(JSONObject jsonObject, Class<T> patchObjectClass, T existingObject) {
```
MartinWahnschaffe, line 16: Add an explanation

MartinWahnschaffe (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/common/PatchHelper.java
```diff
@@ -0,0 +1,67 @@
...
+			attributes.forEach(attribute -> {
+
+				Field existingField = null;
+				try {
+					existingField = existingObject.getClass().getDeclaredField(attribute);
+
+					Field updateField = patchObject.getClass().getDeclaredField(attribute);
+					existingField.setAccessible(true);
+					updateField.setAccessible(true);
+					existingField.set(existingObject, updateField.get(patchObject));
```
MartinWahnschaffe, line 46: What about entity classes? Wouldn't it be necessary to recursively call postUpdate for them? Otherwise you would e.g. override the symptom entity of a case and not keep it's uuids and existing values

sormas-api/pom.xml
```diff
@@ -88,7 +88,13 @@
 			<groupId>org.jsoup</groupId>
 			<artifactId>jsoup</artifactId>
 		</dependency>
+        <dependency>
```
MartinWahnschaffe, line 91: We are already using jackson, so I'd suggest to use that instead

sormas-backend/src/main/java/de/symeda/sormas/backend/common/PatchHelper.java
```diff
@@ -0,0 +1,67 @@
...
+	 *            - the jsonObject contains the values that are intended to be updated,
+	 *            representing a subset of the fields of the object that is intended to be updated
+	 * @param updatingObjectClass
+	 *            - the class of the object that is intended to be updated
+	 * @param existingObject
+	 *            - the existingObject that is intended to be updated
+	 * @param <T>
+	 * @return
+	 */
+	public static <T> T postUpdate(JSONObject jsonObject, Class<T> updatingObjectClass, T existingObject) {
```
MartinWahnschaffe, line 29: Please add a unit test for this. Especially one that passes in an hierarchical dto, like a case and one that covers a DTO with lists. We need to make sure that this does not throw away child objects

MartinWahnschaffe (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/common/PatchHelper.java
marius2301: I will add a unit test. I tested it and it does not throw away child objects.

sormas-rest/src/main/java/de/symeda/sormas/rest/CaseResource.java
```diff
@@ -143,4 +145,21 @@ public Response updateExternalData(List<ExternalDataDto> externalData) {
...
+	 *
+	 * @param uuid
+	 * @param caseDataDtoJson
+	 *            - a subset of caseDataDto fields, same structure as caseDataDto
+	 * @return - the updated caseDataDto
+	 * @throws Exception
+	 */
+	@POST
+	@Path("/postUpdate/{uuid}")
+	public CaseDataDto postUpdate(@PathParam("uuid") String uuid, JsonNode caseDataDtoJson) throws Exception {
```
MartinWahnschaffe, line 161: I'd like to add something like an "experimental" annotation to this, to make clear that this is for now only build for internal usage and still work in progress.
I searched for it, but haven't found something that would be a good fit. Maybe we create our own annotation?
@MateStrysewske Do you have an opinion on it?

sormas-rest/src/main/java/de/symeda/sormas/rest/CaseResource.java
```diff
@@ -143,4 +145,21 @@ public Response updateExternalData(List<ExternalDataDto> externalData) {
...
+		return FacadeProvider.getCaseFacade().postUpdate(uuid, caseDataDtoJson);
```
MartinWahnschaffe, line 162: @marius2301 We have to make sure that the provided json does not override the uuid of existing entities. Also not the uuid of existing child entities (e.g. symptoms of a case).

MartinWahnschaffe (changes requested):
Sorry, a few additional points came to my mind. Since this is laying the foundation for something new we have this. See my comments

marius2301:
1. The dependency regarding jsoup seems to be need it in the CampaignFormMetaFacadeEjb.  
2. I put a check to make sure the uuid in the path is the same as the one in the json. There is no need for internal objects

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,311 @@
...
+
+					} else if (jsonObjectFieldNode.isArray()) {
+						updateObjectList(existingObject, jsonObjectFieldNode, existingObjectField);
+
+					} else {
+						setFieldValue(existingObject, existingObjectField, jsonObjectFieldNode);
+					}
+
+				} catch (IllegalAccessException e) {
+					e.printStackTrace();
```
leventegal-she, line 85: not needed

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,311 @@
...
+					Logger logger = LoggerFactory.getLogger(existingObject.getClass());
+					logger.error(String.format("Failed to update object" + existingObject.getClass().getSimpleName() + ". Error: " + e.getMessage()));
+				}
+
+			}
+
+		} catch (Exception e) {
+			e.printStackTrace();
```
leventegal-she, line 93: not needed

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,311 @@
...
+	 * @throws JsonProcessingException
+	 * @throws IllegalAccessException
+	 */
+	private static <T extends HasUuid> void updateObjectList(T existingObject, JsonNode jsonObjectFieldNode, Field existingObjectField)
+		throws JsonProcessingException, IllegalAccessException {
+		Class listElementClass = getParameterizedType(existingObjectField);
+		ArrayList tempNewElementList = new ArrayList();
+
+		for (JsonNode listElement : jsonObjectFieldNode) {
+			addOrReplaceListElement(existingObject, existingObjectField, listElementClass, tempNewElementList, listElement);
```
leventegal-she, line 114: I don't really like the procedural approach where you start with en empty list and calling a method fills it, it's hard to find and debug where and how the items get added into the list.

Would be cleaner to have a "merge" method that using the existing list and the json list creates a new list that can be set as value on the dto

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,311 @@
...
+			if (existingObjectFieldInstance != null) {
+				existingElementList = (List) existingObjectFieldInstance;
+				existingElementList.clear();
+				existingElementList.addAll(tempNewElementList);
+			} else {
+				existingObjectField.set(existingObject, existingElementList);
+			}
+		} else if (existingObjectField.getType().isArray()) {
+			existingObjectField.set(existingObject, tempNewElementList.toArray());
+		}
```
leventegal-she, line 129: what happens if existingObjectField has another type?

It can happen that the guys using the api send wrong data, `postUpdateOrReplace` fails with exception and in this case it's just ignored

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,311 @@
...
+	 * @param arrayField
+	 *            - the arrayField
+	 * @param uuid
+	 *            - the given uuid
+	 * @return
+	 */
+	private static <T extends HasUuid> T getListElementByUuid(@NotNull Object obj, @NotNull Field arrayField, @NotNull String uuid) {
+		try {
+			return (T) ((List) arrayField.get(obj)).stream().filter(listElement -> {
+				Field uuidField = getFieldByName(EntityDto.UUID, listElement.getClass());
```
leventegal-she, line 292: reflection costs a lot. Looks like you could use isAssignableForm(HasUuid.class) and do a conversion to get the uuid

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,311 @@
...
+				uuidField.setAccessible(true);
+				try {
+					return uuid.equals(uuidField.get(listElement));
+				} catch (IllegalAccessException e) {
+					e.printStackTrace();
+					throw new RuntimeException(e);
+				}
+			}).findFirst().orElse(null);
+		} catch (IllegalAccessException e) {
+			e.printStackTrace();
```
leventegal-she, line 305: not needed

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+	 * @param existingObject
+	 *            - the existing object whose fields are to be updated
+	 * @param <T>
+	 *            - the type of the object
+	 */
+	@Experimental
+	public static <T extends HasUuid> void postUpdate(@NotNull JsonNode jsonObject, @NotNull T existingObject) {
+
+		if (!existingObject.getUuid().equals(jsonObject.get(EntityDto.UUID).textValue())) {
+			throw new RuntimeException("The updated object must have a Uuid");
```
leventegal-she, line 56: please use ValidationRuntimeException with a translated message instead of RuntimeException

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+		} else if (existingObjectField.getType().isArray()) {
+			try {
+				existingObjectField.set(existingObject, tempNewElementList.toArray());
+			} catch (IllegalAccessException e) {
+				e.printStackTrace();
+				throw new RuntimeException(
+					"Cannot access field " + existingObjectField.getName() + " on the object of type: " + existingObject.getClass().getName());
+			}
+		} else {
+			throw new RuntimeException(
```
leventegal-she, line 133: please use ValidationRuntimeException

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+				tempNewElementList.add(updatedObjectList);
+			}
+		}
+
+		if (existingObjectField.getType().isAssignableFrom(List.class)) {
+			Object existingObjectFieldInstance = null;
+			try {
+				existingObjectFieldInstance = existingObjectField.get(existingObject);
+			} catch (IllegalAccessException e) {
+				throw new RuntimeException(
```
leventegal-she, line 108: please use `RuntimeException(String message, Throwable cause)

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+			if (existingObjectFieldInstance != null) {
+				Collection existingElementList = (Collection) existingObjectFieldInstance;
+				existingElementList.clear();
+				existingElementList.addAll(tempNewElementList);
+			} else {
+				try {
+					existingObjectField.set(existingObject, tempNewElementList);
+				} catch (IllegalAccessException e) {
+					e.printStackTrace();
+					throw new RuntimeException(
```
leventegal-she, line 120: please use `RuntimeException(String message, Throwable cause)` instead of `e.printStackTrace(); throw new RuntimeException`

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+						"Cannot access field " + existingObjectField.getName() + " on the object of type: " + existingObject.getClass().getName());
+				}
+			}
+		} else if (existingObjectField.getType().isArray()) {
+			try {
+				existingObjectField.set(existingObject, tempNewElementList.toArray());
+			} catch (IllegalAccessException e) {
+				e.printStackTrace();
+				throw new RuntimeException(
```
leventegal-she, line 129: please use `RuntimeException(String message, Throwable cause)` instead of `e.printStackTrace(); throw new RuntimeException`

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+	 * @throws IllegalAccessException
+	 * @throws JsonProcessingException
+	 */
+	private static <T extends HasUuid> void updateExistingObject(T existingObject, JsonNode jsonObjectFieldNode, Field existingObjectField) {
+		T existingFieldValue = null;
+		try {
+			existingFieldValue = (T) existingObjectField.get(existingObject);
+		} catch (IllegalAccessException e) {
+			e.printStackTrace();
+			throw new RuntimeException(
```
leventegal-she, line 156: please use `RuntimeException(String message, Throwable cause)` instead of `e.printStackTrace(); throw new RuntimeException`

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+	private static <T extends HasUuid> T addOrReplaceListElement(
+		T existingObject,
+		Field existingObjectField,
+		Class listElementClass,
+		JsonNode listElement) {
+		if (HasUuid.class.isAssignableFrom(listElementClass) && listElement.hasNonNull(EntityDto.UUID)) {
+			T existingListElement = (T) getListElementByUuid(existingObject, existingObjectField, listElement.get(EntityDto.UUID).textValue());
+			if (existingListElement != null) {
+				postUpdate(listElement, existingListElement);
+				//tempNewElementList.add(existingListElement);
```
leventegal-she, line 189: commented code should be deleted

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+	 * all the elements are replaced with the values form json
+	 * 
+	 * @param existingObject
+	 * @param existingObjectField
+	 * @param listElementClass
+	 * @param listElement
+	 * @param <T>
+	 * @throws JsonProcessingException
+	 */
+	private static <T extends HasUuid> T addOrReplaceListElement(
```
leventegal-she, line 180: this method should be called `createOrUpdateElement`

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+	 */
+	private static <T> void setFieldValue(@NotNull T existingObject, @NotNull Field existingField, JsonNode fieldJson) {
+		Object fieldValue = null;
+
+		fieldValue = getObjectFromJsonNode(fieldJson, existingField.getType());
+		try {
+			existingField.set(existingObject, fieldValue);
+		} catch (IllegalAccessException e) {
+			e.printStackTrace();
+			throw new RuntimeException(
```
leventegal-she, line 220: please use RuntimeException(String message, Throwable cause) instead of e.printStackTrace(); throw new RuntimeException

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+		if (!fieldJson.isNull()) {
+
+			if (objectClass.isEnum()) {
+				object = Enum.valueOf((Class<Enum>) objectClass, fieldJson.textValue());
+			} else {
+				try {
+					object = objectMapper.treeToValue(fieldJson, objectClass);
+				} catch (JsonProcessingException e) {
+					e.printStackTrace();
+					throw new RuntimeException("Cannot covert JSon" + fieldJson.asText() + " to object of class: " + objectClass.getName());
```
leventegal-she, line 245: please use RuntimeException(String message, Throwable cause) instead of e.printStackTrace(); throw new RuntimeException

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+		}
+	}
+
+	private static @NotNull Field getClassFieldByName(@NotNull String fieldName, @NotNull Class objectClass) {
+		if (classFieldsMap.get(objectClass) == null) {
+			getAllClassFields(objectClass);
+		}
+		Field requiredField = classFieldsMap.get(objectClass).stream().filter(field -> field.getName().equals(fieldName)).findAny().orElse(null);
+		if (requiredField == null) {
+			throw new RuntimeException("No such field exception! " + fieldName + " in the class:" + objectClass.getName());
```
leventegal-she, line 272: should throw a ValidationRunTImeException and use objectClass.getSimpleName

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,330 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import de.symeda.sormas.api.EntityDto;
+import de.symeda.sormas.api.HasUuid;
+import de.symeda.sormas.api.utils.Experimental;
+
+public class PatchHelper {
+
+	private static ObjectMapper objectMapper = new ObjectMapper();
+	private static Map<Class, List<Field>> classFieldsMap = new HashMap<Class, List<Field>>();
```
leventegal-she, line 27: please use `ConcurrentHashMap` to be thread safe

leventegal-she (changes requested):

leventegal-she (approved):

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,335 @@
...
+	 * @param <T>
+	 * @throws JsonProcessingException
+	 * @throws IllegalAccessException
+	 */
+	private static <T extends HasUuid> void updateObjectList(T existingObject, JsonNode jsonObjectFieldNode, Field existingObjectField) {
+		Class listElementClass = getParameterizedType(existingObjectField);
+		List tempNewElementList = new ArrayList();
+
+		for (JsonNode listElement : jsonObjectFieldNode) {
+			T updatedObjectList = createOrUpdateElement(existingObject, existingObjectField, listElementClass, listElement);
```
MateStrysewske, line 101: Just for my understanding, the element returned by createOrUpdateElement here is not a list, but an object that is part of the list, right? In that case, the variable should rather be called updatedListObject for clarification; if I'm mistaken, ignore this suggestion

sormas-backend/src/main/java/de/symeda/sormas/backend/util/PatchHelper.java
```diff
@@ -0,0 +1,335 @@
...
+	 * all the elements are replaced with the values form json
+	 * 
+	 * @param existingObject
+	 * @param existingObjectField
+	 * @param listElementClass
+	 * @param listElement
+	 * @param <T>
+	 * @throws JsonProcessingException
+	 */
+	private static <T extends HasUuid> T createOrUpdateElement(
```
MateStrysewske, line 188: Rename to `createOrUpdateListElement` to make the intended usage more clear

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #5631, opened 2021-06-02
4607 translated data dictionary

### LINKED ISSUE #4607: Translate captions in data dictionary depending on the user language setting [2]
### Situation Description
Thus far, Sormas only provides the data dictionary in the english language. However, users might need the data dictionary captions in another languages to maximize the usability.
### Feature Description
Translate the captions in the data dictionary depending on the user language setting. 

### Additional Information
Examples in data dictionary:
Screen Shot 2021-03-03 at 1 29 33 PM
Screen Shot 2021-03-03 at 1 30 05 PM

* Create new `InfoFacade` **(name should be discussed)** and `InfoFacadeEjb` classes
* Add a method called `generateDataDictionary` to the new facade that returns a File, can take an optional file path and move the generation code from the `DataDictionaryGenerator` class in the test package; the test class should call the facade method with the file path that is currently used there instead; calling this method with an empty file path should create a file in the temp directory with the name "data_dictionary_{random_long}"
* The new method should generate the data dictionary based on the user language (or in default English if it's called from the test class)
* In the AboutView in the UI, don't retrieve the default data dictionary, but call the `generateDataDictionary` method with an empty file path and let the user download/open the returned file instead

leventegal-she:
Fixes #4607


sormas-backend/src/test/java/de/symeda/sormas/backend/share/ExternalShareInfoFacadeEjbEjbTest.java
```diff
@@ -36,7 +36,7 @@
 import de.symeda.sormas.backend.AbstractBeanTest;
 import de.symeda.sormas.backend.TestDataCreator;
 
-public class ExternalShareInfoFacadeEjbTest extends AbstractBeanTest {
+public class ExternalShareInfoFacadeEjbEjbTest extends AbstractBeanTest {
```
Alexandru-Pirvu, line 39: Double Ejb

Alexandru-Pirvu (changes requested):

Alexandru-Pirvu (approved):

PR #5632, opened 2021-06-02
#3621 Display username in visit origin

### LINKED ISSUE #3621: Display username in visit origin [0.5]
### Feature Description
Visit origin column in the case and contact follow-up tabs.
### Problem Description
So far, the visit origin column only differentiates between 'created by user' and 'external journal'. 
### Proposed Change
Instead of just displaying created by user, the username of the user who created the new visit should be displayed. This will help in internal communication between users, especially when there is a lot of staff working on contact/case tracing. Use the `getShortCaption` method of UserReferenceDto to make sure that the role is not included.

### Additional Information
visitOriginUserName

leventegal-she:
Fixes #3621

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/visit/VisitFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/visit/VisitFacadeEjb.java
@@ -308,7 +308,7 @@ public class VisitFacadeEjb implements VisitFacade {
 		CriteriaQuery<VisitIndexDto> cq = cb.createQuery(VisitIndexDto.class);
 		Root<Visit> visit = cq.from(Visit.class);
 		Join<Visit, Symptoms> symptoms = visit.join(Visit.SYMPTOMS, JoinType.LEFT);
-		Join<Object, Object> visitUser = visit.join(Visit.VISIT_USER, JoinType.LEFT);
+		Join<Visit, User> visitUser = visit.join(Visit.VISIT_USER, JoinType.LEFT);
 
 		cq.multiselect(
 			visit.get(Visit.ID),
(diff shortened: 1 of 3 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/visit/VisitFacadeEjb.java
```diff
@@ -308,7 +308,7 @@ public void deleteVisit(String visitUuid) {
 		CriteriaQuery<VisitIndexDto> cq = cb.createQuery(VisitIndexDto.class);
 		Root<Visit> visit = cq.from(Visit.class);
 		Join<Visit, Symptoms> symptoms = visit.join(Visit.SYMPTOMS, JoinType.LEFT);
-		Join<Object, Object> visitUser = visit.join(Visit.VISIT_USER);
+		Join<Object, Object> visitUser = visit.join(Visit.VISIT_USER, JoinType.LEFT);
```
cazacmarin, line 311: Please use appropriate Objects for this joining. 
(IE: Join<Visit, User> instead of Join<Object, Object>)

sormas-ui/src/main/java/de/symeda/sormas/ui/visit/VisitGrid.java
```diff
@@ -59,6 +61,19 @@ public VisitGrid(VisitCriteria criteria) {
...
 			e -> ControllerProvider.getVisitController().editVisit(e.getUuid(), getCriteria().getContact(), getCriteria().getCaze(), r -> reload()));
 
+		removeColumn(VisitIndexDto.ORIGIN);
+		addComponentColumn(visitIndexDto -> {
+			VisitOrigin origin = visitIndexDto.getOrigin();
+
+			String displayText = origin.toString();
+
+			if (origin == VisitOrigin.USER) {
+				displayText += " " + visitIndexDto.getVisitUser().getShortCaption();
```
cazacmarin, line 71: I'm not pretty sure if visitIndexDto.getVisitUser() can be null, but if it can, than NullPointerException will be raised.
In my opinion is better to add as well a check for nonNull for visitIndexDto.getVisitUser().

cazacmarin (changes requested):

cazacmarin (approved):

PR #5639, opened 2021-06-02
Fix of the responsible user not being notified when a participant is added to an event #4817

### LINKED ISSUE #4817: Notification to the responsible user when a participating person is part of other events

aponcon-atolcd:
Fix of the issue encountered during feature validation.

Fixes #4817


BarnaBartha (approved):
sorry for the late reply
looks good, please update it again from dev

PR #5643, opened 2021-06-02
add FeatureConfiguration for event hierarchies (#4527)

### LINKED ISSUE #4527: Add a feature configuration for superordinate and subordinate grouping of events [1]
### Feature Description
- There is a need to group events  in hierarchy (#4239) or in non-hierarchy (#4256)
- It is unlikely that users of an instance may need to have both options activated. Some users do not also need any of them.
### Proposed Change
-  Add a feature configuration for superordinate and subordinate grouping of events
-  The feature configuration fro non-hierarchical grouping would be handled in #4256 
* Add a new feature type `EVENT_HIERARCHIES`; server feature, enabled by default, depending on the EVENT_SURVEILLANCE feature type
* If disabled, don't show the "superordinate" and "subordinate" components in the event view, and also remove the superordinateEvent_id column from the import template (ImportFacadeEjb.generateEventImportTemplateFile)

syntakker:
Fixes #4527

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/importexport/ImportFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/importexport/ImportFacadeEjb.java
@@ -233,7 +233,10 @@ public class ImportFacadeEjb implements ImportFacade {
 
 		char separator = configFacade.getCsvSeparator();
 
-		List<String> columnsToRemove = Arrays.asList(EventDto.SORMAS_TO_SORMAS_ORIGIN_INFO, EventDto.OWNERSHIP_HANDED_OVER);
+		ArrayList<String> columnsToRemove = new ArrayList<>(Arrays.asList(EventDto.SORMAS_TO_SORMAS_ORIGIN_INFO, EventDto.OWNERSHIP_HANDED_OVER));
+		if (featureConfigurationFacade.isFeatureDisabled(FeatureType.EVENT_HIERARCHIES)) {
+			columnsToRemove.add(EventDto.SUPERORDINATE_EVENT);
+		}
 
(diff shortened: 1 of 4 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/importexport/ImportFacadeEjb.java
```diff
@@ -233,7 +233,10 @@ public void generateEventImportTemplateFile() throws IOException {
 
 		char separator = configFacade.getCsvSeparator();
 
-		List<String> columnsToRemove = Arrays.asList(EventDto.SORMAS_TO_SORMAS_ORIGIN_INFO, EventDto.OWNERSHIP_HANDED_OVER);
+		ArrayList<String> columnsToRemove = new ArrayList<>(Arrays.asList(EventDto.SORMAS_TO_SORMAS_ORIGIN_INFO, EventDto.OWNERSHIP_HANDED_OVER));
```
sergiupacurariu, line 236: if no particular reason for doing this, please remove new ArrayList<> and add List to columnsToRemove field declaration

sergiupacurariu (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/importexport/ImportFacadeEjb.java
syntakker: `Arrays.asList` returns a fixed size list, so I cannot add another column after list creation. On the other hand side, I wanted to keep the handy syntax of `asList`, that is why I use the result to create an `ArrayList`.
Any better idea?

sergiupacurariu (approved):

PR #5670, opened 2021-06-04
#5324 Add a checkbox to avoid sending this case to SurvNet

### LINKED ISSUE #5324: Add a checkbox to avoid sending this case to SurvNet [2]
- Add a new boolean field called `dontShareWithReportingTool` to cases
- Extend the case creation and edit forms in the web app with this new field (checkbox); only visible if the SurvNet interface is configured; position it below the external ID; caption: "Don't share this case with the external reporting tool"
- When the checkbox is checked, display a warning label (exclamation icon + text) below it: "This case is actively prevented from being sent to the external reporting tool"
- ~~Add a checkbox filter to the case directory: "Only show cases that are not shared with the external reporting tool"~~ 
- Add a checkbox filter to the case directory: "Only cases marked as don't share with reporting tool"
- Extend the case bulk edit form to include the new field
- Hide the "Send to reporting tool" button in the case view if the checkbox is checked
- Prevent cases from being sent to SurvNet if the checkbox is checked; this includes cases that may be sent as part of event clusters

*** Additional Info
- in GERMAN call this box please "Senden an Meldesoftware erfolgt in anderem Landkreis".

_Originally posted by @maxiheyner in https://github.com/hzi-braunschweig/SORMAS-Project/issues/4770#issuecomment-832503288_

leventegal-she:
Fixes #5324


sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseBulkEditData.java
```diff
@@ -70,6 +71,7 @@
 	private FacilityType facilityType;
 	private FacilityReferenceDto healthFacility;
 	private String healthFacilityDetails;
+	private Boolean dontShareWithReportingTool;
```
FredrikSchaeferVitagroup, line 74: Why not make it primitive? This would be cheaper.

sormas-api/src/main/resources/captions.properties
```diff
@@ -311,6 +312,7 @@ caseFilterCasesWithReinfection=Only cases with reinfection
 caseFilterOnlyCasesNotSharedWithExternalSurvTool=Only cases not yet shared with reporting tool
 caseFilterOnlyCasesSharedWithExternalSurvToo=Only cases already shared with reporting tool
 caseFilterOnlyCasesChangedSinceLastSharedWithExternalSurvTool=Only cases changed since last shared with reporting tool
+caseFilterOnlyCasesWithDontShareWithExternalSurvTool=Only cases marked as don't share with reporting tool
```
FredrikSchaeferVitagroup, line 315:
```suggestion
caseFilterOnlyCasesWithDontShareWithExternalSurvTool=Only cases marked with 'Don't share with reporting tool'
```

sormas-api/src/main/resources/strings.properties
```diff
@@ -313,7 +313,7 @@ errorNoPopulationDataLocations=No population data was found for the following lo
 errorNoPopulationDataFound=There is no population data available. Switching to absolute data view
 errorFormIdPopulationAgeGroup=Both "Form Id" and "Population Age Group" options are set 
 errorExternalSurveillanceToolNonCoronavirusCase=Could not send the selected cases to the reporting tool because the case %s is not a %s case.
-errorExternalSurveillanceToolCaseNotOwned=Could not send the selected cases to the reporting tool because your SORMAS instance is not the owner of the case %s.
+errorExternalSurveillanceToolCaseNotSharable=Could not send the selected cases to the reporting tool because. Please check if the case %s is editable and it is allowed to be be shared with external reporting tool.
```
FredrikSchaeferVitagroup, line 316: This error message is gramatically not correct and looks like it could be simplified.

FredrikSchaeferVitagroup (changes requested):

FredrikSchaeferVitagroup:
I am also wondering wheter it is a good idea to share the Case.dontShareWithReportingTool via S2S?
What about including it into the case export?

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseBulkEditData.java
leventegal-she: Because it should be null when no change wanted for this field

leventegal-she:
> I am also wondering wheter it is a good idea to share the Case.dontShareWithReportingTool via S2S?
> What about including it into the case export?

S2S solved
Export was not requested in the ticket

FredrikSchaeferVitagroup (approved):

FredrikSchaeferVitagroup:
Please resolve merge conflicts before merge.

PR #5688, opened 2021-06-07
Concept/5597 cert sharing s2s

### LINKED ISSUE #5597: Set up SORMAS X (eXchange) cert sharing [CONCEPT]
### Situation Description
In S2S, instances must be able to verify the signed data they receive from other instances. Currently this requires each key to be shared with each participating party which is not maintainable.

### Feature Description
Introduce a PKI infrastructure. Each instance provides an endpoint under `sormasTosormas/cert` which returns the instances certificate which is signed by a CA. The instance receiving request the certificate, checks if it was correctly singed by the CA, and then if the data was signed correctly by the sending instance. With this scheme, each instance only need to have the CA's public key in the keystore.

JonasCir:
Fixes #5597

### DIFF
--- a/sormas-api/pom.xml
+++ b/sormas-api/pom.xml
@@ -11,6 +11,16 @@
 
 	<build>
 		<finalName>${project.artifactId}</finalName>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
+                </configuration>
+            </plugin>
+        </plugins>
         <resources>
 			<resource>
 				<directory>src/main/resources</directory>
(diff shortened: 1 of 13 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/SormasToSormasEncryptionEjb.java
```diff
@@ -44,25 +51,30 @@
...
+import de.symeda.sormas.api.sormastosormas.SormasToSormasEncryptionFacade;
 import de.symeda.sormas.api.sormastosormas.SormasToSormasException;
 import de.symeda.sormas.backend.crypt.CmsCreator;
 import de.symeda.sormas.backend.crypt.CmsReader;
 
-@Stateless
-@LocalBean
-public class SormasToSormasEncryptionService {
+@Stateless(name = "SormasToSormasEncryptionFacade")
+public class SormasToSormasEncryptionEjb implements SormasToSormasEncryptionFacade {
```
MartinWahnschaffe, line 62: To follow the naming of other facades it should be SormasToSormasEncryptionFacadeEjb

JonasCir:
Is now a part of #5866 Closing

Issue #5599, opened 2021-06-01
[S2S] Refactor and deduplicate S2S code


JonasCir:
### Problem Description
While working on #5597, I noticed a lot of potential for code cleanup and deduplication which once resolved  would make my life easier. Also, I want to avoid merge hell along the way :imp: 

### Proposed Change
Not a complete list but:

- [x] `SormasToSormasEncryptionService` has code duplication and can be simplified for my work on #5597 (PR #5626  )
- [x] `SormasToSormasFacadeHelper` only encapsulates the sharing of data but not the receiving which causes a lot of code duplication in `AbstractSormasToSormasInterface` (PR #5672)
- [x] Reformat `MultiAuthenticationMechanism.java ` for smaller diff in concept PR
- [x] ~Private key does not have a dedicated password, but reuses password from keystore which is bad style~
- [x] Classes could be better organized PR up #5804
- [x] more to come...

JonasCir:
All done

JonasCir:
Nothing actionable for QA here, this was just a tracking issue.

PR #5693, opened 2021-06-07
[#5679] move athentication flow to REST client

### LINKED ISSUE #5679: SormasToSormasRestClient should handle authentication automatically
### Problem Description

This is a spin off from #5599, I want to keep the PRs small and easy to review.

While working on #5597, I noticed a lot of potential for code cleanup and deduplication which once resolved would make my life easier.

### Proposed Change

`SormasToSormasRestClient` puts the burden of providing the correct authorization header for the S2S request on the developer at the call site. With the token based approach around the corner, we just want to call the REST API without worrying about the underlying details. Further, the certificate based S2S approach will also talk to an endpoint of S2S, which is yet another call side that has to take care of running authentication correctly. 

Clearly, this introduces code duplication around a security critical feature which we should avoid.

---

First try: #5693 (closed)
First part: https://github.com/hzi-braunschweig/SORMAS-Project/pull/5697
Second part: #5724

JonasCir:
Fixes #5679 

### Info

This is a spin off from #5599, I want to keep the PRs small and easy to review.

While working on #5597, I noticed a lot of potential for code cleanup and deduplication which once resolved would make my life easier.

### Problem Description

`SormasToSormasRestClient` puts the burden of providing the correct authorization header for the S2S request on the developer at the call site. With the token based approach around the corner, we just want to call the REST API without worrying about the underlying details. Further, the certificate based S2S approach will also talk to an endpoint of S2S, which is yet another call side that has to take care of running authentication correctly.  This introduces code duplication around a security critical feature which we should avoid.

### Proposed change

Remove the auth token generation from the call side and move it into the `SormasToSormasRestClient` class. Further, the target of a S2S REST request is only determined by the receiving server's ID. All necessary information (e.g., host name) is derived from the ID.

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/AbstractSormasToSormasInterface.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/AbstractSormasToSormasInterface.java
@@ -96,10 +96,8 @@ public abstract class AbstractSormasToSormasInterface<T extends AbstractDomainOb
 			associatedEntities.addAll(shareData.getAssociatedEntities());
 		}
 
-		sormasToSormasFacadeHelper.sendEntitiesToSormas(
-			entitiesToSend,
-			options,
-			(host, authToken, encryptedData) -> sormasToSormasRestClient.post(host, saveEndpoint, authToken, encryptedData));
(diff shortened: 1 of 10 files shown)


JonasCir:
With Levis changes we need to go a different approach

PR #5708, opened 2021-06-08
#5685 Update dashboard "confirmed" caption

### LINKED ISSUE #5685: Differenciate strings for confirmed cases and confirmed contacts on dashboard
### Feature Description

Dashboard captions for confirmed cases and confirmed contacts.
### Problem Description

The dashboard uses the same string to display confirmed cases and contacts. This is a problem because in some languages, case classification 'confirmed' and confirmed contact do not have the same translation.
### Proposed Change

Split the strings for confirmed cases and confirmed contacts on the dashboard in order to allow differing translations.
### Additional Information

confirmedDashboard
confirmedDashboard2

DavidBaldsiefen:
Closes #5685


sormas-api/src/main/java/de/symeda/sormas/api/i18n/Captions.java
```diff
@@ -718,7 +718,7 @@
 	String dashboardComparedToPreviousPeriod = "dashboardComparedToPreviousPeriod";
 	String dashboardCompletedFollowUp = "dashboardCompletedFollowUp";
 	String dashboardCompletedFollowUpShort = "dashboardCompletedFollowUpShort";
-	String dashboardConfirmed = "dashboardConfirmed";
```
StefanKock, line 721: Removing `dashboardConfirmed` breaks every translation. Just keeping it for cases would only fix the problem, without any need to change anything to translation files.

sormas-api/src/main/resources/captions.properties
```diff
@@ -787,7 +787,8 @@ dashboardComparedTo=compared to
 dashboardComparedToPreviousPeriod=%s compared to %s
 dashboardCompletedFollowUp=Completed follow-up
 dashboardCompletedFollowUpShort=Completed F/U
-dashboardConfirmed=Confirmed
+dashboardConfirmedCase=Confirmed case
```
StefanKock, line 790: The default value should still be "Confirmed" I suppose (that's what the user reads).

StefanKock (changes requested):
General remarks: 

- Please shorten the commit message to fit into 50 chars with readable ticket id. More info can be given in line 3 and following 
- Did you check the android-app? Maybe there is the same problem with mixed keys. And your removal of a variable will for sure break it.

sormas-api/src/main/java/de/symeda/sormas/api/i18n/Captions.java
DavidBaldsiefen: While I agree, I am not sure if it will then be clear enough during translation. After all, just from seeing the "dashboardConfirmed" string the translator would not know if it's for cases or contacts, right?

StefanKock: I understand your point, but breaking translations like this should not happen when fixing a defect like this. I think if questions arise what which captions are for, we need do clarify it somehow different. And think about how to make refactoring possible here at some point in the future.

StefanKock (approved):

PR #5785, opened 2021-06-14
4607_invalid_table_name_excel

### LINKED ISSUE #4607: Translate captions in data dictionary depending on the user language setting [2]

vidi42:
Fixes #4607

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/info/InfoFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/info/InfoFacadeEjb.java
@@ -186,7 +186,7 @@ public class InfoFacadeEjb implements InfoFacade {
 
 		// Create
 		XSSFTable table = sheet.createTable();
-		String safeTableName = safeName.replaceAll("\\s|\\p{Punct}", "_");
+		String safeTableName = getSafeTableName(safeName);
 		table.setName(safeTableName);
 		table.setDisplayName(safeTableName);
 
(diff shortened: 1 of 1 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/info/InfoFacadeEjb.java
```diff
@@ -186,7 +186,7 @@ private void createEntitySheet(XSSFWorkbook workbook, Class<? extends EntityDto>
 
 		// Create
 		XSSFTable table = sheet.createTable();
-		String safeTableName = safeName.replaceAll("\\s", "_");
+		String safeTableName = safeName.replaceAll("\\s|\\p{Punct}", "_");
```
leventegal-she, line 189: Would be better to create a constant for this regex to avoid duplication

leventegal-she (changes requested):

leventegal-she (approved):

PR #5799, opened 2021-06-15
Add disease variant to events #5525

### LINKED ISSUE #5525: Addition of a Variant field in the "EVENT" part
**_Warning : Ticket update_**

### Situation Description
To date, the field for entering the disease variant only exists for cases. Users regret not categorizing events according to the variant strain of the disease.

### Feature Description
Develop a drop-down list allowing the user to have information relating to the major variant of the event.

This variable will be positioned in the event file, to the right of the disease field.
image

This variable can be completed in two ways:
1) Directly and manually by the user who will select the major variant of the event from the drop-down list.

2) Automatically by SORMAS: a "magical wand" button located to the right of the field is used to launch the base calculations to identify the majority variant among the event participant (ep).

image
 
The management rules for this automatic calculation are as follows:
•	if among the ep, no one is the case, the field remains empty and the user can himself select a variant strain from the list configured on the server
•	if among the ep, one or more people are cases and have the same and unique variant strain, the system would switch the drop down list to this variant (the user could, however, modify it if he wanted)
•	if among the ep, several different variant strains are identified, the system proposes the majority variant strain by number or, in the absence of a majority (ie presence of tie variants), use the most recently (based on report date of the associated cases) identified variant. The user will always have the possibility to modify the strain manually by scrolling the drop down list.
•	If among the ep, the majority variant strain is `unknown` , the "variant" field will indicate the "unknown"  if only unknown and NA exist, else use the second majority category over Unknown.
•	If among the ep, the majority variant strain is `NULL` , the "variant" field will indicate the "NULL" 

The result returned by the query will overwrite the information previously entered in the field.

### Additional Information

tbroyer:
Fixes #5525

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/event/EventCriteria.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/event/EventCriteria.java
@@ -1,6 +1,6 @@
 /*******************************************************************************
  * SORMAS® - Surveillance Outbreak Response Management & Analysis System
- * Copyright © 2016-2018 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ * Copyright © 2016-2021 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -25,6 +25,7 @@ import de.symeda.sormas.api.Disease;
 import de.symeda.sormas.api.EntityRelevanceStatus;
 import de.symeda.sormas.api.action.ActionStatus;
 import de.symeda.sormas.api.caze.CaseReferenceDto;
+import de.symeda.sormas.api.disease.DiseaseVariant;
 import de.symeda.sormas.api.facility.FacilityReferenceDto;
 import de.symeda.sormas.api.facility.FacilityType;
 import de.symeda.sormas.api.person.PersonReferenceDto;
@@ -63,6 +64,7 @@ public class EventCriteria extends CriteriaWithDateType implements ExternalShare
 	private RiskLevel riskLevel;
 	private EventInvestigationStatus eventInvestigationStatus;
 	private Disease disease;
+	private DiseaseVariant diseaseVariant;
 	private UserRole reportingUserRole;
 	private Boolean deleted = Boolean.FALSE;
 	private RegionReferenceDto region;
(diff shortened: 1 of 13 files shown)


tbroyer:
Build fail is a false negative (network error), it needs to be re-triggered.

MateStrysewske (changes requested):
Please add the new field to the mobile app as well

MateStrysewske (changes requested):
Can you resolve the merge conflicts, please? :)

MateStrysewske (changes requested):
* Web app: Don't show the disease variant field if no disease has been selected (works after changing selections, but when opening the event creation form or the event edit form for an event without a disease, the field is shown)
* Mobile app: Don't show the disease variant field if the selected disease does not have a disease variant (new event fragment)
* Mobile app: Disease variant field does not become visible when switching around diseases - if the initial event disease has disease variants, the field is shown, but after switching to another disease and then back to a disease that has disease variants, the field stays hidden

tbroyer:
> * Web app: Don't show the disease variant field if no disease has been selected (works after changing selections, but when opening the event creation form or the event edit form for an event without a disease, the field is shown)
> * Mobile app: Don't show the disease variant field if the selected disease does not have a disease variant (new event fragment)
> * Mobile app: Disease variant field does not become visible when switching around diseases - if the initial event disease has disease variants, the field is shown, but after switching to another disease and then back to a disease that has disease variants, the field stays hidden

Most of this is code that was quite literally copy/pasted from the Case screens, so if it behaves differently then there must be some "hidden" thing that triggers the differing behavior, and I'll need some more specific guidance on why that is and how I'm supposed to fix it "properly".

For instance, the only thing that rings a bell for me in `CaseDataForm` vs `EventDataForm` is https://github.com/hzi-braunschweig/SORMAS-Project/blob/6da08cb8a6322e6c33f520bfafce640e53543bdb/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java#L1631-L1633
Would it be acceptable to do the same in `EventDataForm`? (after some testing to make sure this is actually what causes the differing behavior) or should I rather make some code specific to the disease variant? (if so, where/how?)

Similarly, in the mobile app, the behavior doesn't surprise me much given the code (I don't remember seeing it when testing, I possibly didn't test those scenarios), but then if it doesn't happen in the Case screen, I have zero idea why that is, and what I should do to fix it (I'll probably find it myself by poking at the code and testing, but if you have some ideas, they'll be very welcome)

Thanks in advance

MateStrysewske (approved):

PR #5804, opened 2021-06-15
[#5599] refactor S2S directory layout

JonasCir:
Ticks on box in #5599

No functionality was changed, purely cosmetic. Review best by checking out locally and navigating to `backend.sormas2sormas`. This will also nicely play with the upcoming changes for S2S gen. 2.


JonasCir:
Will open a new one

Issue #5705, opened 2021-06-08
Duplicate person detection creates persons but not cases


bernardsilenou:
### Bug Description

- The logic in duplicate person detection when importing cases seam not to work well. During import, when user decide to create a new person, system create a new person and set is as pseudonymize to all users + does not create a case but import message shows that a record was imported
### Steps to Reproduce

1. create 2 cases, one in-country and another POE
2. export these cases and re-import them, do this twice
3. when duplicate person is suggested by system, chose the same person and in the second round, choose create a new person
4. have a look at import message and compare with what you find in case and person directories
5. POE cases returned error message: `Invalid value XXX in column pointOfEntry; Entry does not exist in the database or in the specified district or is archived (Or maybe a wrong facility type was specified)`

### Expected Behavior

- export and re-import of the same case from should work with POE cases
- When user chooses "create new person", then create a new person + new case  + report that one case was inported
- If user choose an existing person, then do not create a new person + call duplicate case detection method
### System Details

* Device:
* SORMAS version: 1.58.3
* Android version/Browser:
* Server URL: https://test-inhp-ci.sormas.netzlink.com/sormas-ui/#!about
* User Role: admin + national user

bernardsilenou:
@kwa20 Can you also have a look at the process flow?
mainly whey persons are created but not entities?

vidi42:
@MateStrysewske @bernardsilenou

✔️ The issue with exporting / importing case with POE is fixed now.  

⚠However for the other issues described here related to duplication checks here are my findings:


When exporting and importing the same case, there are two main possible flows:

#### 1. The user is presented with the duplicate person popup and chooses to create new person

In this case the duplicate case selection popup won't be shown because the system will try to create a new case.
However because the case from the CSV already contains the UUID of the exported case, it will just update the existing case with the same UUID.

The result will be:
* a new person will be created which is not tied to a case
* existing case will be updated

#### 2. The user is presented with the duplicate person popup and chooses and existing person

In this case the duplicate case selection popup will be shown and the user can choose to select a case or create a new case.
If the user choose to create a new case, the same behavior will happen as in the previous point, because the CSV contains a valid UUID of the case, the existing case will be updated.

The result will be:
* existing person will be updated
* existing case will be updated

#### A decision has to be made here on how to handle imports where the UUID is a valid UUID of an existing case:
1. Either the uuid is completely ignored when the user chooses to create a new case and assign a new UUID
[...]

PR #5820, opened 2021-06-16
fix/5705_export_import_with_poe

vidi42:
Fixes export/import issue regarding point of entry for #5705


MateStrysewske (changes requested):
As discussed - adjust CaseImportFacadeEjb to translate the PoE String rather than exporting the database name

MateStrysewske (approved):

PR #5855, opened 2021-06-18
Bug 5763 setting contacts to no contact / cancelling follow up in bulk edit mode can lead to data loss

### LINKED ISSUE #5763: Setting contacts to "no contact" / Cancelling follow up in bulk edit mode can lead to data loss
### Bug Description

When setting a contacts status to "NOT A CONTACT" the field "FOLLOW-UP STATUS COMMENT" is overwritten by the message 
"[System] Follow-up automatically canceled because contact was dropped"
The original contents of this field are lost.
Similar: When cancelling follow up the message "Canceled by User Name using bulk action" overwrites the fields contents.

The field can contain potentially important data and should not be overwritten, especially without a warning. This behaviour can lead to unnoticed and undocumented data loss.

### Expected Behavior

The original contents of the field should be kept in any case.

A) [Preferred] The status message is appended or prepended to the existing contents
B) No message is inserted, the previous contents of the field are left unchanged 

Being able to see in what way the follow-up was cancelled is a "nice to have", but losing potentially important data is a no-go.

### System Details
* SORMAS version: Version: 1.59.2
* User Role: Admin for bulk edit / any other user for "Not a contact"

FredrikSchaeferVitagroup:
Fixes #5763

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactDto.java
@@ -522,7 +522,7 @@ public class ContactDto extends PseudonymizableDto implements SormasToSormasEnti
 	}
 
 	public void addToFollowUpComment(String comment) {
-		followUpComment = followUpComment == null | "".equals(followUpComment) ? comment : followUpComment + "\n" + comment;
+		followUpComment = followUpComment == null || "".equals(followUpComment) ? comment : followUpComment + "\n" + comment;
 	}
 
 	public Float getReportLatLonAccuracy() {
--- a/sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactDto.java
@@ -522,7 +522,7 @@ public class ContactDto extends PseudonymizableDto implements SormasToSormasEnti
 	}
 
 	public void addToFollowUpComment(String comment) {
-		followUpComment = followUpComment == null || "".equals(followUpComment) ? comment : followUpComment + "\n" + comment;
+		followUpComment = DataHelper.joinStrings("\n", followUpComment, comment);
 	}
 
 	public Float getReportLatLonAccuracy() {
--- a/sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactDto.java
@@ -521,6 +521,10 @@ public class ContactDto extends PseudonymizableDto implements SormasToSormasEnti
 		this.followUpComment = followUpComment;
 	}
 
+	public void addToFollowUpComment(String comment) {
+		followUpComment = followUpComment == null | "".equals(followUpComment) ? comment : followUpComment + "\n" + comment;
+	}
+
 	public Float getReportLatLonAccuracy() {
 		return reportLatLonAccuracy;
 	}
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
@@ -1365,6 +1365,12 @@ public class ContactFacadeEjb implements ContactFacade {
 
 			String followUpComment = null;
 			if (dto.isPseudonymized()) {
+				/**
+				 * Usually, pseudonymized values are not edited, so the pseudonymizer can just overwrite them in the dto when restoring.
+				 * One exception is the followUpComment, which can be pseudonymized, but still be edited by automatic system messages,
+				 * e.g. when cancelling follow up.
+				 * This attribute's value is extracted here before pseudonymized vales are restored and then added again.
+				 */
 				followUpComment = dto.getFollowUpComment();
 			}
 			Pseudonymizer pseudonymizer = Pseudonymizer.getDefault(userService::hasRight);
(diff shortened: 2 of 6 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -1362,11 +1362,26 @@ private void restorePseudonymizedDto(ContactDto dto, Contact existingContact, Co
 		if (existingContactDto != null) {
 			boolean isInJurisdiction = contactJurisdictionChecker.isInJurisdictionOrOwned(existingContact);
 			User currentUser = userService.getCurrentUser();
+
+			String followUpComment = null;
+			if (dto.isPseudonymized()) {
```
leventegal-she, line 1367: I don't see this to be a good solution, the followup comment can be changed through the REST API. It could work as a hotfix but we need a new change request ticket for properly handling system generated comments. Maybe store them in a separate column in the db.

the condition `dto.isPseudonymized()` is always not enough, you should check also if the followup field is accessible or not; see the condition in DtoPseudonymizer::restorePseudonymizedValues 
```java
if (!getFieldAccessCheckers(isInJurisdiction).isAccessible(pseudonymizedField, pseudonymizeMandatoryFields) || dto.isPseudonymized()) {
   restoreOriginalValue(dto, pseudonymizedField, originalDto);
}
````

The accessibility check is needed when someone edits a contact that gets out of jurisdiction in the mean time. In this case it will not work to save the followup comment from the DTO and add it after restore because it will duplicate the comment text.

Similar thing should be done for cases too.

**Review by leventegal-she — CHANGES_REQUESTED** [review]

File: sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactService.java [hunk 2]
```diff
@@ -813,7 +813,11 @@ public void updateFollowUpDetails(Contact contact, boolean followUpStatusChanged
 				if (contact.getFollowUpStatus() != FollowUpStatus.COMPLETED && contact.getContactStatus() == ContactStatus.CONVERTED) {
 					// Cancel follow-up if the contact was converted to a case
 					contact.setFollowUpStatus(FollowUpStatus.CANCELED);
-					contact.setFollowUpComment(I18nProperties.getString(Strings.messageSystemFollowUpCanceled));
+					String followUpComment = contact.getFollowUpComment();
```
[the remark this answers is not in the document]
FredrikSchaeferVitagroup: The method deals with a Contact, not with a ContactDto. 
I could add a `addToFollowUpComment` for the Contact class, but so I have not yet seen entity classes containing methods with logic in SORMAS. That's why I chose this implementation.
Does it appear reasonable to you now?

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactService.java
```diff
@@ -832,7 +836,9 @@ public void updateFollowUpDetails(Contact contact, boolean followUpStatusChanged
 
 	public void cancelFollowUp(Contact contact, String comment) {
 		contact.setFollowUpStatus(FollowUpStatus.CANCELED);
-		contact.setFollowUpComment(comment);
+		String followUpComment = contact.getFollowUpComment();
```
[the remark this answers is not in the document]
FredrikSchaeferVitagroup: Same as above.

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactService.java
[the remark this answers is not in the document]
leventegal-she: Then please use `DataHelper.joinStrings`, and also would be better to have a helper function so the logic is not duplicated.

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
FredrikSchaeferVitagroup: I added a field accessibility check.

I could not find system follow up comments for cases, that's why there is nothing similar for cases.

I am also not very happy with this solution. Why are changes via the REST API a problem? They call the same method...

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -1362,11 +1362,27 @@ private void restorePseudonymizedDto(ContactDto dto, Contact existingContact, Co
 		if (existingContactDto != null) {
 			boolean isInJurisdiction = contactJurisdictionChecker.isInJurisdictionOrOwned(existingContact);
 			User currentUser = userService.getCurrentUser();
+
 			Pseudonymizer pseudonymizer = Pseudonymizer.getDefault(userService::hasRight);
 
+			String followUpComment = null;
+			if (dto.isPseudonymized() || !pseudonymizer.isAccessible(ContactDto.class, "followUpComment", isInJurisdiction)) {
```
leventegal-she, line 1369: Please use `CaseDataDto.FOLLOWUP_COMMENT` instead of hardcoded `"followUpComment"`

sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactDto.java
```diff
@@ -521,6 +521,10 @@ public void setFollowUpComment(String followUpComment) {
 		this.followUpComment = followUpComment;
 	}
 
+	public void addToFollowUpComment(String comment) {
+		followUpComment = followUpComment == null || "".equals(followUpComment) ? comment : followUpComment + "\n" + comment;
```
leventegal-she, line 525: could use `DataHelper.joinStrings` here as well

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
leventegal-she: Changes via the REST API could have been a problem without checking the field access. There could have been a similar issue I mentioned in my previous comment: 

> The accessibility check is needed when someone edits a contact that gets out of jurisdiction in the mean time. In this case it will not work to save the followup comment from the DTO and add it after restore because it will duplicate the comment text.

vidi42 (approved):

PR #5856, opened 2021-06-18
fix/5705_export_import_duplicate_check

### LINKED ISSUE #5705: Duplicate person detection creates persons but not cases

vidi42:
Fixes the import of exported entities by ignoring the UUID when new entities should be created.

Affects: cases, contacts and event participants for which exported CSVs could be imported.

Fixes #5705


vidi42:
Related to https://github.com/hzi-braunschweig/SORMAS-Project/issues/5705#issuecomment-863072592

MateStrysewske (changes requested):
The main issue of this bug is not resolved - if I export a case, change its name and then re-import it, the system automatically overwrites the existing case because it has the same UUID. The proposed solution was to completely ignore the UUID column during import, i.e. when building the entity based on the CSV file, the UUID is NOT taken over into the created entity (i.e. there's always a new UUID generated).

MateStrysewske (approved):

PR #5864, opened 2021-06-18
[#5863] allow local external URLs

### LINKED ISSUE #5863: validateExternalUrls in ConfigFacadeEjb does not account for local URLs
### Problem Description
`validateExternalUrls` expects full domains with a TLD, however, if you use for example `localhost` or a docker-compose setup (e.g., my_service) to reference the service, validation will fail. 

### Proposed Change
Use `UrlValidator.ALLOW_LOCAL_URLS` as option to support this case.

JonasCir:
Fixes #5863


JonasCir:
@MateStrysewske Do you have time to review this? Cool with me if not :) Otherwise @DavidBaldsiefen maybe? :D

MateStrysewske:
@JonasCir Not before tomorrow, I'll probably be busy with refinements the whole day

DavidBaldsiefen (approved):

PR #5869, opened 2021-06-18
filter+pagination for coutries, regions, districts, communities

marius2301:
Fixes #


sormas-api/src/main/java/de/symeda/sormas/api/region/CommunityFacade.java
```diff
@@ -36,6 +37,8 @@
 
 	List<CommunityDto> getIndexList(CommunityCriteria criteria, Integer first, Integer max, List<SortProperty> sortProperties);
 
+	public Page<CommunityDto> getIndexPage(CommunityCriteria communityCriteria, Integer offset, Integer size, List<SortProperty> sortProperties);
```
MateStrysewske, line 40: Remove the "public" keyword

MateStrysewske (changes requested):

PR #5872, opened 2021-06-20
Bugfix 5841 contacts with source case not shown

alexcaruntu-vita:
Fixes #

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseController.java
@@ -331,7 +331,10 @@ public class CaseController {
 		convertToCaseSelectComponent.getDiscardButton().setCaption(I18nProperties.getCaption(Captions.actionCancel));
 
 		convertToCaseSelectComponent.addCommitListener(() -> {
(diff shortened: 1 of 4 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseController.java
```diff
@@ -335,7 +331,10 @@ private void showConvertToCaseSelection(
 		convertToCaseSelectComponent.getDiscardButton().setCaption(I18nProperties.getCaption(Captions.actionCancel));
 
 		convertToCaseSelectComponent.addCommitListener(() -> {
-			setResultingCase(caze, convertToCaseSelectionField.getSelectedContacts(), convertToCaseSelectionField.getSelectedEventParticipants());
+			List<SimilarContactDto> selectedContacts = convertToCaseSelectionField.getSelectedContacts();
+			caze.getEpiData().setContactWithSourceCaseKnown(selectedContacts.isEmpty() ? YesNoUnknown.NO : YesNoUnknown.YES);
```
MateStrysewske, line 335: Don't set the value to NO if the list is empty because it's possible that the case already exists and does have a source contact, in which case this code would actually produce a bug; it's sufficient to set it to YES if the list is not empty and leave it like it is otherwise

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #5926, opened 2021-06-23
5852 refine case jurisdiction

### LINKED ISSUE #5852: Refine the split of jurisdiction and place of stay [5]
### Feature Description

The place of stay was recently split into the jurisdiction of the case and the place of stay. This was implemented to allow users to enter case data properly where the cases place of stay differs from its jurisdiction.
### Problem Description

The split of place of stay now makes the reporting district field obsolete because the jurisdiction of the case will always be the district that reports the case. The problem that arises now is that the fallback for the mapping to SurvNet considers the place of stay when reporting district is missing and does not consider the jurisdiction if different from the place of stay. This results in health departments reporting for other districts which is incorrect.

### Proposed Change

- Move the jurisdiction above the place of stay (these should stay compulsory) and put the check boxes underneath the location fields
- Always enable the jurisdiction fields and make them the default entry
- Revert the logic of the checkbox "Responsible jurisdiction of this case differs from its place of stay" and enable additional region, district and community fields under the place of stay if selected. Otherwise, leave them disabled.
- When unchecked, the jurisdiction serves as the location information
- Add the jurisdiction also to the case creation form with the same logic in association with place of stay
- **If feasible (tbd):** Remove the field reporting district and migrate the districts to the jurisdiction districts, fill the region according to the respective region references

**Updated specifications:**
* Move the responsible jurisdiction above the place of stay; move the checkboxes between the jurisdiction fields and the "Place of stay" heading
* Always enable the responsible jurisdiction fields and make them mandatory
* Write database migration script that for each case in the system:
  * If the responsible jurisdiction fields are empty AND "reporting district" is empty OR the same as the place of stay district, copy the place of stay region, district and community over to the responsible jurisdiction fields AND clear the place of stay region, district and community
  * If the responsible jurisdiction fields are empty and "reporting district" is NOT empty AND differs from the place of stay district, copy the reporting district over to the responsible district and also fill in the responsible region with the region of the reporting district (reporting region is not stored in the case, so this requires to consult the district table)
  * If the responsible jurisdiction is filled in, no migration is needed
* Remove the "responsible district" field from cases (no database script needed in the mobile app because SQLite doesn't support DROP COLUMN, just remove the field from the data model)
* Change the caption of the checkbox "Responsible jurisdiction of this case differs from its place of stay" to "Place of stay of this case differs from its responsible jurisdiction"; if unchecked, the place of stay region, district and community should be hidden, cleared, and made optional; if checked, the place of stay region, district and community should be visible, and region and district need to be mandatory;
* Place of stay facility- and point of entry-related fields need to stay as they are right now, i.e. they stay mandatory even if the place of stay does not differ from the responsible jurisdiction, so no need to change anything about those
* Add the responsible jurisdiction to the case creation forms (i.e. case creation forms need to work exactly as the edit forms in regards of jurisdiction fields), both in web and mobile
* Adjust the line listing form and logic - use responsible region, district and community instead of place of stay region, district and community
* Change all backend logic to incorporate these changes, i.e. responsible jurisdiction is always present and adjusted place of stay is optional; if I'm not mistaken, logic about task and user assignments don't need to be changed because they already prioritise users from the responsible jurisdiction over those from the place of stay; however, there's probably a ton of other logic that needs to be adjusted; for example when calculating the jurisdiction for contacts - those fall back to the place of stay of the case if they're emtpy, but now have to fall back to the responsible jurisdiction of the case
* Make sure that case import works accordingly, i.e. it's required to specify a responsible region and district, and a place of stay facility, but it's not necessary to specify a place of stay region and district (probably need to change the validation logic in CaseFacadeEjb)
* If the place of stay region, district or community is referenced in the UI, e.g. in grids (case directory), change it to the responsible region, district or community
 
### Additional Information

Example where place of stay equals jurisdiction:
jursidictionPlaceOfStay2

Examples for differing locations
jursidictionPlaceOfStay

jursidictionPlaceOfStay3

leventegal-she:
Fixes #5852


sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -7433,4 +7433,38 @@ ALTER TABLE cases ADD COLUMN dontsharewithreportingtool boolean DEFAULT false;
...
+                SET responsibleregion_id = (SELECT region_id from district where id = _case.reportingdistrict_id),
+                    responsibledistrict_id = _case.reportingdistrict_id,
+                    reportingdistrict_id = null
+                where id = _case.id;
+            END LOOP;
+    END;
+$$ LANGUAGE plpgsql;
+
+ALTER TABLE cases DROP COLUMN reportingdistrict_id;
+ALTER TABLE cases_history DROP COLUMN reportingdistrict_id;
```
MateStrysewske, line 7467: Don't drop the history table column so that we can restore data in case something goes wrong during migration

MateStrysewske (changes requested):
While looking at the code, I realised that the current filtering mechanism is a bit flawed because it basically does a cross-filtering (responsible region == region && responsibleRegion == responsibleRegion), which is kind of confusing. I've consulted Kay again on this matter and he decided that the filtering logic for now should be:

a) if place of stay region/district/community is filled in, filter by those
b) otherwise, fall back to responsible region/district/community

Can you please adjust the parts of the code that do these filterings accordingly? That means we probably don't need a dedicated responsibleRegion field in CaseCriteria for example. In the future, we will probably have to add additional filters to allow users to differentiate between the two, but that's of course out of scope for this issue.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1233,7 +1230,12 @@ public Long countCasesForMap(
 		Predicate nameSimilarityFilter =
 			cb.gt(cb.function("similarity", double.class, nameSimilarityExpr, nameSimilarityExpr2), configFacade.getNameSimilarityThreshold());
 		Predicate diseaseFilter = cb.equal(root.get(Case.DISEASE), root2.get(Case.DISEASE));
-		Predicate regionFilter = cb.equal(region.get(Region.ID), region2.get(Region.ID));
+		Predicate responsibleRegionFilter = cb.or(
```
MateStrysewske, line 1233: Okay, after seeing this code, I think it actually doesn't make sense to do a cross-check here. We should only check whether responsible jurisdiction == responsible jurisdiction and place of stay == place of stay when searching for duplicates.

MateStrysewske (changes requested):

PR #5933, opened 2021-06-24
#5850 Make sure S2S tests are correctly executed

### LINKED ISSUE #5850: Mocks for S2SRest Client are broken and default to noops at the call site
### Bug Description
Mocks for the S2S Rest client are not working correctly. The lambdas in `thenAnswer` are silently ignored.

Take a look at `SormasToSormasCaseFacadeEjbTest.testShareCase` for example. The arguments of the `post` function have been changed in the past and the test has not been updated (e.g., the first argument of the invocation is now the id, not the URL), so this assertion should fail (what it doesn't). If you place a breakpoint in the lambda, it will never be hit.

TheREST  client gets correctly injected into `AbstractSormasToSormasInterface`, but the call to `post` in `AbstractSormasToSormasInterface.shareEntities` is basically a noop as the `thenAnswer` of the mock is never executed. 

Interestingly, the Mocks for the S2S config (i.e., `mockDefaultServerAccess`) work perfectly fine.

### Steps to Reproduce
1. Run `SormasToSormasCaseFacadeEjbTest.testShareCase`. Although the assertions in `thenAnswer` are wrong, the test passes.

### Expected Behavior
Test passes, but with `thenAnswer` being called.

--- 
@StefanKock any idea?

DavidBaldsiefen:
Fixes #5850

### DIFF
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/sormastosormas/SormasToSormasCaseFacadeEjbTest.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/sormastosormas/SormasToSormasCaseFacadeEjbTest.java
@@ -21,13 +21,10 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.isEmptyString;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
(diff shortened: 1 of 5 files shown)


JonasCir (approved):
lgtm thank you!

PR #5995, opened 2021-06-29
Mass convert event participants to contacts 

### LINKED ISSUE #4969: Transform several PPs into contact from the directory of event participants
### Situation Description
To date, the number of participants in an event and especially a cluster can be very high. The transformation of event participants into contact is very time-consuming and takes place person by person.

### Feature Description
We want to add a bulk-action to the list of event participants plus a  "line-list contact creation" action.
This action opens a pop-up that asks: (inspired by the popup for creating a new contact from the screen of an event participant).

• Common information, namely:

- disease of the source case (mandatory, pre-filled depending on the event, not editable)
- source case with choice button (not compulsory)
- responsible region (mandatory if no source case selected)
- responsible department (mandatory if no source case selected)

• Information specific to each participant (one line per participant, presentation close to the "line-list" entry of cases)

- first name of the contact (pre-filled, not editable)
- contact name (pre-filled, not editable)
- date of the report (required)
- date of last contact (not compulsory) (with possibility of contact over several days as currently)
- type of contact (not compulsory) => in the form of a drop-down list
- relation to the case (not compulsory)

### Additional Information
@ftavin @bernardsilenou  @emond-fuseau

tbroyer:
Fixes #4969


sormas-api/src/main/resources/captions.properties
```diff
@@ -244,6 +244,7 @@ bulkDelete=Delete
 bulkDisease=Change Disease
 bulkEdit=Edit...
 bulkEventInvestigationStatus=Change event investigation status
+bulkEventParticipantsToContacts=Generate contacts
```
MateStrysewske, line 247: I think I'd prefer "Create contacts" here because "Generate" sounds a bit technical to me - open for discussion though

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #6014, opened 2021-06-30
#5622: Added reportId to LabMessage

### LINKED ISSUE #5622: [DEMIS2SORMAS] Handle New Profile: Enable SORMAS to relate lab messages to each other [2]
### Situation Description

The new DEMIS profiles include a lifecycle management of DEMIS notifications (some overview of that is given in #5138). This means DEMIS notifications can soon relate to each other, for example one DEMIS notification can extend information from a previous one. Theoretically, this could possibly be handled by the current duplicate detections called when processing lab messages. For the processing user, it would be much more convenient (and less prone to errors), if SORMAS would propose 'shortcuts'.

For example, if there is a lab message that relates to another lab message already processed, the user should not need to go through the person, case/contact/event-participant, and sample selections. Instead, a popup could propose to directly go to the related sample. The user should be able to follow that proposal, or process the lab message the usual manual way.

### Feature Description

- [x] Add a String attribute called reportId to the lab message dto and entity. 
- [x] Lab messages related to each other should have the same reportId. This should be documented somewhere in the code.
- [x] Whenever a user starts processing a lab message, when this lab messages reportId is not null, SORMAS needs to automatically search for already processed lab messages with the same reportId. Therefore, provide an getByReportId(...) in the labMessageFacade.

### Additional Information

The demis-adapter needs to fill the new reportId attribute, probably with the Composition.identifier.value. The related adapter ticket is [#14](https://github.com/hzi-braunschweig/sormas-demis-adapter/issues/14).

ChristopherRiedel:
Fixes #5622

### DIFF
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjbTest.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjbTest.java
@@ -18,6 +18,12 @@ package de.symeda.sormas.backend.labmessage;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.util.List;
 
 import org.junit.Test;
 
@@ -33,5 +39,28 @@ public class LabMessageFacadeEjbTest extends AbstractBeanTest {
 		getLabMessageFacade().deleteLabMessage(labMessage.getUuid());
 
 		assertThat(getLabMessageFacade().getByUuid(labMessage.getUuid()), is(nullValue()));
+	}
+
+	@Test
+	public void testGetByReportId() {
+		String reportId = "123456789";
+		creator.createLabMessage((lm) -> lm.setReportId(reportId));
+
+		List<LabMessageDto> list = getLabMessageFacade().getByReportId(null);
+
+		assertNotNull(list);
+		assertTrue(list.isEmpty());
+
+		list = getLabMessageFacade().getByReportId("");
+
+		assertNotNull(list);
+		assertTrue(list.isEmpty());
+
+		list = getLabMessageFacade().getByReportId(reportId);
+
+		assertNotNull(list);
+		assertFalse(list.isEmpty());
+		assertEquals(1, list.size());
+		assertEquals(reportId, list.get(0).getReportId());
 	}
 }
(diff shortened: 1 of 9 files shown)


sormas-backend/src/test/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjbTest.java
```diff
@@ -33,4 +39,17 @@ public void deleteLabMessage() {
 
 		assertThat(getLabMessageFacade().getByUuid(labMessage.getUuid()), is(nullValue()));
 	}
+
+	@Test
+	public void testGetByReportId() {
```
FredrikSchaeferVitagroup, line 44: I would like the test to be a little more extensive:
What happens when null or an empty string is given as reportId? What happens with reportId with a space " " at the end or beginning? Is in those cases really the correct set of LabMessageDtos returned? 

Is there a distinction between uppercase and lowercase letters? I think there should not be, at least in some other parts of SORMAS, there is none (e.g. [here](https://github.com/hzi-braunschweig/SORMAS-Project/blob/6d3a033031eae61ba09c75f7f70ebc15e5528f64/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java#L3653), which also is about external Ids of some sort).

FredrikSchaeferVitagroup (changes requested):

sormas-backend/src/test/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjbTest.java
```diff
@@ -33,4 +40,27 @@ public void deleteLabMessage() {
...
+		assertNotNull(list);
+		assertTrue(list.isEmpty());
+
+		list = getLabMessageFacade().getByReportId(reportId);
+
+		assertNotNull(list);
+		assertFalse(list.isEmpty());
+		assertEquals(1, list.size());
+		assertEquals(reportId, list.get(0).getReportId());
+	}
```
FredrikSchaeferVitagroup, line 65:
```suggestion
	@Test
	public void testGetByReportIdWithCornerCaseInput() {
		String reportId = "123456789";
		creator.createLabMessage((lm) -> lm.setReportId(reportId));

		List<LabMessageDto> list = getLabMessageFacade().getByReportId(null);

		assertNotNull(list);
		assertTrue(list.isEmpty());

		list = getLabMessageFacade().getByReportId("");

		assertNotNull(list);
		assertTrue(list.isEmpty());
	}

	@Test
	public void testGetByReportIdWithOneMessage() {

		String reportId = "123456789";
		creator.createLabMessage((lm) -> lm.setReportId(reportId));

		// create noise
		creator.createLabMessage(null);
		creator.createLabMessage((lm) -> lm.setReportId("some-other-id"));

		List<LabMessageDto> list = getLabMessageFacade().getByReportId(reportId);

		assertNotNull(list);
		assertFalse(list.isEmpty());
		assertEquals(1, list.size());
		assertEquals(reportId, list.get(0).getReportId());
	}
```

FredrikSchaeferVitagroup (changes requested):
Also, please update resolve merge conflicts

PR #6036, opened 2021-07-02
#191 When sending data to Sormas during initial survnet import, create the associated shared info data

### LINKED ISSUE #191: Write app synchronization log, including information on overriden fields [2]
In the rare situation when a single field is changed by two users at the same time, the later user should be informed which of the entered data is overridden.

- [x] create a SyncLog entity: entity name, uuid, pushed/pulled, conflict text
- [x] collect all fields that were overriden during merge and write them into conflict text
- [x] display all logs that had conflicts in an alert popup. Group by date (+5 minutes)
- [x] show only 50 entries and add button to load next 50

### LINKED ISSUE #6038: Mark cases as sent to the reporting tool after the initial import
### Feature Description
Complementary task for [Import] Mark cases as sent to the reporting tool after the initial import #191
Initial import upon rollout of SurvNet interface should mark imported entities as shared with SurvNet

### Problem Description
When data is initially imported into Sormas, cases are not marked as sent to the reporting tool which leads to them wrongfully appearing on the directory when filtering with reporting tool associated filters.

### Proposed Change
Enhance the **/push** endpoint for _cases_ and _events_ to be able to also create the necessary surveillance info upon SurvNet import

Alexandru-Pirvu:
Fixes #6038

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/externalsurveillancetool/ExternalSurveillanceToolFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/externalsurveillancetool/ExternalSurveillanceToolFacade.java
@@ -40,6 +40,10 @@ public interface ExternalSurveillanceToolFacade {
 
 	void sendEvents(List<String> eventUuids) throws ExternalSurveillanceToolException;
 
+	void createCaseShareInfo(List<CaseDataDto> cases);
+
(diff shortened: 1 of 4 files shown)


vidi42:
Converted to Draft until the hotfix for 1.61.1 is done to avoid this being merged into development

PR #6061, opened 2021-07-05
#6016 Correctly update POEs based on district

### LINKED ISSUE #6016: Separation of responsible region and districts breaks the logic for  POE cases
### Bug Description

POint of entry cases can not be created without setting "place of stay of this case differs from responsible jurisdiction == True".
- point of entry (POE) of a case is validated using region and district (ie when place of stay of this case differs from responsible jurisdiction == True).  
- when creating POE cases, POE in the responsible region are not suggested as drop down options in "point of entry"
This should be based on the responsible region and district.
### Steps to Reproduce

1. Create a POE case
2. Under POE, the names are POE linked to the responsible jurisdiction are not suggested to user
3. set `place of stay of this case differs from responsible ` to true and choose same region and district, now you have the names of PEO suggested
4. When saving the case, do not choose place of stay since this is not required for POE cases.
6. During this process of editing or referring the case, system keeps showing the warning message "A problem has occurred: data could not be saved because it was edited in the meantime" This error appears when the same user is editing the form.
7. In case edit form, the place of stay (region, district , etc) cannot be edited. It should be possible to edit it. (this one is an old bug I think so)

### Expected Behavior

- Use responsible jurisdiction when filtering enums for "point of entry"
- if "place of stay of this case differs from responsible jurisdiction == True", then suggest POE in both regions in the dropdown for point of entry
### System Details

* Device:
* SORMAS version: 1.61
* Android version/Browser:
* Server URL: https://release-international.sormas.netzlink.com/
* User Role: admin + national user

DavidBaldsiefen:
Fixes #6016

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
@@ -34,8 +34,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.apache.commons.collections.CollectionUtils;
 
@@ -319,21 +317,8 @@ public class CaseCreateForm extends AbstractEditForm<CaseDataDto> {
 					FacadeProvider.getFacilityFacade()
 						.getActiveFacilitiesByDistrictAndType(districtDto, (FacilityType) facilityType.getValue(), true, false));
 			}
-			// show all possible POEs for jurisdiction-district and placeofstay-district
-			if (differentPlaceOfStayJurisdiction.getValue()) {
-				List<PointOfEntryReferenceDto> allPOEs = districtDto == null
-					? Collections.emptyList()
-					: FacadeProvider.getPointOfEntryFacade().getAllActiveByDistrict(districtDto.getUuid(), true);
-				if (responsibleDistrictCombo.getValue() != null && ((DistrictReferenceDto) responsibleDistrictCombo.getValue()) != districtDto) {
-					allPOEs = Stream.concat(
-						allPOEs.stream(),
-						FacadeProvider.getPointOfEntryFacade()
-							.getAllActiveByDistrict(((DistrictReferenceDto) responsibleDistrictCombo.getValue()).getUuid(), allPOEs.size() == 0)
-							.stream())
-						.collect(Collectors.toList());
-				}
-				FieldHelper.updateItems(cbPointOfEntry, allPOEs);
-			}
+
+			updatePOEs();
 		});
 		communityCombo.addValueChangeListener(e -> {
 			if (!TypeOfPlace.HOME.equals(facilityOrHome.getValue())) {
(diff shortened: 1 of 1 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
```diff
@@ -317,9 +319,21 @@ protected void addFields() {
 					FacadeProvider.getFacilityFacade()
 						.getActiveFacilitiesByDistrictAndType(districtDto, (FacilityType) facilityType.getValue(), true, false));
 			}
-			FieldHelper.updateItems(
-				cbPointOfEntry,
-				districtDto != null ? FacadeProvider.getPointOfEntryFacade().getAllActiveByDistrict(districtDto.getUuid(), true) : null);
+			// show all possible POEs for jurisdiction-district and placeofstay-district
+			if (differentPlaceOfStayJurisdiction.getValue()) {
```
leventegal-she, line 323: If `differentPlaceOfStay` is checked then it should show only the POE-s of the place of stay district to be consistent with the health facility selection.

If it needs to be filtered by both districts, then it should be a topic for a new change request and both POE and facility selection should be adopted and also the validation on the backend side.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
```diff
@@ -456,6 +470,22 @@ protected void addFields() {
 			if (differentPlaceOfStay == null || Boolean.FALSE.equals(differentPlaceOfStay)) {
 				updateFacility();
 			}
+			DistrictReferenceDto districtDto = (DistrictReferenceDto) e.getProperty().getValue();
```
leventegal-she, line 473: same as for district change handler

leventegal-she (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
DavidBaldsiefen: I see, I implemented it this way because of @bernardsilenou comment in the related ticket:
> if "place of stay of this case differs from responsible jurisdiction == True", then suggest POE in both regions in the dropdown for point of entry

Should we wait for feedback from @bernardsilenou / @kwa20 ?

leventegal-she: I would to do this similarly to health facility to be consistent and if needed, @bernardsilenou / @kwa20  would create a change request. 

I don't really say the reason for having a POE linked to jurisdiction district when there is a different place of stay district. As I know the POE should be the place where the case is located so if it has different place of stay then the POE must be in the place of stay district and not on the jurisdiction district.

@MateStrysewske what do you think?

kwa20: @DavidBaldsiefen I agree with @leventegal-she. There should not be a need to filter for both regions/districts at the same time because if the POE is in another region/district, it would not make sense to show the POE's in the jurisdiction

DavidBaldsiefen: Alright, then I will filter the POEs only based on the placeofstay (if provided). Otherwise based on jurisdiction.

leventegal-she (approved):

PR #6087, opened 2021-07-07
#6080 persist case completeness without changedate

### LINKED ISSUE #6080: CaseService.updateCompleteness: Persist calculated completeness without updating the Case.changeDate [1]
### Problem Description

Follow up to #5543: When sending data from SurvNet (initial import), the last change imposes two severe problems:

1. When the completeness is calculated, the Case.changeDate is updated. This causes that the case is then implicitly marked to be sent to SurvNet. This needs to be avoided.
2. The asynchronuous calculation can clash with an OptimisticLockException while the import from SurvNet is still running.

### Proposed Change

Discussed with @MartinWahnschaffe and @HolgerReiseVSys: The completeness value is a calculated value that shall not trigger anymore the changeDate to be updated. This avoids unneeded sync roundtrips towards the reporting tool and the android app.

How: Write the completeness value directly via SQL, not via JPA Entity mechanisms.

StefanKock:
Fixes #6080 

Side effect: Auditlog is not triggered (the same with batch archive/dearchive of cases).

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
@@ -1318,8 +1318,21 @@ public class CaseService extends AbstractCoreAdoService<Case> {
 	}
 
 	public void updateCompleteness(Case caze) {
-		caze.setCompleteness(calculateCompleteness(caze));
-		ensurePersisted(caze);
+
+		float completeness = calculateCompleteness(caze);
+
+		/*
+		 * Set the calculated value without updating the changeDate:
+		 * 1. Do not trigger sync mechanisms that compare changeDates
+		 * 2. Avoid optimistic locking problem with parallel running logic like batch imports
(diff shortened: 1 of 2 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
```diff
@@ -1318,8 +1318,20 @@ public void updateCompleteness(List<String> caseUuids) {
 	}
 
 	public void updateCompleteness(Case caze) {
```
MartinWahnschaffe, line 1320: Please add a comment that this does not trigger the audit log

MartinWahnschaffe (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
StefanKock: Added in the code comment

Issue #2332, opened 2020-06-16
Document read-only query hint [1]


StefanKock:
### Problem Description

The query hint `ModelConstants.HINT_HIBERNATE_READ_ONLY` is rarely used, and sometimes pointless.

### Proposed Change
1. Rename constant to `ModelConstants.READ_ONLY` or `ModelConstants.HINT_READ_ONLY` for better readability when using it.
2. Add Javadoc:
```
A query hint to make JPA entities loaded as read-only to avoid caching and dirty checks. Usage:

"List<JpaEntity> resultList = em.createQuery(cq).setHint(ModelConstants.READ_ONLY, true).getResultList();"

Note: This does not have any effect on Multiselect-Queries directly into DTOs.
```
3. Remove usages on non-Entity queries.

### Additional Information
Hibernate documentation: https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html

AndyBakcsy-she:
Verified ticket while looking through the code

PR #6088, opened 2021-07-07
#6076 refactor batched querying

### LINKED ISSUE #6076: Simplify batched query fetching [1]
### Problem Description

We have about 50 times this copy-paste pattern with 6 lines in Services and Facades:
```java
List<TaskIndexDto> tasks;
if (first != null && max != null) {
	tasks = em.createQuery(cq).setFirstResult(first).setMaxResults(max).getResultList();
} else {
	tasks = em.createQuery(cq).getResultList();
}
```

### Proposed Change

Extract this pattern into QueryHelper to shrink down the usage to this one-liner:
```java
List<TaskIndexDto> tasks = QueryHelper.getResultList(em, cq, first, max);
```

### Additional Information

I found method `CaseFacade.getCaseFollowUpList` where this refactoring would avoid exceptions because the `null` checks are currently missing.

StefanKock:
Fixes #6076

I provide now two new functionalities:
- `QueryHelper.getResultList`
- `QueryHelper.getSingleResult`

I accidently mixed it up with finding the first entry (where multiple results exist). I fixed this and will provide a refactoring for `getFirstResult` also.

After finished review I would like to wait for v1.61.1 to be finished. There otherwise might come up merge conflicts on hotfix-finish.

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
@@ -52,6 +52,7 @@ import de.symeda.sormas.backend.region.Community;
 import de.symeda.sormas.backend.region.District;
 import de.symeda.sormas.backend.region.Region;
 import de.symeda.sormas.backend.user.User;
+import de.symeda.sormas.backend.util.QueryHelper;
 
 @Stateless
 @LocalBean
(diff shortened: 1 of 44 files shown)


MateStrysewske (approved):
Looks good to me; the only thing I'm not quite sure about is that some READ_ONLY hints are dismissed with this refactoring. I can't judge whether that makes a difference though.

StefanKock:
> Looks good to me; the only thing I'm not quite sure about is that some READ_ONLY hints are dismissed with this refactoring. I can't judge whether that makes a difference though.

#2332 deals with this: I replaced this only on DTO queries where it doesn't have an effect anyway. But if we would want to use this on entity queries combined with my approach, we need to enhance the getResultList method I created.

PR #6099, opened 2021-07-08
6038 initial import share info

### LINKED ISSUE #6038: Mark cases as sent to the reporting tool after the initial import

HolgerReiseVSys:
Fixes #6038


sormas-rest/src/main/java/de/symeda/sormas/rest/ExternalSurveillanceToolGatewayResource.java
```diff
@@ -12,30 +9,27 @@
...
-    @POST
-    @Path("/import/events")
-    public List<PushResult> importEvents(@Valid List<EventDto> dtos) {
-        List<PushResult> result = savePushedDto(dtos, FacadeProvider.getEventFacade()::saveEvent);
-        FacadeProvider.getExternalSurveillanceToolFacade().createEventShareInfo(dtos);
-        return result;
-    }
+	@POST
+	@Path("/share/cases")
+	public void importCases(@Valid List<String> caseUuids) {
```
HolgerReiseVSys, line 25: naming of methods could be improved

PR #6102, opened 2021-07-09
Feature 6022 dea travel entry entity and backend logic

### LINKED ISSUE #6022: DEA TravelEntry entity and backend logic [3]
### Situation Description

Concept issue for travel entries: #5489

### Feature Description

* Add a new `TravelEntry` entity extending `CoreAdo` and the corresponding Dto class with the following fields:
  * person_id (not null)
  * `reportDate` (not null)
  * `reportingUser_id`
  * `archived` - boolean
  * `deleted` - boolean
  * `disease`
  * `diseaseDetails` - text
  * `diseaseVariant`
  * `responsibleRegion_id` (not null)
  * `responsibleDistrict_id` (not null)
  * `responsibleCommunity_id`
  * `pointOfEntryRegion_id`
  * `pointOfEntryDistrict_id`
  * `pointOfEntry_id` (not null)
  * `pointOfEntryDetails` - text
  * `resultingCase_id`
  * `externalId` - text
  * `recovered` - boolean
  * `vaccinated` - boolean
  * `testedNegative` - boolean
  * `deaContent` - json
  * ALL quarantine-related fields that are also part of cases
* Add basic Facade and Service methods (no synchronization needed)
* Pseudonymization is based on the point of entry or responsible jurisdiction of the travel entry, or the responsible jurisdiction of the associated case
* User access is based on the point of entry or responsible jurisdiction of the travel entry; also, users have access to all travel entries that are linked to cases they can access via the resultingCase field
* Extend the person user filter as well as the person resource methods that allow users to retrieve persons they have access to; users also need to have access to persons that are related to travel entries they can access
* Add new user rights: `TRAVEL_ENTRY_MANAGEMENT_ACCESS` (all user roles that have access to PERSON_VIEW), `TRAVEL_ENTRY_VIEW` (all user roles that have access to PERSON_VIEW), `TRAVEL_ENTRY_CREATE` (all user roles that have access to PERSON_EDIT), `TRAVEL_ENTRY_EDIT` (all user roles that have access to PERSON_EDIT), `TRAVEL_ENTRY_DELETE` (all roles that have access to PERSON_DELETE), `TRAVEL_ENTRY_ARCHIVE` (all roles that have access to CASE_ARCHIVE)

sergiupacurariu:
Fixes #6022


sormas-api/src/main/java/de/symeda/sormas/api/travelentry/TravelEntryDto.java
```diff
@@ -0,0 +1,426 @@
...
+	private String quarantineHomePossibleComment;
+	private YesNoUnknown quarantineHomeSupplyEnsured;
+	@SensitiveData
+	private String quarantineHomeSupplyEnsuredComment;
+	private boolean quarantineExtended;
+	private boolean quarantineReduced;
+	private boolean quarantineOfficialOrderSent;
+	private Date quarantineOfficialOrderSentDate;
+
+	private YesNoUnknown wasInQuarantineBeforeIsolation;
```
MateStrysewske, line 75: Isolation fields are not necessary - sorry, I missed those when I refined the issue

sormas-backend/src/main/java/de/symeda/sormas/backend/person/Person.java
```diff
@@ -135,6 +136,7 @@
 	public static final String EVENT_PARTICIPANTS = "eventParticipants";
 	public static final String IMMUNIZATIONS = "immunizations";
 	public static final String ADDITIONAL_DETAILS = "additionalDetails";
+	public static final String TRAVEL_ENTRY = "travelEntries";
```
MateStrysewske, line 139: Rename to TRAVEL_ENTRIES

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
```diff
@@ -179,29 +184,39 @@ public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<?,
...
 
 		return CriteriaBuilderHelper.or(
 			cb,
 			caseFilter,
 			contactFilter,
 			eventParticipantFilter,
 			immunizationFilter,
-			cb.and(cb.isNull(caseJoin), cb.isNull(contactJoin), cb.isNull(eventParticipantJoin), cb.isNull(immunizationJoin)));
+			travelEntryFilter,
+			cb.and(
```
MateStrysewske, line 214: the `cb.and` part is no longer needed (because it's actually a bug)

sormas-backend/src/main/java/de/symeda/sormas/backend/travelentry/TravelEntry.java
```diff
@@ -0,0 +1,498 @@
...
+	@Convert(converter = DiseaseVariantConverter.class)
+	public DiseaseVariant getDiseaseVariant() {
+		return diseaseVariant;
+	}
+
+	public void setDiseaseVariant(DiseaseVariant diseaseVariant) {
+		this.diseaseVariant = diseaseVariant;
+	}
+
+	@ManyToOne(cascade = {})
```
MateStrysewske, line 182: `cascade = {}` should be unnecessary in all cases because it's the default

alexcaruntu-vita (approved):

sormas-backend/src/main/java/de/symeda/sormas/backend/travelentry/TravelEntryJurisdictionPredicateValidator.java
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	@Override
+	protected Predicate whenNationalLevel() {
+		return cb.conjunction();
+	}
+
+	@Override
+	protected Predicate whenRegionalLevel() {
+		return cb.equal(joins.getResponsibleRegion().get(Region.ID), currentUser.getRegion().getId());
```
MateStrysewske, line 74: The `whenRegionalLevel` and `whenDistrictLevel` methods need to also take into account the point of entry region and district; have a look at CaseJurisdictionPredicateValidator to see how it should work

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -7532,4 +7532,79 @@ ALTER TABLE immunization_history OWNER TO sormas_user;
...
+    quarantinehomepossible varchar(255),
+    quarantinehomepossiblecomment varchar(512),
+    quarantinehomesupplyensured varchar(255),
+    quarantinehomesupplyensuredcomment varchar(512),
+    quarantineextended boolean DEFAULT false,
+    quarantinereduced boolean DEFAULT false,
+    quarantineofficialordersent boolean DEFAULT false,
+    quarantineofficialordersentdate timestamp,
+
+    wasInQuarantineBeforeIsolation varchar(255),
```
MateStrysewske, line 7581: See above, isolation fields are not necessary

MateStrysewske (changes requested):
- [x] PersonService.getAllAfter needs to be extended to include travel entries as well (to make sure that the persons related to the travel entries the user has access to are synchronised to the mobile app)
- [ ] Please use `text` for all database fields where you have used `varchar(512)`, and change the annotation of the entities to `@Column(columnDefinition = "text")` accordingly
- [x] Looking at the way we're working with JSON in e.g. campaigns compared to the new customizable enums, I think the latter should be the preferred way; i.e., add a model class like CustomizableEnumTranslation that is called `DEAConentEntry` and that has two Strings (caption + value); use the annotations that are used in `CustomizableEnumValue` to automatically parse the data to a list of the model type; adjust the entity and DTO accordingly

sormas-backend/src/main/java/de/symeda/sormas/backend/travelentry/TravelEntry.java
```diff
@@ -0,0 +1,498 @@
...
+	private boolean quarantineOfficialOrderSent;
+	private Date quarantineOfficialOrderSentDate;
+
+	private YesNoUnknown wasInQuarantineBeforeIsolation;
+	private QuarantineReason quarantineReasonBeforeIsolation;
+	private String quarantineReasonBeforeIsolationDetails;
+	private EndOfIsolationReason endOfIsolationReason;
+	private String endOfIsolationReasonDetails;
+
+	@ManyToOne(cascade = {}, fetch = FetchType.EAGER)
```
MateStrysewske, line 105: What's the reason to specifically choose fetch type eager here?

PR #6104, opened 2021-07-09
#4754 - Added the missing methods in the immunization facades

alexcaruntu-vita:
Fixes #


sormas-backend/src/main/java/de/symeda/sormas/backend/common/AdoServiceWithUserFilter.java
```diff
@@ -84,6 +85,64 @@ public AdoServiceWithUserFilter(Class<ADO> elementClass) {
 		return em.createQuery(cq).getResultList();
 	}
 
+	public List<String> getArchivedUuidsSince(Date since) {
```
MateStrysewske, line 88: Move the new code in this class to `ImmunizationService`

MateStrysewske (changes requested):

PR #6114, opened 2021-07-13
#6091 make countries grid filterable by i18n name

### LINKED ISSUE #6091: Allow country grid to be filtered and sorted by translated name
### Bug Description

Infrastructure data edit form for country does not get updated on directory after editing.
### Steps to Reproduce

1. Edit the name of a country and save. One name is maintain on the directory and new name on the form
2. When edited the UNO code, it got updated on directory (this is OK)
3. Filter by name of country does not suggest all countries having the specified character, eg "cot" in the the filter does not suggest "Cote_dIvoire" as a country, but "c" does.

### Expected Behavior

- Directory should be reloaded after changing the name
- Check whether there's an easy fix for the filtering problem - likely not
### Screenshots

image

### System Details

* Device:
* SORMAS version: 1.61
* Android version/Browser:
* Server URL: https://release-international.sormas.netzlink.com/sormas-ui/
* User Role: admin

DavidBaldsiefen:
Fixes #6091


MateStrysewske (changes requested):
I think this needs a proper refactoring. Right now, it feels like we have a lazy-loaded grid that is somehow forced into eager loading, but still uses things like backend sorting. I would prefer to use a dedicated eager-loaded grid that is also sorted directly in the UI, and the `getIndexList` method to be rewritten accordingly.

DavidBaldsiefen:
> I think this needs a proper refactoring. Right now, it feels like we have a lazy-loaded grid that is somehow forced into eager loading, but still uses things like backend sorting. I would prefer to use a dedicated eager-loaded grid that is also sorted directly in the UI, and the `getIndexList` method to be rewritten accordingly.

I am not completely sure what you mean by writing a custom grid. I think the current grid model works quite fine, just the eager-mode parameter being used as synonym for the selectionmode is not ideal
I agree on writing a custom `getEagerIndexList`-Method though that does only the filtering, but no sorting.

MateStrysewske:
@DavidBaldsiefen I don't even think we should do the filtering in the backend, shouldn't it be possible to do that directly on the grid data? I also don't really know what you mean by "custom grid" because I didn't use that term :D What I meant is that the country grid currently, with your changes, is still a lazy-loaded grid that is forced into eager loading, but since we only have eager loading and will never have lazy loading, we should get rid of all code that is crafted towards lazy loading.

DavidBaldsiefen:
@MateStrysewske 
I have changed the `CountriesGrid` so that it is completely eager-loaded now (the list only being re-fetched from the DB when entries are added, changed or removed. Then it is also sorted by default, as IMO that does no harm and writing an extra method just for that purpose would make the code unnecessarily bloated).
However, I did not touch the `CountryFacadeEjb`, as the lazy-methods are still used in other parts of the code (specifically for REST in `CountryResource.java`)

MateStrysewske (approved):

PR #6130, opened 2021-07-15
#4756 - change data type for detail fields to text

### LINKED ISSUE #4756: Immunizations I: Entities, DTOs and backend logic [5]
### Situation Description
#4081 introduces the full immunization module 
### Feature Description
- [ ]  Add an immunization entity in db with the following attributes:
- disease: required
- report date:  required
- external id free text
- responsible region (optional), district (optional), community
- immunization facility:  we use the default facility class as for case creation without home as option.  We exclude home here because the goal is to capture the organization providing the immunization.
- - expected number of dose  (positive Integer, required adn same validation as for cases 1:10 )
- immunizationStartDate (required)
- immunizationEndDate (required if number of dose >1 and set to end date if dose == 1)
- number of days between doses  (required if number of dose >1): would be used for  immunization followup, but not in this ticket
- immunizationStatus(required, ongoing as default state): ongoing | canceled| completed |
- has person been infected with this disease before? yes |no | unknown if yes
- date of last infection
- general comment
- 
- [ ] Add a form to create new immunization as in figure
   - Trigger duplicate person detection method when creating a new immunization.
   - Duplicate immunization detection
        - After calling duplicate person detection, check for duplicate immunization using the following:
        ` disease, date of report, immunizationStartDate, region, district, facility`.
    This is the same method for duplicate case detection by just replacing symptom onset date with  immunisationStartDate
- It should be possible to create multiple immunization per person but if the disease is the same, then both should not overlap based on end dates.
- Link and display immunization on the person form, just as we have for cases and contacts
### Additional Information
@kay The positions if the text in figures are just for demo, we can place them right 
image

image

* Add a new `ImmunizationStatus` enum with the following values: `PENDING`, `ACQUIRED`, `NOT_ACQUIRED`, `EXPIRED`
* Add a new `MeansOfImmunization` enum with the following values: `VACCINATION`, `RECOVERY`, `VACCINATION_RECOVERY`, `OTHER`
* Add a new `ImmunizationManagementStatus` enum with the following values: `SCHEDULED`, `ONGOING`, `COMPLETED`, `CANCELED`
* Add a new `Immunization` entity extending `CoreAdo` and the corresponding Dto class with the following fields (in the mobile app, make sure to add the entity to all code in DatabaseHelper that creates, clears or drops the tables):
  * `disease` (not null)
  * `person_id` (not null)
  * `reportDate` (not null)
  * `reportingUser_id`
  * `archived` - boolean
  * `deleted` - boolean
  * `meansOfImmunization` - varchar(255)/MeansOfImmunization (not null)
  * `meansOfImmunizationDetails` - text
  * `immunizationStatus` (not null) - varchar(255)/ImmunizationStatus
  * `managementStatus` - varchar(255)/ImmunizationManagementStatus
  * `externalId` - text
  * `responsibleRegion_id`
  * `responsibleDistrict_id`
  * `responsibleCommunity_id`
  * `country_id`
  * `startDate`
  * `endDate`
  * `validFrom` - date
  * `validUntil` - date
  * `numberOfDoses` - integer
  * `previousInfection` - varchar(255)/YesNoUnknown
  * `lastInfectionDate`
  * `additionalDetails` - text
  * `positiveTestResultDate`
  * `recoveryDate`
  * `relatedCase_id`
* Add basic Facade and Service methods that are necessary for synchronization
* Add a ReST resource with the basic synchronization methods (`getAllImmunizations`, `getByUuids`, `postImmunizations`, `getAllUuids`, `getDeletedUuidsSince`)
* Extend the synchronization logic in the mobile app; always synchronize immunizations AFTER persons (because they have a person reference); don't forget `pullAndRemoveDeletedUuidsSince`
* Pseudonymization is based on the responsible jurisdiction of the immunization
* Users can access all immunizations in their jurisdiction
* Extend the person user filter and getAllAfter method as well as the person resource methods that allow users to retrieve persons they have access to; users also need to have access to persons that are related to immunizations they can access
* Add a new feature configuration: `IMMUNIZATION_MANAGEMENT` , DISABLED by default
* Add new user rights: `IMMUNIZATION_MANAGEMENT_ACCESS` (all user roles that have access to CASE_VIEW), `IMMUNIZATION_VIEW` (all user roles that have access to CASE_VIEW), `IMMUNIZATION_CREATE` (all user roles that have access to CASE_CREATE), `IMMUNIZATION_EDIT` (all user roles that have access to CASE_EDIT), `IMMUNIZATION_DELETE` (all user roles that have access to CASE_DELETE), `IMMUNIZATION_ARCHIVE` (all user roles that have access to CASE_ARCHIVE)

BarnaBartha:
Fixes #4756


MateStrysewske (changes requested):
Data types need to be changed in the mobile app as well; please just change the create table script there because I don't believe retrospective data type changes are possible there

MateStrysewske (approved):

PR #6150, opened 2021-07-16
Introduce feature configurations for cases-contacts-samples, events and lab messages

### LINKED ISSUE #6021: [Sormas2Sormas] Feature configuration for events, cases & contacts, Lab-Messages [1]
### Situation Description

As some health departments might consider to use Sormas 2 Sormas for cases & contacts (including samples) only, we have to add a feature toggle, so that they can disable the functionality to share DEMIS Lab Messages and/or Events. This should be applied to bulk mode as well.

### Feature Description

Add a Feature configuration to enable/disable
1. cases & contacts with samples (default true)
2. Events (default false)
3. DEMIS Lab-Messages (default false)

vidi42:
Closes #6021

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasFacade.java
@@ -44,4 +44,10 @@ public interface SormasToSormasFacade {
 	boolean isFeatureEnabledForUser();
 
 	boolean isFeatureConfigured();
+
+    boolean isSharingCasesContactsAndSamplesEnabledForUser();
+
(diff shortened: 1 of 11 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasFacade.java
```diff
@@ -44,4 +44,10 @@
 	boolean isFeatureEnabledForUser();
 
 	boolean isFeatureConfigured();
+
+    boolean isSharingCasesContactsAndSamplesEnabledForUser();
```
leventegal-she, line 48: you forgot to format the code

leventegal-she (approved):
Pleas format the code in SormasToSormasFacade.java and it can be merged

PR #6155, opened 2021-07-19
#5891 - use json include non default for primitive values and collect…

### LINKED ISSUE #5891: Dto deserialization handles primitives and collections as required attributes
### Bug Description
If property values are null in the serialization of Dto objects, deserialization may fail in the `sormas-rest` module. This is the case for, e.g., `PersonDto.hasCovidApp` or`PersonDto.covidCodeDelivered`. Properties in question seem to have either primitive types (here: `boolean`) or collection types.
This requires these properties to be present in the serialization even if having the default value (e.g., `false`) and not having been intentionally set.
The overhead of obsolete data to be transmitted is relevant in a low bandwidth setting (see #3522), hence the focus on `PersonDto` and `CaseDto`.

### Expected Behavior
Missing property values - unless annotated as `@Required` - should be handled gracefully, e.g., if `PersonDto.hasCovidApp` is missing in the serialization, the property should be instantiated with its default value, but deserialization should not fail in the `sormas-rest` module.

### System Details

* Device:
* SORMAS version: 1.61.0-SNAPSHOT
* Android version/Browser:
* Server URL:
* User Role:

BarnaBartha:
…ions in DTOs used for saving entries

Fixes #5891

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
@@ -23,6 +23,7 @@ import static de.symeda.sormas.api.CountryHelper.COUNTRY_CODE_SWITZERLAND;
 import java.util.Date;
 import java.util.List;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import de.symeda.sormas.api.Disease;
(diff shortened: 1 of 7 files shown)


MartinWahnschaffe:
I'm closing this. If we want to exclude default values from being sent via json, we should do it in a more generic way. We don't have an issue for this at this point though.

PR #6158, opened 2021-07-19
#6156 Update facilityCount whenever necessary

### LINKED ISSUE #6156: Facility count not updated after import
### Bug Description

after manually importing a list of facilities, the imported facilities are shown in the table, but "No. of facilities" is not updated

### Steps to Reproduce

1. Navigate to facilities view
2. Perform import
3. observe "No. of facilities" on the right side above the table

### Expected Behavior

Count is updated and does not read "0" any more

### System Details

* Device:
* SORMAS version: 1.61.1
* Android version/Browser:
* Server URL:
* User Role:

DavidBaldsiefen:
Fixes #6156

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/configuration/infrastructure/FacilitiesView.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/configuration/infrastructure/FacilitiesView.java
@@ -92,6 +92,7 @@ public class FacilitiesView extends AbstractConfigurationView {
 	protected Button createButton;
 	protected PopupButton exportPopupButton;
 	private MenuBar bulkOperationsDropdown;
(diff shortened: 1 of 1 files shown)


syntakker (changes requested):
The grid is still updated without updating the row count in several places (`grid.reload()` in lines 207, 241, 268).
I  would assume this to introduce the behaviour described in the issue in other situations.

DavidBaldsiefen:
> The grid is still updated without updating the row count in several places (`grid.reload()` in lines 207, 241, 268).
> I would assume this to introduce the behaviour described in the issue in other situations.

Yes I noticed that, but I was just not sure what is intended behaviour here. I changed it so that the count should always represent the amount of currently represented rows

syntakker (approved):

PR #6161, opened 2021-07-19
Fix infrastructure import error handling and regions duplicate check

### LINKED ISSUE #6089: Duplicate detection during region import does not work
### Bug Description

#1621 added duplicate detection for infrastructure data.
Importing infrastructure results in duplicate data in the system.
### Steps to Reproduce

1. export region data on a server
2. re-import the exported regions but do not choose `override existing entries with imported data`.  This create duplicate entries even if the name and epid code are the same. 
3. Import a region with the same name as a region in the system but with another country. You get the error message `The database already contains a region with this name.`

### Expected Behavior

- If `override existing entries with imported data` == t and duplicate are identified, merge the two records using the one being imported as the reference
- If `override existing entries with imported data` == f  and duplicate are identified,  do not import and add the entry to the import error report.
- Two regions with the same name but different countries, continents, etc should not be considered as duplicates. Duplicate should consider the parent/ higher administrative unit.

### Screenshots

image

image

### System Details

* Device:
* SORMAS version: 1.61
* Android version/Browser:
* Server URL:  https://release-international.sormas.netzlink.com/sormas-ui/
* User Role: admin

vidi42:
Closes #6089

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/region/RegionFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/region/RegionFacadeEjb.java
@@ -1,6 +1,6 @@
-/*******************************************************************************
+/*
  * SORMAS® - Surveillance Outbreak Response Management & Analysis System
- * Copyright © 2016-2018 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ * Copyright © 2016-2021 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -14,7 +14,8 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with this program. If not, see <https://www.gnu.org/licenses/>.
- *******************************************************************************/
+ */
+
 package de.symeda.sormas.backend.region;
 
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/region/RegionFacadeEjb.java
```diff
@@ -347,7 +348,7 @@ public void saveRegion(RegionDto dto, boolean allowMerge) throws ValidationRunti
 		Region region = regionService.getByUuid(dto.getUuid());
 
 		if (region == null) {
-			List<Region> duplicates = regionService.getByName(dto.getName(), true);
+			List<Region> duplicates = regionService.getByName(dto.getName(), countryService.getByReferenceDto(dto.getCountry()), true);
```
MateStrysewske, line 351: This will result in a NullPointerException for regions that are not assigned to a country

MateStrysewske (changes requested):
The changes made so far will likely fix this problem, but they will potentially lead to subsequent problems during the import of other entities. The reason is that with this change, we will allow multiple regions with the same name, but in different countries, to be added to the system. However, other imports (e.g. case import), when processing the region column, will still throw an error if they find two regions with the same name in the system irrespective of the country. So in order to not produce additional problems, the other importers need to be adjusted to call the getByName method based on the country that was specified if available (e.g. for the event import when importing the location information), or based on the server country OR no country in other cases (because we consider both to be in the server country - see RegionFacadeEjb.getAllActiveByServerCountry where this logic is used). This is more complicated for the imports where there might be a country specified (again, e.g. event import); the logic here would need to be similar to e.g. ImportHelper.getRegionBasedOnDistrict (= getCountryBasedOnRegion).

vidi42:
Opened a new ticket which should also cover this https://github.com/hzi-braunschweig/SORMAS-Project/issues/6190

PR #6185, opened 2021-07-22
Add specific risks property into events #5940

### LINKED ISSUE #5940: Addition of a "Specific risk" field in the events file to define a media or political risk for example
### Situation Description

To date, an event can be defined according to its epidemiological risk (risk level => high, moderate, low). In addition, there is another risk which, to date, cannot be mentioned by the teams: it is a media or political risk for example.

The two risks are completely independent: a cluster with a high epidemiological risk may or may not have any media risk. It is therefore not possible to complete the drop-down list with additional modalities because users need to define both epidemic risk and other types of risk including, for example, media risk.

### Feature Description

We propose :
• rename the “risk level” variable to “epidemiological risk level”
• create an variable "other specific risk" which would take the form of a drop-down list with configurable modalities for each server.

This data should appear on the events directory (but not actions directory) and be used to sort/filter (+ basic and detailed export)

### Possible Alternatives

Create an “other specific risk” variable in drop-down list format with four modalities defined in the code:
- media risk,
- political risk,
- other risk,
- unknown

### Additional Information

aponcon-atolcd:
Fixes #5940


sormas-api/src/main/java/de/symeda/sormas/api/customizableenum/CustomizableEnumFacade.java
```diff
@@ -46,6 +46,19 @@
...
+	 * on the user's language, or the server language as a fallback. If this specific value has not been requested yet,
+	 * the cache is extended with it on demand.
+	 *
+	 * @param type
+	 *            The type for which to retrieve the enum value
+	 * @param <T>
+	 *            The specific extension of {@link CustomizableEnum} for type safety
+	 * @return The enum instances related to the type in parameter, internationalized caption, and optional properties
+	 */
+	<T extends CustomizableEnum> List<T> getEnumValues(CustomizableEnumType type);
```
MateStrysewske, line 60: I don't think this method is necessary; you should simply be able to call getEnumValues(CustomizableEnumType, Disease) with disease == null instead. This new logic actually makes the diseases column unusable, but there might be cases in the future in which people want to associate a specific risk with a specific disease for whichever reason.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/customizableenum/CustomizableEnumValueDao.java
```diff
@@ -72,6 +72,32 @@ public String getTableName() {
 		return CustomizableEnumValue.TABLE_NAME;
 	}
 
+	@SuppressWarnings("unchecked")
+	public <T extends CustomizableEnum> List<T> getEnumValues(CustomizableEnumType type) {
```
MateStrysewske, line 76: See above, use the method with the disease parameter instead and pass null as the disease

sormas-app/app/src/main/java/de/symeda/sormas/app/event/edit/EventEditFragment.java
```diff
@@ -201,6 +203,13 @@ protected void prepareFragmentData() {
 		parenteralTransmissionModeList = DataUtils.getEnumItems(ParenteralTransmissionMode.class, true);
 		medicallyAssociatedTransmissionModeList = DataUtils.getEnumItems(MedicallyAssociatedTransmissionMode.class, true);
 		infectionPathCertaintyList = DataUtils.getEnumItems(InfectionPathCertainty.class, true);
+
+		List<SpecificRisk> specificRisks =
+				DatabaseHelper.getCustomizableEnumValueDao().getEnumValues(CustomizableEnumType.SPECIFIC_RISK);
+		specificRiskList = DataUtils.toItems(specificRisks);
+		if (record.getSpecificRisk() != null && !specificRisks.contains(record.getSpecificRisk())) {
+			specificRiskList.add(DataUtils.toItem(record.getSpecificRisk()));
+		}
```
MateStrysewske, line 212: Make sure to not show the new field when there are no entries in the database; see CaseEditFragment.updateDiseaseVariantsField. You don't have to move the code to a dedicated method, just make sure that the visibility logic is executed (probably best to do that in onLayoutBinding)

MateStrysewske (changes requested):
I would suggest to call the new enum `SPECIFIC_EVENT_RISK` to make the context clearer. Can you do the necessary changes please?

MateStrysewske (changes requested):
* Don't show the "Other specific risk" column if there are no entries in the database
* The new field is still displayed in the mobile app, both in the read and edit forms, when there are no entries in the database

aponcon-atolcd:
Hi @MateStrysewske Thanks for the review.

Indeed it was not ok at all, I'm sorry about that. Should be better now. :)

MateStrysewske:
@aponcon-atolcd Thanks! I'll merge your PR now because I'm aware that you won't be able to work on it next week. There's still something that needs to be done, i.e. changing the disease of an event also needs to trigger a reload of the potential event risks, and to be consistent with the disease variants, we should not show the new field in read fragments in the mobile app at all when it's not filled in, but I will implement that myself.

MateStrysewske (approved):

PR #6194, opened 2021-07-22
#6144 Import module should not consider white spaces for infrastructure data names

### LINKED ISSUE #6144: Import module should not consider white spaces for infrastructure data names
### Bug Description

Region name with white spaces after the name on import data are not mapped with the same region in the system.

### Steps to Reproduce

1. Create a region in the sytem
2. Import a hospital in the region, on import template, use region name with space at the end, eg "Berlin "
3. System does not map the region with the "Berlin" in the system

### Expected Behavior

- Delete white spaces before of after the name before matching
- I think we should not delete white spaces within the name, eg "New Berlin" should not be mapped with "NewBerlin"
- Character case should not also matter berlin == BerLin
### System Details

* Device:
* SORMAS version: 1.60.3
* Android version/Browser:
* Server URL: https://test-inhp-ci.sormas.netzlink.com/sormas-ui/#!about
* User Role:

leventegal-she:
Fixes #6144

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/caseimport/CaseImportFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/caseimport/CaseImportFacadeEjb.java
@@ -351,7 +351,7 @@ public class CaseImportFacadeEjb implements CaseImportFacade {
 		List<String> invalidColumns = new ArrayList<>();
 
 		for (int i = 0; i < values.length; i++) {
-			String value = values[i] != null ? values[i].trim() : values[i];
+			String value = StringUtils.trimToNull(values[i]);
 			if (ignoreEmptyEntries && (value == null || value.isEmpty())) {
 				continue;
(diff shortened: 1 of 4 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/caseimport/CaseImportFacadeEjb.java
```diff
@@ -351,7 +351,7 @@
 		List<String> invalidColumns = new ArrayList<>();
 
 		for (int i = 0; i < values.length; i++) {
-			String value = values[i];
+			String value = values[i] != null ? values[i].trim() : values[i];
```
vidi42, line 354: Just a suggestion. `StringUtils.trimToNull(values[i])` achieves the same and maybe it's easier to read.

Alexandru-Pirvu (approved):

vidi42 (approved):

PR #6234, opened 2021-07-28
#5946 - immunization automation

### LINKED ISSUE #5946: Immunizations III: Immunization automation [1]
### Situation Description

### Feature Description

* Add a new feature configuration: `IMMUNIZATION_STATUS_AUTOMATION`, DISABLED by default and dependent on `IMMUNIZATION_MANAGEMENT`
* Add a method called `updateImmunizationStatuses` that in one query (without loading the JPA entities) updates the immunization status of all immunizations with `immunizationStatus == ACQUIRED && currentDate >= validUntil` to `EXPIRED`
* Write a test for the backend method
* Create a new cronjob that executes the new method at some time in the night

BarnaBartha:
Fixes #5946

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/feature/FeatureType.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/feature/FeatureType.java
@@ -195,7 +195,8 @@ public enum FeatureType {
 		true,
 		new FeatureType[] {
 			TASK_MANAGEMENT }),
-	IMMUNIZATION_MANAGEMENT(true, false, null);
+	IMMUNIZATION_MANAGEMENT(true, false, null),
+	IMMUNIZATION_STATUS_AUTOMATION(true, false, new FeatureType[]{ IMMUNIZATION_MANAGEMENT});
 
 	/**
 	 * Server feature means that the feature only needs to be configured once per server since they define the way the system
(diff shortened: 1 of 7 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/feature/FeatureType.java
```diff
@@ -195,7 +195,8 @@
 		true,
 		new FeatureType[] {
 			TASK_MANAGEMENT }),
-	IMMUNIZATION_MANAGEMENT(true, false, null);
+	IMMUNIZATION_MANAGEMENT(true, false, null),
```
MateStrysewske, line 198: Please move `IMMUNIZATION_MANAGEMENT` to the "Feature Modules" section and `IMMUNIZATION_STATUS_AUTOMATION` to "Feature Extension"

sormas-backend/src/main/java/de/symeda/sormas/backend/immunization/ImmunizationFacadeEjb.java
```diff
@@ -274,6 +282,17 @@ private Immunization fillOrBuildEntity(@NotNull ImmunizationDto source, Immuniza
 		return target;
 	}
 
+	@Override
+	public void updateImmunizationStatuses() {
```
MateStrysewske, line 286: Ah, I guess my development specifications were misleading in this regard - I didn't mean to write a native query, just wanted to point out that the entities should not be loaded, changed, and then persisted again, but instead there should be a direct update. I would still go with CriteriaQuery here unless there's a good reason not to; see e.g. CaseService.updateArchived as an example.

Additionally, please make sure to update the change date of the affected immunizations as well.

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #6242, opened 2021-07-30
Feature 5827 immunization android

### LINKED ISSUE #5827: Immunization II: Overview and add/edit form for mobile app [5]
### Situation Description

### Feature Description

#### Overview
* Add a new "Immunizations" main menu entry after Samples that contains a list of immunizations
* Only accessible for users with the `IMMUNIZATION_MANAGEMENT_ACCESS` user right
* Show the following information:
  * Immunization UUID
  * First name
  * Last name
  * Means of immunization
  * Management status
  * Immunization status
  * Immunization period (defined by start and end date)
* Add filters for disease, immunization status, immunization start and immunization end
* If only one disease is active in the system, pre-select that disease
* The list always needs to be filtered by a single disease and shows the data from the most recent immunization for that disease; if no disease has been selected, don't show the list, but instead a hint with the following text: "Please select a disease to display the immunization overview."
* Add a "+" action that allows to create new immunizations

#### Creation form
Refer to #4764 for a description of the creation form; it should basically look and work identical in the mobile app

#### Edit form
Refer to #4764 for a description of the edit form; it should basically look and work identical in the mobile app; mobile app does not need the delete and archive options

BarnaBartha:
Fixes #5827


sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/AbstractAdoDao.java
```diff
@@ -1315,4 +1316,24 @@ public long countOf() {
 	public ConnectionSource getConnectionSource() {
 		return dao.getConnectionSource();
 	}
+
+	// dao utilities
+
+	protected void addDateFromCriteria(List<Where<Immunization, Long>> whereStatements, Where<Immunization, Long> where, Date reportDateFrom, String reportDate) throws SQLException {
```
MateStrysewske, line 1322: Use more generic parameter names, i.e. `dateFrom` and `date`

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/AbstractAdoDao.java
```diff
@@ -1315,4 +1316,24 @@ public long countOf() {
...
+		if (reportDateFrom != null) {
+			whereStatements.add(where.ge(reportDate, DateHelper.getStartOfDay(reportDateFrom)));
+		}
+	}
+
+	protected void addDateToCriteria(List<Where<Immunization, Long>> whereStatements, Where<Immunization, Long> where, Date endDateTo, String endDate) throws SQLException {
```
MateStrysewske, line 1328: Use more generic parameter names, i.e. `dateTo` and `date`

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/immunization/ImmunizationDao.java
```diff
@@ -1,54 +1,182 @@
...
 
-        // Cancel if not in local database
-        if (immunization == null) {
-            return;
-        }
+			Where<Immunization, Long> where = queryBuilder.where().eq(AbstractDomainObject.SNAPSHOT, false);
+			ImmunizationCriteria immunizationCriteria = criteria.getImmunizationCriteria();
+			where.and().eq(Immunization.DISEASE, immunizationCriteria.getDisease());
+			where.and().eq(Immunization.RESPONSIBLE_REGION + "_id", criteria.getImmunizationCriteria().getResponsibleRegion());
+			if (criteria.getStartDate() != null && criteria.getEndDate() != null) {
```
MateStrysewske, line 167: Date filters are incorrect; the system should search for overlapping immunizations. Here's a more detailed definition - each of these scenarios is valid:

* any combination in which either immunization.startDate or immunization.endDate is between criteria.startDate and criteria.endDate
* immunization.startDate == null && ((criteria.startDate == null && criteria.endDate != null) || criteria.startDate < immunization.endDate)
* immunization.endDate == null && ((criteria.endDate == null && criteria.startDate != null) || criteria.endDate > immunization.startDate)

Since you're already in communication with @alexcaruntu-vita, could you please make sure that this is also reflected in the web backend? Thanks!

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationEditFragment.java
```diff
@@ -0,0 +1,220 @@
...
+
+		contentBinding.immunizationMeansOfImmunization.addValueChangedListener(e -> {
+			if (e.getValue() == MeansOfImmunization.OTHER || e.getValue() == MeansOfImmunization.RECOVERY) {
+				contentBinding.immunizationImmunizationManagementStatus.setValue(ImmunizationManagementStatus.COMPLETED);
+				contentBinding.immunizationImmunizationManagementStatus.setEnabled(false);
+				contentBinding.overwriteImmunizationManagementStatusCheckBox.setVisibility(View.VISIBLE);
+			}
+			if (e.getValue() == MeansOfImmunization.VACCINATION || e.getValue() == MeansOfImmunization.VACCINATION_RECOVERY) {
+				contentBinding.immunizationVaccinationHeading.setVisibility(View.VISIBLE);
+				contentBinding.immunizationRecoveryHeading.setVisibility(View.VISIBLE);
```
MateStrysewske, line 165: Recovery section and its fields are only supposed to be visible when RECOVERY or VACCINATION_RECOVERY is selected, not when VACCINATION is selected; I suggest to add surrounding linear layouts for the vaccination and the recovery fields and headings so that you only have to change its visibility and not the visibility of all fields that depend on this logic

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationEditFragment.java
```diff
@@ -0,0 +1,220 @@
...
+		// Initialize ControlDateFields
+		contentBinding.immunizationReportDate.initializeDateField(getFragmentManager());
+		contentBinding.immunizationRecoveryDate.initializeDateField(getFragmentManager());
+		contentBinding.immunizationStartDate.initializeDateField(getFragmentManager());
+		contentBinding.immunizationEndDate.initializeDateField(getFragmentManager());
+		contentBinding.immunizationLastInfectionDate.initializeDateField(getFragmentManager());
+
+		contentBinding.immunizationMeansOfImmunization.addValueChangedListener(e -> {
+			if (e.getValue() == MeansOfImmunization.OTHER || e.getValue() == MeansOfImmunization.RECOVERY) {
+				contentBinding.immunizationImmunizationManagementStatus.setValue(ImmunizationManagementStatus.COMPLETED);
```
MateStrysewske, line 159: Vaccination fields and headings have to be hidden when the means of immunization is not VACCINATION or VACCINATION_RECOVERY (see next comment)

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationEditFragment.java
```diff
@@ -0,0 +1,220 @@
...
+			}
+			if (e.getValue() == ImmunizationManagementStatus.COMPLETED) {
+				contentBinding.immunizationImmunizationStatus.setValue(ImmunizationStatus.ACQUIRED);
+			}
+			if (e.getValue() == ImmunizationManagementStatus.CANCELED) {
+				contentBinding.immunizationImmunizationStatus.setValue(ImmunizationStatus.NOT_ACQUIRED);
+			}
+		});
+
+		contentBinding.immunizationImmunizationManagementStatus.setValue(ImmunizationManagementStatus.SCHEDULED);
```
MateStrysewske, line 203: This should only be done in the NewFragment (because the edit fragment shouldn't need a default)

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationEditFragment.java
```diff
@@ -0,0 +1,220 @@
...
+		contentBinding.immunizationImmunizationManagementStatus.setEnabled(false);
+		contentBinding.overwriteImmunizationManagementStatusCheckBox.setVisibility(View.VISIBLE);
+		contentBinding.immunizationNumberOfDoses.setVisibility(View.GONE);
```
MateStrysewske, line 206: This should be covered by the listeners above

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationNewFragment.java
```diff
@@ -0,0 +1,227 @@
...
+		contentBinding.immunizationPositiveTestResultDate.initializeDateField(getFragmentManager());
+
+		ValidationHelper
+				.initIntegerValidator(contentBinding.immunizationNumberOfDoses, I18nProperties.getValidationError(Validations.vaccineDosesFormat), 1, 10);
+	}
+
+	@Override
+	protected void onAfterLayoutBinding(FragmentImmunizationNewLayoutBinding contentBinding) {
+
+		contentBinding.immunizationMeansOfImmunization.addValueChangedListener(e -> {
```
MateStrysewske, line 175: See my comments in the EditFragment

sormas-app/app/src/main/res/layout/fragment_immunization_new_layout.xml
```diff
@@ -0,0 +1,260 @@
+<?xml version="1.0" encoding="utf-8"?><!--
```
MateStrysewske, line 1: Please have another look at the development specifications and remove the fields that are only supposed to be part of the edit form

sormas-app/app/src/main/res/values/strings.xml
```diff
@@ -236,6 +236,9 @@
     <string name="caption_year">Year</string>
```
MateStrysewske, line 236: Most, if not all, of these strings should exist in the API because they are needed in both the web and the mobile app; please use those (or create them if they're missing) instead. Only create strings here that are really ONLY used in the mobile app.

sormas-app/app/src/main/res/layout/row_immunization_pick_or_create_layout.xml
```diff
@@ -0,0 +1,112 @@
...
+        <import type="de.symeda.sormas.api.i18n.I18nProperties" />
+        <import type="de.symeda.sormas.api.immunization.ImmunizationDto" />
+        <import type="de.symeda.sormas.api.person.PersonDto" />
+
+        <variable name="data" type="de.symeda.sormas.app.backend.immunization.Immunization" />
+        <variable name="index" type="java.lang.Integer" />
+        <variable name="callback" type="de.symeda.sormas.app.core.IEntryItemOnClickListener" />
+    </data>
+
+    <LinearLayout
```
MateStrysewske, line 32: Please have another look at the issue description; person information are not useful in this popup because there has been person duplicate detection before already; what's more useful here is the information that is also part of the creatio window, i.e. means of immunization, external ID, management status, responsible jurisdiction, start date, end date

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationEditActivity.java
```diff
@@ -0,0 +1,164 @@
...
+		default:
+			throw new IndexOutOfBoundsException(DataHelper.toStringNullable(section));
+		}
+
+		return fragment;
+	}
+
+	@Override
+	public void saveData() {
+		if (saveTask != null) {
```
MateStrysewske, line 98: If I'm not mistaken and not missing anything, duplicate detection is missing here; please have another look at the issue description; there's supposed to be a popup if the user changes the start and end date and then tries to save when there's an existing overlapping immunization.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/immunization/Immunization.java
```diff
@@ -293,4 +305,30 @@ public void setCountry(Country country) {
     public String getI18nPrefix() {
         return I18N_PREFIX;
     }
+
+
+    public void update(Immunization immunization) {
```
MateStrysewske, line 310: What is this method used for?

MateStrysewske (changes requested):

sormas-app/app/src/main/res/layout/fragment_immunization_new_layout.xml
BarnaBartha: can you please confirm that:
the number of doses be in the new layout without the heading
recovery related fields should should not be in the new layout

MateStrysewske: @BarnaBartha Yes, exactly!

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/immunization/Immunization.java
BarnaBartha: to update the immunization in case the user chooses to override the existing data

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/AbstractAdoDao.java
```diff
@@ -1315,4 +1316,24 @@ public long countOf() {
 	public ConnectionSource getConnectionSource() {
 		return dao.getConnectionSource();
 	}
+
+	// dao utilities
+
+	protected void addDateFromCriteria(List<Where<Immunization, Long>> whereStatements, Where<Immunization, Long> where, Date dateFrom, String date) throws SQLException {
```
MateStrysewske, line 1322: Change Immunization references to AbstractDomainObject (since we're in a generic Dao) for all three methods

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/immunization/Immunization.java
```diff
@@ -46,33 +45,50 @@
...
+    private Date reportDate;
+    @DatabaseField(foreign = true, foreignAutoRefresh = true)
+    private User reportingUser;
+    @DatabaseField
+    private boolean archived;
+    @Enumerated(EnumType.STRING)
+    private ImmunizationStatus immunizationStatus;
+    @Enumerated(EnumType.STRING)
+    private MeansOfImmunization meansOfImmunization;
+    @Column(length = COLUMN_LENGTH_BIG)
```
MateStrysewske, line 86: Use `@Column(columnDefinition = "text")` (since the column is a text column)

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/immunization/Immunization.java
MateStrysewske: Please check whether you can use AbstractAdoDao.mergeOrCreate instead because this method would have to be maintained, which will almost certainly lead to problems later; if that's not possible at all, please add the new fields here that are missing (e.g. facility fields)

sormas-app/app/src/main/res/layout/dialog_immunization_overlaps_layout.xml
```diff
@@ -0,0 +1,151 @@
...
+
+                <LinearLayout
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:orientation="horizontal">
+
+                    <de.symeda.sormas.app.component.controls.ControlDateField
+                        android:id="@+id/immunization_startDate"
+                        style="@style/ControlFirstOfTwoColumnsStyle"
+                        app:value="@={data.startDate}" />
```
MateStrysewske, line 86: Remove the = signs from the `app:value` parameters and turn the fields into ControlTextReadFields because they should not be editable

sormas-app/app/src/main/res/layout/fragment_immunization_new_layout.xml
```diff
@@ -0,0 +1,257 @@
...
+                app:dependencyParentValue="@{MeansOfImmunization.OTHER}"
+                app:required="true"
+                app:value="@={data.meansOfImmunizationDetails}" />
+
+            <de.symeda.sormas.app.component.controls.ControlSpinnerField
+                android:id="@+id/immunization_immunizationManagementStatus"
+                style="@style/ControlSingleColumnStyle"
+                app:value="@={data.immunizationManagementStatus}" />
+
+            <de.symeda.sormas.app.component.controls.ControlCheckBoxField
```
MateStrysewske, line 86: This check box should probably be above the management status field

sormas-app/app/src/main/res/layout/row_immunization_pick_or_create_layout.xml
```diff
@@ -0,0 +1,183 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:gravity="top"
+            android:textSize="@dimen/row_entry_secondary_text_size"
+            android:textColor="@color/row_entry_secondary_text"
+            app:dateValue="@{data.reportDate}"
+            app:valueCaption="@{I18nProperties.getPrefixCaption(ImmunizationDto.I18N_PREFIX, ImmunizationDto.REPORT_DATE)}"
+            app:defaultValue='@{@string/value_not_available}' />
+
+        <TextView
```
MateStrysewske, line 57: Disease can be removed because it will always be the same as for the created immunization anyway

sormas-app/app/src/main/res/layout/row_immunization_pick_or_create_layout.xml
```diff
@@ -0,0 +1,183 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:gravity="top"
+            android:textSize="@dimen/row_entry_secondary_text_size"
+            android:textColor="@color/row_entry_secondary_text"
+            app:value="@{data.externalId}"
+            app:valueCaption="@{I18nProperties.getPrefixCaption(ImmunizationDto.I18N_PREFIX, ImmunizationDto.EXTERNAL_ID)}"
+            app:defaultValue='@{@string/value_not_available}' />
+
+<!--        <LinearLayout-->
```
MateStrysewske, line 112: What's the reason for these fields being commented?

MateStrysewske (changes requested):
* validFrom and validUntil fields are missing in the mobile app
* I'd like you to move any logic that can also be achieved by using attributes in the layouts from the fragments to the layouts; e.g. whether or not validDoses is shown can be handled by adding the meansOfImmunization field as a dependency field and then use dependencyParentValue and dependencyParentValue2 for Vaccination and Vaccination/Recovery, respectively

BarnaBartha:
@MateStrysewske where can i find how exactly are the validFrom and validUntil are used or where should they be? What is their role given that we have already a startDate and endDate ?

sormas-app/app/src/main/res/layout/row_immunization_pick_or_create_layout.xml
BarnaBartha: was not sure if we need them or not, seems like showing too much information there for a row entry
the mechanism that fetches the similar immunization considers the responsibleRegion for now, i am not sure if this is wanted - to be onest i don't get how immunizations are jurisdiction dependent
basically i forgot the commented code there - so the question would be do we want to show in this row entry also the responsible jurisdiction or not?

MateStrysewske:
@BarnaBartha You can have a look at the issue that implements the create/edit forms in the UI, Alexandru already added them there. The `startDate` and `endDate` are used to document when the immunization starts and ends (e.g. the date of first vaccination and the date of last vaccination), `validFrom` and `validUntil` track from when to when the immunization is valid (e.g. if you're vaccinated with Comirnaty for Covid-19, validFrom would be 14 days after the the endDate, and validUntil might be in six months from now).

sormas-app/app/src/main/res/layout/row_immunization_pick_or_create_layout.xml
MateStrysewske: Ah okay, yeah, makes sense. The jurisdiction fields are there because we somehow have to define which users are allowed to access the immunizations (since persons don't have a mandatory jurisdiction themselves). However, the jurisdictions should not be used in the mechanism that retrieves potential duplicates (only disease and startDate/endDate period are relevant here), so that should be changed. I still think that we don't have to display the information here though because mobile app users won't have access to immunizations from other regions and districts anyway, so it's pretty redundant. You can remove the commented code.

MateStrysewske (changes requested):
See my comment from above:

- [ ] I'd like you to move any logic that can also be achieved by using attributes in the layouts from the fragments to the layouts; e.g. whether or not validDoses is shown can be handled by adding the meansOfImmunization field as a dependency field and then use dependencyParentValue and dependencyParentValue2 for Vaccination and Vaccination/Recovery, respectively
- [ ] Also, I'd like you to re-style the row items in the list like this:

image

- [ ] Management status and external ID do not have a proper caption
- [x] Add a page heading called "Immunization details" to the immunization read and edit fragment (between the menu bar and the form):

image

- [x] Disease details field is missing
- [x]  Move the checkbox to overwrite the management status to above the management status
- [ ]  Facility details field is not displayed when selecting place of stay "Home" in the creation fragment (it works properly in the edit fragment)
- [ ]  Make person sex required in the creation fragment
- [ ] The string in the "Pick or create immunization" dialog is not formatted correctly
- [ ] The "Pick or create immunization" dialog
- [ ] Means of immunization needs to be required
- [ ]  Checkbox to overwrite the management status is missing in the creation fragment
- [ ] The logic to retrieve similar immunizations does not work correctly; it retrieves immunizations that don't have a date interception with the created immunization

BarnaBartha:
@MateStrysewske on update of existing immunization that needs to be overwritten, we override the diseaseDetails right?

MateStrysewske:
@BarnaBartha Yes, we basically overwrite everything :)

BarnaBartha:
@MateStrysewske moving logic to the layout i cannot hide the heading, without adding a new control field or making some extension to the ControlTextReadField, or do you know of an example on how we do this in the mobile app. that is hide a section from within the fragment layout ?

MateStrysewske:
@BarnaBartha True, I forgot about that. Any logic that involves the heading can stay in the fragment then. I believe there's still some logic that can be moved to the layout though, right?

MateStrysewske (changes requested):
- [x] Immunization management status has no proper caption (the problem is in captions.properties here, it should be `Immunization.immunizationManagementStatus` instead of `Immunization.managementStatus`
- [x] Immunization management status field is not disabled when entering the edit form with the Overwrite checkbox not checked (it's working properly in the creation form)
- [x] Opening the immunization creation form and then navigating back directs the user to the settings screen; should nagivate them to the immunization list instead
- [ ] Concerning the place of stay field - I just realised that this is not needed at all; immunizations always have to be assigned to a facility, so the place of stay field needs to be removed and the facility field needs to be visible all the time
- [ ] Valid from and valid to fields are not functional
- [ ] Valid until does not have a proper caption in the creation form
- [x] Disease details needs to be required when "Other disease" is selected (in both forms)

PR #6247, opened 2021-07-30
#6244 - Use translated data dictionary in import popups

### LINKED ISSUE #6244: Data dictionary downloads are not tranlsated when getting them from import pop-ups
### Bug Description

When downloading the data dictionary not from the info page but e.g. from import pop-ups, the data dctionary is not translated depending on the user language as implemented in https://github.com/hzi-braunschweig/SORMAS-Project/issues/4607.
### Steps to Reproduce

1. Set the user language to another language than english
2. Go to e.g. the case directory
3. Select detailed import
4. Download the data dictionary in the subsequent pop-up

### Expected Behavior

Independent of where the data dictionary is downloaded, it will be translated according to the user language setting.
### Screenshots

datadictTranslation

### System Details

* Device:
* SORMAS version: 1.61.2

MateStrysewske:
Fixes #6244

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/importer/AbstractImportLayout.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/importer/AbstractImportLayout.java
@@ -1,7 +1,15 @@
 package de.symeda.sormas.ui.importer;
 
+import static de.symeda.sormas.ui.utils.DownloadUtil.createFileNameWithCurrentDate;
+
+import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
 import java.util.function.Function;
+
+import org.apache.commons.io.IOUtils;
+import org.slf4j.LoggerFactory;
 
 import com.vaadin.icons.VaadinIcons;
 import com.vaadin.server.ClassResource;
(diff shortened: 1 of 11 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/importer/AbstractImportLayout.java
```diff
@@ -66,8 +76,19 @@ protected void addDownloadResourcesComponent(int step, ClassResource importGuide
...
+				IOUtils.copy(Files.newInputStream(new File(documentPath).toPath()), out);
+			} catch (IOException e) {
+				LoggerFactory.getLogger(AboutView.class).error("Failed to generate data dictionary", e);
+
+				// fall back to pre-generated document
+				InputStream preGeneratedDocumentStream = new ClassResource("/doc/SORMAS_Data_Dictionary.xlsx").getStream().getStream();
+				IOUtils.copy(preGeneratedDocumentStream, out);
+			}
+		}, (e) -> {
+		}), createFileNameWithCurrentDate(ExportEntityName.DATA_DICTIONARY, ".xlsx"))).extend(dataDictionaryButton);
```
DavidBaldsiefen, line 91: This is duplicate to parts of `AboutView.createDocumentsSection()`; It would be better to move it to a seperate method, e.g. `DownloadUtil.attachDataDictionaryDownloader(Button button)`

DavidBaldsiefen (changes requested):
Looks good to me, however we should try to reduce duplication to avoid similar problems in the future

DavidBaldsiefen (approved):

PR #6258, opened 2021-08-02
#5184 : do not log entry to sync log when use has no longer access

### LINKED ISSUE #5184: Synchronisation after jurisdiction change
### Bug Description
Changing the case jurisdiction in mobile app fails synchronisation.

### Steps to Reproduce
1. Login if with SurvOff to mobile app 
2. Edit a case this user has access too and change the district.
3. Synchronise with server
4. Synchronisation of the case is successful but there is an error and sync log says :
image

### Expected Behavior
There shouldn't be any error (discuss if user should not be able to update or sync process needs to be updated)
Case list on mobile app should no longer show that case (as web UI doesn't show it any more) - these should be consistent!

Note: this issue is most probably in place for other entities as well.

BarnaBartha:
Fixes #5184


MateStrysewske:
@BarnaBartha I just realised that it's probably a bad idea to remove these messages from the sync log because at least they give the user a hint that changes have been dropped; I haven't really though this through properly, I guess. My feeling is that we actually don't have to do anything here. I'll get in touch with you.

PR #6264, opened 2021-08-03
#5496 set default value for disease fields

### LINKED ISSUE #5496: Default value for disease if only one is used by the server [2]
### Feature Description

Setting the deafult value for all fields, where the desease is entered to Covid19, if the server is used for this desease only.

### Problem Description

At a lot of points the user has to enter the disease, even if the server is only used for C19. This is produces a lot of clicks every day for every user.

### Proposed Change

Set the value for every disease-field to Covid19 if the system has only this desease enabled.

I know that this could be handled also in the issue #3507, but there is a) no progress and b) has every user to set this in every field by himself. More user-friendly is this variant.

DavidBaldsiefen:
Closes #5496


MateStrysewske (changes requested):
If I'm not mistaken, what happens right now with these changes is that EXISTING data is manipulated once the user enters a page (e.g. an event page) by setting the default disease if no disease has been selected before. The default selection should only be done for creation forms.

@DavidBaldsiefen I can take care of the necessary changes.

DavidBaldsiefen:
> If I'm not mistaken, what happens right now with these changes is that EXISTING data is manipulated once the user enters a page (e.g. an event page) by setting the default disease if no disease has been selected before. The default selection should only be done for creation forms.
> 
> @DavidBaldsiefen I can take care of the necessary changes.

That *should* not happen because I (think I) always checked if the disease field is already set, and only changed it if it was `null` before. But yea better check twice!

MateStrysewske:
@DavidBaldsiefen Yes, but at least for events (where the disease is not mandatory), that would actually manipulate data

DavidBaldsiefen:
> @DavidBaldsiefen Yes, but at least for events (where the disease is not mandatory), that would actually manipulate data

Oh no I completely missed that!

syntakker (approved):

PR #6268, opened 2021-08-03
#6121 Add sql script to check history tables for differences

### LINKED ISSUE #6121: Provide a script or test that compares tables with their _history equivalent
### Situation Description
Most tables have a `_history` equivalent, which should contain the same columns (e.g. `cases` and `cases_history`). However, when the base tables are changed, the `_history` tables are frequently missed leading to inconsistencies ([example](https://github.com/hzi-braunschweig/SORMAS-Project/issues/6078)).

### Feature Description

Provide a script or test that checks all tables for such inconsistencies, and lists the differing columns afterwards. The script would probably need to be executed manually, as it requires a fully running database.

DavidBaldsiefen:
Closes #6121


JonasCir:
@syntakker With #6978 being merged, this can be closed right?

syntakker:
@JonasCir  yes.

PR #6311, opened 2021-08-09
#5553 Init testReports to empty list

### LINKED ISSUE #5553: [DEMIS2SORMAS] Handle Conversion Errors Gracefully [2]
### Situation Description

Right now, when the DEMIS adapter is not able to convert a DEMIS notification to a labMessage, it returns an error message which is shown to the user. This cancels the lab message fetch and does not lead to any new messages transferred to SORMAS.

This means that whenever a DEMIS notification can not be transferred to a lab message, it constantly blocks fetching new lab messages. With the increased complexity of the new DEMIS profiles, the likelihood of something like that happening increases.

### Feature Description

The interaction between adapter and SORMAS should be more resilient against conversion errors.
The adapter should try to transform as much as possible from a DEMIS notification to a lab message, and return it even when not everything could properly be converted. The labMessageDetails should always be filled, though (if that is not possible, return a conversion error as it is currently done). This is described in [#10](https://github.com/hzi-braunschweig/sormas-demis-adapter/issues/10).

In SORMAS, it must technically be possible to process incomplete lab messages. The most extreme example would be a lab message with just the UUID and labMessageDetails filled (all other attributes == null). It is not clear if processing something like that is currently possible. The task for this issue is to try that out and implement the changes necessary to be able do that.

@AndyBakcsy-she Can there be an automatted UI test for that?

UPDATE: The automatted UI test was considered out of scope and moved to #6213

Alexandru-Pirvu:
Fixes #5553


FredrikSchaeferVitagroup (approved):

FredrikSchaeferVitagroup:
Please resolve the merge conflict.

PR #6320, opened 2021-08-09
#6180 [SORMAS2SORMAS] Shares > Share requests > Rejected items should delete preview data

### LINKED ISSUE #6180: [SORMAS2SORMAS] Shares > Share requests > Rejected items should delete preview data [1]
### Feature Description

Delete preview data from rejected entities in the Shares queue (on the receiving SORMAS instance's side).

### Problem Description

An entity is sent from SORMAS instance A to SORMAS instance B.
SORMAS B gets a preview to view before accepting/rejecting the entity.
If SORMAS B rejects the entity, the preview data is still available, despite the rejection.

image
image

### Proposed Change

Delete an item's preview data on the receiving entity when it is rejected by the receiving entity.

### Possible Alternatives

Changing visibility of the 'view' icon (i.e. hiding the icon upon rejection) may be a temporary workaround, but from a data minimization perspective may not be sufficient in the long run.

leventegal-she:
Fixes #6180


sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/AbstractSormasToSormasInterface.java
```diff
@@ -226,10 +226,10 @@ public void sendRejectShareRequest(String uuid) throws SormasToSormasException {
 		SormasToSormasShareRequestDto shareRequest = shareRequestFacade.getShareRequestByUuid(uuid);
 
 		String organizationId = shareRequest.getOriginInfo().getOrganizationId();
-		sormasToSormasRestClient.post(organizationId, requestRejectEndpoint, uuid, null);
+//		sormasToSormasRestClient.post(organizationId, requestRejectEndpoint, uuid, null);
```
vidi42, line 229: Hmm, I think this should be uncommented otherwise the status won't be updated on the original instance

vidi42 (changes requested):

vidi42 (approved):

PR #6326, opened 2021-08-10
Feature 4991 bulk linking cases contacts to event

### LINKED ISSUE #4991: Cases & Contacts > Bulk mode for linking/adding to Events [3]
### Situation Description

As an outbreak management team, I would like to use events to group related cases/contacts and to optimally accompany and support affected facilities in on-site outbreak management. An important part of this work is to link the people affected by an outbreak (via cases and via contacts) with the corresponding event.

So that not only the person, but also the corresponding case or contact is correctly linked to the event, must

the case mask or contact mask for each individual case or contact is opened,
clicked on the button "+ link event",
the event to be linked can be searched for and selected.
This procedure means a very high expenditure of time for the employees.
Linking the case to the event is important because the MA in outbreak management depends on the information while working on the phone.

If event participants are linked directly to the event with the "+ Add event participant" button, only the person is linked, but not the case or the contact. If you then click on "Create" in the "Case ID" column, a completely new case will be created for the linked person.

image

The case register can be used to search for cases that are assigned to a specific facility.
image

At the top right under "MORE" you can switch on the bulk processing mode, select all cases and on the right above the case table under "Bulk processing" select the option "Edit". In the pop-up window that opens, it is not possible to link all selected cases to an event.

image

Analogous to contacts:
image

### Feature Description
In these pop-up windows, an option to link the selected cases or contacts to an event would be very helpful. Components: "Link event" tick box, search field to find the event, option to select the desired event.

* Add a new element to the bulk edit dropdown for cases: "Link cases to an event"
* If clicked, bring up a popup that is identical to the popup opened when clicking on "link event" inside a single case (only exception: the info text should read "selected cases" instead of "current case" and "these cases belong to" instead of "this case belongs to")
* Don't show the "create event participant" popup that opens after the user has selected/created an event
* If a case is already linked to the event (i.e. it's the resulting case of an event participant in the event), skip it; if the case person is part of the event as an event participant WITHOUT a resulting case, update the resulting case of the event participant
* If cases with multiple diseases are selected, don't open the popup and instead display a warning notification: "You can't link cases with different diseases to the same event. Please only select cases of the same disease."
* Hide the new option when the `EVENT_SURVEILLANCE` feature is turned off

syntakker:
Fixes #4991


ChristopherRiedel (changes requested):
Everything that has been implemented looks good so far. But I would have a suggestion for improvement: If cases or contacts were connected to the event, we show in a notification the sentence '%s cases/contacts were already linked to the event. All others have been linked'. I think we should not show this sentence with a 0 in front. It looks kind of strange. We should show an extra sentence ('All cases/contacts were linked') instead.

PR #6334, opened 2021-08-11
#3706 Consider person changed date when computing order for cases and…

### LINKED ISSUE #3706: Contacts > Set "Last update date" in DB of Contact if changes have been made on the Contact Person [1]
### Feature Description
As a contact tracing reporter, I would like to keep track of Contacts and when changes are made to their information (Contact and Contact Person)(i.e. when this is updated) so that external systems may be implemented to automatically log and keep track of these changes (for communication purposes for example telephone call scheduling, SMS, automated emails etc.).

### Problem Description
When saving a mutation on an existing contact person, only the contact person is updated, not the contact. The behavior is different in the cases. If a "case person" is changed, the case is also updated and is at the top of the case directory list.

This is requested for Contacts too - please see attachment for detailed example: [Fehlerdokumentation_Kontaktperson.docx](https://github.com/hzi-braunschweig/SORMAS-Project/files/5649789/Fehlerdokumentation_Kontaktperson.docx)

This is requested so that the automation of the contact process may be done. With the current state, it is not possible  to determine subsequent changes on the contact person.

### Proposed Change
Please handle the behavior of the last update date in the same way as in the cases. In this way, we can also determine mutations on a contact that were only made on the person of the contact using the update date.

### Additional Information
Original request: https://github.com/hzi-braunschweig/SORMAS-Switzerland/issues/157

* In the contact index list, use the change date of the associated person if it's newer than the contact change date

Alexandru-Pirvu:
… contacts

Fixes #3706

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseIndexDetailedDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseIndexDetailedDto.java
@@ -86,7 +86,7 @@ public class CaseIndexDetailedDto extends CaseIndexDto {
 								YesNoUnknown reInfection, String city, String street, String houseNumber, String additionalInformation, String postalCode, String phone,
 								String reportingUserUuid, String reportingUserFirstName, String reportingUserLastName, Date symptomOnsetDate,
 								String responsibleRegion, String responsibleCommunity,
-								int visitCount, long eventCount, Date latestSampleDateTime, long sampleCount) {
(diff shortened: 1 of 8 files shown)


MateStrysewske (changes requested):
Trying to access the case detailed index list results in the following error:

```
org.hibernate.hql.internal.ast.QuerySyntaxException: Unable to locate appropriate constructor on class [de.symeda.sormas.api.caze.CaseIndexDetailedDto]
```

MateStrysewske (approved):

PR #6342, opened 2021-08-12
Feature 5594 introduce a reference definition for cases

### LINKED ISSUE #5594: Introduce a reference definition for cases [2]
### Situation Description

German health departments use a reference definition in order to further classify cases. This reference definition is used for later analysis by higher health authorities. 
### Feature Description

Introduce a **'Reference definition'** field for COVID-19 cases that is uneditable and computed on save (DE only). This field will have the values
- fulfilled
- not fulfilled

Computation follows closely what is specified for confirmed case classifications in #4770. What defines whether the reference definition is fulfilled or not is the pathogen test types. Therefore, `fulfilled` is automatically applied if the classification criteria for the respective case classification CONFIRMED, CONFIRMED_NO_SYMPTOMS or CONFIRMED_UNKNOWN_SYMPTOMS is met; however, the test types `ANTIGEN_DETECTION` and `RAPID_TEST` are NOT to be taken into account here. 

Implementation suggestion: In CaseFacadeEjb.onCaseChanged, after calculating the case classification and only on German instances and for CORONAVIRUS cases, check whether it's one of the three CONFIRMED classification AND the case has at least one positive pathogen test result for CORONAVIRUS of the test types PCR_RT_PCR, ISOLATION or SEQUENCING. If so, set the reference definition to fulfilled, otherwise to not fulfilled.

referenceDefinition

Further, add a checkbox filter on the case directory that filters cases according to their reference definition (DE only): 
- "Only show cases with fulfilled reference definition"
    - if unchecked, show all cases as usual
    - if checked, only show cases with fulfilled reference definition

Add an additional number to the surveillance dashboard (DE only) in the disease carousel layout for CORONAVIRUS only, underneath "Cases resulting from contacts" (see Kay's screenshot below); call it "Reference definition fulfilled" and count the number of cases where reference definition == fulfilled.

### Additional Information

@JaquM

sergiupacurariu:
Fixes #5594


sormas-api/src/main/resources/captions.properties
```diff
@@ -358,6 +359,7 @@ caseSearchSpecificCase=Search specific case
 caseSearchCase=Search case
 caseSelect= Select case
 caseCreateNew=Create new case
+caseReferenceDefinition=Reference definition
```
MateStrysewske, line 362: Since this is a new actual field, please add it as a proper caption (=> CaseData.caseReferenceDefinition)

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1981,6 +1991,32 @@ public void onCaseChanged(CaseDataDto existingCase, Case newCase, boolean syncSh
 		}
 	}
 
+	public boolean evaluateFulfilledCondition(CaseDataDto newCase, CaseClassification caseClassification) {
+
+		if (newCase.getCaseClassification() != CaseClassification.NO_CASE) {
```
MateStrysewske, line 1996: This if condition is not necessary because the code in the if block will return false for a case classified as NO_CASE anyway, right?

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
```diff
@@ -1628,6 +1632,12 @@ public void setValue(CaseDataDto newFieldValue) throws ReadOnlyException, Conver
 			differentPlaceOfStayJurisdiction.setValue(Boolean.TRUE);
 		}
 
+//		if (referenceDefinition != null) {
```
MateStrysewske, line 1635: Remove commented code

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
sergiupacurariu: The code in the if block might return Confirmed. The condition is that if NO_CASE is set the Reference Definition will be always NOT_FULFILLED

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseDataForm.java
sergiupacurariu: removed

sormas-api/src/main/resources/captions.properties
sergiupacurariu: updated

MateStrysewske (approved):

PR #6356, opened 2021-08-12
#6241 [S2S] Revoking accepted case results in inconsistent status of cases

### LINKED ISSUE #6241: [S2S] Revoking accepted case results in inconsistent status of cases
### Bug Description
When sharing a case/event without ownership and the Receiver accepts it, if the Sender does not refresh the UI they can still revoke the share which results in the case still being present on the Receiver end and shown as revoked on the Sender end.

### Steps to Reproduce
1. Login on SORMAS A and share a case with SORMAS B (from the case details and without ownership)
2. Don't close, navigate to other page or refresh the page from where the case was shared
3. In another window/browser login on SORMAS B and go to Shares
4. On SORMAS B accept the case
5. On SORMAS A now click the Revoke

### Actual behavior
* On SORMAS A the case shared status will be shown as revoked
* On SORMAS B in the Shares view, the request will be shown as revoked
* On SORMAS B in the case directory the case will still be present

### Expected Behavior
When clicking revoke, SORMAS B should check if the request was not already accepted or rejected and return an error.

### System Details
* SORMAS version: 1.62
* Android version/Browser:
* Server URL: local using https://github.com/hzi-braunschweig/S2S-Testbed
* User Role: NatUser

leventegal-she:
Fixes #6241

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/i18n/Strings.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/i18n/Strings.java
@@ -281,6 +281,7 @@ public interface Strings {
 	String errorSormasToSormasOwnershipAlreadyHandedOver = "errorSormasToSormasOwnershipAlreadyHandedOver";
 	String errorSormasToSormasRequestToken = "errorSormasToSormasRequestToken";
 	String errorSormasToSormasResult = "errorSormasToSormasResult";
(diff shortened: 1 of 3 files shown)


vidi42 (approved):

JonasCir (approved):
lgtm

Issue #4763, opened 2021-03-15
Immunizations II: Vaccination entity [3]


bernardsilenou:
### Situation Description
Issue #4081 introduces the general immunization entity. An immunization is associated to a person, and has one or more vaccination entities. If we assume that immunization is like contacts, then vaccinations are like followup visits of the contact.
Would be good to read the associated issues before working on this.
### Feature Description
- [x]  Add a vaccination entity in the db and link it to Immunization. The relationship is 1 immunization -> n vaccination
- Activate this for covid only for now and for all instances. But this is a module to be later activated for all vaccine preventable diseases.
- [x]  Attributes of vaccination
- Use the `additional medical information section`, exactly as for cases. We can use the same ‘class/ object’ 
- Pregnancy: yes no unknown (remember the validation that exist)
- Postpartum
- Trimester 
- `date of report` (validation: should not be in future of current date) 
- `date of dose expiration` ( can be current date or in future but not after expected end of immunization)
- vaccine name
- manufacturer
- vaccine type
- vaccine dose
- inn (international nonpropriotory name)
- batch number
- uni code
- atc code
- source of vaccination

### Development Specifications
* Add a new `Vaccination` entity (since there is already a Vaccination enum that will be removed in the refactoring issue, call the class VaccinationEntity for now, table should already be called Vaccination though) extending AbstractDomainObject and the corresponding Dto class with the following fields (in the mobile app, make sure to add the entity to all code in DatabaseHelper that creates, clears or drops the tables); most of them can be copied over from the VaccinationInfo class
  * immunization_id (not null)
  * healthConditions_id (not null)
  * reportDate (not null)
  * reportingUser_id
  * vaccinationDate
  * vaccineName - varchar(255)/Vaccine
  *  otherVaccineName - text
  * vaccineNameDetails - text
  * vaccineManufacturer - varchar(255)/VaccineManufacturer
  * vaccineManufacturerDetails - text
  * otherVaccineManufacturer - text
  * vaccineType - text
  * vaccineDose - text
  * vaccineInn - text
  * vaccineBatchNumber - text
  * vaccineUniiCode - text
  * vaccineAtcCode - text
  * vaccinationInfoSource - varchar(255)/VaccinationInfoSource
  * pregnant - varchar(255)/YesNoUnknown
  * trimester - varchar(255)/Trimester
* Mobile app entity needs the annotation `@EmbeddedAdo` to synchronize them automatically alongside immunizations
[...]

PR #6365, opened 2021-08-13
#4763 - Make reporting user not null in vaccination tables

alexcaruntu-vita:
Fixes #


alexcaruntu-vita:
Talked to @MateStrysewske and concluded that this is not the solution that we need, reporting user can be `null`.

Issue #6363, opened 2021-08-13
Create immunization creation request in automation framework in order to prepare data for person populating script 


MartinWahnschaffe:
This task is related to:
#5986

### Problem Description
For #5986 we need to be able to create immunizations in order to populate the test system with it

### Proposed Solution

Create needed pojo and service. Relevant fields are:

* disease
* person
* reportDate
* reportingUser
* responsibleRegion
* responsibleDistrict
* immunizationStatus
* meansOfImmunization
* immunizationManagementStatus
* startDate
* endDate
* externalId

rdutu-vg:
Updated description to include externalId field as well in the request, if we don't add it, the response is 204

rdutu-vg:
Test is done, updated the current script with the immunization part.
Have changed the design for the test to run 5 times.

![executionFromScript](https://user-images.githubusercontent.com/83643843/129753550-2b470a5d-fd7f-417d-9bfd-0a1c8b1e9cef.PNG)

rdutu-vg:
Performed checks on resources used by test execution.

1. Running the script (1-10 loops) 
![resources during run](https://user-images.githubusercontent.com/83643843/130073943-6422e037-2d0c-46e0-a25e-7e1139c3bb5f.PNG)

2. Running the script (1-5 loops) in parallel (3 executions)
![issues](https://user-images.githubusercontent.com/83643843/130074054-508d59b2-4db9-45f6-89ac-00bdb0f2ce99.PNG)

Resources wise, the test is a big consumer, and based on the provided analysis we need to find the best approach to run it in order to get as many runs as possible to populate the database quicker.

PR #6418, opened 2021-08-16
#6363-Create-Immunization-Request

rdutu-vg:
Updated Endpoints interface, created Immunization POJO, updated TestData users to include user's uuid, created immunization helper with POST immmunzation request, corrected mistakes from PersonHelper class, updated ApiState class, created Immunization steps to glue the code there under BDD layer, updated existent test with new immunization steps


sormas-e2e-tests/src/main/java/org/sormas/e2etests/enums/DiseaseApiValues.java
```diff
@@ -0,0 +1,48 @@
...
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package org.sormas.e2etests.enums;
+
+import java.util.Random;
+import lombok.Getter;
+
+@Getter
+public enum DiseaseApiValues {
```
BarnaBartha, line 25: i am not sure i understand why there are 2 disease enums
i think would be better to have one Disease enum and maybe two fields (apiValue & uiValue)

BarnaBartha (changes requested):

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/api/ImmunizationSteps.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class ImmunizationSteps implements En {
+
+  @Inject
+  public ImmunizationSteps(ImmunizationHelper immunizationHelper, ApiState apiState, Faker faker) {
+
+    When(
+        "API: I create {int} new immunizations for last created person",
+        (Integer numberOfImmunizations) -> {
+          String lastCreatedUserUUID = apiState.getCreatedPerson().getUuid();
```
MateStrysewske, line 44: Should be called lastCreatedPersonUuid

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/api/ImmunizationSteps.java
```diff
@@ -0,0 +1,89 @@
...
+          Person person = Person.builder().uuid(lastCreatedUserUUID).build();
+          int immunizationAttempts = 1;
+          while (immunizationAttempts <= numberOfImmunizations) {
```
MateStrysewske, line 47: Any reason to not use a for loop here?

MateStrysewske (changes requested):
Please have another look at the development setup guide, specifically the import formatting rules (asterisk imports shouldn't be used)

sormas-e2e-tests/src/main/java/org/sormas/e2etests/enums/DiseasesValues.java
```diff
@@ -0,0 +1,79 @@
...
+  CSM_API_VALUE("CSM"),
+  YELLOW_FEVER_API_VALUE("YELLOW_FEVER"),
+  RABIES_API_VALUE("RABIES"),
+  UNSPECIFIED_VHF_API_VALUE("UNSPECIFIED_VHF"),
+  OTHER_API_VALUE("OTHER");
+
+  private final String disease;
+  private static Random random = new Random();
+
+  DiseasesValues(String aDisease) {
```
BarnaBartha, line 60: i was thinking about something like this:

`CORONAVIRUS("CORONAVIRUS", "COVID-19")`
and for the fields/ constructor
```
private final String disease;
private final String diseaseCaption;
Disease(String disease, String diseaseCaption) {
...
}
```
to avoid maintaining two sets of enums for the desieases

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/api/ImmunizationSteps.java
rdutu-vg: Foreach cannot be used and I considered while more simpler for our needs :D if we the for loop is more performant, we can change it, sure

sormas-e2e-tests/src/main/java/org/sormas/e2etests/enums/DiseaseApiValues.java
rdutu-vg: agree, refactored this

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/api/ImmunizationSteps.java
rdutu-vg: Yep, will refactor this

MateStrysewske (approved):

BarnaBartha (approved):

PR #6419, opened 2021-08-17
Feature 5843 vaccinations lists and forms

### LINKED ISSUE #5843: Immunizations III: Vaccination lists and forms [3]
### Situation Description

### Feature Description

#### List (web app)
* Extend the immunization form in the web app with a list of vaccinations (use the same kind of lists as exposures or previous hospitalizations)
* Show the following columns:
  * Edit icon column (no caption)
  * Vaccination ID (representing the UUID)
  * Vaccination date
  * Vaccine name (+ details)
  * Manufacturer name (+ details)
  * Vaccine type
  * Dose
* Allow the user to open the edit form of the vaccination in a popup by clicking on the edit icon
* Add a "Create vaccination" button that opens the vaccination creation form (only accessible for users with the IMMUNIZATION_EDIT user right)

#### List (mobile app)
* Same as above

#### Create/edit form
* Create/edit form is identical
* Create form contains the following fields:
  * Report date (required)
  * Reporting user (read only)
  * Vaccination date
  * Vaccine name
  * Vaccine name details (when "Other" vaccine name is selected)
  * Manufacturer
  * Manufacturer details (when "Other" manufacturer is selected)
  * Vaccine type
  * Vaccination info source
  * Dose
  * INN
  * Batch number
  * UNII code
  * ATC code
  * Pregnant
  * Trimester
  * Health conditions form
* When opening this form for an existing vaccination in the web app, it needs a delete option for users with the `IMMUNIZATION_DELETE` user right
* When creating, editing or deleting a vaccination, update the management and immunization status of the associated immunization like this:
  * If `currentDate >= startDate AND vaccination count >= 1 AND vaccination count < numberOfDoses`, update management status to `ONGOING` and immunization status to `PENDING`
  * If `vaccination count >= numberOfDoses`, update management status to `COMPLETED` and immunization status to `ACQUIRED`

syntakker:
Fixes #5843


MateStrysewske (changes requested):
**Web App**
- [ ] Vaccine name details is always visible; also move it to the same line as vaccine name
- [ ] Vaccine manufacturer details is always visible; also move it to the same line as vaccine manufacturer
- [x] Add the automation between vaccine name and vaccine manufacturer (like it's done in the case view right now for example)
- [x] Remove the `vaccineNameDetails` and `vaccineManufacturerDetails` fields from the entity, database, form and table (they are fully redundant to `otherVaccineName` and `otherVaccineManufacturer`; no idea why I specified them in the vaccination entity issue) - adjust the captions file accordingly
- [x] Don't show the time in the vaccination date column because we don't allow users to enter it in the first place
- [x] Only show the short UUID in the ID column

- [x] When the user changes the means of immunization to `RECOVERY` or `OTHER` and there is at least one vaccination in the vaccination list, show a popup with the following text: "Changing the means of immunization will delete all vaccinations that have been added to this immunization. Are you sure you want to proceed with the change?" and two buttons ("Confirm" and "Cancel")
  - [x] Clicking on Confirm sets the field to the new value and removes all vaccinations from the vaccination list (i.e. when the user saves afterwards, the vaccinations get deleted, if they discard the form or don't save otherwise, they remain in the system)
  - [x] Clicking on Cancel resets the means of immunization to the former value

**Android App**
- [x] Remove the `vaccineNameDetails` and `vaccineManufacturerDetails` fields from the entity (they are fully redundant to `otherVaccineName` and `otherVaccineManufacturer`; no idea why I specified them in the vaccination entity issue)
- [ ] Make sure that the chevron right icon is displayed to the right side of the list items, not at the bottom right
- [ ] Display more items in the same line in the vaccination read layout (e.g. pregnant and trimester, unii code and atc code, etc. - only display items in a separate line if the caption or expected value is long, so e.g. vaccine name and manufacturer)
- [ ] Vaccinations are not properly synchronized from the mobile app to the web app; the other way around works

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationEditFragment.java
```diff
@@ -177,33 +192,66 @@ public void onAfterLayoutBinding(final FragmentImmunizationEditLayoutBinding con
...
+			public void onChange(ControlPropertyField e) {
+				MeansOfImmunization meansOfImmunization = (MeansOfImmunization) e.getValue();
+
+				if (currentMeansOfImm != meansOfImmunization) {
+					if (meansOfImmunization == MeansOfImmunization.OTHER || meansOfImmunization == MeansOfImmunization.RECOVERY) {
+						contentBinding.immunizationImmunizationManagementStatus.setValue(ImmunizationManagementStatus.COMPLETED);
+						contentBinding.immunizationImmunizationManagementStatus.setEnabled(false);
+
+						removeVaccinationsConfirmation();
+						if (meansOfImmunization == MeansOfImmunization.VACCINATION || meansOfImmunization == MeansOfImmunization.VACCINATION_RECOVERY) {
```
MateStrysewske, line 210: There's something off with this code. This if condition is always false because it's nested in the if clause in line 205.

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationNewFragment.java
```diff
@@ -132,6 +132,7 @@ protected void prepareFragmentData() {
 	protected void onLayoutBinding(FragmentImmunizationNewLayoutBinding contentBinding) {
 		contentBinding.setData(record);
 
+		contentBinding.setYesNoUnknownClass(YesNoUnknown.class);
```
MateStrysewske, line 135: Duplicate code line

MateStrysewske (changes requested):

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationEditFragment.java
BarnaBartha: good catch, not sure how i managed this one
corrected it

MateStrysewske (changes requested):
* [Web] The popup that asks the user to confirm changing the means of immunization to a non-vaccination value does not reliably appear for me; there seems to be a bug there
* [App] Don't use a date time field for the vaccination date, only a simple date field
* [App] Vaccinations are only synchronized when the immunization has also been changed; the immunization synchronization should also be triggered when ONLY its vaccinations have been changed (=> change date filter doesn't seem to be working correctly)

sormas-app/app/src/main/java/de/symeda/sormas/app/BaseEditActivity.java
```diff
@@ -178,6 +178,13 @@ protected void setStoredRootEntity(ActivityRootEntity entity) {
 		this.storedRootEntity = entity;
 	}
 
+	/**
+	 *
```
MateStrysewske, line 182: Please add a comment to the comment (describing what the method is supposed to be used for) :)

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/vaccination/VaccinationEditActivity.java
```diff
@@ -0,0 +1,181 @@
...
+
+            @Override
+            public void doInBackground(TaskResultHolder resultHolder) throws DaoException {
+                final VaccinationEntity savedVaccination = DatabaseHelper.getVaccinationDao().saveAndSnapshot(vaccinationEntity);
+                final Immunization immunization = DatabaseHelper.getImmunizationDao().queryUuid(vaccinationEntity.getImmunization().getUuid());
+                final List<VaccinationEntity> vaccinations = immunization.getVaccinations();
+                if (vaccinationEntity.getId() == null) {
+                    vaccinations.add(savedVaccination);
+                } else {
+                    VaccinationEntity vaccinationToReplace = null;
```
MateStrysewske, line 137: You don't have to create a variable for this purpose, simply move the `remove` call to inside the for loop and then add a break.

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/vaccination/VaccinationEditActivity.java
```diff
@@ -0,0 +1,181 @@
...
+        saveTask = new SavingAsyncTask(getRootView(), vaccinationEntity) {
+
+            @Override
+            protected void onPreExecute() {
+                showPreloader();
+            }
+
+            @Override
+            public void doInBackground(TaskResultHolder resultHolder) throws DaoException {
+                final VaccinationEntity savedVaccination = DatabaseHelper.getVaccinationDao().saveAndSnapshot(vaccinationEntity);
```
MateStrysewske, line 131: You shouldn't be required to save the vaccination manually here because it should automatically happen when you save the immunization below.

MateStrysewske (changes requested):
Vaccinations that are created in the mobile app are not synchronized to the backend for me; changes to existing vaccinations are working now.

MateStrysewske (approved):

Issue #5987, opened 2021-06-29
[Angular][Backend] The UUID needs to be accepted as NULL


laurentiustanvitagroup:
### Situation Description

"For the moment we need the uuid in order to cerate entities. This will change and the uuid would be accepted null and the backend will generate it." @marius2301 

### Feature Description

- We need to make the backend accept the UUID as NULL as per https://github.com/hzi-braunschweig/SORMAS-Project/issues/5735

### Possible Alternatives

N/A

### Additional Information

N/A

marius2301:
I add the restriction that the filter criteria to not be NULL for  Case and Event - because otherwise the Jurisdiction criteria is not added. 
For Sample the jurisdiction is applied even if the criteria is null. 
For Treatment, Prescriptions, Person, Task, Country, Region, Community, District, Contact, User is seems that is no need it.

PR #6429, opened 2021-08-17
#5987 no null filer

marius2301:
Fixes #


sormas-rest/src/main/java/de/symeda/sormas/rest/CaseResource.java
```diff
@@ -119,6 +119,14 @@
 		return FacadeProvider.getCaseFacade().getIndexPage(criteriaWithSorting.getCriteria(), offset, size, criteriaWithSorting.getSortProperties());
 	}
 
+	/**
+	 * 
+	 * @param criteriaWithSorting
+	 *            - The criteria object inside criteriaWithSorting the cannot be null! Use instead an empty criteria!
```
MateStrysewske, line 125: Text correction (also in EventResource): "The criteria object inside criteriaWithSorting cannot be null. Use an empty criteria instead."

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #6479, opened 2021-08-23
#6466 set symptomatic to false for unsuccessful visits

### LINKED ISSUE #6466: Visits documented as unavailable are marked as 'no' with symptomatic
### Bug Description

Visits documented as unavailable are marked as 'no' for symptomatic. This should rather be null or at least 
### Steps to Reproduce

1. Enter a new visit as unavailable 
2. Check the symptomatic column

### Expected Behavior

Column value should be conform to the entered data in the visit. Therefore, it should be null
### Screenshots

visitsSymptoms

### System Details

* Device:
* SORMAS version: 1.61.2
* User Role:

DavidBaldsiefen:
Fixes #6466 

I checked the android app and it seems like no changes are required there


MateStrysewske (changes requested):
Please change the logic in CaseFacadeEjb and ContactFacadeEjb that calculates the symptomatic status of the last cooperative visit to return YesNoUnknown.UNKNOWN when visitSymptoms.getSymptomatic() is null. The current code looks like this: `exportDto.setLastCooperativeVisitSymptomatic(visitSymptoms.getSymptomatic() ? YesNoUnknown.YES : YesNoUnknown.NO);`

MateStrysewske (changes requested):
Looks good - sorry for missing that earlier, but we need to perform the same database operation in the Android app as well to avoid data inconsistencies

MateStrysewske (approved):

PR #6541, opened 2021-08-30
6149, 6484 improve constraints for api csv documents

### LINKED ISSUE #6149: Improve constraints for pushing data trough the API, inserting trough CSV and uploading documents [5]
### Situation Description

SORMAS offers multiple ways of inserting data into the system: REST API, CSV Import, Web UI.

As a result of #5919 we found a few issues with some of the data insert mechanisms which could expose SORMAS to some attacks.

#### REST API length constraints
- [ ] Introduce character length validation for DTOs used in the REST API. 
Character length should be based on the length accepted by the DB or the UI.
In situations where no length restriction is imposed at DB or UI level, use a max allowed field-size of 1.000.000 chars.

Affected fields:
* all fields of type `varchar(x)`
* all fields of type `text`
---
#### CSV Import length constraints
- [ ] Introduced validations for data imported trough CSV.
Introduce length validation similarly to those from the DB or UI to the free text fields accepted trough the 
> At the moment the only CSV constraints imposed are related to ENUMs, Dates, Number and References, but not to any free text fields.

Affected fields:
* all fields of type `varchar(x)`
* all fields of type `text`
---
#### GPS Coordinates constraints
- [ ] Introduce validation of ranges for latitude and longitude (latitude between -90 to 90, longitude between -180 and 180)
> Was only able to test the limits manually on the [Open Street Map website](https://www.openstreetmap.org/). I assume the API accepts higher values, but maps them to the max values

Affected fields:
* `latitude` and `longitude` fields of `LocationDto`
---
#### UUID constrains
- [ ] Introduce pattern constraint for UUIDs in the CSV Import similar to the REST APIs
The REST API requests validate UUIDs against this pattern `^[0-9a-zA-Z-]*$`
- [ ] Introduce a length validation for UUIDs to have between 20 and 36 characters

Affected fields:
* `uuid` field of `EntityDto`
---
#### Documents size for uploading to cases/contacts and events
- [ ] Introduce a configurable size limit for uploading documents with a default of 20 MB
> At the moment the built in size limit is 2147483647 bytes (~2GB) per document

Affected classes:
- `DocumentMultiFileUpload`

### Additional Information
See #4959 for similar restrictions applied to DTOs used in the REST API.

For details about the findings, see https://github.com/hzi-braunschweig/SORMAS-Security/issues/2

### LINKED ISSUE #6484: Introduce file limit size for CSV files imported in the system [2]
### Situation Description

As described in hzi-braunschweig/SORMAS-Security#3 the system is exposed to DDOS Attacks with huge CSV files.

### Feature Description

Introduce file limits for CSV imports system wide.
The file limits should be configurable trough sormas.properties (default 20MB).

### Additional Information

`nginx` and `apache` should also be configured with such properties, however this is out of scope for this ticket.

See hzi-braunschweig/SORMAS-Security#3 for more details.

leventegal-she:
Fixes #6149
Fixes #6484


sormas-api/src/main/java/de/symeda/sormas/api/utils/ConstrainValidationHelper.java
```diff
@@ -0,0 +1,50 @@
...
+						int pathSize = path.size();
+						path.set(pathSize - 1, path.get(pathSize - 1) + "[" + propertyNode.getIndex() + "]");
+					}
+
+					path.add(propertyNode.getName());
+				}
+			}
+		}
+
+		return errors;
```
vidi42, line 48: Violation error message not added to `errors` and errors is always returned empty

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
```diff
@@ -461,51 +480,72 @@
...
+	@Size(max = COLUMN_LENGTH_DEFAULT, message = Validations.textTooLong)
 	private String additionalDetails;
 	@HideForCountriesExcept(countries = {
 		COUNTRY_CODE_GERMANY,
 		COUNTRY_CODE_SWITZERLAND })
+	@Size(max = COLUMN_LENGTH_DEFAULT, message = Validations.textTooLong)
 	private String externalID;
+	@Size(max = COLUMN_LENGTH_DEFAULT, message = Validations.textTooLong)
 	private String externalToken;
+	@Size(max = COLUMN_LENGTH_TEXT, message = Validations.textTooLong)
```
vidi42, line 548: Internal Token UI validation is still set at 512 which means that after import the users won't be able to change and save a case until they adjust the internal token as well.
I suggest changing the length validation on the UI as well to have them aligned.

vidi42 (changes requested):
Not done, but found until now

vidi42 (changes requested):
This was not specified in the ticket description, but was recorded in the findings from  https://github.com/hzi-braunschweig/SORMAS-Security/issues/2 that the UI fields for Latitude and longitude should also be updated with validator so the user is shown a message in case of invalid values for the coordinates.

At the moment the ValidationConstraintViolationInterceptor prevents the user from saving invalid coordinates, but the user is shown a generic error without any indication on what is wrong.

vidi42 (changes requested):
UI validations should also be added to

vidi42 (approved):

PR #6550, opened 2021-08-31
Introduce a property to the SORMAS properties to disable manually editing of infrastructure data

### LINKED ISSUE #6471: Introduce a property to the SORMAS properties to disable manually editing of infrastructure data [1]
### Situation Description
Right now, GAs are allowed to manually edit and tamper with their local data. 

### Feature Description
Introduce a server property which disables the manual editing of infrastructure data. While this is up to debate, I think that there should be at least three modes

1. Enabled
2. Archive-only: where old data can be archived 
3. Disabled: no access to what so ever, data cannot be de-archived and no manual edits

The affected data would include:
        - continents
        - subcontinents
        - countries
        - regions
        - districts
        - communities

Note: facilities and points of entries need to stay editable

- [x] When the manual editing of infrastructure data is disabled, also the import should be disabled.
- [x] Default value: Enabled

vidi42:
Closes #6471


JonasCir (approved):
lgtm thank you!

PR #6554, opened 2021-08-31
#6510 - Show reference definition only for Germany

### LINKED ISSUE #6510: Reference definition is visible on the dashboard of international servers
### Bug Description

On international servers, the German specific reference definition is visible.
### Steps to Reproduce

1. Open the surveillance dashboard
2. Check the Disease slides section for 'Case with reference definition fulfilled'

### Expected Behavior

This field should only be shown on German servers
### Screenshots

referenceDefinitionInt

### System Details

* Device:
* SORMAS version: 1.63.0
* Server URL: release-international.sormas.netzlink.com

alexcaruntu-vita:
Fixes #6510

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/surveillance/components/statistics/summary/DiseaseSummaryComponent.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/dashboard/surveillance/components/statistics/summary/DiseaseSummaryComponent.java
@@ -1,5 +1,6 @@
 package de.symeda.sormas.ui.dashboard.surveillance.components.statistics.summary;
 
+import de.symeda.sormas.api.CountryHelper;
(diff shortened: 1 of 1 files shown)


DavidBaldsiefen (changes requested):
Looks good so far, but please add the same check to `refreshDataForCasesWithReferenceDefinitionFulfilled` in `DashboardDataProvider`.java (could save some performance on international servers)

alexcaruntu-vita:
> Looks good so far, but please add the same check to `refreshDataForCasesWithReferenceDefinitionFulfilled` in `DashboardDataProvider.java` (could save some performance on international servers)

The actual performance gain would be if we won't fetch those cases for Germany, but the cases are already fetched there. I preferred to make changes only in the ui layer, without affecting the data processing layer.

DavidBaldsiefen (approved):

PR #6588, opened 2021-09-02
fix display of duplicate person selection #6491

### LINKED ISSUE #6491: Improving the duplicate detection pop-up at the import
### Feature Description

The fields of the imported person are cut-off at the end of the duplicate detection. For the desk officer it is not possible to check all possible fields.

Even though the fields should be in the same order at the imported person and the possible duplicates to make it easier for the health offices.

image

syntakker:
Fixes #6491


MateStrysewske (changes requested):
I think the current width of the popup should make it possible to get rid of the third line and move the two properties to the second line

syntakker:
I would advocate not to drop the third line, long values could still push columns out of the visible area.
In this example eith two columns, passport number column is gone, with value lengths being still in a realistic range:

image

MateStrysewske:
@syntakker Good point!

MateStrysewske (approved):

PR #6594, opened 2021-09-02
#6472 A user viewing a case containing e.g., an archived region has no indication about the fact the region was archived

### LINKED ISSUE #6472: A user viewing a case containing e.g., an archived region has no indication about the fact the region was archived [2]
### Feature Description
All infrastructure drop-down menus in the UI or any part of the application the user sees infra. data, including export.

### Problem Description
If an infrastructure entity is archived, there is zero indication for the user looking e.g., at  a case that this specific entity is archived and only exists for historic reasons. 

### Proposed Change
Once an infra. entity gets archived, append "(inactive)" to the name/caption such that the user looking at a details page gets a clear indication that this data is out-of-date. 
In the mobile app check for every fragment. We should do it also for disease fields.

leventegal-she:
Fixes #6472


vidi42 (changes requested):
On the Web-UI it seems not all the places where infrastructure fields were updated (ex. Case Edit > Responsible Region / District / Community).
Changes similarly to those in the PersonEditForm have to be applied to all the infrastructure field.

On the app it looks ok when editing data, but the `(Inactive)` suffix is not shown when in the read-only mode, only in the edit mode.

leventegal-she:
Reviewed and fixed the creation all infrastructure data selectors in the web UI.

I don't think we need to add (Inactive) suffix in the read-only views, It doesn't really help the user who just wants to see the data. But doing that it would require to add the suffix in all other readonly places like list, texts to be consistent but that is not the scope of this ticket, at least it was not included in the estimation.

@MateStrysewske what do you think?

vidi42 (approved):

PR #6617, opened 2021-09-03
#6592 [S2S] Revoking before accepting a case results in server error

### LINKED ISSUE #6592: [S2S] Revoking before accepting a case results in server error
### Bug Description
When sharing a case/event without ownership and the Receiver accepts it, if the Sender does not refresh the UI they can still revoke the share which results in the case still being present on the Receiver end and shown as revoked on the Sender end.

### Steps to Reproduce

1. Login on SORMAS A and share a case with SORMAS B (from the case details and without ownership)
2. Don't close, navigate to other page or refresh the page from where the case was shared
3. In another window/browser login on SORMAS B and go to Shares
4. On SORMAS A now click the Revoke
5. On SORMAS B accept the case

### Expected Behavior

- On SORMAS A the case shared status will be shown as revoked
- On SORMAS B in the Shares view, the request will be shown as revoked
- On SORMAS B in the case directory the case will still be present

### Screenshots
image

### System Details

* Device: Windows
* SORMAS version: 1.64.0-snapshoot
* Server URL: test-de1 + test-de2
* User Role: NatUser

leventegal-she:
Fixes #6592


JonasCir:
@leventegal-she I'm off for one week, can you please reassign this PR? :)

vidi42 (approved):

Issue #6140, opened 2021-07-15
[Angular][Backend] Event profile : event actions endpoint [2]


laurentiustanvitagroup:
### Situation Description

- In view of the future replacement of the current FE with the Angular component, the Angular team will focus on the required changes.

### Feature Description

- We need to implement an endpoint for the events actions section which should return all the required data as presented in the cards from: https://www.sketch.com/s/fd061796-65a3-4991-886a-c8a172667c91/a/8yApaY3 (including the documents links)

### Possible Alternatives

- N/A

### Additional Information

- N/A

marius2301:
This is partially implemented and it is in review. In the current implementation the (name of) documents are not in the return DTO. I need to talk to Martin to see how we should do it.

PR #6619, opened 2021-09-03
#6140 pagination and filtering of EventActions and Actions

marius2301:
Fixes #


sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
```diff
@@ -191,6 +192,70 @@ public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<?,
...
+
+		if (filter != null) {
+			cq.where(filter);
+		}
+		if (sortProperties != null && sortProperties.size() > 0) {
+			List<Order> order = new ArrayList<>(sortProperties.size());
+			for (SortProperty sortProperty : sortProperties) {
+				Expression<?> expression;
+				switch (sortProperty.propertyName) {
+				case ActionDto.UUID:
```
MateStrysewske, line 219: I think you can simplify all of these (except the title because of the lowercase addition) by writing all the cases underneath each other and then just writing `action.get(sortProperty.propertyName)` because they're direct properties of actions and aren't computed; see e.g. EventFacadeEjb.getIndexList for an example

sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
```diff
@@ -191,6 +192,70 @@ public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<?,
 		return QueryHelper.getResultList(em, cq, first, max);
 	}
 
+	public List<Action> getActionList(ActionCriteria actionCriteria, Integer first, Integer max, List<SortProperty> sortProperties) {
```
MateStrysewske, line 195: Please unify this method with the existing `getActionList` method (i.e., remove the code from the existing method and let it call this method with null as sortProperties) to avoid duplicate code

sormas-api/src/main/java/de/symeda/sormas/api/action/ActionFacade.java
```diff
@@ -47,9 +48,17 @@
...
 
+	Page<EventActionIndexDto> getEventActionIndexPage(EventCriteria criteria, Integer offset, Integer size, List<SortProperty> sortProperties);
+
+	Page<ActionDto> getActionPage(ActionCriteria criteria, Integer offset, Integer size, List<SortProperty> sortProperties);
+
 	List<EventActionExportDto> getEventActionExportList(EventCriteria criteria, Integer first, Integer max);
 
 	long countEventAction(EventCriteria criteria);
+
+	long countAction(ActionCriteria criteria);
```
MateStrysewske, line 63: Rename to `countActions`

MateStrysewske (changes requested):

MateStrysewske (changes requested):
Renaming of the countAction method is missing :)

MateStrysewske (approved):

PR #6620, opened 2021-09-03
first draft

### LINKED ISSUE #6469: Check (outbound) entities for correct infrastructure data [3]
### Feature Description

### Problem Description

### Proposed Change

---
### Tasklist
- [x] #6496 
- [x] #6500
- [x] #6503
- [x] #6528
- [x] #6573
- [x] #6587
- [x] #6758
- [x] #7047
- [x] #7050
- [x] #7058
- [x] #7070
- [x] #7202

JonasCir:
Fixes #6469 

This is just a first draft. I removed the complicated handling logic in between and process the different types directly (see e.g., `processContinent`, other methods are stubs right now). All classes that implement `ReceivedDataProcessor` will go through the received objects and call the corresponding processing functions of `InfrastructureValidator` everytime they encounter an infrastructure reference DTO. The processing functions in `InfrastructureValidator`will in the future check back with central etcd and see if for the given (globally unique UUID) a matching entity is stored. If not, a validation error is added and returned.


JonasCir:
This PR is paused for now, see comment in the ticket.

PR #6652, opened 2021-09-07
fixed #6579

### LINKED ISSUE #6579: Additional tests cannot be deleted
### Bug Description

Trying to delete additional tests does not work properly.
### Steps to Reproduce

1. Create or open an additional test in any sample
2. Try deleting the additional test
3. Confirm the deletion

### Expected Behavior

The additional test should get deleted
### Screenshots

additionalTestDeleteBug

### System Details

* Device:
* SORMAS version: 1.63.0

ChristopherRiedel:
Fixes #6579

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sample/AdditionalTestService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sample/AdditionalTestService.java
@@ -42,7 +42,7 @@ public class AdditionalTestService extends AdoServiceWithUserFilter<AdditionalTe
 		CriteriaQuery<AdditionalTest> cq = cb.createQuery(getElementClass());
 		Root<AdditionalTest> from = cq.from(getElementClass());
 
-		Predicate filter = createActiveSamplesFilter(cb,from);
+		Predicate filter = createActiveSamplesFilter(cb, from);
 
 		if (user != null) {
 			Predicate userFilter = createUserFilter(cb, cq, from);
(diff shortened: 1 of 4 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/samples/AdditionalTestController.java
```diff
@@ -57,9 +57,8 @@ public void onCommit() {
 
 	public void openEditComponent(AdditionalTestDto dto, Runnable callback) {
 
-		AdditionalTestDto newDto = FacadeProvider.getAdditionalTestFacade().getByUuid(dto.getUuid());
 		AdditionalTestForm form = new AdditionalTestForm(FacadeProvider.getSampleFacade().getSampleByUuid(dto.getSample().getUuid()), false);
-		form.setValue(newDto);
+		form.setValue(dto);
```
MartinWahnschaffe, line 61: Why was this switched to dto instead of newDto?
Getting the latest version of the dto is what we are always doing when starting to edit an entry, so if there is a reason to do it differently here, we need to at least have a comment. In this case it would also not be necessary to do the getByUuid call, done above.

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/AdditionalTestService.java
```diff
@@ -141,4 +138,10 @@ public void delete(List<String> additionalTestUuids) {
 
 		em.createQuery(cd).executeUpdate();
 	}
+
+	@Override
+	public void delete(AdditionalTest deleteme) {
+		em.remove(em.contains(deleteme) ? deleteme : em.merge(deleteme));
```
MartinWahnschaffe, line 144: It's super strange. I also did not manage to find out why deleteme is no longer attached.
Since we don't understand it, it could also happen for other entities, so I think we should move this code to the BaseAdoService and a comment there that the passed entity may be detached.

MartinWahnschaffe (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/AdditionalTestController.java
BarnaBartha: we were doing some testing if it affects the detached entity issues and forgot to revert it
will revert it now

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/AdditionalTestService.java
BarnaBartha: ok

Issue #6379, opened 2021-08-13
[Angular][Backend] Delete end-points [5]


laurentiustanvitagroup:
### Situation Description

- In view of the future replacement of the current FE with the Angular component, the Angular team will focus on the required changes.

### Feature Description

- The standard for the call should  be on "POST" -> "entity/delete"  (and in the body of the request we will have an array of ID's)
- The endpoint should be implemented for:
- The response (200) will be returned (i.e json -> Message, 200 OK), not an empty response

1. cases
2. tasks
3. events
4. samples
5. contacts
6. persons
7. users

### Possible Alternatives

- N/A

### Additional Information

- N/A

marius2301:
I made a sample for cases to discuss it with mate before doing for all the other entities. I also need to discuss with the frontend team after. In the sample for delete cases - the delete endpoint accepts a list of uuids and returns a list of uuids of the cases that have been deleted (the cases that were not deleted for different reasons will not be in the return list).

laurentiustanvitagroup:
The implementation was done only for the case. All other entities will be handled in a separate task.
Note: The response will return a list of deleted entities ID.

PR #6656, opened 2021-09-07
#6379 delete cases sample

marius2301:
Fixes #

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -2365,13 +2365,14 @@ public class CaseFacadeEjb implements CaseFacade {
 	}
 
 	public List<String> deleteCases(List<String> caseUuids) {
+		if (!userService.hasRight(UserRight.CASE_DELETE)) {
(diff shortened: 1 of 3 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2364,6 +2364,21 @@ public void deleteCase(String caseUuid) throws ExternalSurveillanceToolException
 		caseService.delete(caze);
 	}
 
+	public List<String> deleteCases(List<String> caseUuids) {
```
MateStrysewske, line 2367: Return void instead of a list of Strings and don't execute `isDeleted` for every case because that means one additional database call and query each time; if something goes wrong during the deletion, there will likely be an error anyway; also, please add the user right check from the `deleteCase` method at the start of this method as well

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
marius2301: The guys from the frontend requested a return. What about if I return similar as in case of push instead of PushResult, and Im not checking I return error only if that is an error?

MateStrysewske (approved):

PR #6697, opened 2021-09-14
Feature 4764 immunization changes

### LINKED ISSUE #4764: Immunizations II: Immunization overview and add/edit form [5]
### Situation Description
#4081 introduces the full immunization module. It would be good to read it before working on this.
### Feature Description
- [ ] Immunization directory
- Add a menu  item below samples 
- column variables: uuid, disease, first name, last name, ImmunizationStatus, immunidationStartDate, immunidationEndDate, 
- [ ] Immunization directory filter options
- followup status: all, under followup, ….. 
- general filter based id, person identifiers
- immunization person month, day and year of birth
- facility filters: region, districts, community, facility, ….
- checkbox called "Overdue immunization": computed on the fly as true if (current date > immunidationEndDate and sum (vaccination) < expected number of dose)
- disease (no multi-select)
- date:
-  date type: report date | immunization start, this should be enough, no need for (immunization end)
-  date ranger filter: from … to
- [ ] Immunization directory link to person search
-  add a box on the right below events to show all immunizations that the person is associated to.
- Variables to show in the box are: date of last vaccination, immunization end date, followup status , see immunization of this person link to the immunization directory filtered for the specific immunizations of the person.
### Development Specifications (WIP)

#### Overview
* Add a new main menu entry called "Immunizations" below Samples
* Only accessible for users with the `IMMUNIZATION_MANAGEMENT_ACCESS` user right
* Add a grid that contains all persons that the user has access to (similar to the already existing person directory) and that have at least one immunization entity connected, but with the following columns:
  * Immunization ID (containing the UUID)
  * Person ID (containing the UUID)
  * First name
  * Last name
  * Age and birth date
  * Sex
  * District (from the home address)
  * Means of immunization
  * Management status
  * Immunization status
  * Immunization start
  * Immunization end
  * Last vaccine type (the vaccineType of the latest vaccination)
  * Date of recovery
* The grid always needs to be filtered by a single disease and shows the data from the most recent immunization for that disease; if no disease has been selected, don't show the grid, but instead a label with bigger font size with the following text: "Please select a disease to display the immunization overview."
* Add the following filters:
  * Disease (if there is only one active disease on the server, set it to that disease by default)
  * General text filter to search for persons (as in person directory)
  * Birth date filters for persons (as in person directory)
  * Home address filters for persons (as in person directory)
  * Means of immunization
  * Management status
  * Immunization status
  * Jurisdiction filters for the most recent immunization (down to facility level)
  * Date filters (date type: report date, immunization end, validUntil, date of recovery, last vaccination date (= the date of the latest vaccination), first vaccination date (= the date of the first vaccination))
  * Checkbox: "Only show persons with overdue immunziation" (should show persons where the latest immunziation is ONGOING and the current or filtered date is after the immunization end date)
* Add a "New immunization" button to the top right of the screen that opens the immunization creation form

#### Creation form
* Add a new `ImmunizationCreationForm` that can, for now, be accessed by clicking the "New immunziation" button in the immunziation overview
* Only accessible for users with the `IMMUNIZATION_CREATE` user right
* The form consists of two parts - information about the immunization itself and about the person that it is associated to; person fields can be similar to those in the case creation form and should be at the bottom of the form
* Add the following fields for the immunization:
  * Report date (required)
  * Disease (required)
  * Means of immunization + details if "Other"; if "Recovery" or "Other" is selected, set the management status to `COMPLETED`
  * Checkbox "Overwrite immunization management status" - if checked, management status becomes editable
  * Management status (read only; set to SCHEDULED) - automatically updates the immunization status: if scheduled or ongoing, immunization status = pending; if completed, immunization status = acquired; if canceled, immunization status = not acquired
  * Immunization status (read only; set to PENDING)
  * External ID
  * Responsible region, district, community and facility (required)
  * Start date (required)
  * End date
  * Number of doses (text field, validation that makes sure only integers are added; only visible if means of immunization == Vaccination or Vaccination/Recovery)
* When clicking on "Save" in the creation form, we need a two-part duplicate detection process (person duplicate detection is similar to e.g. when creating new cases):
  * First, check for potential duplicates of the specified person
  * Secondly, if an existing person has been selected in the first step, search for immunziations associated with that person according to the following criteria:
    * Same disease
    * Overlapping immunization period defined by start and end date
  * If a duplicate immunization is found (there should never be more than one), show a popup (use the dupliate case popup as a reference) with the following contents:
    * An info text (and an info icon to its left side): "The system already contains an immunization for {disease name} for this person that overlaps with the specified immunziation period. Please compare the existing immunization with the one you created and decide whether to discard the new immunization or use its information to update the existing one."
    * An overview of the existing immunization, displaying the same information as in the creation form (without person information)
    * A radio button below this line: "Keep the existing information and discard the new immunization"
    * An overview of the immunziation information that was just entered, same as for the existing immunization
    * A radio button below this line: "Overwrite the existing immunization with this data"
    * None of the radio buttons should be selected by default
    * Popup has two options: Discard and Confirm; Confirm should only be enabled when one of the radio buttons is selected

#### Edit form
* The edit form should be displayed on an individual page after clicking on the UUID of the immunization or double clicking on a row in the immunization overview
* Read-only for users without the `IMMUNIZATION_EDIT` user right
* Edit rights are based on the responsible jurisdiction of the immunization
* The page has two tabs: Immunization and Person 
* The immunization tab contains the same information as in the creation form (without the person information, of course), plus the following fields:
  * Reporting user (read only)
  * Previous infection with this disease (Yes/No/Unknown)
  * Date of last infection (only visible if previous infection == Yes)
  * Additional details (text area)
  * Heading "Vaccination" (this and its dependent fields are only visible if means of immunization == vaccination or vaccination/recovery)
    * Number of doses (text field, validation that makes sure only integers are added)
  * Heading "Recovery" (this and its dependent fields are only visible if means of immunization == recovery or vaccination/recovery)
    * Date of first positive test result
    * Date of recovery
  * Immunization country (not related to the responsible jurisdiction fields)
* When saving, check whether there is any other immunization of this person and disease in the system that has an overlapping time period (defined by start and end date); if that's the case, DON'T save and display a popup instead:
  * Show the following message: "The system already contains an immunization for {disease name} for this person that overlaps with the specified immunization period. Please adjust the immunization period and save again."
  * Show two read-only text fields below: "Immunization period of this immunization" and "Immunization period of existing immunization"
  * Caption for the button that closes the popup: "Okay"
* Use the same logic and automation for means of immunization, management and immunization status as in the create form
* Edit form needs to include archive and delete option (only for users with the `IMMUNIZATION_ARCHIVE` and `IMMUNIZATION_DELETE` user right, respectively)

alexcaruntu-vita:
Fixes #4764


sormas-api/src/main/java/de/symeda/sormas/api/immunization/ImmunizationIndexDto.java
```diff
@@ -211,12 +208,4 @@ public Date getRecoveryDate() {
 	public void setRecoveryDate(Date recoveryDate) {
 		this.recoveryDate = recoveryDate;
 	}
-
-	public Disease getDisease() {
```
MateStrysewske, line 215: The disease still has to be a part of the directory and actually needs to be added as a column (I didn't think about that when writing down that the "select a disease" hint needs to be removed)

sormas-api/src/main/resources/strings.properties
```diff
@@ -725,15 +725,15 @@ infoCaseIncidenceIncompatible = No population data is available for communities
...
 infoPickOrCreateCase = There are existing cases in the database that seem very similar to the one you are about to create. Please have a look at the list of existing cases and verify that the case you want to create is not a duplicate of one of them. If you find a case that is most likely the same as yours, please <b>click on it in the list</b> and confirm.
 infoPickOrCreateCaseNewCase = Newly added case information
-infoPickOrCreateImmunization = The system already contains an immunization for %s for this person that overlaps with the specified immunziation period. Please compare the existing immunization with the one you created and decide whether to discard the new immunization or use its information to update the existing one.
+infoPickOrCreateImmunization = The system already contains at least one immunization for COVID-19 for this person and means of immunization that overlaps with the specified immunziation period. Please compare the most recent existing immunization with the one you created and decide whether to discard the new immunization, use its information to update the existing one, or create it anyway.
```
MateStrysewske, line 728: Remove the explicit disease name (the text in the issue was just an example) and replace it with a placeholder again (same in the string below)

sormas-app/app/src/main/java/de/symeda/sormas/app/immunization/edit/ImmunizationEditFragment.java
```diff
@@ -238,6 +233,12 @@ public void onAfterLayoutBinding(final FragmentImmunizationEditLayoutBinding con
 				contentBinding.facilityTypeGroup.setValue(facilityType.getFacilityTypeGroup());
 			}
 		}
+//
```
MateStrysewske, line 236: If this is no longer needed, remove the code

MateStrysewske (changes requested):

MateStrysewske (changes requested):
Mobile app:
- [x] Facility type dropdown contains no items for non-accommodation facility categories
- [x] Add the disease to the immunization list items
- [x] Add some spacing to the top and bottom of the popup when saving an overlapping existing immunization
- [x] "Save changes" button is too small to contain its caption; maybe setting slim to false could fix this, in the worst case you'll have to use one row per button

PR #6707, opened 2021-09-15
[#6696] use constant UUIDs for region, district, community

### LINKED ISSUE #6696: Constify UUID generation for default entities
### Feature Description
Default entity/infra generation

### Problem Description
Default infrastructure data has random UUID on every machine. To support testing of the central infra feature, it is beneficial to change this.

### Proposed Change
Deterministic UUIDs used in  `createDefaultInfrastructureData`

JonasCir:
Fixes #6696 
This is intentionally leaving out point of entry and facility. This is because they are (not yet included) in central, meaning UUIDs of these two types will also vary in the wild. To mimic this behavior while testing, both of the corresponding default entities will have random UUIDs.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/utils/DefaultEntityHelper.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/utils/DefaultEntityHelper.java
@@ -61,7 +61,7 @@ public class DefaultEntityHelper {
 			POE_INF_USERNAME_AND_PASSWORD);
 	}
 
-	public enum Infrastructure {
+	public enum DefaultInfrastructureUuidSeed {
 		REGION,
 		DISTRICT,
 		COMMUNITY,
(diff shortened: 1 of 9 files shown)


JonasCir:
This seems to trigger an issue with the default users

JonasCir:
Martin had some comments about using a seed to generate different UUIDs. Will take care of it tomorrow + I found some issues with the default users which I'm going to fix here as well.

JonasCir:
This is now ready for review. I created an enum and use the ordinal values of each variant to generate the corresponding UUID. By this we can arbitrary extend the enum (e.g., `SECOND_DEFAULT_REGION`) and we will also get a constant UUID for it.

sormas-api/src/main/java/de/symeda/sormas/api/utils/DefaultEntityHelper.java
```diff
@@ -61,6 +61,14 @@
 			POE_INF_USERNAME_AND_PASSWORD);
 	}
 
+	public enum Infrastructure {
```
MartinWahnschaffe, line 64: Better DefaultInfrastructureUuidSeed

MartinWahnschaffe (changes requested):
Just one rename - afterwards you can merge it.

PR #6711, opened 2021-09-15
Fixed #6327 Race condition populating CustomizableEnum caches

### LINKED ISSUE #6327: Race condition populating CustomizableEnum caches
### Bug Description

In `CustomizableEnumFacadeEjb` the values are cached in several static `HashMap`s.
They are cleared and populated in `CustomizableEnumFacadeEjb::loadData`.
The pool of `CustomizableEnumFacade`s is adjusted dynamically, especially when under load.
Spawning new instances may result in a 
`RuntimeException: Invalid enum value [...] for customizable enum type DISEASE_VARIANT`.

### Steps to Reproduce

1.
2.

### System Details

* Device:
* SORMAS version: 1.61.2
* Android version/Browser:
* Server URL:
* User Role:

### Additional Information

Please get rid of current violation of Java naming convention for variables when fixing this ticket.

MartinWahnschaffe:
Fixes #6327


HolgerReiseVSys:
Sharing the cache between the bean instances would make a lot of sense, especially when programmatically resetting, but also for performance reasons.
The maps handled in loadData() could be encapsulated in a cache object, maybe held in a static AtomicReference.
If the maps are not modified afterwards, synchronization should not be an issue when accessing the data.

Lazy loading of the translations in initCaches() could maybe rely on ConcurrentMaps in the cache object.
Entries should not be added to the map before the value object has been fully constructed.

MateStrysewske (approved):

HolgerReiseVSys (approved):
I have quickly looked through the changes and did not notice any obvious blockers

PR #6721, opened 2021-09-16
Feature 5713 demis2sormas handle new profile map observation.identifier

### LINKED ISSUE #5713: [DEMIS2SORMAS] Handle New Profile: Map Observation.identifier
### Situation Description

Preriquisite #5539 

The new DEMIS profiles introduce lifecycle management of DEMIS notifications. This means one notification can relate to previously sent notifications. They can, among other references, reference previously trasmitted observations, which are something like pathogen tests in SORMAS terminology. As a user, I want SORMAS to automatically relate such demis notifications to the same pathogen test(s).

### Feature Description

For that, it is necessary to store the [Observation.Identifier](https://simplifier.net/demis/pathogendetectioncvdp) to a pathogen test.
- [ ] Add a externalId String attribute to the TestReport dto and entity.
- [ ] Add this attribute also to the PathogenTest dto and entity. 
- [ ] Show this new attribute for the PathogenTest in the UI when creating or editing pathogenTests.
- [ ] When processing lab messages, always transfer the externalId from the related test report to pathogen test.

### Possible Alternatives

The PathogenTest.typingId can not be used here, or?

### Additional Information

Out of scope for this ticket is to fill the testReport.externalId when converting demis notifications to lab messages in the demis-adapter. This is to be done in [TICKET TO BE CREATED].

FredrikSchaeferVitagroup:
Fixes #5713. In contrast to the specification, I made the externalId field in the pathogen test form editable. Reasons are:

- it does not make sense to show it when creating a new pathogen test (if not processing lab messages), when it is not editable
- maybe users want to manually add a pathogen test with an external id (e.g. when that was recieved via FAX)
- pathogen test external ids do not need to stay in sync with something else. It is helpful if it they do match the Observation.Identifier.value used in DEMIS, but nothing seriously breaks if they don't. Do you have objections regarding this point, @HolgerReiseVSys?

This is a feature, so it sould be merged after 1.64 was released.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/labmessage/TestReportDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/labmessage/TestReportDto.java
@@ -44,6 +44,8 @@ public class TestReportDto extends EntityDto {
 	private String testResultText;
 	@Size(max = COLUMN_LENGTH_TEXT, message = Validations.textTooLong)
 	private String typingId;
+	@Size(max = COLUMN_LENGTH_SMALL, message = Validations.textTooLong)
+	private String externalId;
 
 	private PathogenTestReferenceDto pathogenTest;
 
@@ -147,5 +149,13 @@ public class TestReportDto extends EntityDto {
 
 	public void setTypingId(String typingId) {
 		this.typingId = typingId;
+	}
+
+	public String getExternalId() {
+		return externalId;
+	}
+
+	public void setExternalId(String externalId) {
+		this.externalId = externalId;
 	}
 }
--- a/sormas-api/src/main/java/de/symeda/sormas/api/sample/PathogenTestDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/sample/PathogenTestDto.java
@@ -62,6 +62,7 @@ public class PathogenTestDto extends PseudonymizableDto {
 	public static final String CQ_VALUE = "cqValue";
 	public static final String REPORT_DATE = "reportDate";
 	public static final String VIA_LIMS = "viaLims";
+	public static final String EXTERNAL_ID = "externalId";
 
 	@Required
 	private SampleReferenceDto sample;
@@ -105,6 +106,9 @@ public class PathogenTestDto extends PseudonymizableDto {
 	private Date reportDate;
 	@HideForCountriesExcept(countries = CountryHelper.COUNTRY_CODE_GERMANY)
 	private boolean viaLims;
+	@HideForCountriesExcept(countries = CountryHelper.COUNTRY_CODE_GERMANY)
+	@Size(max = COLUMN_LENGTH_SMALL, message = Validations.textTooLong)
+	private String externalId;
 
 	public static PathogenTestDto build(SampleDto sample, UserDto currentUser) {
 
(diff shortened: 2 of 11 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/labmessage/TestReportDto.java
```diff
@@ -44,6 +44,8 @@
 	private String testResultText;
 	@Size(max = COLUMN_LENGTH_TEXT, message = Validations.textTooLong)
 	private String typingId;
+	@Size(max = COLUMN_LENGTH_SMALL, message = Validations.textTooLong)
+	private String externalId;
```
leventegal-she, line 48: The externalId in other entities has max size COLUMN_LENGTH_DEFAULT, I think this one should also be the same

sormas-api/src/main/java/de/symeda/sormas/api/sample/PathogenTestDto.java
```diff
@@ -105,6 +106,9 @@
 	private Date reportDate;
 	@HideForCountriesExcept(countries = CountryHelper.COUNTRY_CODE_GERMANY)
 	private boolean viaLims;
+	@HideForCountriesExcept(countries = CountryHelper.COUNTRY_CODE_GERMANY)
+	@Size(max = COLUMN_LENGTH_SMALL, message = Validations.textTooLong)
+	private String externalId;
```
leventegal-she, line 111: same as for TestReport: COLUMN_LENGTH_DEFAULT

sormas-backend/src/main/java/de/symeda/sormas/backend/labmessage/TestReport.java
```diff
@@ -157,11 +159,21 @@ public void setPathogenTest(PathogenTest pathogenTest) {
...
+	@Column(length = COLUMN_LENGTH_SMALL)
 	public String getTypingId() {
 		return typingId;
 	}
 
 	public void setTypingId(String typingId) {
 		this.typingId = typingId;
 	}
+
+	@Column(length = COLUMN_LENGTH_SMALL)
```
leventegal-she, line 171: COLUMN_LENGTH_DEFAULT

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/PathogenTest.java
```diff
@@ -281,6 +283,15 @@ public void setViaLims(boolean viaLims) {
 		this.viaLims = viaLims;
 	}
 
+	@Column(length = COLUMN_LENGTH_SMALL)
```
leventegal-she, line 286: COLUMN_LENGTH_DEFAULT

leventegal-she (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/labmessage/TestReportDto.java
FredrikSchaeferVitagroup: In the databse, it is introduced as `varchar(255)`, so I think it should point to something of size 255. 
It seems that this was changed recently, I suggest to use [FieldContraints.CHARACTER_LIMIT_SMALL](https://github.com/hzi-braunschweig/SORMAS-Project/blob/0ae2c81f7e75bf4d69a261e3de3ed8ad59051420/sormas-api/src/main/java/de/symeda/sormas/api/utils/FieldConstraints.java#L7). Okay?

leventegal-she: It is introduced by the PR as `varchar(255)` so it can be changed and I think it's better to keep similar to other externalId fields that are `varchar(512)`

leventegal-she (approved):

PR #6724, opened 2021-09-16
[#6712] add uuids to continents, subcontinents, and countries

### LINKED ISSUE #6712: Consitfy UUIDs of continents, subcontinents and countries
### Feature Description
Importing the "global" Infra. data through csv/default infra import

### Problem Description
If you import the default continents, subcontinents, and countries, all entities will receive random uuids. This does not work for central infra distribution, as e.g., countries needs to be referenced through an DTO from regions and that UUID needs to be known beforehand.

### Proposed Change
Extend the CSV parser in such a way that it can handle an additional uuid column to specify the uuid of an entity beforehand.
Also, add uuids to all default CSV files [here](https://github.com/hzi-braunschweig/SORMAS-Project/tree/development/sormas-backend/src/main/resources).

JonasCir:
Fixes #6712 
Turns out the existing logic can figure out the property path on its own, so just adding the UUID column was sufficient :)
Import works and I checked the UUIDs they are written to the DB as they should.

### DIFF
--- a/sormas-backend/src/main/resources/sormas_import_all_continents.csv
+++ b/sormas-backend/src/main/resources/sormas_import_all_continents.csv
@@ -1,8 +1,7 @@
-## All continents
-"defaultName","externalId","archived"
-"Africa","31000005","FALSE"
-"America","31000006","FALSE"
-"Asia","31000004","FALSE"
-"Australia (Continent)","31000007","FALSE"
-"Europe","31000003","FALSE"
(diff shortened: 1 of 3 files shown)


JonasCir:
I have to convert it to a draft. Just found out, that the country data are different for Germany and I need to investigate further...

JonasCir:
Ready for review, files can also be found [here](https://github.com/hzi-braunschweig/SORMAS-central-data)

MateStrysewske (approved):

PR #6731, opened 2021-09-16
fix travel entry conversion #6283

### LINKED ISSUE #6283: Travel entry conversions to cases shows jurisdictional complications
### Bug Description

When converting travel entries to cases, jurisdiction and place of stay contradictions arise. This seems to be due to place of stay or jurisdiction being applicable to facilities as well as points of entries.
### Steps to Reproduce

1. Convert a travel entry to a case (the travel entry should have a point of entry outside the jurisdiction)
2. Select Home or Facility as place of stay
3. Enter compulsory data and save

### Expected Behavior

When converting travel entries to cases, cases should be treated as being referred to a facility (which includes home). This way, place of stay fields for facilities and points of entries should not compete with one another.  
### Screenshots

travelEntriesConvertError2
travelEntriesConvertError
travelEntriesConvertError3

### System Details

* Device:
* SORMAS version: 1.62.1
* User Role: admin, national user, supervisors, officers

syntakker:
Fixes #6283


MartinWahnschaffe (changes requested):
@syntakker Just a request to add a comment to that "temporary" variable added to the case create form. We need to make clear that the convertedTravelEntry variable is not supposed to stay there for long.

Afterwards you can merge.

PR #6734, opened 2021-09-17
6684 send samples automatically with ownership

### LINKED ISSUE #6684: [S2S] Samples have to be send when ownership is handed over
### Feature Description

Sharing cases and handing over the ownership
### Problem Description

Thus far, users are not obligated to share samples when handing over cases. This is very problematic because the samples' pathogen tests are accounted for when calculating the case classification. This cannot differ depending between instances because it could break the reporting process and miss-align important data. 
### Proposed Change

- When a case is being handed over, the samples always have to be selected as well to make sure that all relevant data is shared.
- As long as "Hand over the ownership" is selected, "Share samples" should not be able to be deselected 
### Additional Information

HandOverCases

vidi42:
Fixes #6684

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/SormasToSormasOptionsForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/SormasToSormasOptionsForm.java
@@ -56,15 +56,24 @@ public class SormasToSormasOptionsForm extends AbstractEditForm<SormasToSormasOp
 
 	public static SormasToSormasOptionsForm forCase(List<String> excludedOrganizationIds) {
 		return new SormasToSormasOptionsForm(
-			excludedOrganizationIds,
-			true,
-			Arrays.asList(SormasToSormasOptionsDto.WITH_ASSOCIATED_CONTACTS, SormasToSormasOptionsDto.WITH_SAMPLES),
-			(form) -> FieldHelper.setEnabledWhen(
-				form.getFieldGroup(),
-				SormasToSormasOptionsDto.HAND_OVER_OWNERSHIP,
-				Boolean.TRUE,
-				SormasToSormasOptionsDto.WITH_SAMPLES,
-				false));
+				excludedOrganizationIds,
+				true,
+				Arrays.asList(SormasToSormasOptionsDto.WITH_ASSOCIATED_CONTACTS, SormasToSormasOptionsDto.WITH_SAMPLES),
+				(form) -> {
+					FieldHelper.setEnabledWhen(
+							form.getFieldGroup(),
+							SormasToSormasOptionsDto.HAND_OVER_OWNERSHIP,
+							Boolean.FALSE,
+							SormasToSormasOptionsDto.WITH_SAMPLES,
+							false);
+
+					FieldHelper.setValueWhen(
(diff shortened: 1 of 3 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/SormasToSormasOptionsForm.java
```diff
@@ -121,12 +135,16 @@ protected void addFields() {
...
 			CheckBox pseudonymizeSensitiveData = addField(SormasToSormasOptionsDto.PSEUDONYMIZE_SENSITIVE_DATA);
+			
 			pseudonymizeSensitiveData.addStyleNames(CssStyles.VSPACE_3);
 
 			handoverOwnership.addValueChangeListener(e -> {
 				boolean ownershipHandedOver = (boolean) e.getProperty().getValue();
 				pseudonimyzePersonalData.setEnabled(!ownershipHandedOver);
 				pseudonymizeSensitiveData.setEnabled(!ownershipHandedOver);
+				
+				if (customOptions.contains(SormasToSormasOptionsDto.WITH_SAMPLES))
```
leventegal-she, line 146: I think this if statement is not needed

sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/SormasToSormasOptionsForm.java
```diff
@@ -56,10 +56,24 @@
...
 		return new SormasToSormasOptionsForm(
-			excludedOrganizationIds,
-			true,
-			Arrays.asList(SormasToSormasOptionsDto.WITH_ASSOCIATED_CONTACTS, SormasToSormasOptionsDto.WITH_SAMPLES),
-			null);
+				excludedOrganizationIds,
+				true,
+				Arrays.asList(SormasToSormasOptionsDto.WITH_ASSOCIATED_CONTACTS, SormasToSormasOptionsDto.WITH_SAMPLES),
+				(form) -> {
+					FieldHelper.setEnabledWhen(
```
leventegal-she, line 63: similar thing would be needed for sharing events with event participants too.

Also I would change a bit the order of the options on the dialog to have Handover as a first option because almost all other options depend on it.


Also on the backend would be nice to force sending samples and cleanup pseudonymization option when handing over the ownership, for the in case something doesn't work as expected on the IU. 
But this could go into a separate ticket

leventegal-she (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/SormasToSormasOptionsForm.java
vidi42: I asked if this has to be applied to events with participants as well (https://github.com/hzi-braunschweig/SORMAS-Project/issues/6684#issuecomment-922676797), but the ticket states that it concerns cases because it affects calculating the case classification, so I'm not sure if it's relevant for events as well.

I changed the order and moved the ownership and pseudonymization options to the top.

I've also added a backend method to ensure the options are correctly matching the ownership selection.

leventegal-she (approved):

Issue #6758, opened 2021-09-21
Cleanup InfraValidator


JonasCir:
### Feature Description
InfrastructureValidator class
### Problem Description
Currently, there is a lot of code duplication and unclear flows in this class, improve the status quo step by step.

### Proposed Change
- [x] Unify lookup logic of local entities
- [x] Rename some functions such that it is clearer what they do
- [ ] Move validation logic to dedicated file
- [ ] Unify return types and call sites of the validation functions

PR #6759, opened 2021-09-21
[#6758] unify local entitiy loading and rename functions

JonasCir:
Ticks boxes in #6758 
If you review this, you should enable the hiding of white space changes in GitHub, InfrastructureValidator was not formatted correctly.


vidi42:
LGTM, however some tests have to be fixed.

Also, since we are going with the approach of having constant IDs for infrastructure data, maybe we should also consider updating the scripts from https://github.com/hzi-braunschweig/SORMAS-Project/tree/development/sormas-backend/utils to support importing with UUID, or provide new scripts for this

vidi42 (approved):

JonasCir:
@vidi42 I made a tiny change as I thought it was missed before, however, apparently it makes a difference though and tests rely on it.

PR #6760, opened 2021-09-21
#6327 fixed NPE and initialization of unspecifc custom enums + unit test

### LINKED ISSUE #6327: Race condition populating CustomizableEnum caches

MartinWahnschaffe:
Fixes #6327

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/customizableenum/CustomizableEnumFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/customizableenum/CustomizableEnumFacadeEjb.java
@@ -133,7 +133,7 @@ public class CustomizableEnumFacadeEjb implements CustomizableEnumFacade {
 	public <T extends CustomizableEnum> List<T> getEnumValues(CustomizableEnumType type, Disease disease) {
 		Language language = I18nProperties.getUserLanguage();
 		Class<T> enumClass = (Class<T>) type.getEnumClass();
-		Optional<Disease> innerDisease = Optional.of(disease);
+		Optional<Disease> innerDisease = disease != null ? Optional.of(disease) : Optional.empty();
 
 		if (!enumValuesByLanguage.get(enumClass).containsKey(language)) {
 			fillLanguageCache(type, enumClass, language);
@@ -204,7 +204,7 @@ public class CustomizableEnumFacadeEjb implements CustomizableEnumFacade {
 				.stream()
 				.filter(
 					e -> !disease.isPresent() && CollectionUtils.isEmpty(e.getDiseases())
-							|| e.getDiseases() != null && e.getDiseases().contains(disease.get()))
+							|| disease.isPresent() && e.getDiseases() != null && e.getDiseases().contains(disease.get()))
 				.map(CustomizableEnumValue::getValue)
 				.collect(Collectors.toList());
 			enumValuesByDisease.get(enumClass).put(disease, filteredEnumValues);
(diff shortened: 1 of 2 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/customizableenum/CustomizableEnumFacadeEjb.java
```diff
@@ -204,7 +204,7 @@ public boolean hasEnumValues(CustomizableEnumType type, Disease disease) {
 				.stream()
 				.filter(
 					e -> !disease.isPresent() && CollectionUtils.isEmpty(e.getDiseases())
-							|| e.getDiseases() != null && e.getDiseases().contains(disease.get()))
+							|| disease.isPresent() && e.getDiseases() != null && e.getDiseases().contains(disease.get()))
```
HolgerReiseVSys, line 207: Seems to be e bit too much logic for one expression
`if (disease.isPresent())…`?

sormas-backend/src/main/java/de/symeda/sormas/backend/customizableenum/CustomizableEnumFacadeEjb.java
```diff
@@ -143,24 +138,19 @@
...
-				enumValue.setProperties(enumProperties.get(type).get(value));
-				enumValues.add(enumValue);
-			} catch (InstantiationException | IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
-				throw new RuntimeException(e);
-			}
-		});
+		Stream<String> diseaseValuesStream;
+		if (innerDisease.isPresent()) {
+			// combine specific and unspecific values
+			diseaseValuesStream = Stream.concat(
```
HolgerReiseVSys, line 144: Frankly, I do not completely understand what how this is working, but can't we get two enties for the same value now?

sormas-backend/src/main/java/de/symeda/sormas/backend/customizableenum/CustomizableEnumFacadeEjb.java
```diff
@@ -169,46 +159,71 @@ public boolean hasEnumValues(CustomizableEnumType type, Disease disease) {
...
-					enumValues.putIfAbsent(customizableEnumValue.getValue(), customizableEnumValue.getCaption());
+		// may have already been put by another thread while this one was waiting (synchronized)
+		if (enumValuesByLanguage.get(enumClass).containsKey(language)) {
+			return;
+		}
+
+		boolean isCountryLanguage = StringUtils.equals(configFacade.getCountryLocale(), language.getLocale().toString());
+
+		Map<String, String> languageEnumValues = new HashMap<>();
+		for (CustomizableEnumValue customizableEnumValue : enumValueEntities.get(type)) {
```
HolgerReiseVSys, line 184: It might be a bit cleaner to stream enumValueEntities then collected in a map

MateStrysewske (changes requested):
~~Disease variants with diseases == null are not shown in the UI~~

sormas-backend/src/main/java/de/symeda/sormas/backend/customizableenum/CustomizableEnumFacadeEjb.java
MartinWahnschaffe: I'd leave it as-is.

sormas-backend/src/main/java/de/symeda/sormas/backend/customizableenum/CustomizableEnumFacadeEjb.java
MartinWahnschaffe: No, enum values are unique per data type, so it's not possible that the same value is in the steam twice.

MateStrysewske (approved):

PR #6784, opened 2021-09-22
fix android tests #6714

### LINKED ISSUE #6714: Fix failing AndroidTests
### Bug Description
A number of android tests fail
androidTests2
androidTests1

Some of the tests do not reflect the current state of the implementation:
- `CaseBackendTest..shouldUpdateCaseAndAssociatedEntitiesOnMove()` manipulates `Case.district`. However, task reassignment is triggered by a change of `Case.responsibleDistrict` or - if `responsibleDistrict` is null, which is not the case here - by a change of `district`. Thus, neither assignment of a new `surveillanceOfficer` nor a reassignment of tasks occur in the course of the test. Equallly, hospitalizations are not updated as expected in the test.

Some of the tests fail depending of other tests run before:
- `CaseBackendTest.shouldCreateCase()` succeeds in the first run with no tests run before, but fails in a second run.

### Steps to Reproduce

Run android tests in android studio, either one by one or all together

### Expected Behavior
Android tests should succeed, test result should be independent of other tests run  before.
Database and other persistent states should be cleaned and prepared before each test.

### System Details

* Device:
* SORMAS version: 1.64.0-SNAPSHOT
* Android version/Browser:
* Server URL:
* User Role:

syntakker:
Fixes #6714

To the reviewer: This fix affects a number of processes in SORMAS. Please have a close eye on possible alterations of processes and double check if implementation and test expectations still meet the intended behaviour.


sormas-app/app/src/androidTest/java/de/symeda/sormas/app/CaseBackendTest.java
```diff
@@ -272,9 +274,14 @@ public void shouldUpdateCaseAndAssociatedEntitiesOnMove() throws DaoException {
 		Task doneTask = TestEntityCreator.createCaseTask(caze, TaskStatus.DONE, user);
 
 		Case existingCase = caseDao.queryUuidWithEmbedded(caze.getUuid());
-		caze.setDistrict(DatabaseHelper.getDistrictDao().queryUuid(TestHelper.SECOND_DISTRICT_UUID));
-		caze.setCommunity(DatabaseHelper.getCommunityDao().queryUuid(TestHelper.SECOND_COMMUNITY_UUID));
+		District secondDistrict = DatabaseHelper.getDistrictDao().queryUuid(TestHelper.SECOND_DISTRICT_UUID);
```
MartinWahnschaffe, line 277: I'd suggest to only set the responsible... instead of both.

This means that we also have to fix the TestEntityCreator.createCase() method to assign responsible... instead of the old infrastructure region,district,community fields that are only relevant when the home/place of stay differs from the responsible region/district/community.

sormas-app/app/src/androidTest/java/de/symeda/sormas/app/CaseBackendTest.java
```diff
@@ -420,19 +427,23 @@ public void testEditCasePermissionWhenRegionMatch() throws DaoException, SQLExce
 	}
 
 	@Test
-	public void testEditCasePermissionWhenRegionDoesNotMatchAndCaseNotCreatedByUser() throws DaoException, SQLException {
+	public void testEditCasePermissionWhenDistrictDoesNotMatchAndCaseNotCreatedByUser() throws DaoException, SQLException {
 		Case caze = TestEntityCreator.createCase();
-		caze.setRegion(null);
-		caze.setDistrict(null);
+		caze.setRegion(DatabaseHelper.getRegionDao().queryUuid(TestHelper.REGION_UUID));
```
MartinWahnschaffe, line 432: No need to set this again

sormas-app/app/src/androidTest/java/de/symeda/sormas/app/CaseBackendTest.java
```diff
@@ -420,19 +427,23 @@ public void testEditCasePermissionWhenRegionMatch() throws DaoException, SQLExce
...
+		District secondDistrict = DatabaseHelper.getDistrictDao().queryUuid(TestHelper.SECOND_DISTRICT_UUID);
+		caze.setDistrict(secondDistrict);
+		caze.setResponsibleDistrict(secondDistrict);
 		caze.setHealthFacility(null);
 
-		UserRole userRole = UserRole.HOSPITAL_INFORMANT;
```
MartinWahnschaffe, line 429: Not sure why it's necessary to switch this - should also work with the HOSPITAL_INFORMANT

sormas-app/app/src/androidTest/java/de/symeda/sormas/app/FacilityBackendTest.java
```diff
@@ -45,16 +52,45 @@ public void initTest() {
...
-			DatabaseHelper.getFacilityDao().getActiveHealthFacilitiesByDistrictAndType(district, FacilityType.HOSPITAL, true, true).size(),
-			is(3));
+			DatabaseHelper.getFacilityDao().getActiveHealthFacilitiesByDistrictAndType(district, FacilityType.HOSPITAL, false, false).size(),
+			is(1));
+
+		List<Facility> activeHealthFacilitiesByDistrictAndType =
+			DatabaseHelper.getFacilityDao().getActiveHealthFacilitiesByDistrictAndType(district, null, true, true);
+		assertThat(activeHealthFacilitiesByDistrictAndType.size(), is(2));
+
+		Facility otherFacility = new Facility();
```
MartinWahnschaffe, line 72: Please move the creation of OTHER/NONE facilities into a new method in TestHelper that is called by TestHelper.insertInfrastructureData. To be realistic, we should always have those available.

I'd like to have one additional assertion that tests (district, FacilityType.HOSPITAL, true, true) -> should be 3

sormas-app/app/src/androidTest/java/de/symeda/sormas/app/TestHelper.java
```diff
@@ -53,9 +53,10 @@
 
 	public static void initTestEnvironment(boolean setInformantAsActiveUser) {
 		// Initialize a testing context to not operate on the actual database
-		RenamingDelegatingContext context = new RenamingDelegatingContext(InstrumentationRegistry.getTargetContext(), "test_");
+		Context context = InstrumentationRegistry.getInstrumentation().getTargetContext();
```
MartinWahnschaffe, line 56: This will always use the existing db on the emulator and afterwards delete it. Would be good to somehow wrap this to use another database location.

MartinWahnschaffe: In the worst case we could mock DatabaseHelper.DATABASE_NAME to provide a different name when testing,

MartinWahnschaffe (changes requested):

sormas-app/app/src/androidTest/java/de/symeda/sormas/app/CaseBackendTest.java
syntakker: removed.

sormas-app/app/src/androidTest/java/de/symeda/sormas/app/TestHelper.java
syntakker: DatabaseHelper is used with a prefixed test database now.

sormas-app/app/src/androidTest/java/de/symeda/sormas/app/FacilityBackendTest.java
syntakker: done.

PR #6796, opened 2021-09-23
[#6587] abstract class for shared functionality

### LINKED ISSUE #6587: Create abstract base class for all infra. EJBs
### Feature Description
Sormas Infrastructure/Location

### Problem Description
We have a LOT of code duplication in the different EJBs which makes fixing and introducing new features a pain.

### Proposed Change
Create an abstract class for all EJBs and implement shared functionality there.

JonasCir:
Fixes #6587 
@MateStrysewske @MartinWahnschaffe @StefanKock you might want to take a look at this :)

What I did is the following:

I created an abstract class `AbstractBaseEjb` which in the future should implement default versions for all methods in `BaseFacade`. There is also `AbstractInfrastructureEjb` which currently works around some issue `AbstractBaseEjb` has. Every infrastructure ejb now extends the `AbstractInfrastructureEjb` and uses constructor injection to instantiate the used service.

In this PR, I only cleanup up (de)archiving, but you can do the same for e.g., `save` as well as, at least in the infra facades, its just copy pasted. Prospectively, I think we should apply this to more facades/ejbs and let them extend `BaseFacade` and `AbstractBaseEjb` respectively. 

Doing this will help us to enforce clear data processing flows and validation. Right now there is a lot of inconsistency in the ejbs where methods are named differently, but do roughly the same across the system. I see the chance to greatly reduce LOC here by introducing more generic code.

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractBaseEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractBaseEjb.java
@@ -1,7 +1,4 @@
-package de.symeda.sormas.backend;
-
-import de.symeda.sormas.backend.common.AbstractDomainObject;
-import de.symeda.sormas.backend.common.AdoServiceWithUserFilter;
+package de.symeda.sormas.backend.common;
 
 // todo should we use BaseAdoService?
 public abstract class AbstractBaseEjb<ADO extends AbstractDomainObject, SRV extends AdoServiceWithUserFilter<ADO>> {
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractBaseEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractBaseEjb.java
@@ -1,6 +1,5 @@
 package de.symeda.sormas.backend.common;
 
-// todo should we use BaseAdoService?
 public abstract class AbstractBaseEjb<ADO extends AbstractDomainObject, SRV extends AdoServiceWithUserFilter<ADO>> {
 
 	protected SRV service;
@@ -17,4 +16,6 @@ public abstract class AbstractBaseEjb<ADO extends AbstractDomainObject, SRV exte
 	public abstract void archive(String uuid);
 
 	public abstract void dearchive(String uuid);
+
+	// FIXME(@JonasCir) #6821: Add missing functions like save, getByUuid etc
 }
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/AbstractInfrastructureEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/AbstractInfrastructureEjb.java
@@ -36,24 +36,17 @@ public abstract class AbstractInfrastructureEjb<ADO extends InfrastructureAdo, S
 
 	public void dearchive(String uuid) {
 		checkInfraDataLocked();
-		doDearchive(uuid);
-	}
 
-	protected void dearchiveUnchecked(String uuid) {
-		doDearchive(uuid);
-	}
-
-	private void checkInfraDataLocked() {
-		if (!featureConfiguration.isFeatureEnabled(FeatureType.EDIT_INFRASTRUCTURE_DATA)) {
-			throw new ValidationRuntimeException(I18nProperties.getValidationError(Validations.infrastructureDataLocked));
-		}
-	}
-
-	private void doDearchive(String uuid) {
 		ADO ado = service.getByUuid(uuid);
 		if (ado != null) {
 			ado.setArchived(false);
 			service.ensurePersisted(ado);
+		}
+	}
+
+	protected void checkInfraDataLocked() {
+		if (!featureConfiguration.isFeatureEnabled(FeatureType.EDIT_INFRASTRUCTURE_DATA)) {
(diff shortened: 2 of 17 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/area/AreaFacadeEjb.java
```diff
@@ -125,50 +132,38 @@ public AreaDto save(@Valid AreaDto dto, boolean allowMerge) {
...
-		return areaService.isUsedInInfrastructureData(areaUuids, Region.AREA, Region.class);
-	}
-
-	@Override
-	public void archive(String areaUuid) {
-		Area area = areaService.getByUuid(areaUuid);
-		area.setArchived(true);
-		areaService.ensurePersisted(area);
+	public void dearchive(String uuid) {
+		super.dearchiveUnchecked(uuid);
```
MateStrysewske, line 141: I don't think you need to override this here, the missing check for `EDIT_INFRASTRUCTURE_DATA` essentially looks like a bug to me (i.e. it should be there)

sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/AbstractInfrastructureEjb.java
```diff
@@ -0,0 +1,55 @@
...
+
+	protected AbstractInfrastructureEjb(SRV service, FeatureConfigurationFacadeEjb featureConfiguration) {
+		super(service);
+		this.featureConfiguration = featureConfiguration;
+	}
+
+	@Override
+	public void archive(String uuid) {
+		// todo this should be really in the parent but right now there the setter for archived is not available there
+		DTO dto = service.getByUuid(uuid);
```
MateStrysewske, line 29: Shouldn't archiving also check the `EDIT_INFRASTRUCTURE_DATA` right?

MateStrysewske (changes requested):
Martin and Stefan definitely should also have a look at this; lgtm, but I don't know whether EJBs should have constructors in the first place for example

sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/area/AreaFacadeEjb.java
JonasCir: good catch

sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/AbstractInfrastructureEjb.java
JonasCir: fixed

JonasCir:
https://stackoverflow.com/questions/53376733/dependency-injection-through-constructor-does-not-work-for-ejb-bean

This is the reason why we still need the default c'tors.

sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/AbstractInfrastructureEjb.java
```diff
@@ -0,0 +1,59 @@
...
+			service.ensurePersisted(ado);
+		}
+	}
+
+	public void dearchive(String uuid) {
+		checkInfraDataLocked();
+		doDearchive(uuid);
+	}
+
+	protected void dearchiveUnchecked(String uuid) {
```
MartinWahnschaffe, line 42: remove. Make checkInfraDataLocked protected and override it e.g. in PointOfEntryFacadeEjb

sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/AbstractInfrastructureEjb.java
```diff
@@ -0,0 +1,59 @@
...
+		doDearchive(uuid);
+	}
+
+	private void checkInfraDataLocked() {
+		if (!featureConfiguration.isFeatureEnabled(FeatureType.EDIT_INFRASTRUCTURE_DATA)) {
```
MartinWahnschaffe, line 47: Also use this method in the save method in all classes touched in this PR

MartinWahnschaffe (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractBaseEjb.java
```diff
@@ -0,0 +1,20 @@
+package de.symeda.sormas.backend.common;
+
+// todo should we use BaseAdoService?
```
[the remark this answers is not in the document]
JonasCir: Done

sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/AbstractInfrastructureEjb.java
JonasCir: removed

JonasCir: done

sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/AbstractInfrastructureEjb.java
JonasCir: done

PR #6826, opened 2021-09-29
disable overwrite follow-up when follow-up canceled or lost #6803

### LINKED ISSUE #6803: Overwrite follow-up until date checkbox is still displayed if the follow-up is canceled
### Bug Description
Overwrite follow-up until date checkbox is still displayed if the follow-up is canceled

### Steps to Reproduce
1. Open Sormas app
2. Create a new Contact or edit any existent one
3. Make sure that Follow-Up Status is set to:  Canceled follow-up
4. Check **Overwrite follow-up until date checkbox**

### Expected Behavior
Overwrite follow-up until date checkbox should be disabled if the follow up status is canceled.

### Screenshots
followUpdate

### System Details

* Device: Chrome
* SORMAS version: 1.64.0
* Android version/Browser: 93.0.45
* Server URL: https://test-performance.sormas.netzlink.com/sormas-ui/#!contacts/data/2d6e40fd-75d3-49c1-bb7a-1539421441ed
* User Role: National User

syntakker:
Fixes #6803

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactDataForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactDataForm.java
@@ -178,6 +178,8 @@ public class ContactDataForm extends AbstractEditForm<ContactDto> {
 	private NullableOptionGroup contactCategory;
 	private boolean quarantineChangedByFollowUpUntilChange = false;
 	private TextField tfExpectedFollowUpUntilDate;
+	private CheckBox cbOverwriteFollowUpUntil;
+	private DateField dfFollowUpUntil;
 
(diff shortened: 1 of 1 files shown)


MateStrysewske (changes requested):
While testing I realised that setting the checkbox to read-only isn't really a good solution because the read-only state looks like the writable state. I'd suggest to simply hide the checkbox if the follow-up status is canceled or lost.

syntakker:
@MateStrysewske shouldn't the user be informed here that the expected follow-up date has been intentionally overwritten? Otherwise there are only two different folow-up dates visible, but no explanation at all why they are different.

MateStrysewske:
@syntakker Good point - let's leave it like it is then :)

MateStrysewske (approved):

Issue #6676, opened 2021-09-09
[Angular][Backend] Create / Edit endpoints for adding Countries, Communities, Districts, Regions [3]


laurentiustanvitagroup:
### Situation Description

- In view of the future replacement of the current FE with the Angular component, the Angular team will focus on the required changes.

### Feature Description

-We need to implement the endpoints for adding Countries, Communities, Districts, Regions
-This task needs to also handle the edit for this entities.

### Possible Alternatives

- N/A

### Additional Information

- N/A

PR #6838, opened 2021-09-29
#6676 create edit country region district communities

marius2301:
Fixes #

### DIFF
--- a/sormas-rest/src/main/java/de/symeda/sormas/rest/CommunityResource.java
+++ b/sormas-rest/src/main/java/de/symeda/sormas/rest/CommunityResource.java
@@ -21,6 +21,7 @@ import java.util.Date;
 import java.util.List;
 
 import javax.annotation.security.RolesAllowed;
+import javax.validation.Valid;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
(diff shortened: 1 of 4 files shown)


JonasCir:
General question: Was there a specific reason why this was not support until now? :)

sormas-backend/src/main/java/de/symeda/sormas/backend/infrastructure/community/CommunityFacadeEjb.java
```diff
@@ -286,6 +287,15 @@ public CommunityDto save(@Valid CommunityDto dto, boolean allowMerge) throws Val
 		if (!featureConfiguration.isFeatureEnabled(FeatureType.EDIT_INFRASTRUCTURE_DATA)) {
 			throw new ValidationRuntimeException(I18nProperties.getValidationError(Validations.infrastructureDataLocked));
 		}
+		if (dto.getUuid() != null) {
```
MateStrysewske, line 290: Please
* remove the unnecessary nesting (i.e. combine the two statements in the if and the else block)
* move the checks to below the getByUuid methods and check whether the entity is null or not; DTOs will always have a UUID when they come from the UI, so the current logic will not work properly

MateStrysewske (changes requested):

MateStrysewske (approved):

MateStrysewske:
@marius2301 Can you please fix the remaining merge conflict? :)

MateStrysewske:
@JonasCir Because there is no way to edit infrastructure data in the mobile app

PR #6846, opened 2021-09-30
Bugfix 6815 synchronization not work when lang changed on web and app

### LINKED ISSUE #6815: Synchronization does not work anymore when user language is changed both on web and app
### Steps to Reproduce

1. Create a Surveillance Officer and select English for language
2. Open the app and change the language to anything else
3. Go to web app and change the language to a different one
4. Update a contact. Synchronization works.
5. Update a case.
6. Synchronize the app. The following message is displayed: "Some entries could not be synchronized. Please try again later"

### Expected Behavior

Synchronization should work fine in this case

### Screenshots

image

### System Details

* Device: Android
* SORMAS version: 1.65.0
* Android version/Browser: 1.65.0
* Server URL:

sergiupacurariu:
Fixes #6815


MateStrysewske (commented):
I would also request a review by @MartinWahnschaffe here because I'm not sure whether we have to move users out of the infrastructure synchronization loop and move everything into the usual data synchronization loop (i.e. also the pulling) since we now allow pushing as well.

sergiupacurariu:
the fix is in this PR https://github.com/hzi-braunschweig/SORMAS-Project/pull/6882

PR #6849, opened 2021-09-30
#6833 - fix merging of activityAsCase on case merge

### LINKED ISSUE #6833: Activiy as case data does not get merged properly
### Bug Description

When merging cases, activity as case data is not merged with available data as the leading data sets. Specifically, when the leading case has `activityAsCaseDetailsKnown` set to `YES` and the table has content
### Steps to Reproduce

1. Create two duplicate cases
2. In the first case (leading), set `activityAsCaseKnown` to `NO`
3. In the second case (dublicate), set `activityAsCaseKnown` to `YES` and add an entry to the table
4. Merge the cases with the first one remaining
5. Check the activity as case section of the remainging case

### Expected Behavior

Independent of whether a case is leading or subsequently discarded, available information should always be adopted over entries like NO or UNKNOWN

(cf. merging of exposures for reference)
### Screenshots

Preconditions:
mergeActivityAsCaseError4

After merging:
mergeActivityAsCaseError

### System Details

* Device:
* SORMAS version: 1.63.1
* Server URL: int-sormas-x.sormas.netzlink.com

BarnaBartha:
Fixes #6833

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -191,6 +191,7 @@ import de.symeda.sormas.api.utils.fieldvisibility.FieldVisibilityCheckers;
 import de.symeda.sormas.api.visit.VisitDto;
 import de.symeda.sormas.api.visit.VisitResultDto;
 import de.symeda.sormas.api.visit.VisitStatus;
+import de.symeda.sormas.backend.activityascase.ActivityAsCaseService;
 import de.symeda.sormas.backend.caze.classification.CaseClassificationFacadeEjb.CaseClassificationFacadeEjbLocal;
 import de.symeda.sormas.backend.caze.maternalhistory.MaternalHistoryFacadeEjb;
 import de.symeda.sormas.backend.caze.maternalhistory.MaternalHistoryFacadeEjb.MaternalHistoryFacadeEjbLocal;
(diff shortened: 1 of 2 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -3454,6 +3457,21 @@ private void mergeCase(CaseDataDto leadCaseData, CaseDataDto otherCaseData, bool
...
+		final EpiData otherEpiData = otherCase.getEpiData();
+		if (otherEpiData != null && YesNoUnknown.YES == otherEpiData.getActivityAsCaseDetailsKnown()) {
+			final EpiData leadEpiData = leadCase.getEpiData();
+			if (leadEpiData != null) {
+				if (YesNoUnknown.YES != leadEpiData.getActivityAsCaseDetailsKnown()) {
+					leadEpiData.setActivityAsCaseDetailsKnown(YesNoUnknown.YES);
+					epiDataService.ensurePersisted(leadEpiData);
+				}
+			} else {
+				leadCase.setEpiData(otherEpiData);
```
MateStrysewske, line 3471: This is a **very** bad idea because it will overwrite ANY epi data changes of the lead case which is not what we want to do. Instead of the logic implemented here, please make sure that
a) if otherCase has at least one activity as case, set activityAsCaseDetailsKnown of leadCase to YES and save the epi data
b) if not cloning, set the epi data of any activity as case of otherCase to the epi data of leadCase and save the activity as case; if cloning, create a copy of each activity as case of otherCase and assign that copy to the epi data of leadCase; see e.g. how it's done for clinical visits above

MateStrysewske (changes requested):

PR #7009, opened 2021-10-19
fix import case overwrite #6975

### LINKED ISSUE #6975: Re-imported cases produce duplicates if the existing case is supposed to be overwritten
### Bug Description

When exporting and re-importing a case, a duplicate case is created when trying to overwrite the existing case.
### Steps to Reproduce

1. Create or choose an existing case
2. Export the case in a detailed export
3. Change something in the file (don't do changes that would exclude them in the duplicate detection)
4. Re-import the case and select the same person
5. In the case detection window, select overwrite and confirm

### Expected Behavior

The existing case should get updated according to the imported data
### Screenshots

reImportOverwriteDuplicate

### System Details

* Device:
* SORMAS version: 1.63.1 / 1.64.0
* Server URL: int-sormas-x.sormas.netzlink.com / test1401.sormas-oegd.de

### Additional Information

This issue might be caused by changes in https://github.com/hzi-braunschweig/SORMAS-Project/issues/5705

syntakker:
Fixes #6975


syntakker:
@MateStrysewske Note: the changes introduced in #5705 (regeneration of the Uuid to import) affect contacts and event participants, too. However this does not result in the problem described here, since:

- contact import do not allow to overwrite data in existing contacts when a duplicate contact is selected (no overwrite checkbox)
- event participant import does a deduplication of persons, but not of event participants. So importing a duplicate event participant always results in creating a new event participant, even if a duplicate person is selected in the deduplication

It might make sense to implement the import procedure uniformly for all three entities, but this should be subject to a change request of its own.

PR #7020, opened 2021-10-20
change/6469-check-infra

### LINKED ISSUE #6469: Check (outbound) entities for correct infrastructure data [3]

JonasCir:
Fixes #6469


JonasCir:
this has been done in a couple of other Prs

PR #7067, opened 2021-10-27
#4201 Errors repeatedly logged for minor configuration issues

### LINKED ISSUE #4201: Errors repeatedly logged for minor configuration issues
### Bug Description
Docker sets properties in `sormas.properties` by replacing the lines in the file,
see https://github.com/hzi-braunschweig/SORMAS-Docker/blob/master/sormas/start-server.sh (Line 140ff).
When an env variable is missing, or empty to signal default behavior, the property is not removed or commented out, but the value is left blank.

ConfigFacadeEjb returns the defaultValue, but each time logs an error:
```
Could not parse integer value of property 'daysAfterCaseGetsArchived': For input string: "" de.symeda.sormas.backend.common.ConfigFacadeEjb.java:172)
````

### Steps to Reproduce
1. Start docker image with `CASEARCHIVEDAYS` not set
2. Open cases view
3. Observer log

### Expected Behavior
Empty property entries should be handled just like missing ones.

Alternatively, the log level could be reduced to `debug`, 
or `start-server.sh` could remove or comment out the lines in `sormas.properties`.

### System Details
* SORMAS version: 1.54.1

leventegal-she:
Fixes #4201

Changed logs to debug as suggested in the ticket


vidi42 (changes requested):
I think we should better check and only parse the property values if they are not empty and if they are empty we could add a debug log mentioning this.

For the catch we should still log an error because some values might be set incorrectly and those should be pointed out.

vidi42 (approved):

PR #7093, opened 2021-10-29
Change 1773 directly load similar person dto

### LINKED ISSUE #1773: Remove PersonNameDto and directly load SimilarPersonDto instead
### Feature Description
The person duplicate detection has been refactored some time ago in order to do the similarity detection directly in the database.

### Problem Description
We are still retrieving PersonNameDtos and then use them to immediately query for SimilarPersonDtos based on the UUIDs. The PersonNameDtos were used before to match the names, but they are not necessary anymore.

### Proposed Change
Get rid of the PersonNameDto, rename PersonFacade.getMatchingNameDtos to getSimilarPersonDtos and let it directly query for SimilarPersonDtos; PersonFacade.getSimilarPersonByUuids can also be removed then.

### Additional information
https://github.com/hzi-braunschweig/SORMAS-Project/issues/2782

alexcaruntu-vita:
Fixes #1773


sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
```diff
@@ -565,14 +565,61 @@ public Predicate inJurisdictionOrOwned(PersonQueryContext personQueryContext) {
...
-		TypedQuery<PersonNameDto> query = em.createQuery(personQuery);
+		TypedQuery<Person> query = em.createQuery(personQuery);
 		if (limit != null) {
 			query.setMaxResults(limit);
 		}
-		return query.getResultList();
+		return query.getResultList().stream().map(this::toSimilarPersonDto).collect(Collectors.toList());
+	}
+
+	public List<SimilarPersonDto> getSimilarPersonsByUuids(List<String> personUuids) {
```
MartinWahnschaffe, line 575: This method is no longer needed. Please remove.

PR #7108, opened 2021-11-01
Feature 5899 demis2sormas process multiple test reports

### LINKED ISSUE #5899: [DEMIS2SORMAS] Handle New Profile: Several PathogenTests: Process multiple testReports [3]
### Situation Description

Preriquisite: #5539 

In the future profile of DEMIS notifications, there may be several Observations (which map to PathogenTestResults).
SORMAS needs to be able to handle that. 

### Feature Description

Adapt the processing:
- [x] When processing a labMessage with just one testReport, use the same workflow as before
- [x] When a new sample is created while processing a labMessage with several testReports, create a prefilled PathogenTest for each testReport. Allow the user to add additional PathogenTests or to remove some, but require at least one PathogenTest to be created (see mockup below)
- [x] When an existing sample is selected while processing a labMessage with several test reports, forward to a view similar to the mockup, but just with all existing PathogenTests shown. The testReports from the labMessage must be transformed manually. This is shall be facilitated in the future (#5713 is needed for that, the facilitation is in #5729).

### Additional Information

Mockup_Multiple_Tests

FredrikSchaeferVitagroup:
Fixes #5899.


FredrikSchaeferVitagroup:
One thing I am not certain about is in the `sormas_schema.sql`: the `deleted` column in the `labmessage_history` should not default to `false`, or?

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleFacadeEjb.java
```diff
@@ -1191,6 +1196,18 @@ private void handleAssotiatedObjectChanges(Sample newSample, boolean syncShares)
 		}
 	}
 
+	@Override
+	public int caseSampleCountOf(SampleDto sample) {
```
leventegal-she, line 1200: Could be moved to the UI module

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleFacadeEjb.java
```diff
@@ -1215,4 +1232,25 @@ public Boolean isSampleEditAllowed(String sampleUuid) {
 
 		return sampleService.isSampleEditAllowed(sample);
 	}
+
+	@Override
+	public Disease getDiseaseOf(SampleDto sample) {
```
leventegal-she, line 1237: Could be moved to the UI module, don't really see a reason sending SampleDto back to backend

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
```diff
@@ -508,71 +499,39 @@ private void pickOrCreateSample(PseudonymizableDto dto, LabMessageDto labMessage
...
-					callback.run();
-				}
-			} else {
-				callback.run();
-			}
-			finishProcessingLabMessage(labMessageDto, pathogenTestDto);
-		};
+		// add newly submitted tests to sample edit component
+		List<String> existingTestExternalIds = existingTests.stream().map(t -> t.getExternalId()).collect(Collectors.toList());
+		while (existingTestExternalIds.remove(null)); // otherwise tests without external id will seen as match
```
leventegal-she, line 518: you could filter the stream instead of this one liner that is hard to understand

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
```diff
@@ -693,118 +655,79 @@ private FacilityReferenceDto getLabReference(LabMessageDto labMessageDto) {
...
-			((DateField) sampleCreateComponent.getWrappedComponent().getField(PathogenTestDto.REPORT_DATE))
-				.setValue(labMessageDto.getMessageDateTime());
+		sampleCreateComponent.addDoneListener(() -> finishProcessingLabMessage(labMessageDto, sample));
+
+		// add pathogen test create components
+		//********************
+		List<PathogenTestDto> pathogenTests = buildPathogenTests(sample, labMessageDto);
+		int caseSampleCount = FacadeProvider.getSampleFacade().caseSampleCountOf(sample);
+		// the first pathogenTestCreateComponent must not be removable. It is expected to exist because the buildPathogenTests shall always return at least one.
+		PathogenTestForm pathogenTestCreateComponent =
```
leventegal-she, line 673: It would be better to don't let to delete the last remaining pathogen test then forcing the first one to be not deletable. What if the user would like to keep the third one only?

leventegal-she (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
FredrikSchaeferVitagroup: With the current approach, the user would need to copy the values from the third pathogen test to the first one. There are not too many fields... I'm open to both approaches, but I think the one proposed by you also has ownsides.
- it is a little more complex and puts a little more load on the UI. 
- the remove pathogen test button would show for all attached pathogen test forms (when there are more than one). That may suggest to some users that all pathogen tests are removable. 

Considering this, do you still prefer your proposal?

leventegal-she: Yes, I prefer my proposal because it's a better UX

sormas-backend/src/main/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjb.java
```diff
@@ -179,13 +179,31 @@ public LabMessageDto toDto(LabMessage source) {
...
+		}
 
 		return target;
 	}
 
 	@Override
-	public LabMessageDto getByUuid(String uuid) {
-		return toDto(labMessageService.getByUuid(uuid));
+	public LabMessageDto getByUuid(String uuid, boolean getIfDeleted) {
+		if (getIfDeleted) {
```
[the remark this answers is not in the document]
FredrikSchaeferVitagroup: Okay, so I shall remove the getIfDeleted parameter again. Do you want the LabMessageFacadeEjb.getByUuid(...) to return deleted lab messages? If so, I'm a little worried that developers will assume a different behaviour or completely overlook that lab messages are a CoreADO.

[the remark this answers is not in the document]
leventegal-she: Yes, I would say to remove it. It will just work the same as all other CoreADO entities so it would be consistent with other facades.

Maybe it would be better to read the labmessage by uuid and throw an exception if it is marked as deleted so the developers would be forced to handle it so it wouldn't happen that developers "overlook that lab messages are a CoreADO" and also there won't be NullpointerExceptions because of deleted entries.

If you want to keep this parameter I would suggest to refactor a but the method and kepp only the critarieFilter approach but add the "not deleted" filter based on the that, this would it won't look like 2 methonds in one. But anyway the return value of the method should be checked against null because excluding deleted entries increases the possibility of null pointer exceptions.

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
FredrikSchaeferVitagroup: The lifecycle scenarios of the new profiles allow lab messages to only correct data. Processing this should not require a pathogen test to exist for a related sample. 
The restraint to always keep at least one pathogen test was therefore removed by me completely.

leventegal-she (changes requested):
Can you please resolve the conflict in sormas_schema.sql so I can do test in my local setup?

FredrikSchaeferVitagroup:
> Can you please resolve the conflict in sormas_schema.sql so I can do test in my local setup?

Sure, done.

leventegal-she (changes requested):
- please keep the `INSERT INTO schema_version ` statement on a single line otherwise the migration doesn't run
- `pathogenTest` field of `LabMessage` should be deleted because it causes sql error
- clicking Discard on the sample creation step changes the labmessage status to PROCESED
- when clicking save the invalid fields are not marked with red border in the pathogen test sections, having multiple test it is hard to find which fields are invalid
- When editing a sample there is no possibility to add pathogen tests, the "Add new test" button is missing but there are other buttons that I'm not sure should be there
Screenshot 2021-11-05 at 12 10 02

- after editing a sample and save the labmessage doesn't change to PROCESSED
- the tests added by the user has no `Via LIMS` checked by default, I think it should
- I miss a label between sample section and the pathogen tests section, it's not really clear where the sample form ands and the pathogen test starts
- the remove button is misaligned

Maybe for a followup ticket for improving the UI:
- When editing an existing sample it's not clear for the user which tests are the new ones coming from the labmessage, maybe test sections collapsible and collapse existing tests by default
- When editing a sample a better distinction between existing tests and the ones to be created form the labmessage would be needed
- A better distinction between test sections would be needed

FredrikSchaeferVitagroup:
> * please keep the `INSERT INTO schema_version ` statement on a single line otherwise the migration doesn't run

Okay

> * `pathogenTest` field of `LabMessage` should be deleted because it causes sql error

Done

> * clicking Discard on the sample creation step changes the labmessage status to PROCESED

Should be fixed

> * when clicking save the invalid fields are not marked with red border in the pathogen test sections, having multiple test it is hard to find which fields are invalid

Should be fixed

> * When editing a sample there is no possibility to add pathogen tests, the "Add new test" button is missing but there are other buttons that I'm not sure should be there
> 
> Screenshot 2021-11-05 at 12 10 02

It should be possible to add pathogen tests now. I guess, the other buttons can remain. Lab messages can soon also transmit corrections of data.
> 
> * after editing a sample and save the labmessage doesn't change to PROCESSED

Should be fixed

> * the tests added by the user has no `Via LIMS` checked by default, I think it should

Done. I have also removed that the viaLims field gets non editable. I think this fits the higher flexibility necessary for the more complex processing here.


> * I miss a label between sample section and the pathogen tests section, it's not really clear where the sample form ands and the pathogen test starts

As specified in the issue, there is a thin line separator between sample and pathogen test. It's the same as the one between pathogen tests. I think the view can not easily be divided into a sample and a pathogen test section, because the buttons shown on the bottom of the commitDiscardWrapperComponent (Refer to another laboratory, DELETE) don't relate to the pathogen tests.
Therefore I would consider this request out of scope. Or do you have an easy solution?

> * the remove button is misaligned

That was actually done be me intentionally. This button is very different from the fields above and below. First, it's a button instead of a field, second it does not add information to a pathogen test. That's why I think it has to stand out a little.

> 
> Maybe for a followup ticket for improving the UI:
> 
> * When editing an existing sample it's not clear for the user which tests are the new ones coming from the labmessage, maybe test sections collapsible and collapse existing tests by default
> * When editing a sample a better distinction between existing tests and the ones to be created form the labmessage would be needed
> * A better distinction between test sections would be needed

I would like to wait for user feedback before creating such a ticket. We do not yet know whether the whole way of processing the lab messages with new profile will really be accepted by the users, so it may change a lot before we get to improve what you mentioned.

leventegal-she (changes requested):
> * `pathogenTest` field of `LabMessage` should be deleted because it causes sql error

Still not removed

leventegal-she (changes requested):
I'm not sure the red delete button is needed when editing sample with tests. I did the following:
- clicked delete and confirmed the deletion, the only thing happened is that the while page to shifted to the right around half a display size, was able to scroll to right to see the whole form but opening combo boxes scrolled back to the left; then if I try to save a notification is displayed
image
image
image

- same thing happens with the "Refer to another lab" button
> I'm not sure these buttons are useful during the lab message processing, it mostly confuses the users

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -316,7 +320,7 @@ public void onDiscard() {
 			}
 		});
 
-		if (UserProvider.getCurrent().hasUserRight(UserRight.SAMPLE_DELETE)) {
+		if (!hideReferAndDeleteButton && UserProvider.getCurrent().hasUserRight(UserRight.SAMPLE_DELETE)) {
```
leventegal-she, line 323: would be better to have all the code below inside an if statement instead of adding the same condition in multiple places:

```
if (!hideReferAndDeleteButton) {
....
}
```

Also I would prefer to use a variable name that must not be negated in the condition

leventegal-she (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
FredrikSchaeferVitagroup: This is supposed to be temporary, so I wanted to change as little as possible.
Also I would argue against a nested if statement, because it is easier to overlook in this rather large method with rather large sections.
Do you insist on this change?

FredrikSchaeferVitagroup:
> > * `pathogenTest` field of `LabMessage` should be deleted because it causes sql error
> 
> Still not removed

After i understood that this is about the `TestReport.pathogenTest`, this is now removed.

FredrikSchaeferVitagroup:
> I'm not sure the red delete button is needed when editing sample with tests. I did the following:
> 
> * clicked delete and confirmed the deletion, the only thing happened is that the while page to shifted to the right around half a display size, was able to scroll to right to see the whole form but opening combo boxes scrolled back to the left; then if I try to save a notification is displayed
>   image
>   image
>   image
> * same thing happens with the "Refer to another lab" button
> 
> > I'm not sure these buttons are useful during the lab message processing, it mostly confuses the users

As we discussed, they are probably necessary, but out of scope for this ticket. I will create a follow up ticket.

leventegal-she (approved):

PR #7109, opened 2021-11-01
#7068 #6989 case external data

### LINKED ISSUE #6989: Display all data that is not mappable from SurvNet transport file in SORMAS [2]
### Situation Description
In SurvNet there are different schema versions with different fields and corresponding values. 
There is the possibility of importing cases, which are created with the schema version 7.0
In the schema version 7 there are fields, that are no longer available in SORMAS - as SORMAS is currentyl mapping schema 8.

Depends on #7068 (backend implementation).

### Feature Description
With this issue we are aiming for a new section that shows also those fields and values which are not mapped in SORMAS.
This approach was already implemented when we imported the travel entries from DEA into SORMAS. 

In the corresponding entities we should add a tab with additional information when there are any, like in the cases details view there would be another tab next to - clinical course | contacts | "additional information"

When we take a look at the current schema publication from the RKI we can see that the fields:

- StandardAdditionalAttribute
- StatusPatientSetting
- GuidPatientSetting
- PatientSettingDetail
- P112InfCol
- StatusContact
- StatusExpo
- Expo0001
- Expo0002
- Expo0003
- Expo0005

were available in Schema 7 and no longer available in schema 8.
In this case, when importing one of the cases with this information, we should display "additional info" and the emlpoyee of the health department can access this information in readonly mode.

It's worth taking a look at the implementation where we imported DEA travel entries and display not mappable fields and values in a SORMAS form in the UI.

### LINKED ISSUE #7068: Allow to store external data for a case [1]
### Situation Description
As described in #6989 we need to be able to display additional data provided by SurvNet and other external reporting systems.

### Feature Description
In order to display the information we need to be able to store them and push them via ReST first.

- [ ] Add externalData field to the case and dto - see below for details
- [ ] <del>Consider validation of the data - e.g. white listing allowed characters</del>
- [ ] When an existing case is updated via ReST (e.g. sync with mobile app) and does not include externalData (null instead of empty map), don't override existing data. This is needed so we don't have to implement the externalData field in the mobile app for now
- [ ] <del>Add unit test to store externalData</del>

#### externalData field
This is similar to what we are doing for the TravelEntry.deaContent, but I'd suggest to use a Map<String,String> (see https://vladmihalcea.com/java-map-json-jpa-hibernate/) as data type so that the stored JSON looks like this:

```
{
  "StatusPatientSetting": "...",
  "PatientSettingDetail": "..."
}
```

Instead of:

```
[{ 
  "caption": "StatusPatientSetting",
  "value": "..."
},
{ 
  "caption": "PatientSettingDetail",
  "value": "..."
}]
```

leventegal-she:
Fixes #7068
Fixes #6989


sormas-rest/src/main/java/de/symeda/sormas/rest/EntityDtoResource.java
```diff
@@ -32,6 +31,7 @@
 				dto = transactionWrapper.execute(saveEntityDto, dto);
 				result = PushResult.OK;
 			} catch (Exception e) {
+				e.printStackTrace();
```
vidi42, line 34:
```suggestion

```

vidi42 (changes requested):

vidi42 (approved):

PR #7118, opened 2021-11-02
#6867 add not null constraint to user jurisdiction level and make sur…

### LINKED ISSUE #6867: Assigned to user list of task should consider related entities jurisdiction
### Situation Description
Assigned to user list of task depends on some old logic based on district and region, this should be updated that it considers the jurisdiction of the related entity (case / contact / event).
Beside doing this on the UI this should also be validated upon save in the backend.
See bug report for cases : #6858

MartinWahnschaffe:
…e it's always updated

Fixes #6867


JonasCir (approved):
lgtm thanks!

PR #7138, opened 2021-11-03
Change/7136 s2s validation classes

### LINKED ISSUE #7136: Create S2S validator classes and move validation logic there
### Situation description
Right, now validation and accessing the structure of S2S Dtos is coupled and only done in incoming data. To enable independent checks of incoming and outgoing data, we need come to a point where we access a certain field of a dto and pass a predicate to validate the accessed field. By this we can realize outgoing checks more easily. 

### Propose change
Introduce a generic class and subclasses extending it for each S2S shareable DTO.
Proposed methods are:
1. `walkDto` functions to walk the given Dto and applies validation functions to it
2. `validateIncoming` to validate incoming Dtos (as we do now)
3. `validateOutgoing` to validate Dtos before sending

JonasCir:
Fixes #7136 
This is an important step to enable outgoing validation in the future. The mid-term idea is to separate the the accessing of infrastructure fields in a dto from the validation (i.e., the the access function for a field also takes a function which checks the value).

In the future, we therefore can reuse the access functions for both outgoing and incoming validation, but for this, we need to establish dedicated places in the code where this can happen :) 

---
@vidi42 Not certain if `SormasToSormasDtoValidator` is the best naming here, feel free to propose something else as you see fit :rocket:


JonasCir:
@vidi42 Question: As you can see, the validation functions currently all call into the data validator which I want/need to get rid of. Should I copy over the code to the call site in this PR or should I do it in a followup.

vidi42:
> @vidi42 Question: As you can see, the validation functions currently all call into the data validator which I want/need to get rid of. Should I copy over the code to the call site in this PR or should I do it in a followup.

I think it makes sense to do it in this PR so we can test the changes together

vidi42 (changes requested):
Looks good to me, but I suggest moving entity specific validation logic into the new validators and only leave the common part into the SormasToSormasDtoValidator or some other common place.

JonasCir:
Great idea will do it now...

JonasCir:
I really like the approach now :)

vidi42 (changes requested):

vidi42 (changes requested):

vidi42 (approved):

PR #7173, opened 2021-11-05
#7111 - Added properties to feature configurations

### LINKED ISSUE #7111: Add properties to feature configurations [2]
### Situation Description

In the course of #7105 we need to extend the feature configurations with additional properties that differ from configuration to configuration. Explicitely, the `IMMUNIZATIONS` feature type needs a `reduced` property that toggles the detail level of the immunization module.

### Feature Description

* Add a `properties` field to the `FeatureConfiguration` class (similar to the properties of CustomizableEnumValue)
* Add a `FeatureTypeProperty` enum that contains all supported properties of feature types
* Add a `properties` string array to the `FeatureType` class and add a varargs constructor parameter to fill it; add "reduced" as a parameter to the `IMMUNIZATIONS` definition
* Add a `getPropertyValue` method to `FeatureConfigurationFacadeEjb` that returns the value of a property of a given feature type after checking whether that feature type supports the given property; if it doesn't, throw an error

MateStrysewske:
Fixes #7111


MateStrysewske:
@StefanKock @MartinWahnschaffe Please have a look and see whether this approach is fine.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfiguration.java
```diff
@@ -101,4 +108,14 @@ public boolean isEnabled() {
 	public void setEnabled(boolean enabled) {
 		this.enabled = enabled;
 	}
+
+	@Type(type = ModelConstants.HIBERNATE_TYPE_JSON)
+	@Column(columnDefinition = ModelConstants.COLUMN_DEFINITION_JSON)
+	public Map<FeatureTypeProperty, Object> getProperties() {
```
StefanKock, line 114: How are Maps working in the auditlog?

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
```diff
@@ -254,6 +257,35 @@ public boolean isFeatureDisabled(FeatureType featureType) {
...
+		CriteriaBuilder cb = em.getCriteriaBuilder();
+		CriteriaQuery<Object> cq = cb.createQuery(Object.class);
+		Root<FeatureConfiguration> root = cq.from(FeatureConfiguration.class);
+
+		cq.where(cb.and(cb.equal(root.get(FeatureConfiguration.FEATURE_TYPE), featureType)));
+		cq.select(root.get(FeatureConfiguration.PROPERTIES));
+
+		Map<FeatureTypeProperty, Object> properties = (Map<FeatureTypeProperty, Object>) em.createQuery(cq).getSingleResult();
+
+		if (properties != null && properties.containsKey(property)) {
```
StefanKock, line 282: Initialize only one return value, suppress warning locally on the line that causes it, not for the whole method.
Before if-else:
```
final boolean result;
```

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
```diff
@@ -254,6 +257,35 @@ public boolean isFeatureDisabled(FeatureType featureType) {
...
+			return (boolean) properties.get(property) == expectedPropertyValue;
```
StefanKock, line 283: Properly sanitize boolean value, check `ConfigFacadeEjb.getBoolean`. It also shows how to deal with the `expectedPropertyValue`.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
```diff
@@ -254,6 +257,35 @@ public boolean isFeatureDisabled(FeatureType featureType) {
...
+		} else {
+			return (boolean) property.getDefaultValue() == expectedPropertyValue;
```
StefanKock, line 285: But what's the `property.getDefaultValue()` for?
Caution: `false == false` results in `true`, but is not what I would have expected!

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
```diff
@@ -254,6 +257,35 @@ public boolean isFeatureDisabled(FeatureType featureType) {
 		return em.createQuery(cq).getSingleResult() > 0;
 	}
 
+	@Override
+	@SuppressWarnings("unchecked")
+	public boolean isPropertyValue(FeatureType featureType, FeatureTypeProperty property, boolean expectedPropertyValue) {
```
StefanKock, line 262: 1. I would name this method `isPropertyEnabled`
2. Missing unit test
3. Also define `getPropertyValue:String` or `getPropertyValue(..., Class<P> propertyClass):P` or wait for that use cases (4. bullet point of the ticket)? That would be needed in my opinion to completly deliver the feature (not only for `boolean`).

StefanKock:
> If you want to test it, you'll have to enter the following into the properties column in your database:
> 
> `{"REDUCED": true}`
> 
> The reason is that I directly map the database content to the new enum, but that also means that the mapping fails when writing reduced in lowercase.

Wasn't the idea to define the value already in the database, to define what exists and to easily reconfigure it? I don't see documentation which `FeatureTypeProperty`s are possible for which `FeatureType`. I would expect that also checked in somewhere.

StefanKock (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
MateStrysewske: I'm not exactly sure how you want the code to look like; something like `Boolean.parseBoolean` isn't possible because we're working with an Object, not a String. Catching a ClassCastException here and doing something else with it also isn't a good idea in my opinion because we should rather let the error happen here so that users are aware that something's wrong. Can you add a specific code example here please?

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
MateStrysewske: See above. `isPropertyEnabled` would not be an appropriate name for what the method is supposed to do. I'll add Javadoc. `getPropertyValue` is no longer needed because I replaced it with the `isPropertyValue` method which, in my opinion, is more valuable for this use case right here. We can add `getPropertyValue` later when we need it, but right now it would be dead code.

Unit test has been added, but it only covers the default values. Adding explicit JSON values to the database leads to a data conversion error because H2 has issues with the JSON converting that we use for PostgreSQL. This should be tackled separately from this issue.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
MateStrysewske: `property.getDefaultValue()` is supposed to return the value we want this property to have if it's not specified in the database. And `false == false` returning in `true` is exactly what is desired because the method is supposed to check whether the actual value matches the given `expectedPropertyValue`. I'll add Javadoc to the method to describe what it's supposed to do.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
MateStrysewske: Moved the annotation; can you please explain the advantage of initialising a result variable? It doesn't really make the code more readable in my opinion.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfiguration.java
MateStrysewske: I don't know, but the only way to change these values is to directly do it in the database anyway. If I'm not totally mistaken, there won't ever be audit log entries created as long as we don't provide a user interface for it.

MartinWahnschaffe: I agree. If we don't check this now, we will likely forget about it later.

sormas-api/src/main/java/de/symeda/sormas/api/feature/FeatureTypeProperty.java
```diff
@@ -0,0 +1,37 @@
...
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package de.symeda.sormas.api.feature;
+
+public enum FeatureTypeProperty {
+
+	REDUCED(Boolean.class, false);
```
MartinWahnschaffe, line 20: Default value should be defined for combination of FeatureType and FeatureTypeProperty instead.

sormas-api/src/main/java/de/symeda/sormas/api/feature/FeatureType.java
```diff
@@ -45,7 +45,7 @@
 			SAMPLES_LAB }),
 	TASK_MANAGEMENT(true, true, null),
 	WEEKLY_REPORTING(true, true, null),
-	IMMUNIZATION_MANAGEMENT(true, true, null),
+	IMMUNIZATION_MANAGEMENT(true, true, null, FeatureTypeProperty.REDUCED),
```
MartinWahnschaffe, line 48: Please extend FeatureService.createMissingFeatureConfigurations to also persist the default feature type properties. Otherwise it would be inconsistent (creating entries with the default enabled values of feature configurations in the database, but not including the default values for feature type properties).

MartinWahnschaffe (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfiguration.java
MateStrysewske: There's currently no way to check it specifically for this use case, but I can confirm that the epidemiological evidence criteria of events, which is a map annotated with the same annotations, is logged as well.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
StefanKock: I think the approach that the caller needs to provide the `expectedPropertyValue` that basically only inverts the boolean argument is superfluous and too complicated.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
StefanKock: I assume that casting whatever comes from the JSON to boolean (`null` will result in a NPE!) is probably okay then.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
StefanKock: Not being able to configure this in a unit test might result in other features not properly tested then. But it seems there is a solution: Update _hibernate-types-52_ (see developer comment here: https://stackoverflow.com/questions/39620317/how-to-map-a-json-column-with-h2-jpa-and-hibernate/67467069#67467069)

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
StefanKock: It's easier to debug, review and appendable when you don't have several `return` statements.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
MateStrysewske: I disagree. Moving the comparison logic to the backend instead of executing it in the UI in this scenario even makes the call in the UI less complicated in my opinion. If we already know which value we expect, why should we have to make the comparison in the UI instead of doing it directly in the backend, i.e. moving as much logic as possible there?

StefanKock: I think having a parameter less and inverting it with `!` is the usual approach, better to read and less to type.
- `if(FeatureConfigurationFacade.isPropertyValue(FeatureType.IMMUNIZATION, FeatureTypeProperty.REDUCED)` ~ `if(FeatureConfigurationFacade.isPropertyValue(FeatureType.IMMUNIZATION, FeatureTypeProperty.REDUCED, true)`
- `if(!FeatureConfigurationFacade.isPropertyValue(FeatureType.IMMUNIZATION, FeatureTypeProperty.REDUCED)` ~ `if(FeatureConfigurationFacade.isPropertyValue(FeatureType.IMMUNIZATION, FeatureTypeProperty.REDUCED, false)`

This pattern gets even more complicated if would be about something like a String (comparison doesn't match completely, I know).

MateStrysewske:
@StefanKock Concerning your initial comment, I will extend the Wiki guide on feature configurations accordingly when 1.66 is released.

sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationFacadeEjb.java
MateStrysewske: Done, and also renamed the method to `isPropertyValueTrue` because the original name doesn't make much sense with this changed signature anymore.

MateStrysewske:
@MartinWahnschaffe @StefanKock Please have another look.

MartinWahnschaffe (approved):

StefanKock (commented):
Like `isAnySurveillanceEnabled` you could name the new method `isPropertyEnabled` instead of `isPropertyValueTrue`, but that's only a bit cosmetic.

There was a remark if we don't want to predefine REDUCED = false via SQL. If that's out of scope/intent, then it's also okay (but might be inconvenient for developers and operations to configure the system).

StefanKock (approved):

PR #7210, opened 2021-11-09
6939 change facadeprovider api

### LINKED ISSUE #6939: Don't use FacadeProvider in sormas-api [2]
### Feature Description
The FacadeProvider is a convenience class that provides facade for usage in web applications (sormas-ui and sormas-rest) without having to inject them.

### Problem Description
We are currently also using the FacadeProvider at some places in the sormas-api. This can lead to problems with the class loader and may be the cause for our re-deploment issues #2511 #2251.
I don't have the details on why this is problematic at hand. Maybe @HolgerReiseVSys can point this out.

### Proposed Change
Don't use the FacadeProvider in sormas-api. Instead provide the needed facade or where applicable more detailed objects as method parameter.

- [x] AuthProvider
- [x] ClassificationHtmlRenderer
- [x] ImportExportUtils
- [x] ValueSeparator
- [x] StatisticsHelper

In addition we should make sure that the FacadeProvider is not used that way. 

- [x] Use ArchUnit (https://www.archunit.org/use-cases) to introduce an automatic check that forbids the usage of FacadeProvider in sormas-api and sormas-backend

MartinWahnschaffe:
Fixes #6939

@StefanKock Unfortunately it doesn't seem to fix the re-deployment issue related to the inject of the properties into ConfigFacade.
Restarting the server without having to undeploy first is working for me know, but this seems to be related to something else - at least it also works on the current dev branch.


sormas-base/pom.xml
```diff
@@ -845,6 +845,21 @@
 				<version>2.27.2</version>
 				<scope>test</scope>
 			</dependency>
+
+			<dependency>
+				<groupId>com.tngtech.archunit</groupId>
+				<artifactId>archunit</artifactId>
+				<version>0.22.0</version>
```
StefanKock, line 852: Unify as property `<archunit.version>0.22.0</archunit.version>`

sormas-base/pom.xml
```diff
@@ -845,6 +845,21 @@
...
+				<scope>test</scope>
+			</dependency>
+
```
StefanKock, line 855: Remove empty line (block of dependencies belonging together)

StefanKock (changes requested):
I reviewed the code, only two trivial findings.

PR #7236, opened 2021-11-12
Feature 5729 demis2 sormas handle new profile shortcuts

### LINKED ISSUE #5729: [DEMIS2SORMAS] Handle New Profile: Shortcuts [3]
### Situation Description

Preriquisite: #5622, #5713, #5539

The new DEMIS profiles introduce lifecycle management, so that demis notifications can relate to each other. This means also lab messages need to relate to each other (foundation for that is #5622). When lab messages get processed, entities like samples and pathogen tests are created out of that. When a lab messages gets processed that relates to a previously processed lab message, it can mostly be related to the previously created entities. 
As a user, when processing such a lab messages, I want to automatically be proposed to jump to those related entities, if there are any. This reduces the risk of errorneous processing and the number of clicks necessary. 

### Feature Description

For the demis lifecycle management, there are five scenarios (+sub scenarios) defined (see #5138). It it not clear whether they will in practice be mixed up or not. Anyway, they should boil down to three kinds of shortcuts in SORMAS:

1. There is a procesed lab message with the same reportId and a sample with the same labSampleId as the in the lab message. The sample and the processed lab message are related. There are no matching pathogen tests (a matching pathogen test has the same externalId as one of the (non processed) lab messages test report). This is the case for scenario 2 (all sub scenarios), 3b, and 3c. In this case, when the process button for a lab message is clicked, a popup should show asking `A lab message with the same report ID was already processed. It is related to a sample also referenced in this lab message. Do you want to directly open that sample?`. It should be possible to either chose `YES`, `NO` or cancel the processing. If chosen `YES`, a sample form should open with all the existing and new pathogen tests attached (a form very similar to the one specified in #5899 ). If chose `NO`, the processing should continue as if no already processed lab message with the same reportId and no matching sample was found.
2. Identical to 1., but there is at least one matching pathogen test found. In this case, popup question should be `A lab message with the same report ID was already processed. It is related to one or more pathogen tests also referenced in this lab message. Do you want to directly open those pathogen tests?`. This shortcut applies in scenario 4.
3. There is no related processed lab message, but a sample with the same lab sample id as in the lab message. There are no matching pathogen tests (a matching pathogen test has the same externalId as one of the (non processed) lab messages test report). In this case the popup question should be `The sample referenced in this lab message already exists. Do you want to directly open that sample?`. This shortcut applies in scenario 2b, 3a and 3b (if second test is positive).

The correction of data, as specified in #5714, should always be done first, before taking one of those shortcuts.

### Open questions

- When lab messages get completely deleted, some shortcuts may get lost. Should lab messages instead be archieved? This would allow another shortcut for messages forwarded via S2S. **ANSWER**: lab message should not get completely deleted anymore (but archived instead). When a user starts processing a lab message related to another lab message marked as forwarded (either manually or via S2S, a popup shall inform the user about this. `Another related lab message was marked as forwarded/manually forwarded. Do you want to continue processing this lab message?` Answer options should be `YES` and `CANCEL`
- When demis notifications don't stick to the definition of lifecycle scenarios or lab messages get deleted, there are other shortcuts possible. For example one like 3., but with matching pathogen tests. Should they also be taken into consideration? Answer: No.

### Possible Alternatives

Suggestions welcome!

### Additional Information

Out of scope for this ticket: https://github.com/hzi-braunschweig/SORMAS-Project/issues/5729#issuecomment-860528432

FredrikSchaeferVitagroup:
Fixes #5729


sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
```diff
@@ -145,49 +146,71 @@ public void showLabMessagesSlider(List<LabMessageDto> labMessages) {
...
+								&& LabMessageStatus.PROCESSED.equals(otherLabMessage.getStatus()))
+						.collect(Collectors.toList());
 				}
-			}, false);
+
+				// propose to take shortcut
+				String message = (relatedLabMessages != null && !relatedLabMessages.isEmpty())
+					? I18nProperties.getString(Strings.messageRelatedSampleAndLabMessagesFound)
+					: I18nProperties.getString(Strings.messageRelatedSampleFound);
+				VaadinUiUtil.showChooseOptionPopup(
```
leventegal-she, line 196: There are the same options for both with and without related lab messages, does it make sense to get the related lab messages from the backend just for showing a slightly different message?

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
```diff
@@ -233,6 +256,51 @@ private PersonDto buildPerson(LabMessageDto labMessageDto) {
 		return personDto;
 	}
 
+	private void pickOrCreatePerson(LabMessageDto labMessage) {
+		final PersonDto personDto = buildPerson(labMessage);
+
+		if (FacadeProvider.getLabMessageFacade().isProcessed(labMessage.getUuid())) {
```
leventegal-she, line 262: this check should be done before shortcuts too

leventegal-she (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
FredrikSchaeferVitagroup: I wondered about that, too. My conclusion was that it is necessary to show different messages here. Users might not want to directly jump to that sample when no related processed lab message could be found. I the sample edit  form, they don't have a way of seeing the related person data.

I also thought about creating some kind of extended LabMessageReferenceDto containing the reportId and the labMessageStatus, but discarded this because it requires quite some code for a small gain.

Are you okay with leaving it like that?

leventegal-she: It will probably be needed for correction too so it can be kept

leventegal-she (approved):

PR #7264, opened 2021-11-15
#7208 Update `central.location.sync` property description

### LINKED ISSUE #7208: Improve description of s2s property `central.location.sync`
### Problem Description
A new property was introduced with 1.65 which is described poorly
```
# Enable location data sync with central
#central.location.sync=
```

### Proposed Change
Update the description to include more information about the supported values and the default value and also a better description.
```
# If set to true, all infrastructure data from the central location will be synchronized into the local SORMAS database at startup and on a nightly basis.
# Default: false
#central.location.sync=
```

vidi42:
Fixes #7208

### DIFF
--- a/sormas-base/setup/sormas.properties
+++ b/sormas-base/setup/sormas.properties
@@ -269,7 +269,9 @@ app.url=
 # The path to the CA cert trusted by etcd clients.
 #central.etcd.caPath=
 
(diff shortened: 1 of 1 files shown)


JonasCir (approved):
lgtm thanks

PR #7280, opened 2021-11-16
#5144 Fix sormas rest deploy inconsistency

### LINKED ISSUE #5144: Deploying `sormas-rest` before `sormas-ear` results in issues with the API communication
### Problem Description
Since the introduction of Keycloak, the `sormas-rest` has to define a [KeycloakFilter ](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-rest/src/main/java/de/symeda/sormas/rest/security/KeycloakFilter.java) which handles the Authorization check of API calls.

At the moment SORMAS supports 2 Authentication Mechanisms:
* Legacy - built directly into SORMAS
* Keycloak - introduced with #2111

Only one mechanism can be active at a time which is done by configuring into `sormas.properties` the `authentication.provider` property.

The issue is that for `sormas-rest` to access this property it has to call at startup the ConfigFacade EJB which is part of `sormas-ear`, and in the situation that the `sormas-ear` is not yet deployed, it will fail and the `MultiAuthenticationMechanism` will not be correctly setup and some or all API calls with fail with 403.

### Steps to Reproduce
1. Undeploy all applications from the server
2. Deploy `sormas-rest` first
3. Deploy `sormas-ear` second
4. Try to access any API endpoint with an existing user

### Expected Behavior
Expected behavior would be that in the situation that `sormas-rest` cannot read the `authentication.provider` from the ConfigFacade it should either fail at deploy or default to the Legacy Authentication Provider.

vidi42:
With this change the `sormas-rest` artifact cannot be deployed without having the `sormas-ear` deployed first.
I chose this instead of falling back to the legacy login because that might introduce more uncertainty that simply failing with an error message.

Fixes #5144

### DIFF
--- a/sormas-rest/src/main/java/de/symeda/sormas/rest/RestConfig.java
+++ b/sormas-rest/src/main/java/de/symeda/sormas/rest/RestConfig.java
@@ -1,26 +1,32 @@
-/*******************************************************************************
+/*
  * SORMAS® - Surveillance Outbreak Response Management & Analysis System
- * Copyright © 2016-2018 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
- *
+ * Copyright © 2016-2021 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


JonasCir (approved):
lgtm thanks.

Issue #5714, opened 2021-06-08
[DEMIS2SORMAS] Handle New Profile: Automatically propose to correct entities related to a reportId [5]


FredrikSchaeferVitagroup:
### Situation Description

With the new DEMIS profiles, it becomes possible to send correction of data previously sent via DEMIS notifications. As a user, I want to be able to process this correction in SORMAS. I want to do this with as few clicks as possible while always seeing the relevant data for making decisions about which values are the correct ones to save and how to proceed processing lab messages.

### Feature Description

Preriquisite: #5622 , #5539, #5713

When a user starts processing a lab message, SORMAS should check if there are already processed lab messages with the same reportId. If there are any, SORMAS should (1.) determine the related person, samples and pathogenTests and (2.) try to closer relate them to the lab message currently to be processed. 
About (1.): Pathogen tests associated to the reportId are stored in processed lab messages. Samples related can be found through those pathogen tests. Persons can be found via the cases /contacts /event participants related to those samples. There should always just be one person related to one reportId. 
To (2.) closer relate one sample to the currently to be processed lab message, SORMAS should check if one of the found sample's labSampleId is identical to the labMessage's labSampleId. This would indicate a close relation. To closer relate pathogen tests to the lab message, SORMAS should compare the pathogen tests externalId to the test results externalId.

When entities with a close relation according to (2.) are found, SORMAS should check whether there are differences between values in the entities and values in the lab message. This would indicate that there are corrections transmitted via DEMIS.

When there are differences found, show a panel with two forms for the changed entity, side by side. The left one with the caption "Previous [Entity] Information" (e.g. Previous Person Information) and all fields disabled. None of the fields should be highlighted, they should all have the same frame color.
The right one with the caption "Updated [Entity] Information" (e.g. Updated Person Information) and all fields editable. None of the fields should be highlighted (with darker blue frame), except the ones changed. When the user edits a fields, it should become highlighted. When the user reverts changes in a field, it should become non-highlighted.

![Entity_correction](https://user-images.githubusercontent.com/67001822/121135976-61872200-c835-11eb-8ac9-9227c0d1dbca.png)

At the bottom of the panel, there should be 4 buttons: `CANCEL`, `DISCARD ALL AND CONTINUE`,  `SAVE AND CONTINUE` and `VIEW LAB MESSAGE` (use the info icon as in other places).
[...]

PR #7316, opened 2021-11-18
Feature 7204 demis2sormas handle new profile: allow to create sample referral when processing lab messages

### LINKED ISSUE #7204: [DEMIS2SORMAS] Handle New Profile: Allow to create sample referral when processing lab messages [2]
### Situation Description

Imagine the following scenario: A lab 1 receives a sample. It performs a preliminary test and the sends the sample to another lab 2, which performs additional tests.

In SORMAS, a user would first create a sample for lab 1. When the sample gets sent to lab 2, the user would open the sample, and create a referral for it. This would create another sample, which will be related to the first sample created. Both samples are similar, but the facilities differ -> The same physical sample has two sample entities in SORMAS.

### Feature Description

The scenario described above can also be communicated via DEMIS, once the new profiles are used.
When a user edits a sample while processing a lab message, it must be possible use the functionality of the `referOrLinkToOtherLabButton` or the [SampleController](https://github.com/hzi-braunschweig/SORMAS-Project/blob/fc8d3c07da4ccace1f6ed0c569357ba6d8ff3fac/sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java#L300).

### Possible Alternatives

Another option might be to refactor the sample entity, so that it can have multiple facilities. This would probable require some more fields to exist multiple times, e.q. the `fieldSampleId` which might be different for each laboratory.

FredrikSchaeferVitagroup:
Fixes #7204


FredrikSchaeferVitagroup:
@leventegal-she As you seem have some other tickets to deal with, I did not assign you as a reviewer here. Maybe it would still be good if you had a look at those changes to see if they somehow influnence your work, e.g.  for #5714.

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleDto.java
```diff
@@ -481,19 +481,26 @@ public static SampleDto buildReferral(UserReferenceDto userRef, SampleDto referr
...
-		sample.setAdditionalTestingRequested(referredSample.getAdditionalTestingRequested());
-		sample.setRequestedPathogenTests(referredSample.getRequestedPathogenTests());
-		sample.setRequestedAdditionalTests(referredSample.getRequestedAdditionalTests());
-		sample.setPathogenTestResult(PathogenTestResultType.PENDING);
+		migrateAttributesOfPhysicalSample(referredSample, sample);
 
 		return sample;
 	}
 
+	public static void migrateAttributesOfPhysicalSample(SampleDto source, SampleDto target) {
```
leventegal-she, line 489: I don't really see a reason to have this function extracted from the above one. At least is must not be public and/or maybe a better name, I don't really see the meaning of physicalSample on the context of source and target, which one is the physical one and why is it physical. Maybe `migrateAttributesOfReferredSample(source, referred)` would be more meaningful.

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
```diff
@@ -622,13 +627,118 @@ private void editSample(SampleDto sample, LabMessageDto labMessage) {
...
+		addReferredFromButton(sampleEditComponent, labMessage, window);
+
 		sampleEditComponent.addCommitListener(window::close);
 		sampleEditComponent.addDiscardListener(window::close);
 
 		showFormWithLabMessage(labMessage, sampleEditComponent, window, I18nProperties.getString(Strings.headingEditSample), false);
 
 	}
 
+	private void addReferOrLinkToOtherLabButton(
```
leventegal-she, line 642: seems to be a duplicate of `SampleController::addReferOrLinkToOtherLabButton`

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
```diff
@@ -622,13 +627,118 @@ private void editSample(SampleDto sample, LabMessageDto labMessage) {
...
+		}
+
+		if (referOrLinkToOtherLabButton != null) {
+			editForm.getButtonsPanel().addComponentAsFirst(referOrLinkToOtherLabButton);
+			editForm.getButtonsPanel().setComponentAlignment(referOrLinkToOtherLabButton, Alignment.BOTTOM_LEFT);
+		}
+	}
+
+	public void addReferredFromButton(CommitDiscardWrapperComponent<SampleEditForm> editForm, LabMessageDto labMessage, Window window) {
+		SampleReferenceDto referredFromRef = FacadeProvider.getSampleFacade().getReferredFrom(editForm.getWrappedComponent().getValue().getUuid());
```
leventegal-she, line 698: seems to be a duplicate of SampleController:: addReferredFromButton

leventegal-she (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleDto.java
FredrikSchaeferVitagroup: The `public` declaration is a leftover from a state where attributes should sometimes be migrated from the lab mesage instead of the sample. I will change that.

The physical sample is neither the source, nor the target. This method is about migrating the attributes that belong to the real (physical) sample out there in the labs. Source and target both have attributes that somehow have to do with the physical sample, but may vary. For example, the `specimenCondition` may be different in source and target. In one lab (source), the `specimenCondition` may be `ADEQUATE`. But then during transport to another lab (target) the `specimenCondition` might change to `NOT_ADEQUATE`.

In contrast, the attributes of the physical sample don't change (e.g. `samplingReason`) and thus should be migrated when a sample referral is created in SORMAS.

I extracted this function to make it more visible. In the past, adding attributes in the migration was often forgotten (that's why I included `fieldSampleId`, `samplingReason` etc. into this method.

So are you okay with making the method private and adding a comment more or less explaining the points above?

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
FredrikSchaeferVitagroup: This method is different from the `SampleController::addReferOrLinkToOtherLabButton`. It includes handling a lab message.

In other words, it replicates the behaviour of `SampleController::addReferOrLinkToOtherLabButton` while processing lab messages.

One could probably write a method in the SampleController that works for both, normal sample editing and lab message processing. I decided against this, because it would add solely lab message related dependencies to the SampleController. Also the complexity of the SampleController would be increased for a quite specific purpose.

Knowing this, is the current approach acceptable for you?

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
FredrikSchaeferVitagroup: Same as above.

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleDto.java
leventegal-she: okay

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
leventegal-she: Adding Runnable or Consumer argument(s) on `SampleController::addReferOrLinkToOtherLabButton` to have different handlers would not increase the complexity but instead there won't be duplicated code for the same logic.

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
leventegal-she: Adding Runnable or Consumer argument(s) on SampleController::addReferOrLinkToOtherLabButton to have different handlers would not increase the complexity but instead there won't be duplicated code for the same logic.

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
FredrikSchaeferVitagroup: I am not (yet) convinced. The problem I see is that Runnables would depend on an if statement in the method.
Consumers additionally would need a different amount of parameters for sample editing and lab message processing.
Maybe we should wait for a third opinion from @vidi42 ?

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
FredrikSchaeferVitagroup: The problem I see with this is that both Runnables would need to fetch the `referredFrom` SampleDto, so most of the code lines would still be duplicated.
Consumers would need a different amount of parameters for sample editing and lab message processing. Maybe we should wait for a third opinion from @vidi42 ?

FredrikSchaeferVitagroup: Sorry, I was wrong here. Consumers should work in this case. I will change to @leventegal-she suggestion.

sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/LabMessageController.java
vidi42: I'm not sure. It looks like `SampleController::addReferOrLinkToOtherLabButton` could be made a bit more generic with using Runnables or Consumers so that it can be called from the `LabMessageController`, however the whole logic seems to be in the `if` statement and `ClickListeners` so I'm not sure if this will be simplified in the end.

@FredrikSchaeferVitagroup maybe you can try to make it more generic and if it looks ok to you then you can leave it like that, otherwise I wouldn't insist on making it generic.

vidi42 (approved):
Looks good to me

PR #7331, opened 2021-11-19
#5144 Extend development environment documentation

### LINKED ISSUE #5144: Deploying `sormas-rest` before `sormas-ear` results in issues with the API communication

vidi42:
Fixes #5144


docs/DEVELOPMENT_ENVIRONMENT.md
```diff
@@ -38,7 +38,7 @@ If you plan to work on the Android App as well, you will also need the **Java 8
   - Enter `http://localhost:6080/sormas-ui` into the `URL` field
   - Make sure that the correct JRE is specified (your Java 11 JDK)
   - Enter the path to the SORMAS domain and the credentials that you've specified when setting up the server
-  - Open the `Deployment` tab and add the artifacts `sormas-ear`, `sormas-rest` and `sormas-ui`
+  - Open the `Deployment` tab and add the artifacts `sormas-ear`, `sormas-rest` and `sormas-ui` (make sure the respect the order as specified here as there are depdendencies between artifacts at startup)
```
MateStrysewske, line 41:
```suggestion
  - Open the `Deployment` tab and add the artifacts `sormas-ear`, `sormas-rest` and `sormas-ui` (make sure to respect this order as there are dependencies between artifacts at startup)
```

docs/DEVELOPMENT_ENVIRONMENT.md
```diff
@@ -62,7 +62,7 @@ If you plan to work on the Android App as well, you will also need the **Java 8
 - Either run `mvn install` on the `sormas-base` project or execute the `install [default]` Ant script (this needs a Maven installation on your system with the M2_HOME variable set)
 - Execute the `deploy-serverlibs` Ant script
 - Highlight all Eclipse projects and choose `Maven -> Update Project` from the right-click menu; perform the update for all projects
-- Start the Glassfish server and deploy `sormas-ear`, `sormas-rest` and `sormas-ui` by dragging the respective projects onto it, or use the `Add and Remove...` function by right-clicking on the server
+- Start the Glassfish server and deploy `sormas-ear`, `sormas-rest` and `sormas-ui` by dragging the respective projects onto it, or use the `Add and Remove...` function by right-clicking on the server (make sure the respect the order as specified here as there are depdendencies between artifacts at startup)
```
MateStrysewske, line 65:
```suggestion
- Start the Glassfish server and deploy `sormas-ear`, `sormas-rest` and `sormas-ui` by dragging the respective projects onto it, or use the `Add and Remove...` function by right-clicking on the server (make sure to respect this order as there are depdendencies between artifacts at startup)
```

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #7345, opened 2021-11-22
#5714 demis2 sormas correct entities

### LINKED ISSUE #5714: [DEMIS2SORMAS] Handle New Profile: Automatically propose to correct entities related to a reportId [5]

leventegal-she:
Fixes #5714


FredrikSchaeferVitagroup (changes requested):
While testing, I found some additional issues:

In the example below, I did not change any values by hand:

image


1. Unchanges fields are marked as changed (first name, last name...).
2. Changed fields are not marked as changed (date of birth).
3. The age calculation is not updated

Furthermore, when a value in a lab message or test report is null, and the related value in the entity is not null, it should not show as a changed value. Otherwise, the correction would currently popup for every lab message.


image

1. How come this testReport is matched to this pathogen test? 
2. The `RESULT VERIFIED BY LAB SUPERVISOR` is not greyed out on the left hand side.

Generally, fields on the left hand side don't have the same border color. I think it would be better if thay all had the same color (as specified in the issue), because this is less diverting to the user.

FredrikSchaeferVitagroup (approved):

PR #7390, opened 2021-11-29
Feature 7366 vaccination migration

### LINKED ISSUE #7366: [Immunization] Vaccination migration leads to improper vaccination data
### Feature Description

Immunizations > Vaccinations > Migration https://github.com/hzi-braunschweig/SORMAS-Project/issues/5909
### Problem Description

Some flaws in the immunization migration have been identified that lead to improper migration results for vaccinations. This will also have subsequent problems for interface mappings and vaccination display in the reduced immunization module.

One major point is that the `vaccinationDoses` in the previous fields is not equal to the `numberOfDoses` in immunizations because in immunizations it is the expected number of doses whereas in the previous fields it described the administered doses. The administered doses in immunizations are now reflected by the number of entries in the vacinations table.
### Proposed Change

- If `vaccinationDoses` are entered, create vaccinations equal to the number provided in the field.
    - If 1 is entered, use the `lastVaccinationDate` to set the `vaccinationDate`, if missing, use the `firstVaccinationDate` 
    - If 2 is entered, use both dates to set in the two respective vaccination entries
    - If the number is set to >2, create the respective number of vaccinations but only fill the first and last vaccination with the `lastVaccinationDate` and `firstVaccinationDate`
    - If the `vaccinationDoses` is missing, follow the specifications in #5909 

> - Create a new vaccination entity for the date of first vaccination, if available (if no date of vaccination is available, create a vaccination entity anyway, but leave the date empty); set the immunization_id, create a new health conditions entity and attach it to the vaccination; set the report date to the report date of the case/contact/ep; fill the remaining fields with the content of the VaccinationInfo entity of the contact/ep or the vaccination fields of the case; if the vaccine is ASTRA_ZENECA_COMIRNATY or ASTRA_ZENECA_MRNA_1273, set the vaccine of the new entity to OXFORD_ASTRA_ZENECA and the vaccine manufacturer to ASTRA_ZENECA
> - Create a new vaccination entity for the date of last vaccination, if available; same logic as for the first vaccination entity, except: Don't create this entity if NO vaccination date has been specified at all UNLESS the specified vaccine is ASTRA_ZENECA_COMIRNATY or ASTRA_ZENECA_MRNA_1273; in that case, create a new vaccination anyway and set the vaccine to COMIRNATY and the manufacturer to BIONTECH_PFIZER if the specified vaccine is ASTRA_ZENECA_COMIRNATY or the vaccine to MRNA_1273 and the manufacturer to MODERNA if the specified vaccine is ASTRA_ZENECA_MRNA_1273

- If the number is set to >2 and the vaccine is either `ASTRA_ZENECA_COMIRNATY` or `ASTRA_ZENECA_MRNA_1273`, create a first vaccination with `vaccineName` set to `OXFORD_ASTRA_ZENECA` including the `firstVaccinationDate` (if available) and all subsequent entries as either `COMIRNATY` or `MRNA_1273` respectively with one of them including the `lastVaccinationDate` (if available)

- If the person has multiple entities with available vaccination information, compare the `vaccinationDoses` and if earlier cases, contacts or participants have the same or a lower number of doses, set the report date of the number of created vaccinations (number dependent on the compared `vaccinationDoses`) according to the earliest available report date of associated entities.
    - Example:
A person has a case and a contact where the contact happened before they became a case. Let's say that while the person was a contact, they were vaccinated once already but when the case was docmented, they were vaccinated twice. Note that these entities could be independent, meaning the contact does not have to be converted.

|case report date|case vaccination doses|contact report date|contact vaccination doses|
|:-|:-|:-|:-|
|11/26/21|2|10/15/21|1|

Since in the general migration, we draw the information from the latest entity (the case in this example), the report date would be set to the report date of the case. However, the vaccination is also relevant to the contact and we need to fall back onto the report date if the vaccination date is missing. Therefore, the report date should be set to that of the contact.

If we e.g. find two COMIRNATY doses in the case, the first vaccination should have a report date of the contact.

### Additional Information

- The migration is done in 1.64.x, so every system that is already updated probably maybe cannot be changed/fixed. This most likely does not apply to PROD systems, but all stages/instances before.

BarnaBartha:
Fixes #7366


sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -2992,6 +2991,16 @@ private void migrateVaccinationInfo() throws SQLException {
 			formatRawResultDate(caseInfo, 9);
 			formatRawResultDate(caseInfo, 10);
 			formatRawResultDate(caseInfo, 11);
+
+			if (DataHelper.isNullOrEmpty((String) caseInfo[12])) {
```
MartinWahnschaffe, line 3019: Replace with formatRawResultInteger -> I'll create a follow-up refactoring issue for this

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -3014,13 +3023,15 @@ private void migrateVaccinationInfo() throws SQLException {
...
-						&& new Integer((String) earliestObject[12]) <= new Integer((String) objects[12])) {
-						objects[10] = earliestObject[10] == null ? earliestObject[4] : earliestObject[10];
-					}
-				});
+				caseInfoByDisease.get(Disease.valueOf((String) objects[2]))
+					.stream()
+					.sorted(comparator.reversed())
+					.findFirst()
+					.ifPresent(earliestObject -> {
+						if (earliestObject[12] != null && objects[12] != null && (int) earliestObject[12] <= (int) objects[12]) {
```
MartinWahnschaffe, line 3055: I guess I know what this is doing, but this definitely deserves a comment....

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -2898,6 +2898,30 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 			case 327:
 				currentVersion = 327;
 				getDao(FeatureConfiguration.class).executeRaw("ALTER TABLE featureconfiguration ADD COLUMN properties text;");
+
+			case 328:
+				currentVersion = 328;
+				getDao(Immunization.class).executeRaw("ALTER TABLE immunization RENAME TO tmp_immunization;");
```
MartinWahnschaffe, line 2904: Add a comment: Re-create table with correct unique constraint

MartinWahnschaffe (changes requested):
Just add two comments. I'll create a refactoring issue for the other parts that I found.
Note: I have NOT tested this on my Android device.

PR #7417, opened 2021-12-03
#7413 - Show ep samples for national users with limited disease

### LINKED ISSUE #7413: Samples for event participants are not shown again
### Bug Description

The same bug described here https://github.com/hzi-braunschweig/SORMAS-Project/issues/3965 is present again.

When a user has a disease limitation, the samples for event participants are neither shown in the card or sample directory.
### Steps to Reproduce

1. Set the limitation of a user (e.g. admin + national user) to COVID-19
2. open an event and add an event participant
3. Create a sample for this participant
4. Check the sample card and the sample directory for the created sample
5. Remove the limitation for the disease and try again

### Expected Behavior

The user should see the samples everywhere
### Screenshots

Limited user:

adminNationalUserMissingSamples4

adminNationalUserMissingSamples

adminNationalUserMissingSamples5

adminNationalUserMissingSamples7

Unlimited user:

adminNationalUserMissingSamples3

adminNationalUserMissingSamples2

adminNationalUserMissingSamples6

### System Details

* Device:
* SORMAS version: 1.66.0

MateStrysewske:
Fixes #7413

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
@@ -353,13 +353,6 @@ public class EventService extends AbstractCoreAdoService<Event> {
 		final From<?, EventParticipant> eventParticipantPath,
 		final EventUserFilterCriteria eventUserFilterCriteria) {
 
-		final User currentUser = getCurrentUser();
-		final JurisdictionLevel jurisdictionLevel = currentUser.getCalculatedJurisdictionLevel();
-		if (jurisdictionLevel == JurisdictionLevel.NATION || currentUser.hasAnyUserRole(UserRole.REST_USER)) {
-			return null;
-		}
-
-		Predicate filter = null;
 		final EventJoins eventJoins;
 		final From<?, Event> eventJoin;
 		final From<?, EventParticipant> eventParticipantJoin;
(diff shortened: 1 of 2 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
```diff
@@ -375,6 +368,18 @@ public Predicate createUserFilter(
 			eventParticipantJoin = eventParticipantPath;
 		}
 
+		final User currentUser = getCurrentUser();
+		final JurisdictionLevel jurisdictionLevel = currentUser.getCalculatedJurisdictionLevel();
+		if (jurisdictionLevel == JurisdictionLevel.NATION || currentUser.hasAnyUserRole(UserRole.REST_USER)) {
+			if (currentUser.getLimitedDisease() != null) {
+				return cb.or(cb.equal(eventJoin.get(Event.DISEASE), currentUser.getLimitedDisease()), cb.isNull(eventJoin.get(Event.DISEASE)));
```
MartinWahnschaffe, line 375: Calling return in the mid of a method can easily lead to unwanted behaviour. It should only be called at the top, related to validations checks for the method call or at the end, containing the result.

In this setting I'd suggest to do it the same we are doing it in CaseService.createUserFilter: For users without jurisidiction, we simply keep the Predicate filter unchanged (null). The limited disease check is done for all users below the jurisdiction check and uses CriteriaBuilderHelper.and.

MartinWahnschaffe (changes requested):

MartinWahnschaffe (approved):
Please create an issue to refactor/fix the SampleService.createUserFilter method

PR #7421, opened 2021-12-03
#7276 Fix person details update when importing contacts or event aparticipants

### LINKED ISSUE #7276: Import of Contacts are not updating person details
### Bug Description

When importing cases the details of the person (address, sex, salutation) are updated. If you import contacts the details are NOT updated.

I would expect that they are always updated by the imported details for cases, contacts and ep.

I will add screenshots during the next days.

vidi42:
The ticket only reffered to contact person, but I noticed the same issue was present on event participants as well.

Basically if an existing person was picked, the Importers discared the new person details and used only the existing person, so I changed it to merge the new person with the existing person if a person was picked.

Fixes #7276


vidi42:
Duplicate created by IntelliJ

PR #7422, opened 2021-12-03
#7276 Fix person details update when importing contacts or event aparticipants

### LINKED ISSUE #7276: Import of Contacts are not updating person details

vidi42:
The ticket only reffered to contact person, but I noticed the same issue was present on event participants as well.

Basically if an existing person was picked, the Importers discared the new person details and used only the existing person, so I changed it to merge the new person with the existing person if a person was picked.

Fixes #7276


sormas-ui/src/main/java/de/symeda/sormas/ui/contact/importer/ContactImporter.java
```diff
@@ -235,7 +238,15 @@ protected ImportLineResult importDataFromCsvLine(
 					}
 
 					// Workaround: Reset the change date to avoid OutdatedEntityExceptions
```
MateStrysewske, line 240: Move this comment back to above the `setChangeDate` line

sormas-ui/src/main/java/de/symeda/sormas/ui/events/eventparticipantimporter/EventParticipantImporter.java
```diff
@@ -233,11 +209,54 @@ protected ImportLineResult importDataFromCsvLine(
...
 					return ImportLineResult.SKIPPED;
 				} else {
-					// Workaround: Reset the change date to avoid OutdatedEntityExceptions
-					newPerson.setChangeDate(new Date());
+
 					boolean skipPersonValidation = ImportSimilarityResultOption.PICK.equals(resultOption);
-					final PersonDto savedPerson = personFacade.savePerson(newPerson, skipPersonValidation);
-					newEventParticipant.setPerson(savedPerson);
+
+
```
MateStrysewske, line 215: There are some empty lines here and below, please get rid of them

sormas-ui/src/test/java/de/symeda/sormas/ui/event/eventparticipantimporter/EventParticipantImporterTest.java
```diff
@@ -177,7 +177,7 @@ public void testImportEventParticipantSimilarityPick()
 		assertEquals(1, eventParticipantFacade.count(new EventParticipantCriteria().withEvent(eventRef)));
 		assertEquals(person.getUuid(), importedEventParticipant.getPersonUuid());
 		assertEquals(person.getFirstName(), importedPerson.getFirstName());
-		assertEquals(person.getLastName(), importedPerson.getLastName());
+		assertEquals("Heinze", importedPerson.getLastName());
```
MateStrysewske, line 180: Using an explicit string here might not be a good idea because it will make the test fail when someone changes the test .csv file. What's the reason for this change?

MateStrysewske (changes requested):
Haven't tested the event participant import yet, but the contact import has some major problems. I suspect those might also occur for event participants:

* Selecting an existing person skips the contact duplicate detection for me, i.e. the contact is immediately created although it's a duplicate of an existing one
* Person contact details aren't correctly taken over; e.g. I had a contact with a person that had a primary mail address and phone number; I imported a file where I specified a different mail address, but no phone number; I ended up with THREE new contact details, two new primary contact details that are empty (one phone number, one email address), and the specified email address from the import file that should've been the new primary email address, but was imported as a regular email address instead.

PR #7428, opened 2021-12-03
Change/6469 check oubound s2s entities

### LINKED ISSUE #6469: Check (outbound) entities for correct infrastructure data [3]

JonasCir:
Fixes #6469


sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/ShareDataBuilder.java
```diff
@@ -127,21 +162,39 @@ public ShareRequestPreviews buildShareDataPreview(ShareRequestInfo requestInfo)
 		List<SormasToSormasEventPreview> events = new ArrayList<>();
 		List<SormasToSormasEventParticipantPreview> eventParticipants = new ArrayList<>();
 
+		List<ValidationErrors> validationErrors = new ArrayList<>();
```
vidi42, line 165: `valiadtionErrors` are never checked and the preview request is sent empty

vidi42 (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/ShareDataBuilder.java
JonasCir: fixed

JonasCir:
@vidi42 fixed, I added the missing throw. I tested with normal shares (not preview) that's the reason it worked for me :)

vidi42 (approved):

PR #7429, opened 2021-12-03
Change/7132 validate outgoing s2s

### LINKED ISSUE #7070: Make validation a first-class member of the S2S processing pipeline
## Situation description
Right now we do validation of incoming data in the classes that extend `ReceivedDataProcessor` (more specifically the `processReceivedData` in these classes calls data validation logic of `Sormas2SormasDataValidator`. 

Validation happens by going through the received DTO _structure_ and checking the fields on by one for valid or reasonable values.

Since we plan to do this for _outgoing_ data as well in order to provide the user better error messages, we are now in a position where we (as we walk the structure of dto) need to access the same field twice, once while sending and another time while receiving it. However, we want to avoid code duplication. 

## How to resolve this
Ideally, we have have a set of acessor functions which can read a certain property of the DTO, which takes a validation function as second argument. By this, it is much simpler to realize incoming and outgoing checks.

## Implementation steps

- [x] #7071
- [x] #7132
- [ ] Split validation from field access in S2S validators
- [ ] Implement outgoing validation

### LINKED ISSUE #7132: Move the validation logic of core S2S entities to S2S validator classes
### Reasoning
Validation happens by going through the received DTO structure and checking the fields on by one for valid or reasonable values.

Since we plan to do this for outgoing data as well in order to provide the user better error messages, we are now in a position where we (as we walk the structure of dto) need to access the same field twice, once while sending and another time while receiving it. However, we want to avoid code duplication.

Therefore we want to introduce a from which supports this.

### Todos

- [x] #7133
- [x] #7136
- [x] #7139

JonasCir:
Fixes #7132 
Fixes #7070 

This is not yet wired up to the share logic and will not be called, this will be done separately in #7428.


sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/community/CommunityDto.java
```diff
@@ -1,34 +1,31 @@
-/*******************************************************************************
+/*
```
vidi42, line 1: Please also update the copyright year

sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/district/DistrictDto.java
```diff
@@ -1,33 +1,30 @@
-/*******************************************************************************
+/*
```
vidi42, line 1: Copyright year :)

sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/facility/FacilityDto.java
```diff
@@ -1,20 +1,17 @@
-/*******************************************************************************
+/*
```
vidi42, line 1: Copyright year :)

sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/region/RegionDto.java
```diff
@@ -1,37 +1,34 @@
-/*******************************************************************************
+/*
```
vidi42, line 1: Copyright year :)

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -9146,4 +9146,17 @@ ALTER TABLE testreport_history ADD COLUMN preliminary boolean;
 
 INSERT INTO schema_version (version_number, comment) VALUES (431, '[DEMIS2SORMAS] Handle New Profile: Preliminary Test Results #5551');
 
+-- 2021-12-03 [S2S] Implement outgoing S2S entity validation #7070
+ALTER TABLE areas ADD COLUMN centrally_managed boolean DEFAULT false;
```
vidi42, line 9150: should probably add them to the history tables as well

vidi42 (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/community/CommunityDto.java
JonasCir: done

sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/district/DistrictDto.java
JonasCir: done

sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/region/RegionDto.java
JonasCir: done

sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/facility/FacilityDto.java
JonasCir: done

sormas-backend/src/main/resources/sql/sormas_schema.sql
JonasCir: the history tables are not yet included

sormas-api/src/main/java/de/symeda/sormas/api/utils/DataHelper.java
```diff
@@ -71,8 +71,8 @@ public static String createUuid() {
 		return uuid;
 	}
 
-	public static String createConstantUuid(int seed) {
-		return new UUID(0, seed).toString();
+	public static String createConstantUuid() {
```
vidi42, line 74: I think we should leave the possibility of having different constant uuids, otherwise this method will always result in `00000000-0000-0000-0000-000000000000` which in the end could just return this hardcoded string.

We should still keep different constant uuids for the default data, so for example it's ok to always have the default region uuid `00000000-0000-0000-0000-000000000001` while the default district uuid is `00000000-0000-0000-0000-000000000002` for example.

vidi42 (changes requested):
Looks good. I'm just not sure about the changes done to default entities created.
I think the uuids for each default entity should still be "unique"

sormas-api/src/main/java/de/symeda/sormas/api/utils/DataHelper.java
JonasCir: done

JonasCir:
@vidi42 this is ready for review 
Only thing right now is that I have commented the history tables in `sormas_schema.sql`. Not certain what to do here.

vidi42:
> @vidi42 this is ready for review Only thing right now is that I have commented the history tables in `sormas_schema.sql`. Not certain what to do here.

@JonasCir Since we only have a history table for `areas` I think it's enough for now if we update that one. This ticket shouldn't be concerned about providing history tables for infrastructure data. We can open a different ticket for that.

Also since the history tables will be added in a later schema version you can remove the comments as they won't be done there.

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -9393,4 +9393,34 @@ ALTER TABLE testreport_history ADD COLUMN preliminary boolean;
...
+ALTER TABLE areas_history ADD COLUMN centrally_managed boolean DEFAULT false;
+
+ALTER TABLE community ADD COLUMN centrally_managed boolean DEFAULT false;
+/*ALTER TABLE community_history ADD COLUMN centrally_managed boolean DEFAULT false;*/
```
vidi42, line 9401: No need to comment this here since the missing tables will most likely be added in a different schema version

vidi42 (changes requested):

sormas-backend/src/main/resources/sql/sormas_schema.sql
JonasCir: fixed

vidi42 (approved):

PR #7447, opened 2021-12-06
#7263 dashboard disease burden

marius2301:
Fixes #


MateStrysewske (changes requested):
I'd like you to use the existing DashboardCriteria and update it with the new fields you need instead of creating a new one that is only used for the communication with the Angular app, but doesn't serve any other purpose.

MateStrysewske (approved):

PR #7453, opened 2021-12-06
Update README.md

Candice-Louw:
Added:
"We herewith explicitly would like to draw your attention to the fact, that the respective public health agency running SORMAS is in charge of data security and data protection and has to ensure compliance with national data protection and data security regulations in their respective jurisdiction. It has to ensure that state-of-the art requirements for data protection and data security are fulfilled. All those prerequisites and examinations have to be done in the context of the country and its respective legal framework. For these reasons, HZI cannot take the responsibility from the respective public health agency running the SORMAS systems and is not liable for any violation of data protection of the agency as the data generated by SORMAS belong to that very agency."
under the "Who is responsible for Data Protection and Data Security?" section.

Please feel free to move or rename if necessary. Text from GK.

No related issue.


MateStrysewske (approved):

MateStrysewske (changes requested):
Please look at the test failures - lint came up with an error there that needs to be fixed

PR #7473, opened 2021-12-08
#7400 - Set contact disease based on source case disease when it is n…

### LINKED ISSUE #7400: [Contacts] In the 'Create New Contact' form, if the Disease is not selected before choosing the Source Case, the form cannot be saved
### Bug Description
When creating a New Contact, if the Disease is not selected before choosing the Source Case, the form cannot be saved. It can only be saved if the Disease is selected  before choosing the Source Case.

### Steps to Reproduce

1. Go to the Contacts page.
2. Open the New Contact form.
3. Fill in the name fields and select the sex, but leave the Disease field empty.
4. Choose a Source Case.
5. Click on Save.

### Actual Behavior
The form cannot be saved and a message stating that the user needs to specify a valid disease appears.

### Expected Behavior
The form should be saved and created.

### Screenshots

new contact

### System Details

* Device: Windows 10
* SORMAS version: 1.66.0-SNAPSHOT
* Server URL: test-de.sormas.netzlink.com
* User Role: NatUser

alexcaruntu-vita:
…ot available in the creation form

Fixes #7400


MateStrysewske (changes requested):
I'm not really happy with these changes. Contacts ALWAYS have to have a disease, even if they have a source case with a disease, so the extension of the `validate` method will potentially lead to problems. Also, the `toDto` and `fromDto` methods should contain as few logic as possible (i.e. only what's really needed to convert an entity to a DTO and vice versa).

I would expect the fix to set the contact disease in the contact data form when selecting a source case, so that the contact is in a state that is accepted by the backend in its current state.

MateStrysewske (approved):

PR #7479, opened 2021-12-08
Feature 6981 add new userflow for lab message processing

### LINKED ISSUE #6981: [Lab Messages] Add new userflow for lab message processing [1]
### Situation Description
After the processing of the lab sample it is necessary to change to the case and further to edit the case (case classification, check personal data, create tasks). In order to continue the next steps it would be rational to continue data entry for person (firstname, lastname, birthdate (DDMMYYYY), personcontactdetail, address) and in the next step the creation of tasks in the next window.

### Feature Description
During the process of a lab message (DEMIS) is a guide who lead to single steps of each process (e.g. Choose Person\Create new case; Choose case\contact\event participant\; Choose lab sample\create new lab sample; Choose lab sample ID\Create new lab sample; Create new sample result). After entry of all fields it makes sense to continue guide for next steps:
Check personal data, choose case classification, create tasks.

### Possible Alternatives
Enter alternative if possible. Possibly implementation of flow after lab sample processing (SORMAS guide for case creation).

### Additional Information
Discussed in concept meeting (Jan Böhme). I issue a visual work flow after creation

FredrikSchaeferVitagroup:
Fixes #6981


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/AbstractCaseView.java
```diff
@@ -137,6 +138,12 @@ public void refreshMenu(SubMenu menu, String params) {
 
 		menu.removeAllViews();
 		menu.addView(CasesView.VIEW_NAME, I18nProperties.getPrefixCaption(CaseDataDto.I18N_PREFIX, Captions.caseCasesList));
+
+		if (FacadeProvider.getFeatureConfigurationFacade().isFeatureEnabled(FeatureType.LAB_MESSAGES)
```
vidi42, line 142: Would it be possible to only show this navigation option if the case originates from a lab message and to now show it for every case?

vidi42 (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/AbstractCaseView.java
FredrikSchaeferVitagroup: It would be possible. Lab message are related to samples. One would therefore need to retrieve all samples for the case, and then check whether at least one lab message in the database relates to one of those samples. 

Do you think that this effort and additional system load is worth it? I'm not sure...

vidi42: Probably the check can just be a database query and we don't need to retrieve the samples from the db, just a yes or a no if the case can be joined with a sample and a lab message.

Since it will be ran for just a case I don't think it will put much pressure on the system and would probably be more helpful for the users then, as the number of cases created from lab messages probably won't outmatch those which are created directly.

So from my point of view it would make sense

FredrikSchaeferVitagroup: Okay. Wouldn't it be for the contact and event participant view as well?

vidi42: Yes, would be needed for those as well.

sormas-backend/src/main/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjb.java
```diff
@@ -382,6 +387,27 @@ public boolean exists(String uuid) {
 		return labMessageService.exists(uuid);
 	}
 
+	@Override
+	public boolean existsLabMessageForEntity(ReferenceDto entityRef) {
+		List<Sample> samples;
+		if (CaseReferenceDto.class.equals(entityRef.getClass())) {
+			samples = sampleService.getByCaseUuids(Arrays.asList(entityRef.getUuid()));
```
vidi42, line 394: I was thinking to have a check directly in the DB and not retrieve all the samples for a case. I think in the past there were performance problems because of the number of samples.

I tried with a new method which I added in `LabMessageService` and the tests passed.

`LabMessageSerice.java`
```java
public long countForCase(String caseUuid) {
		CriteriaBuilder cb = em.getCriteriaBuilder();
		CriteriaQuery<Long> cq = cb.createQuery(Long.class);
		Root<LabMessage> labMessageRoot = cq.from(LabMessage.class);
		Join<LabMessage, Sample> sampleJoin = labMessageRoot.join(LabMessage.SAMPLE, JoinType.LEFT);
		Join<Sample, Case> caseJoin = sampleJoin.join(Sample.ASSOCIATED_CASE, JoinType.LEFT);

		Predicate filter = cb.and(createDefaultFilter(cb, labMessageRoot), caseJoin.get(AbstractDomainObject.UUID).in(Collections.singleton(caseUuid)));

		cq.where(filter);
		cq.select(cb.countDistinct(labMessageRoot));

		return em.createQuery(cq).getSingleResult();
	}
```

Then I changed this line to this
```suggestion
	return labMessageService.countForCase(entityRef.getUuid()) > 0;
```

The `countForCase` method can be replicated for Contact and EventParticipant

vidi42 (changes requested):

vidi42 (approved):

PR #7495, opened 2021-12-10
#7491 - Only create tmp_healthconditions once to avoid too many trans…

### LINKED ISSUE #7491: Update vaccination migration script to make it executable on environments with significant vaccination data
### Feature Description

With 1.64.0 we introduced the new immunization module and an accompanying migration script that moves the vaccination information from cases, contacts, and event participants to the new tables.

### Problem Description

Within this script, we clone the health conditions table for each vaccination that we create. This results in a lot of temporary tables to be created, which quickly exceeds the maximum available locks in the same transaction on systems where a significant amount of cases, contacts, or event participants have vaccination information.

### Proposed Change

Only create the temporary table once and only fill it with new data in the `clone_healthconditions` function.

MateStrysewske:
…actions

Fixes #7491


BarnaBartha (approved):

StefanKock (approved):
The changes on the script are as we discussed, so no remarks on that.
I think there should not be white space changes on not affected code (separation of concerns for commits), that's polluting the needed changes

PR #7509, opened 2021-12-13
[#7504] add CodeQL scanner

### LINKED ISSUE #7504: Add `CodeQL` to GitHub actions
## Proposed change
Add a [CodeQl](https://codeql.github.com/) pass after CI build task.

JonasCir:
Fixes #7504


MateStrysewske:
@JonasCir The CodeQL checks find issues in files that we didn't create ourselves (e.g. highcharts.js). I don't know whether these files should make the checks fail - if they do, we either have to change code that we didn't write ourselves or we'll end up with failing CI all the time. In any way, I would probably expect this PR to also fix these problems in order to keep CI succeeding on the development branch.

JonasCir:
@MateStrysewske Fair point, we should fix it in this PR. I'm pretty sure the highchar.jst code hasn't been updated in a while so it might be worth to check for a recent version. Also, I'm somewhat confused that this script is not pulled from a CDN, but provided plain in the code.

MateStrysewske (changes requested):
Just a formal change request so that you can re-request a review after you've excluded js files :)

JonasCir:
@MateStrysewske removed unnecessary Long type which was caused by the previous `Collectors.counting()` call. Also fixed a couple of low hanging sonar lint warnings.

JonasCir:
@MateStrysewske this is ready for review :)

MateStrysewske (approved):

PR #7511, opened 2021-12-14
Change 7302 push new data before pulling

### LINKED ISSUE #7302: Push new data to the server before pulling changed data [5]
### Feature Description

The current synchronisation logic used in the mobile app pulls changed data from the server before pushing new and updated data.

### Problem Description

In areas with bad internet connection or on servers with a huge amount of data in the jurisdiction of the app user, pulling changed data might take a long time or sometimes not succeed at all. Pushing new data to the server should always be prioritised because it's more essential to report new cases, contacts, etc. to the server than getting updated or new data back to the mobile device.

### Proposed Change

During the synchronisation process, push new data first, then pull changes from the server, then push changed data to the server afterwards.

See https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-base/doc/SormasSyncProcess.md
Update this documentation accordingly. I'd suggest to move the files from there to the root docs folder.

### Additional Information
Since this add's even more server-request to the sync mechanism, we need to define a follow-up issue that collects those requests into a single one, to reduce the time the requests themselves take (without considering any payload). We already have this implemented for pulling changed infrastructure.

In addition to that the different sync modes currently have a lot of redundant overlapping activity. We should create an issue to address this.

MateStrysewske:
Fixes #7302


sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/AdoDtoHelper.java
```diff
@@ -175,7 +175,7 @@ public boolean pushEntities(boolean onlyNewEntities)
 		throws DaoException, ServerConnectionException, ServerCommunicationException, NoConnectionException {
 		final AbstractAdoDao<ADO> dao = DatabaseHelper.getAdoDao(getAdoClass());
 
-		final List<ADO> modifiedAdos = onlyNewEntities ? dao.queryForNull(ADO.CHANGE_DATE) : dao.queryForEq(ADO.MODIFIED, true);
+		final List<ADO> modifiedAdos = onlyNewEntities ? dao.queryForEq(ADO.CHANGE_DATE, new Date(0)) : dao.queryForEq(ADO.MODIFIED, true);
```
MartinWahnschaffe, line 178: Move queryForEq... logic into the AbstractAdoDao class: queryForNew and queryForModified
Add unit test for those methods

MartinWahnschaffe (changes requested):

PR #7525, opened 2021-12-15
Feature 7310 add option to assign lab messages to users

### LINKED ISSUE #7310: [DEMIS2SORMAS] Add the option to assign lab messages to users [3]
### Situation Description

Currently, users have a hard time managing large numbers of incoming lab message. They can't process them fast enough and they struggle to report cases to higher health institutions in time. Thus, they should be able to assign lab messages to users to manage and distribute work more efficiently.

concept issue: #7308
### Feature Description

- [x] Add a column to the lab message directory in between 'Status' and the 'Process' column
    - [x] Heading: 'Assignee'
    - [x] If there is no assignee set for a lab message, display the text 'Assign' in the respective cell
        - [ ] While hovering, indicate that this text is clickable by changing the colour slightly and underlining the text (cf. uuid's for reference)
        - [x] Upon clicking on 'Assign', open a pop-up
            - [x] Heading: 'Edit assignee'
            - [x] Add a drop-down field for the user selection; heading: 'Assignee'
            - [x] Add a clickable text: 'Assign to me'
                - [x] Upon clicking, automatically select the current user as the assignee
            - [x] Add 'Discard' and 'Save' buttons
                - [x] Pressing 'Discard' should discard the user selection and close the pop-up
                - [x] Pressing 'Save' should save the set user as the assigne and close the window
    - [x] When a user was selected as the assignee, display the username in the respective cell (username text not clickable anymore)
         - [x] In the same cell, display a 'More' icon that fills the cell but does not get overlayed by the username
             - [x] The icon must be present in the cell at all times and should not float out of the visible cell when adjusting the width
             - [ ] If the cell gets clinched, the icon must overlay the username text. When clinched, the username text should float out of the cell behind the icon but the icon should remain it's white background so it doen't overlap with the text.
             - [x] Upon clicking on the icon, open the 'Edit assignee' pop-up
            - [x] If there is an assignment in place the user should be able to unassign the user 

- [x] Add a drop-down filter next to the free text filter. Placeholder: 'Assignee'
    - [x] opening the drop-down should list the assigned users
    - [x] entering content in the field should also search for users in the list
    - [x] Upon selecting a user and applying filters, the list should get filtered to only display messages assigned to the filtered user
    - [x] The filter should include a value called 'Unassigned'
        - [x] Upon selection, the directory should filter for all lab messages that do not have an assignee yet

- [x] The users that are able to be assigned to lab messages should be any users having the `LAB_MESSAGES` user right
### Additional Information

Column and filter
Lab messages Refinements_assignee

Edit assignee pop-up
Lab messages Refinements_editAssignee

Clinched assignee cell
Cell link_assignee

ChristopherRiedel:
Fixes #7310


FredrikSchaeferVitagroup (changes requested):
I saw that two of the tickets requirements are not yet fully met:

`While hovering, indicate that this text [Assign] is clickable by changing the colour slightly and underlining the text (cf. uuid's for reference)`

and 

`If the cell gets clinched, the icon [edit assignee] must overlay the username text. When clinched, the username text should float out of the cell behind the icon but the icon should remain it's white background so it doen't overlap with the text.`

Is there are reason? ANSWER: Yes, those two requirements are not doable with the currently used vaadin version. Implementation has come as near as possible.

Also, when users are assigned, the `edit assignee` is not horizontally aligned:
image 
Would it not be clearer to have the buttons aligned rightmost to the column? ANSWER: No, because some names may be very long, and then all buttons would be far away from their related names.

The changes are completely not covered by unit tests. I think a bare minimum would be to extend the [LabMessageFacadeEjbMappingUnitTest](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-backend/src/test/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjbMappingTest.java).

Please also update to latest state of development.

FredrikSchaeferVitagroup (changes requested):

FredrikSchaeferVitagroup (approved):

Issue #6062, opened 2021-07-05
Hospital Informant: Create a case leads to an unexpected error 


VitaliHZI:
### Bug Description
If you try to create a new case with a hospital informant, an expected error occured. 

### Steps to Reproduce

1. Go to cases
2. Click on "new case" 

### Expected Behavior
No error occurs. 

### Screenshots
![image](https://user-images.githubusercontent.com/79860634/124451908-2ae5ee00-dd86-11eb-8bb6-1c8d568291fe.png)

### System Details

* Device:
* SORMAS version:
* Android version/Browser:
* Server URL:
* User Role:

### Additional Information
![image](https://user-images.githubusercontent.com/79860634/124451908-2ae5ee00-dd86-11eb-8bb6-1c8d568291fe.png)

DavidBaldsiefen:
As discussed with @MartinWahnschaffe we will solve this problem (or followup problems) by automatically setting the case jurisdiction for hospital informants to the facility-jurisdiction, and disabling the fields so they can not be edited. Should this behavior be changed it should be part of a separate issue and not hotfix relevant.

AndyBakcsy-she:
Validated ticket on the latest SORMAS Version: 1.61.1 deployed on https://release-international.sormas.netzlink.com/ 
using Hospital informant user: TestUserInt1  /  8WawGQZwz5fG

PR #7526, opened 2021-12-15
#6062 - add feature type property to add soft/warning validation for …

### LINKED ISSUE #6862: Add a server property that allows to reduce the follow-up period below the usual threshold [2]
Please note that the development specifications differ from the original specification of this issue.

### Feature Description

Make "Overwrite follow-up until date"  remove the restriction for the minimum length of the follow up period.

### Problem Description

At the moment the local health department  in germany do not always use a 14 day quarantine period. for example if someone can provide a negative test result after five days or the quarantine itself is only scheduled for less than 14 days.

As the people entering the data and setting the quarantine dates are most of the time not the ones doing the follow-up calls they don't really know if the person scheduled for a call really needs to be called or if that person already finished quarantine. 
Due to this reason the people setting the quarantine dates tried to set the follow-up period to a shorter time than 14 days but even when using the "override follow-up until date" checkbox the field "follow-up until" does not allow for a persod shorter than the one configured for the disease server-side.

### Proposed Change

If the user activates the checkbox "override follow-up until date" the restriction on the length of the follow-up period should be deactivated. As the user actively clicks on the checkbox it is safe to assume that setting the period to a shorter time than it should theoretically be is an active decision by the user and not an error.

If the user sets a length that is shorter than the server config states the filed should become blue and a mouse over information should appear with the following text:
"The follow up period is shorter than expected for the selected disease."

Checks for dates in the past or invalid dates should stay in place.

### Additional Information

2021-10-01_14h49_13

* Add a boolean feature type property called `ALLOW_FREE_FOLLOW_UP_OVERWRITE` and add it to the `CONTACT_TRACING` and `CASE_FOLLOWUP` features; default false
* When the new property == true, for the respective section (contact/case), when the user overwrites the follow-up period and sets a date that is before the default minimum follow-up date, instead of showing a hard error and preventing the user from saving the form, show a soft error (i.e. the light blue field outline and exclamation mark in the web app that is e.g. displayed when setting a temperature in the symptoms tab and not specifying the fever field) and change the text that is displayed to "The follow-up until date should at least be on the day the standard follow-up duration for the disease would end."

BarnaBartha:
…follow up until field

Fixes #6862


MateStrysewske (changes requested):
UI part looks good, but it seems like when you overwrite the follow-up until date with an earlier date, it is immediately reset to the expected follow-up until date again, so there apparently is some automation there that interfers with this new logic.

MateStrysewske (approved):

PR #7540, opened 2021-12-17
fix TravelEntry different jurisdiction discard #7245

### LINKED ISSUE #7245: Pressing dicard in travel entry forms does not uncheck the differing point o entry jurisdiction
### Bug Description

When discarding entries on the travel entry form, the point of entry jurisdiction difference checkbox does not get unchecked. 
### Steps to Reproduce

1. Create a new travel entry but leave the jurisdiction the same for the point of entry.
2. In the form, check the checkbox "Point of entry differs from responsible jurisdiction"
3. Press discard

### Expected Behavior

The checkbox becomes unchecked

### Screenshots

discardTravelEntryForm
discardTravelEntryForm2

### System Details

* Device:
* SORMAS version: 1.65.0

syntakker:
Fixes #7245

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/travelentry/TravelEntryDataForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/travelentry/TravelEntryDataForm.java
@@ -57,6 +57,8 @@ import de.symeda.sormas.ui.utils.NullableOptionGroup;
 
 public class TravelEntryDataForm extends AbstractEditForm<TravelEntryDto> {
 
+	private static final long serialVersionUID = 4516334613016719068L;
+
 	private static final String RESPONSIBLE_JURISDICTION_HEADING_LOC = "responsibleJurisdictionHeadingLoc";
 	private static final String DIFFERENT_POINT_OF_ENTRY_JURISDICTION = "differentPointOfEntryJurisdiction";
 	private static final String POINT_OF_ENTRY_HEADING_LOC = "pointOfEntryHeadingLoc";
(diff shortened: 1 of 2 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/travelentry/TravelEntryDataForm.java
```diff
@@ -171,22 +173,24 @@ protected void addFields() {
...
-			DistrictReferenceDto districtDto;
-			if (differentPointOfEntryJurisdiction.getValue().booleanValue()) {
-				districtCombo.setValue(null);
-				getPointsOfEntryForDistrict(null);
-			} else {
-				cbPointOfEntry.setValue(null);
-				districtDto = (DistrictReferenceDto) responsibleDistrict.getValue();
-				getPointsOfEntryForDistrict(districtDto);
+			if (!ignoreDifferentPointOfEntryJurisdiction) {
+				DistrictReferenceDto districtDto;
```
alexcaruntu-vita, line 185: You can combine this definition with the initialisation from line 191 and get rid of this line.

sormas-ui/src/main/java/de/symeda/sormas/ui/travelentry/TravelEntryDataForm.java
```diff
@@ -171,22 +173,24 @@ protected void addFields() {
...
+				if (differentPointOfEntryJurisdiction.getValue().booleanValue()) {
```
alexcaruntu-vita, line 186: The unboxing is not needed.

alexcaruntu-vita (approved):

sormas-ui/src/main/java/de/symeda/sormas/ui/travelentry/TravelEntryDataForm.java
syntakker: True, the unboxing fails if `differentPointOfEntryJurisdiction.getValue()` is `null`. However, the `CheckBox` has its own `booleanValue()` method to cope with `null` values.

PR #7544, opened 2021-12-19
Bugfix 7519 pathogen test info not shown in sample card

### LINKED ISSUE #7519: Pathogen test information not initially shown in sample card
### Bug Description

When entering a sample directly with a pathogen test, the pathogen test information is not initially shown.
### Steps to Reproduce

1. Open e.g. a case (or contact, event participant) and press "new sample"
2. enter a sample and directly enter pathogen test information, save.
3. Check the information displayed in the sample card
4. Refresh the page and check the cards content again

### Expected Behavior

The pathogen test information should show immediately with other information in the card.

### Screenshots

samplePathogenTestDisplayBug2
samplePathogenTestDisplayBug3

### System Details

* Device:
* SORMAS version: 1.66.3
* Server URL: test1402.sormas.netzlink.com
* User Role: admin, naitonal user

alexcaruntu-vita:
Fixes #7519


sormas-ui/src/main/java/de/symeda/sormas/ui/utils/components/MultilineLabel.java
```diff
@@ -0,0 +1,17 @@
+package de.symeda.sormas.ui.utils.components;
```
syntakker, line 1: Introducing a new class to set label width to 100% seems suspicious, especially since this is there is only one usage of this class.
I would suggest to just set the width for the label.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/sampleLink/SampleListComponent.java
```diff
@@ -45,33 +38,7 @@
...
-			new SampleList(new SampleListCriteria.Builder().withEventParticipant(eventParticipantRef).build()),
-			e -> ControllerProvider.getSampleController().create(eventParticipantRef, disease, SormasUI::refreshView),
-			view);
-	}
-
-	private void createSampleListComponent(
-		SampleList sampleList,
-		Consumer<Button.ClickEvent> clickListener,
-		AbstractDetailView<? extends ReferenceDto> view) {
+	public SampleListComponent(SampleList sampleList, Consumer<Button.ClickEvent> clickListener, AbstractDetailView<? extends ReferenceDto> view) {
```
syntakker: I advocate to revert the refactoring in commit 51ad5421629cb20d76ed37861d1148740f1cd92c since the details of the instantiation of a SampleListComponent are now spread across several classes (`CaseDataView`, `ContactDataView`, `EventParticipantDataView`)

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleListCriteria.java
```diff
@@ -0,0 +1,70 @@
+package de.symeda.sormas.api.sample;
```
syntakker, line 1: `SampleListCriteria`'s properties are a subset of `SampleCriteria`'s. The introduction of an extra class here creates additional overhead to maintain or extend these criteria.

syntakker:
In general: the combination of a bugfix with a major refactoring is a bit unlucky here. Please consider opening an extra issue for refactorings.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -129,16 +129,19 @@ private void createSample(Runnable callback, SampleDto sampleDto, Disease diseas
...
-		CommitDiscardWrapperComponent<? extends AbstractSampleForm> sampleComponent,
-		PathogenTestDto pathogenTest) {
+	public PathogenTestForm addPathogenTestComponent(CommitDiscardWrapperComponent<? extends AbstractSampleForm> sampleComponent) {
 
 		int caseSampleCount = caseSampleCountOf(sampleComponent.getWrappedComponent().getValue());
-		return addPathogenTestComponent(sampleComponent, pathogenTest, caseSampleCount);
+		return addPathogenTestComponent(sampleComponent, null, caseSampleCount, SormasUI::refreshView);
+	}
+
+	public PathogenTestForm addPathogenTestComponent(
```
syntakker, line 140: Please add the respective Javadoc section here, uniformly with the other definitions of `addPathogenTestComponent`.

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleListCriteria.java
alexcaruntu-vita: `SampleListCriteria` it is used strictly for the sample list. It uses only 3 of the 23 members of `SampleCriteria`. I would say that it reduces the overhead :).

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/components/MultilineLabel.java
alexcaruntu-vita: We have this problem in many places in the code. For instance, here: https://github.com/hzi-braunschweig/SORMAS-Project/pull/7341/files. Vaadin guide also suggest to have small components for situations when what they provide is not enough. But in this particular place we can remove the labels completely, the new flow with showing the discard or save changes popup makes them obsolete. @MateStrysewske, what do you think?

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/sampleLink/SampleListComponent.java
alexcaruntu-vita: This is just an intermediary commit. Take a look at the final version of `SampleListComponent` and its usages.

alexcaruntu-vita:
> In general: the combination of a bugfix with a major refactoring is a bit unlucky here. Please consider opening an extra issue for refactorings.

I agree, but we have the same implementation for side lists in other places as well, so it was an opportunity to align to that paradigm.

MartinWahnschaffe:
I had a look and a few comments. I'm not convinced the re-factoring was necessary here and would request to either revert the refactoring or to address the following points:

1. Services are meant to to give access to the data held by the JPA entities. There should be one service per entity. If there are multiple DTO per entity, the service should provide separate methods for those. 
2. The criteria and buildCriteraQuery method that implements the query logic should be used for all the dtos. 

I noticed that the services for TravelEntries have been separated. This is not the way it's supposed to be, so please don't use this as an example.

I have a todo on my list to define a refactoring for the naming of backend facades and services and to introduce architecture unit tests that make it a bit easier to understand the way, the architecture is supposed to be. I'm aware that there is currently no explanation on how the ideal architecture of those backend classes is supposed to be.

MartinWahnschaffe:
@alexcaruntu-vita My points have been addressed. As far as I can tell there is at least one comment made by Frank that is not addressed yet.
@MateStrysewske @syntakker Can either of you do the final review?

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
```diff
@@ -137,6 +158,287 @@ public SampleService() {
...
+				sample.get(Sample.SAMPLING_REASON),
+				sample.get(Sample.SAMPLING_REASON_DETAILS),
+				sample.get(Sample.ADDITIONAL_TESTING_REQUESTED),
+				cb.isNotEmpty(sample.get(Sample.ADDITIONAL_TESTS))));
+
+		// Tests count subquery
+		Subquery<Long> testCountSq = cq.subquery(Long.class);
+		Root<PathogenTest> testCountRoot = testCountSq.from(PathogenTest.class);
+		testCountSq.where(
+			cb.equal(testCountRoot.join(PathogenTest.SAMPLE, JoinType.LEFT).get(Sample.ID), sample.get(Sample.ID)),
```
MateStrysewske, line 418: Can you test whether you can also write this line like this:

`cb.equal(testCountRoot.get(PathogenTest.SAMPLE), sample)`

Not sure whether that works, but if it does, it would spare us additional joins.

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
alexcaruntu-vita: Seems to be working, I have made the change for sample directory as well.

MateStrysewske (approved):

PR #7550, opened 2021-12-20
[#7547] bump logback

### LINKED ISSUE #7547: Bump logback to 1.2.9
### Proposed change
Title says it all. Read [here](http://logback.qos.ch/news.html) for more information.

### Additional information
For already set up Payara domains, this configuration needs to be added manually to not lose the logging after undeployment of articfacts:
Manual change in `domain.xml`: `<jvm-options>-DlogbackDisableServletContainerInitializer=true</jvm-options>` in `server-config` paragraph.

JonasCir:
Fixes #7547

Removed the scanning attribute as auto-reload is a quite nice gadget for attackers to use.

It was also a prerequisite for the most recent attack against this library as we can see [here](https://github.com/cn-panda/logbackRceDemo), so I think practicing "defense in depth" is a reasonable approach here.


StefanKock (changes requested):
1. Changing logging config on a running instance is a desirable feature. I don't think we should disable it.
2. I tested on a comparable setup that logback 1.2.9 still disables logging by default when a servlet is undeployed. This snippet needs to be inserted into `server-setup.sh` to avoid this: `${ASADMIN} create-jvm-options -DlogbackDisableServletContainerInitializer=true` (after `logback.configurationFile` line). But all already set up domains (all DEV instances, not dockerized setups) will suffer from log losses when redeployments occur. Needed manual change in `domain.xml`: `<jvm-options>-DlogbackDisableServletContainerInitializer=true</jvm-options>` in `server-config` paragraph.
3. CVE-2017-5929 in `check-suppressions.xml` should be removed.

PR #7556, opened 2021-12-21
Feature 7434 do not validate case save from visit

### LINKED ISSUE #7434: Visits for contacts cannot be saved if users don't have access to the converted case
### Bug Description

When a restricted user tries to enter visits in hindsight for contacts that have already been converted to cases, an error message prevents them from doing so when they don't have access to the converted case.
### Steps to Reproduce

1. Create a contact in jurisdiction 1 and convert that contact to a case in jurisdiction 2 (do that with e.g. admin + national user)
2. Log in as a user in jurisdiction 1 and try entering a visit for the converted contact
3. Create a control contact and convert it into a case that is in the same jurisdiction
4. Try entering a visit again with the user in jurisdiction 1

### Expected Behavior

The user should be able to create visits for contacts that they have access to.

### Screenshots

changedInTheMeantimeErrorVisitsContacts

### System Details

* Device:
* SORMAS version: 1.66.0
* Server URL: release-international.sormas.netzlink.com

* When a contact has been converted to a case, disable the button to create a new visit (but still display it)
* Add an info icon next to it with the following description on hover: "This contact has already been converted to a case. Please add new visits to the case instead."

BarnaBartha:
Fixes #7434


MateStrysewske (changes requested):
You would've spared yourself a lot of refactoring work by adding a second `saveCase` method with the additional parameter, and by extending the current method to call the new method with `false` :) I'd suggest to still do this because it makes the code a lot cleaner (i.e. far less unnecessary parameters).

BarnaBartha:
Actually the IDE changed the signature automatically so no trouble there.
But I do agree that exposing it that way shows the actual problem in an enhanced way, that there is no true separation or the concept of system action vs user action. It is a specific flag for this scenario, and after some time a similar flag will appear for another entity's saving/validation logic and we'll just increase the unnecessary logic that complicates developers lives later as we never truly fix these issues.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1335,7 +1335,12 @@ public CaseReferenceDto getReferenceByUuid(String uuid) {
 
 	@Override
 	public CaseDataDto saveCase(@Valid CaseDataDto dto) throws ValidationRuntimeException {
-		return saveCase(dto, true, true);
+		return saveCase(dto, true, true, false, false);
```
MateStrysewske, line 1338: `internal` should not be set to false here; see line 1455, the method there sets it to `true` as well. If I understand it correctly, `internal` should always be true unless the case comes from an external source (e.g. sormas2sormas). I suggest to ommit the argument here and let the saveCase method in 1455 handle it.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1335,7 +1335,12 @@ public CaseReferenceDto getReferenceByUuid(String uuid) {
...
+	}
+
+	@Override
+	public CaseDataDto saveCase(@Valid CaseDataDto dto, Boolean systemSave) throws ValidationRuntimeException {
+		return saveCase(dto, true, true, false, systemSave);
```
MateStrysewske, line 1343: Same here

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -3167,7 +3172,7 @@ private void mergeCase(CaseDataDto leadCaseData, CaseDataDto otherCaseData, bool
 		// 1.1 Case
 
 		copyDtoValues(leadCaseData, otherCaseData, cloning);
-		saveCase(leadCaseData, !cloning, true);
+		saveCase(leadCaseData, !cloning, true, false, false);
```
MateStrysewske, line 3175: Same here

sormas-rest/src/main/java/de/symeda/sormas/rest/CaseResource.java
```diff
@@ -77,7 +77,7 @@
 	@POST
 	@Path("/push")
 	public List<PushResult> postCases(@Valid List<CaseDataDto> dtos) {
-		return savePushedDto(dtos, FacadeProvider.getCaseFacade()::saveCase);
+		return savePushedDto(dtos, dto -> FacadeProvider.getCaseFacade().saveCase(dto));
```
MateStrysewske, line 80: Strange refactoring - the `::` notation should actually be the preferred one by the IDE

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #7560, opened 2021-12-22
Feature 7521 sync in batches

### LINKED ISSUE #7303: Enable getAllAfter methods to be used to pull data in batches [5]
### Feature Description

Functional data (such as cases, contacts, persons, samples, etc.) is currently synchronised in one single request.

### Problem Description

Especially countries that use a single central SORMAS server run into significant problems here - not only during the initial synchronisation, but potentially also during other syncs, because there's too much functional data. Trying to pull this data from the server with low bandwidth can still lead to timeouts.

### Proposed Change

Extend the getAllAfter methods to optionally not return all data, but instead only chunks of a specified size.

* For each type of entity that is synchronized in `SynchronizeDataAsync::synchronizeChangedData` (except outbreaks, disease configurations and customizable enum values), extend the `getAllXXXAfter` methods in the respective facade and service classes with additional parameters:
  * `size`: The size of the chunk of data that is requested
  * `lastSynchronizedUuidSameTimestamp`: The UUID of the entity that was last synchronized when the change date passed to the method is identical to the change date of the last synchronized UUID
* Adjust the ReST resource and add the additional parameters as optional parameters; call the facade methods with these parameters set to null if they're not specified
* Add comments to the ReST methods that explains under which circumstances the last synchronized UUID may be specified and under which circumstances it must NOT be specified
* In the new methods, only retrieve as many entities as specified by `size`; order the query primarily by change date and, if specified, secondarily by UUID (to make sure that we don't synchronize the same entity twice, or skip an entity because the Android app database has less precise dates than the web backend database).
  * IF a lastSynchronizedUuidSameTimestamp is specified, calculate the timestamp including nanoseconds at the start of the specified change date, and the timestamp including nanoseconds at the end of the specified change date; synchronize all data where a) change date > timestamp at the end of the specified change date and b) change date between the two timestamps and UUID > lastSynchronizedUuidSameTimestamp
* Add extensive tests to make sure that this logic works as intended

### LINKED ISSUE #7521: Pull data to the mobile app in batches [5]
### Feature Description

Functional data (such as cases, contacts, persons, samples, etc.) is currently synchronised in one single request.

### Problem Description

Especially countries that use a single central SORMAS server run into significant problems here - not only during the initial synchronisation, but potentially also during other syncs, because there's too much functional data. Trying to pull this data from the server with low bandwidth can still lead to timeouts.

### Proposed Change (WIP)

Synchronise all production data in batches.

#### Calculate the approximate size of every entity
* For each entity that is synchronized this way (see `SychronizeDataAsync#synchronizeChangedData`, excluding outbreaks, disease configurations and customizable enum values), create an example entity on your local server (or a test server) with as much data filled in as possible; if there are sublists that are synchronized alongside the main entity (e.g. contact details for persons, exposures for cases or contacts), create a few entries in these lists as well.
* Do a ReST call that only retrieves this example entity (via Postman or in the browser); copy the returned JSON into a text file and note down the size of the text file in bytes.
* For each of these entities, add a final String to the DTO class that is called `APPROXIMATE_JSON_SIZE`, and set it to the size that you noted down for the respective entity.

#### Calculate the gzip compression factor
* Create ten cases on your local server (or a test server); the detail level of these cases doesn't matter, so you can use the dev mode data generator
* Do a ReST call that only retrieves these ten cases; copy the returned JSON into a text file and note down the size of the text file.
* Compress the text file using gzip with compression level 1; note down the size of the resulting compressed file.
* Calculate the compression factor by comparing the two file sizes, and store it in a final String called `JSON_COMPRESSION_FACTOR` in `SynchronizeDataAsync`.

#### Prepare the Android backend
* Change the synchronization methods to pull data for each individual entity until there's no more data left to be pulled
  * This can be determined by checking whether the number of returned entities < the requested number of entities
* When inserting pulled data, keep the change date of the last request, as well as the change date and UUID of the entity that was inserted last
  * When change date of the last request != change date of the last synchronized entity (i.e. change date of the new request), pull the next batch of data WITHOUT specifying the last synchronized UUID
  * When change date of the last request == change date of the last synchronized entity (i.e. change date of the new request), pull the next batch of data INCLUDING the last synchronized UUID
* Reduce the synchronization timeout to 5 minutes

#### Determine the size of the requested batch
To determine the `size` of the individual requests, perform the following steps:

* Request the current available bandwidth in kbps (see https://stackoverflow.com/questions/55159487/android-how-to-determine-network-speed-in-android-programmatically) and multiply it by 60 to see how many kb can approximately be transferred within one minute.
* Determine how many entities can be synchronized within one minute by dividing the available kbpm (kilobits per minute) by the APPROXIMATE_JSON_SIZE of the entity (make sure to convert this to kilobits) multiplied with the JSON_COMPRESSION_FACTOR; always use 10 as the minimum.

BarnaBartha:
Fixes #7521, fixes #7303


sormas-api/src/test/java/de/symeda/sormas/api/utils/SizeGenerator.java
```diff
@@ -0,0 +1,105 @@
...
+		});
+		factory.setStrategy(strategy);
+
+		for (Class<? extends EntityDto> dtoClass : dtoClasses) {
+			getSizeOfFilledDto(factory, dtoClass);
+		}
+	}
+
+	private static void getSizeOfFilledDto(PodamFactory factory, Class<? extends EntityDto> dtoClass) throws JsonProcessingException {
+		final EntityDto dto = factory.manufacturePojo(dtoClass);
```
MateStrysewske, line 99: Do you take compression factor into account? If not, that needs to be added because it reduces the size of the DTOs when they are transfered to the mobile app.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/AdoDtoHelper.java
```diff
@@ -61,62 +67,94 @@
...
+				: Integer.MAX_VALUE;
+			Integer lastBatchSize = batchSize;
+
+			lastSyncedEntityDate = maxModifiedDate;
+
+			while (lastBatchSize == batchSize) {
+				Call<List<DTO>> dtoCall = pullAllSince(
+					lastSyncedEntityDate.getTime(),
+					batchSize,
+					lastSyncedEntityDate.equals(previousLastSyncedEntityDate) ? EntityDto.NO_LAST_SYNCED_UUID : lastSyncedEntityUuid);
```
MateStrysewske, line 104: Do we really need to pass a constant here instead of null?

MateStrysewske (changes requested):
Please rename the `lastUuid` parameter in all ReST resources to `lastSynchronizedUuid` - `lastUuid` sounds like it specifies the last UUID that is supposed to be retrieved with the current request.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/AdoDtoHelper.java
BarnaBartha: yes, as it is a path parameter it cannot be optional

sormas-api/src/main/java/de/symeda/sormas/api/EntityDto.java
```diff
@@ -46,6 +46,7 @@
 	public static final String CREATION_DATE = "creationDate";
 	public static final String CHANGE_DATE = "changeDate";
 	public static final String UUID = "uuid";
+	public static final String NO_LAST_SYNCED_UUID = "NO_LAST_SYNCED_UUID";
```
MartinWahnschaffe, line 49: Why not use null and/or an empty string?

sormas-backend/src/main/java/de/symeda/sormas/backend/common/BaseAdoService.java
```diff
@@ -165,6 +169,16 @@ public Timestamp getLatestChangeDate() {
 		return em.createQuery(cq).getResultList();
 	}
 
+	public List<ADO> getBatchedQueryResults(CriteriaBuilder cb, CriteriaQuery<ADO> cq, From<?, ADO> from, Integer batchSize) {
+
+		if (batchSize != null) {
```
MartinWahnschaffe, line 174: Add a reference/explanation on why the sorting by uuid is needed.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/BaseAdoService.java
```diff
@@ -202,6 +216,20 @@ public Predicate createChangeDateFilter(CriteriaBuilder cb, From<?, ADO> from, D
...
 
+	public Predicate createChangeDateFilter(CriteriaBuilder cb, From<?, ADO> from, Date date, String lastSynchronizedUuidSameTimestamp) {
+		if (lastSynchronizedUuidSameTimestamp == null || EntityDto.NO_LAST_SYNCED_UUID.equals(lastSynchronizedUuidSameTimestamp)) {
+			return createChangeDateFilter(cb, from, date);
+		} else {
+			Timestamp timestamp = DateHelper.toTimestampUpper(date);
+			Predicate predicate = cb.or(
+				cb.greaterThan(from.get(AbstractDomainObject.CHANGE_DATE), timestamp),
+				cb.and(
+					cb.equal(from.get(AbstractDomainObject.CHANGE_DATE), timestamp),
```
MartinWahnschaffe, line 227: From my understanding this is not correct. Instead we need to take the "TimestampLower" (so basically the date) and say that the change date of the ADO is between lower and upper.
Here it's sufficient to check wether it's greaterOrEqual timestampLower

sormas-backend/src/main/java/de/symeda/sormas/backend/common/ChangeDateFilterBuilder.java
```diff
@@ -24,21 +25,32 @@
...
+
 	private ChangeDateFilterBuilder(CriteriaBuilder cb) {
 		this.cb = cb;
 		this.filters = Stream.builder();
 	}
 
 	public ChangeDateFilterBuilder(CriteriaBuilder cb, Date date) {
 		this(cb);
-		this.date = date;
+		this.date = new Timestamp(date.getTime());
```
MartinWahnschaffe, line 47: The date is passed in as timestamp (e.g. ContactService), but here only interpreted as date and then again cast to timestamp. Quite confusing to me. I think the date field should keep the Date type.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/ChangeDateFilterBuilder.java
```diff
@@ -61,10 +73,23 @@ public Predicate build() {
 			parent = parent.join(joinField, JoinType.LEFT);
 		}
 
+		Predicate filter;
 		if (dateExpression == null) {
```
MartinWahnschaffe, line 77: Same as in BaseAdoService createChangeDateFilter.

### First scenario: no uuid passed
change date has to be greater than change date (lower, so as it currently is)

### Second scenario: uuid passed
Behave as described in issue

sormas-backend/src/main/java/de/symeda/sormas/backend/common/BaseAdoService.java
```diff
@@ -165,6 +169,16 @@ public Timestamp getLatestChangeDate() {
 		return em.createQuery(cq).getResultList();
 	}
 
+	public List<ADO> getBatchedQueryResults(CriteriaBuilder cb, CriteriaQuery<ADO> cq, From<?, ADO> from, Integer batchSize) {
+
+		if (batchSize != null) {
+			cq.orderBy(cb.asc(from.get(AbstractDomainObject.CHANGE_DATE)), cb.asc(from.get(AbstractDomainObject.UUID)));
+		} else {
+			cq.orderBy(cb.desc(from.get(AbstractDomainObject.CHANGE_DATE)));
```
MartinWahnschaffe, line 177: Please also change this to ascending. I'm aware that this will change the old behaviour, but the old behaviour was acutally wrong and only worked because we have always synced the whole list.

MateStrysewske (changes requested):
see Martin's comments

sormas-api/src/main/java/de/symeda/sormas/api/EntityDto.java
syntakker: Empty or null Strings are not compatible with the REST API here.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/BaseAdoService.java
syntakker: Date comparisons in the `CriteriaBuilder` have precision milliseconds, so this fulfills the requirement.
Edge cases are modeled in `TaskFacadeEjbTest.testArchivedTaskNotGettingTransfered`.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/BaseAdoService.java
syntakker: done

sormas-backend/src/main/java/de/symeda/sormas/backend/common/ChangeDateFilterBuilder.java
syntakker: Date comparisons in the `CriteriaBuilder` have precision milliseconds, see above.
Edge cases are modeled in `ImmunizationFacadeEjbTest.testGetAllSince`.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/ChangeDateFilterBuilder.java
syntakker: The conversation to Timestamp is necessary and should be computed only once. Rename the date field?

sormas-backend/src/main/java/de/symeda/sormas/backend/common/BaseAdoService.java
syntakker: Refactored.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/ChangeDateFilterBuilder.java
syntakker: Refactored to use `timestampUpper` and `timestampLower`.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/ChangeDateFilterBuilder.java
syntakker: Refactored.

MartinWahnschaffe:
Code looks good to me now.

MateStrysewske (approved):

PR #7578, opened 2022-01-03
#6478 add frontend token retrieval

### LINKED ISSUE #6478: [External Journal] Add configurable frontend token authentication [2]
### Situation Description

For authorization in climedo, SORMAS fetches a token.

Now, PIA wants to use the external journal interface the same way climedo does, which includes opening a new tab when accessing PIA from within SORMAS (currently, there is an iFrame approach). 

The problem the PIA developers have with the current situation is that the token fetched as described above is written to the URL when opening the journal in a new tab. The token can be copied from there and then used for backend requests via ReST.

### Feature Description

To solve this problem, it was decided based on [this documentation](https://github.com/hzi-braunschweig/SORMAS-Project/issues/6175), that there shall be a frontend token introduced. 

For a maximum of compatibility, there shall be an option to configure the frontend authentication in the [sormas.properties](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-base/setup/sormas.properties):

- [ ] Modify the current properties file to show a `#interface.patientdiary.frontendAuthurl=`
- [ ] When this property is configured (not `null` and not empty), use this URL to authorize for frontend acces in the [ExternalJournalUtil.openPatientDiaryPage(...)](https://github.com/hzi-braunschweig/SORMAS-Project/blob/6afc9b772392dbb72f26cdc0c182512a547bc4e3/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/ExternalJournalUtil.java#L234). This should work work exactly as currently with the [ExternalJournalService.getPatientDiaryAuthToken()](https://github.com/hzi-braunschweig/SORMAS-Project/blob/6afc9b772392dbb72f26cdc0c182512a547bc4e3/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/ExternalJournalService.java#L142), just with the different URL. 
- [ ] Update the [documentation](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-rest/src/main/resources/workflow-documentation.properties)

When this property is not configured, use the current way to authenticate.
For backend requests, keep the current backend authentication.

### Additional Information

If I undestood correctly, the climedo developers also want to avoid using the backend token for frontend access in the future. Instead of using the frontend token, they want to use no token and instead provide a page with very limited rights to SORMAS. @DraganMileski, @mariusclimedo: Could you please confirm this or correct me if I'm wrong?
The easiest way to go is to keep the backend-token authentication approach when the `#interface.patientdiary.frontendAuthurl=`  is not configured, and then later completely disable this approach for frontend access. This would require a minimum of configuration changes on running instances, could be done once climedo is ready and would avoid undesired backend token retrieval even in misconfigured instances.

FredrikSchaeferVitagroup:
Fixes #6478

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/PatientDiaryClient.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/PatientDiaryClient.java
@@ -60,7 +60,8 @@ public class PatientDiaryClient {
 	public static final String MOBILE_PHONE_QUERY_PARAM = "Mobile phone";
 
 	private static final String PATIENT_DIARY_KEY = "patientDiary";
-	private static final Cache<String, String> authTokenCache = CacheBuilder.newBuilder().expireAfterWrite(6, TimeUnit.HOURS).build();
+	private static final Cache<String, String> backendAuthTokenCache = CacheBuilder.newBuilder().expireAfterWrite(6, TimeUnit.HOURS).build();
+	private static final Cache<String, String> frontendAuthTokenCache = CacheBuilder.newBuilder().expireAfterWrite(6, TimeUnit.HOURS).build();
 
 	@EJB
 	private ConfigFacadeEjb.ConfigFacadeEjbLocal configFacade;
@@ -197,7 +198,8 @@ public class PatientDiaryClient {
 			String encodedParams = URLEncoder.encode(queryParam, StandardCharsets.UTF_8.toString());
 			String fullUrl = probandsUrl + "?q=" + encodedParams;
 			Client client = ClientHelper.newBuilderWithProxy().build();
-			Response response = client.target(fullUrl).request(MediaType.APPLICATION_JSON).header("x-access-token", getPatientDiaryAuthToken()).get();
+			Response response =
+				client.target(fullUrl).request(MediaType.APPLICATION_JSON).header("x-access-token", getPatientDiaryAuthToken(false)).get();
 			if (response.getStatus() == HttpStatus.SC_NOT_FOUND) {
 				return Optional.empty();
 			}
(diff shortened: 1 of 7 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/externaljournal/PatientDiaryClient.java
```diff
@@ -211,31 +213,47 @@ public PatientDiaryResult deletePatientDiaryPerson(String personUuid, Runnable c
...
+	 * @param frontendRequest
+	 *            if true && a interface.patientdiary.frontendAuthurl is configured, this url is used to fetch the token.
+	 *            Otherwise, the interface.patientdiary.authurl is used.
 	 * @return the authentication token
 	 */
-	public String getPatientDiaryAuthToken() {
+	public String getPatientDiaryAuthToken(boolean frontendRequest) {
 		try {
-			return authTokenCache.get(PATIENT_DIARY_KEY, this::getPatientDiaryAuthTokenInternal);
+			if (frontendRequest && StringUtils.isNotBlank(configFacade.getPatientDiaryConfig().getFrontendAuthUrl())) {
```
vidi42, line 229: if  the request is made with `frontendRequest` set to `true`, but the `frontendAuthUrl` is not configured, this will fallback to use the `backendAuthToken`. Is this intended or should the call fail and report that the `frontendAuthUrl` is not configured?

I'm just not sure if this fallback is correct, because I guess there are different scenarios when the frontend authentication is used comparted to the backend authentication

FredrikSchaeferVitagroup: This is intended behaviour. It also says in the issue: `When this property is not configured, use the current way to authenticate.`

Background is that only the pia guys want the frontend token. Climedo does not. To not break the already existing instances, SORMAS needs to default to the previous behaviour when the frontend token is not configured...

vidi42: Yep, saw that comment too, but was not sure why this was needed. I get it now, thanks

vidi42 (approved):

PR #7596, opened 2022-01-05
Feature 7313 extend the bulk edit feature on the lab message

### LINKED ISSUE #7313: [DEMIS2SORMAS] Extend the bulk edit feature on the lab message directory [1]
### Situation Description

Currently, users have a hard time managing large numbers of incoming lab message. They can't process them fast enough and they struggle to report cases to higher health institutions in time. Thus, they should be able to assign lab messages to users to manage and distribute work more efficiently. #7310 will implement the overall assignment feature. However, it would be very useful to assign lab message in bulk because it saves time.

concept issue: #7308
### Feature Description

- Extend the bulk actions drop down on the lab message directory with the option 'Edit Assignee'
- When one or multiple lab messages have been selected and 'Assign' pressed, open the 'Edit assignee' pop-up.
    - After saving a selected assignee, set the assignee of all selected lab messages to the chosen assignee.
    - A selected assignee could also be "no assignee"? [first option]
- If the 'Assign' value is pressed but no lab messages was selected beforehand, display the following error message to the user: "_You have not selected any lab messages_"
### Additional Information

Lab messages Refinements_bulkAction

FredrikSchaeferVitagroup:
Fixes #7313


sormas-api/src/main/resources/strings.properties
```diff
@@ -986,6 +986,7 @@ messageImportPartiallySuccessful = <b>Import partially successful!</b><br/>The i
 messageImportSuccessful = <b>Import successful!</b><br/>All rows have been imported. You can now close this window.
 messageUploadSuccessful = Upload successful! You can now close this window.
 messageIncompleteGpsCoordinates = GPS coordinates are incomplete
+messageLabMessagesAssigned = All selected lab messages have been (un)assigned
```
leventegal-she, line 989: `(un)assigned` could be misleading, I think it would be better something like `The assignee has been changed for all selected lab messages`

leventegal-she (changes requested):

leventegal-she (approved):

PR #7601, opened 2022-01-05
#7300 epidemiological curve data endpoint

### LINKED ISSUE #7300: [Angular][Backend] Dashboard: Epidemiological curve [5]
### Situation Description

- In view of the future replacement of the current FE with the Angular component, the Angular team will focus on the required changes.

### Feature Description

- Reference: https://www.sketch.com/s/fd061796-65a3-4991-886a-c8a172667c91/a/Gm43r4D#Inspector
- The FE will send the BE one of (day, week, month) together with a start and end date
- The response should be grouped based on what param was sent (day, week, month) - this should already have the totals (counts) calculated 

### Possible Alternatives

- N/A

### Additional Information

- N/A

marius2301:
Fixes #7300


sormas-api/src/main/java/de/symeda/sormas/api/dashboard/EpiCurveGrouping.java
```diff
@@ -0,0 +1,7 @@
+package de.symeda.sormas.api.dashboard;
+
+public enum EpiCurveGrouping {
```
MateStrysewske, line 3: Please remove the enum from the UI and use this new enum instead (don't forget to add the toString method here)

MateStrysewske (changes requested):
I'm not a big fan of the introduced interface tbh because it doesn't add any real semantic logic (and is kind of misleading even, because PresentCondition does not have anything to do with case classifications) since the backend method returns the elements based on the selected epi curve mode anyway; I'd prefer to have two separate methods and endpoints instead, one for the case classification epi curve, and one for the present condition epi curve.

MateStrysewske (approved):

PR #7626, opened 2022-01-07
Feature #5703 option upload documents upon generation

### LINKED ISSUE #5703: Add option to directly upload documents upon generation [3]
### Situation Description

Users would like to immediately upload documents that they have generated for a respective entity.
### Feature Description

Add the option to directly upload the documents that are generated during the document generation for respective entities and bulk edit mode.
- Add a checkbox to the creation pop-up saying "Also upload the generated document to this entity" (Might need slight adjustment for bulk action: "Also upload the generated documents to the selected entities")
    - If selected, upload to respective entity
    - If unchecked, only generate
- If the document is larger than the file size limit, after the document has been generated, display a warning message to the user: "The document was successfully generated, but could not be uploaded to the entity because its file size exceeds the specified file size limit of {SIZE_LIMIT}."; in bulk mode, if at least one document exceeds the file size limit, display an actual popup window that contains the following elements:
  - Window caption: "Document upload warning"
  - Label: "The documents were successfully generated, but at least one document could not be uploaded to its entity because its file size exceeds the specified file size limit of {SIZE_LIMIT}."
  - Below that, another label: "Documents could not be uploaded to the following entities:"
  - Below that, a list of the UUIDs of the entities for which the upload failed
  - The window only needs one primary button with the caption "Okay" that closes the window

Default: false
### Additional Information

documentGenerationUpload

cazacmarin:
Fixes #5703


sormas-ui/src/main/java/de/symeda/sormas/ui/importer/DocumentUploadFinishedHandler.java
```diff
@@ -108,8 +110,10 @@ private void saveDocument(
 		String relatedEntityUuid,
 		byte[] bytes)
 		throws Exception {
+		if (isFileSizeLimitExceeded(length))
```
leventegal-she, line 113: would be better to make `FacadeProvider.getDocumentFacade().saveDocument` throw an exception if file is too big so there would be no need to check the file size in several places

sormas-backend/src/main/java/de/symeda/sormas/backend/docgeneration/DocGenerationHelper.java
```diff
@@ -0,0 +1,85 @@
...
+import de.symeda.sormas.api.caze.CaseReferenceDto;
+import de.symeda.sormas.api.contact.ContactReferenceDto;
+import de.symeda.sormas.api.document.DocumentDto;
+import de.symeda.sormas.api.document.DocumentRelatedEntityType;
+import de.symeda.sormas.api.event.EventReferenceDto;
+import de.symeda.sormas.api.utils.DataHelper;
+import de.symeda.sormas.backend.document.DocumentFacadeEjb;
+import de.symeda.sormas.backend.user.UserService;
+
+public class DocGenerationHelper {
```
leventegal-she, line 18: would be better to make this a stateless local bean so `userService` and `documentFacade` could be injected as EJB and no constructor would be needed

leventegal-she (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/importer/DocumentUploadFinishedHandler.java
cazacmarin: It cannot be done as the one remark from this feature is asking to upload all files and for failed one to raise a specific dialog where all UUIDs are present. As a generic implementation, same logic was left for a single file.

sormas-backend/src/main/java/de/symeda/sormas/backend/docgeneration/DocGenerationHelper.java
cazacmarin: It cannot be done due to the point of how CGLIB is working. They will always be null if it will not be called through an interface. Initially I did it in that way, but afterwards @EJB was not created and `null pointer exception` was raised.
image

leventegal-she (approved):

PR #7628, opened 2022-01-10
Bug 7496 prompt to update case disease variant is not shown when processing lab messages

### LINKED ISSUE #7496: Prompt to update case disease variant is not shown when processing lab messages 2
### Bug Description

This bug already occurred before in #5190

The pop-up that asks users to update the disease variant of a case when a positive pathogen test with a different disease variant is saved does not work when processing lab messages.

### Steps to Reproduce

1. Process a lab message
2. Enter a pathogen test with a different disease variant than the selected case and save

### Expected Behavior

The users should be asked to update the disease variant of the case
### Screenshots

Expected pop-up:
caseDiseaseVariantUpdateCaseMissing

caseDiseaseVariantUpdateCaseMissing2

### System Details

* Device:
* SORMAS version: 1.66.2
* Server URL: test1402.sormas.netzlink.com

FredrikSchaeferVitagroup:
Fixes #7496


sormas-ui/src/main/java/de/symeda/sormas/ui/samples/PathogenTestController.java
```diff
@@ -211,27 +220,38 @@ private void handleAssociatedCase(
...
-									() -> {
-										// Retrieve the case again because it might have changed
-										showConfirmCaseDialog(FacadeProvider.getCaseFacade().getByUuid(caze.getUuid()));
-									});
-							} else {
-								showConfirmCaseDialog(caze);
-							}
+
+				// the check is supposed to be skipped when a sample result update is denied
+				Consumer<Boolean> checkForDiseaseVariantUpdate = (skip) -> {
```
leventegal-she, line 231: it's a bit hard to read. I would move `checkForDiseaseVariantUpdate` into a separate method without the skip param. and add the skip check in a lambda when calling `showChangeAssociatedSampleResultDialog(dto, checkForDiseaseVariantUpdate);`

leventegal-she (changes requested):

leventegal-she (approved):

PR #7636, opened 2022-01-10
[#5065] remove jcenter

### LINKED ISSUE #5065: Replace jcenter Maven repository with an alternative
### Feature Description

jcenter is one of the repositories we're using both for the web app (sormas-base/pom.xml) as well as in the mobile application (gradle.build).

### Problem Description

Unfortunately, the repository will shut down by the end of April. While it will remain functional until February 2022, there will be no updates or new submissions to the repository.

### Proposed Change

We need to migrate to a different Maven repository that can replace jcenter. Before that, we need to make a decision which repository we can use instead.

### Additional Information

Some resources:
https://medium.com/codechai/jcenter-bintray-shutting-down-2e029d98a812
https://developer.android.com/studio/build/jcenter-migration
https://blog.gradle.org/jcenter-shutdown

### LINKED ISSUE #7513: Fix code scanning alert - Depending upon JCenter/Bintray as an artifact repository
Tracking issue for:
- [ ] https://github.com/hzi-braunschweig/SORMAS-Project/security/code-scanning/7

JonasCir:
Fixes #5065
Fixes #7513


MateStrysewske:
@JonasCir I'll review and test this PR after the next release because I don't want to replace the repository as early in a sprint as possible (in case we encounter any issues).

JonasCir:
@MateStrysewske Sounds good, perfectly fine with me :)

JonasCir:
If testing this, you might want to remove your .m2 folder in your home directory to simulate a clean build.

JonasCir:
@MateStrysewske ping :)

JonasCir:
@MateStrysewske Do you have time to take the review for this PR? :)

MateStrysewske:
@JonasCir Sorry for the late update; the problem is that we can't downgrade a dependency, especially not one that we can't test (which is the case here); A possible solution proposed by @StefanKock would be to still use jcenter as a fallback for this specific dependency.

JonasCir:
The `openbeans` dependency does not look to be maintained at all, last update from the original Google package is a couple of years old. Maybe its worth to search for a replacement than depending on legacy code.

JonasCir:
I think it is worth to replace the dependency then, also read [here](https://github.com/dm77/barcodescanner#project-archived). The project is not maintained and alternatives are proposed. I think we  should invest in healthy dependencies instead of depending on unmaintained ones.

JonasCir:
@MateStrysewske Not certain if this is the correct repo associated with the library: https://github.com/dm77/barcodescanner .
If yes, it is not maintained anymore and it proposes alternatives. I think we should did this dependency then as it is legacy software + it is served from legacy registry.

MateStrysewske:
@JonasCir The dependency is actively used in the mobile app and we don't have the tools to find an alternative for it ourselves, but would need to get in touch with the developers from Ghana to have a look at it. In any way we can't downgrade it in the meantime because we have no way of ensuring that the scanning mechanic will still work afterwards.

vidi42:
The repository for barcode scanning is not actively supported anymore https://github.com/dm77/barcodescanner so I think there's not much luck in finding the dependency elsewhere then jcenter.

Could we just leave `jcenter` as well in the declared repositories alongside `mavenCentral` and open a new ticket to replace this library with an actively supported one?

JonasCir:
Sure I'm also in favor of replacing it :+1:

StefanKock:
Then let's postpone this ticket and replace the barcodescanner functionality first.

JonasCir:
See #7872

JonasCir:
I'll keep the branch though till this is resolved :D

PR #7655, opened 2022-01-12
#7575: Documented how to Migrate to new Payara Server

### LINKED ISSUE #7575: Simplify Payara domain migration
### Feature Description

Without using `SORMAS-Docker`, a Payara domain is set up and updated with shell scripts.

### Problem Description

With #4601, we need to set up a fresh Payara domain **and** migrate existing data and configuration. But this approach is not covered by the existing `server-setup.sh` and `server-update.sh` nor with existing documentation.

### Proposed Change

* Document what needs to be considered when setting a new Payara domain for an existing instance: What and how to migrate.
  * Stop the running payara domain, move existing domain to another path.
  * Set up the fresh payara domain with `server-setup.sh`. Reuse the same domain name and paths to directories as before, also databases.
  * Transfer your settings from `sormas.properties`, `logback.xml` or changes in the domain setup. Use the new provided files and copy your changes in, don't reuse old files!
* Have this ticket done when #4601 is merged to ease the migration.
* Integrate the migration hints into SERVER_UPDATE.md, link to relevant parts of SERVER_SETUP.md .

### Additional Information

Scripting this or having proper documentation helps every developer and every hoster who is not using the dockerised setup.

StefanKock:
Fixes #7575

### DIFF
--- a/docs/SERVER_UPDATE.md
+++ b/docs/SERVER_UPDATE.md
@@ -2,6 +2,33 @@
 
 SORMAS releases starting from 1.21.0 contain a script that automatically updates and deploys the server. If you are using an older version and therefore need to do a manual server update, please download the 1.21.0 release files and use the commands specified in the server-update.sh script.
 
+## Migrate to new Payara Server
+At some versions it is mandatory to switch to a new Payara Server:
+* Switching from <=v1.66.4 to v1.67.0 or newer
+
+If your version bump does not apply to the listing above, please proceed with [Preparations](SERVER_UPDATE.md#preparations).
+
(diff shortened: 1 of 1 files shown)


MartinWahnschaffe (changes requested):
The migration guide looks good to me.

I don't think it should be at the top of the server update guide, though.

Please move it down and I'd suggest to add one point "migrate to new payara server" to the start of the update guide. This point should link to the migration chapter down in the document. I'd also keep lines 6 and 7 at the top of the document (versions that need the migration).

PR #7707, opened 2022-01-18
Change/7698 final tweaks

### LINKED ISSUE #7698: S2S/Central final tweaks
### Feature Description
S2S and central infra sync.

### Problem Description
While testing, I found two issues.

- [x] We need to support merging while syncing as we have existing data in the DB.
- [x] `toString` is missing on some infra Dtos which makes the logs less readable
- [x] Do not consider archive data for duplicates 

### Proposed Change
Use merging whale sync saving and adjust toString

JonasCir:
Fixes #7698 
This PR excludes archived entities from the merge step (we run a python script before which archives all invalid data). Further, we ignore the change data check as it does not make sense in our scenario. Merging an entity will always bump the change data what causes it to be always greater than the value in central.


JonasCir:
I tested this PR against a test-de2 DB dump locally and all the migration etc. worked :tada:

MateStrysewske (approved):

PR #7715, opened 2022-01-19
#7521 - Introduced upper batch limit and batching for person queries

### LINKED ISSUE #7521: Pull data to the mobile app in batches [5]

MateStrysewske:
Fixes #7521


sormas-app/app/src/main/java/de/symeda/sormas/app/rest/RetroProvider.java
```diff
@@ -308,7 +309,10 @@ public static Integer getNumberOfEntitiesToBePulledInOneBatch(long approximateJs
 		double compressedJsonSizeInBits = approximateJsonSizeInBytes * 8 / JSON_COMPRESSION_FACTOR;
 		int batchSize =
 			Math.toIntExact(Math.round(getNetworkDownloadSpeedInKbps(context) * ASSUMED_TRANSFER_TIME_IN_SECONDS * 1024 / compressedJsonSizeInBits));
-		return batchSize < 10 ? 10 : batchSize;
+		// Restrict the batch size to a maximum value to avoid backend queries leading to a timeout
+		batchSize = Math.min(MAX_BATCH_SIZE, Math.max(10, batchSize));
+		batchSize = (int) (10 * Math.sqrt(batchSize / 10f));
+		return batchSize;
```
MartinWahnschaffe, line 315: Limiting the batch size to 500 has to be done last!

MateStrysewske: True, no idea why I wrote it like that :D

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
```diff
@@ -479,7 +479,7 @@ public Predicate buildCriteriaFilter(PersonCriteria personCriteria, PersonQueryC
 				immunizationPersonsQuery.where(immunizationPersonsFilter);
 			}
 			immunizationPersonsQuery.distinct(true);
-			immunizationPersonsResultList = em.createQuery(immunizationPersonsQuery).getResultList();
+			immunizationPersonsResultList = getBatchedQueryResults(cb, immunizationPersonsQuery, personsRoot, batchSize);
 		}
 
 		List<Person> travelEntryPersonsResultList = new ArrayList<>();
```
MartinWahnschaffe, line 485: please also fix the travel entry batch size comment

MartinWahnschaffe (approved):

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
```diff
@@ -409,7 +409,7 @@ public Predicate buildCriteriaFilter(PersonCriteria personCriteria, PersonQueryC
 			casePersonsQuery.where(casePersonsFilter);
 		}
 		casePersonsQuery.distinct(true);
-		List<Person> casePersonsResultList = em.createQuery(casePersonsQuery).getResultList();
+		List<Person> casePersonsResultList = getBatchedQueryResults(cb, casePersonsQuery, personsRoot, batchSize);
```
syntakker, line 412: `personsRoot` => `casePersonsRoot`

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
```diff
@@ -431,7 +431,7 @@ public Predicate buildCriteriaFilter(PersonCriteria personCriteria, PersonQueryC
 			contactPersonsQuery.where(contactPersonsFilter);
 		}
 		contactPersonsQuery.distinct(true);
-		List<Person> contactPersonsResultList = em.createQuery(contactPersonsQuery).getResultList();
+		List<Person> contactPersonsResultList = getBatchedQueryResults(cb, contactPersonsQuery, personsRoot, batchSize);
```
syntakker, line 434: `personsRoot` => `contactPersonsRoot`

MartinWahnschaffe (approved):

Issue #7694, opened 2022-01-17
[Angular][Backend] Map endpoints [2]


laurentiustanvitagroup:
### Situation Description

- In view of the future replacement of the current FE with the Angular component, the Angular team will focus on the required changes.

### Feature Description

- Reference: 
- We need to create an endpoint for the map section in the Dashboard and to return the relevant information for the Angular component.

### Possible Alternatives

- N/A

### Additional Information

- N/A

PR #7742, opened 2022-01-21
#7694 endpoints for map

marius2301:
Fixes #


sormas-api/src/main/java/de/symeda/sormas/api/dashboard/CaseDataTypeDashboardCriteria.java
```diff
@@ -0,0 +1,13 @@
+package de.symeda.sormas.api.dashboard;
+
+import de.symeda.sormas.api.caze.NewCaseDateType;
+
+public class CaseDataTypeDashboardCriteria extends DashboardCriteria {
```
MateStrysewske, line 5: I think it would make more sense to call this `SurveillanceDashboardCriteria` so that we could use it for additional properties that are only relevant for the surveillance dashboard in the future.

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #7755, opened 2022-01-24
#[7752] fix warning about array type inference

### LINKED ISSUE #7752: CI failure on development
### Bug Description
See last [run](https://github.com/hzi-braunschweig/SORMAS-Project/runs/4899566809?check_suite_focus=true#step:7:882) on `development`.

JonasCir:
Fixes #7752 
Interestingly, I could not reproduce the failure locally, but if IIRC I saw this error last week in one of my branches.

IDE showed some warnings that missing array tape might cause trouble, so refactored the call site a little bit.
While I was at it, I also fixed some low hanging warnings.

Cheers

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleFacadeEjb.java
@@ -1,24 +1,20 @@
-/*******************************************************************************
+/*
  * SORMAS® - Surveillance Outbreak Response Management & Analysis System
- * Copyright © 2016-2018 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
- *
+ * Copyright © 2016-2022 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleFacadeEjb.java
```diff
@@ -979,18 +979,18 @@ private void onSampleChanged(SampleDto existingSample, Sample newSample, boolean
 
 	private void handleAssotiatedObjectChanges(Sample newSample, boolean syncShares) {
 		if (newSample.getAssociatedCase() != null) {
-			caseFacade.onCaseChanged(CaseFacadeEjbLocal.toDto(newSample.getAssociatedCase()), newSample.getAssociatedCase(), syncShares);
+			caseFacade.onCaseChanged(CaseFacadeEjb.toDto(newSample.getAssociatedCase()), newSample.getAssociatedCase(), syncShares);
```
MateStrysewske, line 982: Afaik there's a reason we're using the local bean here; @MartinWahnschaffe can you share some insight?

JonasCir: I think the warning originated from the usage of a static function on the extended class.

MartinWahnschaffe: What you have in mind is that when using an instance of an EJB in the backend, you have to use the "local" instance.

The toDto method is a static method though, so there is no difference between using the static method via the class name of the ejb or any inherited class. As Jonas points out, the system will give a warning that says we should use the class that acutally defines the static method.

TL;DR: The refactoring is good the way it is.

JonasCir: :+1:

MateStrysewske (approved):

PR #7778, opened 2022-01-26
Added test scenario for creating case from event participant

mp-sgent:


mp-sgent:
image

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CreateNewCaseSteps.java
```diff
@@ -141,6 +141,30 @@ public CreateNewCaseSteps(WebDriverHelpers webDriverHelpers, CaseService caseSer
...
+          selectPlaceOfStay(caze.getPlaceOfStay());
+
+          fillDateOfBirth(caze.getDateOfBirth());
+          selectPresentConditionOfPerson(caze.getPresentConditionOfPerson());
+          fillDateOfSymptomOnset(caze.getDateOfSymptomOnset());
+          fillPrimaryPhoneNumber(caze.getPrimaryPhoneNumber());
+          fillPrimaryEmailAddress(caze.getPrimaryEmailAddress());
+          fillDateOfReport(caze.getDateOfReport());
+          fillPlaceDescription(caze.getPlaceDescription());
+          webDriverHelpers.clickOnWebElementBySelector(SAVE_BUTTON);
```
rdutu-vg, line 163: let's move out the save logic in order to make the code re-usable (is in plan to do with all BDD classes)
And rename the name of this method, like I fill all fields for a new case created for event participant

sormas-e2e-tests/src/main/java/org/sormas/e2etests/pages/application/events/EventDirectoryPage.java
```diff
@@ -37,6 +37,7 @@
       By.cssSelector("div.v-grid-tablewrapper tbody tr");
   public static final By EVENT_ACTIONS_TABLE_DATA = By.tagName("td");
   public static final By FIRST_EVENT_ID_BUTTON = By.cssSelector(".v-grid-row-has-data a[title]");
+  public static final By CREATE_CASE_BUTTON= By.xpath("//a[@title='Create new case']");
```
rdutu-vg, line 40: I know that this approach is common among the project, but we can improve it a bit by adding other identifier, because if we'll start using the tests on different markets, the language will be different and will fail

mp-sgent:
image
New allure report with separated steps

PR #7797, opened 2022-01-27
Field ID for Data Dictionary

### LINKED ISSUE #7680: Data dictionary should have a new column "fieldId" that is the combination of dto and field name [1]
### Feature Description
Currently the data dictionary has one sheet per entity that is using the caption of the dto and one column with the technial field name

### Problem Description
This makes it a bit difficult to read the data dictionary and map the provided information to the related field, as we would like to do in #7679. Also it would be good to allow users to put information together into one sheet (e.g. case data + symptoms of a case), but still have the ability the match each field technically.

### Proposed Change

- [x] Add a new first column called "fieldId" to the data dictionary.
- [x] The values are the combination of the DTO name (without the Dto post-fix) with the field name. E.g. CaseData.caseClassification
- [ ] If easily doable, the fieldId should be hidden by default

### Possible Alternatives

Discuss wether using the DTO name is the best approach here.

vidi42:
Hiding the column is easy, however there are multiple tables in one sheet regarding enums and facility references which get hidden as well. And shifting these tables one cell to the right it's not that easy (I couldn't find a way). So the "Field ID" column is visible.

Fixes #7680


sormas-backend/src/main/java/de/symeda/sormas/backend/info/InfoFacadeEjb.java
```diff
@@ -241,6 +243,10 @@ private void createEntitySheet(XSSFWorkbook workbook, Class<? extends EntityDto>
 				continue;
 			XSSFRow row = sheet.createRow(rowNumber++);
 
+			// filedId
+			XSSFCell fieldIdCell = row.createCell(EntityColumn.FIELD_ID.ordinal());
+			fieldIdCell.setCellValue(i18nPrefix + "." + field.getName());
```
leventegal-she, line 248: I would say `DataHelper.getHumanClassName(entityClass)` instead of `i18nPrefix + "." + field.getName()` because `getHumanClassName` is used when generating the import template files and it's also used in other import related code.

leventegal-she (changes requested):

leventegal-she (approved):

PR #7806, opened 2022-01-28
Feature 7008 add configurations to entities for automated deletion

### LINKED ISSUE #7008: Add configurations for each entity to trigger automated deletion [5]
### Situation Description
In order to comply with the requirements of data protection SORMAS has to provide a process to automatically delete entities and referenced objects. To be able to decide, when an entity and the referenced object are deleted, we have to create certain new fields.

### Feature Description

**Deletion** is done based on the SORMAS core entities:
**Case, Contact, Event, Event Participant, Immunization, Travel Entry**

Deletion takes place **relative to one of the following three events/dates**

1. Entity was created in the system
2. Date of origin (e.g. date of arrival for a travel entry) 
3. End of processing (last change date for now)

A **deletion period** in days relative to this date is defined for each core entity (later it can also be defined for each field individually).

A **nightly job will identify all entities** that need to be deleted and **trigger their deletion**. The actual deletion will be defined in a follow-up issue.

The entity configuration in regards to deletion will not be done in the UI but in the server configurations.

### Examples

* A travel entry is deleted 14 days after it's date of origin
* A case is deleted 10 years after it's end of processing

### Implementation

- [x] New DeletionConfiguration ADO (with respective Service and Facade) with the following fields:
  * entity: Name of the entity to be deleted. An enum called CoreEntityType: CASE, CONTACT, EVENT, EVENT_PARTICIPANT, IMMUNIZATION, TRAVEL_ENTRY
  * deletionReference enum value. One of: CREATION, ORIGIN, END
    * ORIGIN is currently only supported for TravelEntry. Use the report date now. There will be a separate issue to include the date of arrival in SORMAS
  * deletionPeriod: Number of days until deletion, validate to be [7,MAXINT]
- [x] A new abstract class "CoreEntityFacade" (feel free to find a better name) that provides an executeAutomaticDeletion method with a generic implementation. The method identifies all instances of one entity type that need to be deleted and calls the delete method. For deletion use the existing delete... methods for now, based on uuid or ReferenceDto. 
- [x] A job in the cron service class that executes this for all entities named above once every night.

*Note:* Make sure to implement this as generic as possible.

### Follow-up issues / out of scope

* UI that displays information on when an entity is going to be deleted #7240 
* Introduce an "end of processing date" #7247 and re-fine the process of archiving/closing #7249
* For deletion we re-use the existing soft-delete and will have a separate issue to do an actual deletion
* Field-based deletion. For entity types it must be possible to document several deletion configurations.
* Deletion after hand-over to other health-departments

sergiupacurariu:
Fixes #7008


sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/CoreEntityType.java
```diff
@@ -0,0 +1,30 @@
+package de.symeda.sormas.backend.deletionconfiguration;
+
+import de.symeda.sormas.api.deletionconfiguration.DeletionReference;
+
+public enum CoreEntityType {
+
+	//ENTITY LIST
+	CASE(DeletionReference.CREATION, 3650),
```
MartinWahnschaffe, line 8: Don't define default deletion periods & references.
Automatic deletion will have to be configured by someone on database level in order to work.

sergiupacurariu: updated

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/AbstractCoreEntityFacade.java
```diff
@@ -0,0 +1,55 @@
...
+		cq.where(cb.lessThanOrEqualTo(from.get(getDeleteReferenceField(deletionReference)), referenceDeletionDate));
+
+		List<T> toDeleteEntities = QueryHelper.getResultList(em, cq, null, null);
+
+		toDeleteEntities.forEach(entity -> {
+			permanentDelete(entity);
+		});
+	}
+
+	protected abstract void permanentDelete(T entity);
```
MartinWahnschaffe, line 43: I think this method should just be called delete. Whether it's doing a soft-delete or later a permant (hard) delete should not be part of this.

sergiupacurariu: changed

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/DeletionConfigurationService.java
```diff
@@ -0,0 +1,53 @@
...
+
+		CriteriaBuilder cb = em.getCriteriaBuilder();
+		CriteriaQuery<DeletionConfiguration> cq = cb.createQuery(getElementClass());
+		Root<DeletionConfiguration> from = cq.from(getElementClass());
+		cq.where(cb.equal(from.get(DeletionConfiguration.ENTITY_TYPE), coreEntityType));
+
+		return em.createQuery(cq).getSingleResult();
+	}
+
+	public void createMissingDeletionConfiguration() {
```
MateStrysewske, line 36: Not quite sure
a) how this is supposed to work since the `build` method doesn't set a reference and period, but both are supposed to be not nullable
b) whether we want something like this in the first place because in that case, we'd also need to define default values ourselves and I'm not sure whether that's a good idea (since countries might not want to use this feature at all, or might have other regulations than those we define)

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -9798,4 +9798,27 @@ ALTER TABLE weeklyreportentry_history ALTER COLUMN creationdate TYPE timestamp(3
...
 INSERT INTO schema_version (version_number, comment) VALUES (436, 'Set timestamp precision to milliseconds #7303');
 
+-- 2022-01-25 Add configurations for each entity to trigger automated deletion #7008
+CREATE TABLE deletionconfiguration(
+                                     id bigint not null,
+                                     uuid varchar(36) not null unique,
+                                     changedate timestamp not null,
+                                     creationdate timestamp not null,
+                                     entityType varchar(255),
+                                     deletionReference varchar(255),
```
MateStrysewske, line 9808: Inconsistent not null constraint definitions compared to the entity definition for both deletionReference and deletionPeriod

sormas-backend/src/test/java/de/symeda/sormas/backend/AbstractBeanTest.java
```diff
@@ -786,4 +788,12 @@ public SormasToSormasLabMessageDtoValidator getSormasToSormasLabMessageDtoValida
 	public DefaultEntitiesCreator getDefaultEntitiesCreator() {
 		return getBean(DefaultEntitiesCreator.class);
 	}
+
+	public CoreEntityDeletionService getCoreEntityDeletionService(){
```
MateStrysewske, line 792: Formatting is off here as well

sormas-backend/src/test/java/de/symeda/sormas/backend/TestDataCreator.java
```diff
@@ -445,6 +445,8 @@ public CaseDataDto createCase(
 		return caze;
 	}
 
+
```
MateStrysewske, line 448: Unnecessary new lines

sormas-backend/src/test/java/de/symeda/sormas/backend/deletionconfiguration/CoreEntityDeletionServiceTest.java
```diff
@@ -0,0 +1,89 @@
...
+	private void createDeletionConfiguration() {
+		build(CoreEntityType.CASE);
+		build(CoreEntityType.CONTACT);
+		build(CoreEntityType.EVENT);
+		build(CoreEntityType.EVENT_PARTICIPANT);
+		build(CoreEntityType.IMMUNIZATION);
+		build(CoreEntityType.TRAVEL_ENTRY);
+	}
+
+	private DeletionConfiguration build(CoreEntityType coreEntityType) {
```
MateStrysewske, line 79: Rename to `create` because `build` suggests that the ADO is only built, but not persisted

sormas-backend/src/test/java/de/symeda/sormas/backend/deletionconfiguration/CoreEntityDeletionServiceTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+		assertEquals(1, getCaseFacade().count(caseCriteria));
+
+		getCoreEntityDeletionService().executeAutomaticDeletion();
+
+		assertEquals(0, getCaseFacade().count(caseCriteria));
+
+	}
+
+	private void createDeletionConfiguration() {
```
MateStrysewske, line 70:
```suggestion
	private void createDeletionConfigurations() {
```

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/CoreEntityDeletionService.java
```diff
@@ -0,0 +1,75 @@
...
+		coreEntityFacades.add(EntityTypeFacadePair.of(CoreEntityType.TRAVEL_ENTRY, travelEntryFacadeEjb));
+	}
+
+	public void executeAutomaticDeletion() {
+
+		coreEntityFacades.forEach(coreEntityType -> {
+			DeletionConfiguration coreEntityTypeConfig = deletionConfigurationService.getCoreEntityTypeConfig(coreEntityType.coreEntityType);
+
+			if (coreEntityTypeConfig.deletionReference!=null && coreEntityTypeConfig.deletionPeriod != null) {
+				Date referenceDeletionDate = DateHelper.subtractDays(new Date(), coreEntityTypeConfig.deletionPeriod);
```
MateStrysewske, line 54: This logic seems a bit unintuitive for me, and I actually think that calling the result of this calculation referenceDeletionDate is incorrect. Wouldn't it make more sense to pass the deletionPeriod to executeAutomaticDeletion, and then do the calculation in that method?

MateStrysewske (changes requested):
@MartinWahnschaffe Please have a look at my change requests and check whether they seem legit to you.

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/AbstractCoreEntityFacade.java
```diff
@@ -0,0 +1,55 @@
...
+		this.entityClass = entityClass;
+	}
+
+	public void executeAutomaticDeletion(DeletionReference deletionReference, Date referenceDeletionDate) {
+
+		CriteriaBuilder cb = em.getCriteriaBuilder();
+		CriteriaQuery<T> cq = cb.createQuery(entityClass);
+		Root<T> from = cq.from(entityClass);
+
+		cq.where(cb.lessThanOrEqualTo(from.get(getDeleteReferenceField(deletionReference)), referenceDeletionDate));
```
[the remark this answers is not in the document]
MartinWahnschaffe: No, before - it has to be older.

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/AbstractCoreEntityFacade.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
+
+import de.symeda.sormas.api.deletionconfiguration.DeletionReference;
+import de.symeda.sormas.backend.common.AbstractDomainObject;
+import de.symeda.sormas.backend.util.ModelConstants;
+import de.symeda.sormas.backend.util.QueryHelper;
+
+public abstract class AbstractCoreEntityFacade<T extends AbstractDomainObject> {
```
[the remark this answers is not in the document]
sergiupacurariu: I changed so T extends CoreAdo. I think that will remain even after refactoring as CoreAdo extends DeletableAdo

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/AbstractCoreEntityFacade.java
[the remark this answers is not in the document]
sergiupacurariu: here lessThanOrEqualTo means that let's say the case creation date is before the reference date. So we need to delete the cases older than the reference date. I retested the expression and seems right to me.

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/CoreEntityDeletionService.java
```diff
@@ -0,0 +1,75 @@
...
+		coreEntityFacades.add(EntityTypeFacadePair.of(CoreEntityType.IMMUNIZATION, immunizationFacadeEjb));
+		coreEntityFacades.add(EntityTypeFacadePair.of(CoreEntityType.TRAVEL_ENTRY, travelEntryFacadeEjb));
+	}
+
+	public void executeAutomaticDeletion() {
+
+		coreEntityFacades.forEach(coreEntityType -> {
+			DeletionConfiguration coreEntityTypeConfig = deletionConfigurationService.getCoreEntityTypeConfig(coreEntityType.coreEntityType);
+
+			if (coreEntityTypeConfig.deletionReference!=null && coreEntityTypeConfig.deletionPeriod != null) {
```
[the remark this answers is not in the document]
sergiupacurariu: formatted

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/DeletionConfigurationService.java
sergiupacurariu: Initially, the implementation was to create a default DeleteConfiguration for each Core Entity. In the review notes above I received the recommendation to remove default deletion reference and deletion period. 
Now it works like this, on startup if a DeletetionConfiguration is missing of a CoreEntity it will be added have DeletionReference and DeletionPeriod null. When these 2 values are not defined the deletion code is not executed. This check is in CoreEntityDeletionService.

sormas-backend/src/main/resources/sql/sormas_schema.sql
sergiupacurariu: since the deletion reference and deletion period could be undefined "nullable = false" was removed from the respective fields in the entity class DeletionConfiguration

sormas-backend/src/test/java/de/symeda/sormas/backend/AbstractBeanTest.java
sergiupacurariu: formatted

sormas-backend/src/test/java/de/symeda/sormas/backend/deletionconfiguration/CoreEntityDeletionServiceTest.java
sergiupacurariu: renamed

sormas-backend/src/test/java/de/symeda/sormas/backend/deletionconfiguration/CoreEntityDeletionServiceTest.java
sergiupacurariu: renamed

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/AbstractCoreEntityFacade.java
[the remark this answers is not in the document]
MateStrysewske: Understood now, thanks!

MateStrysewske (approved):

PR #7809, opened 2022-01-28
Bugfix 7793 multiple resulting cases for same contact

### LINKED ISSUE #7793: Multiple 'Resulting Cases' can be created for the same contact
### Bug Description
A user is able to create multiple 'Resulting Cases' for a contact that has a sample with a positive pathogen test result.

### Steps to Reproduce

1. Create a contact with disease COVID-19 (report and last contact date 1/1/22);
2. Add a sample with a positive final laboratory result and positive pathogen test result for COVID-19 (date sample was collected 1/1/22);
3. Saving the sample triggers a pop-up that asks the user whether they would like to create a 'Resulting Case' for this contact;
5. Click 'Yes' and 'Create new case' form appears;
6. Create the new case ( date of report 1/1/22) and save the case;
7. Go to the 'Contacts directory' and select the contact from step 1;
8. Go to the sample and test result that were created on step 2;
9. Click to edit the result, do not edit any field and save the test again (or create a new positive test);
10. Pop-up is triggered again asking the user whether they would like to create a 'Resulting Case' for this contact - click 'Yes';
11. A second pop-up is triggered - select 'Create new case' and click 'Confirm'.

### Actual Behavior
User is able to create multiple 'Resulting Cases' for the same contact.

### Expected Behavior
User is not able to create more than one 'Resulting Case' for a contact.

### Screenshots
GIF of the steps to reproduce:
Resulting Case1

### System Details

* Device: Windows 10
* SORMAS version: 1.68.0-SNAPSHOT
* Server URL: test.sormas.netzlink.com; test-de2.sormas.netzlink.com
* User Role: Admin, NatUser

### Additional Information
Issue was observed while doing exploratory testing regarding the triggering of the duplicate detection.

alexcaruntu-vita:
Fixes #7793


MateStrysewske (changes requested):
The bug isn't really fixed because users can still set the contact classification back to unconfirmed and then create another resulting case by saving the pathogen test; the fix should make sure that it is absolutely impossible to convert a contact to a case if it already has a resulting case set.

MateStrysewske (changes requested):
The popup that asks me to convert the contact to a case still appears for me (e.g. when simply saving a positive pathogen test again without doing any changes), even if the contact is confirmed and has contact status converted to case. Please make the necessary changes and make sure to test that the bug doesn't appear anymore.

PR #7810, opened 2022-01-31
#7458 remove profile with keycloak

### LINKED ISSUE #7458: Remove maven profile `with-keycloak` [1]
### Problem Description

With #3602 the module `sormas-keycloak-service-provider` was introduced, and not with automatic integration but with a distinct profile `with-keycloak`. Since this is not considered in all release processes, the POM of that module does not get updated in hotfixes.

Additionally, I don't see a point why this module is not included per default.

### Proposed Change

- [x] Add `sormas-keycloak-service-provider` as default module and remove the profile `with-keycloak` in the base POM
- [ ] Remove usage of profile in Jenkins job `sormas-Build`, document for `sormas-Release` that profile is removed with 1.68.0
- [x] Check that removed profile is not mentioned in any documentation (anymore).

### Additional Information

The flawed dependency was changed in hotfix 1.66.1 manually and caused the release build to fail.

StefanKock:
Fixes #7458


JonasCir:
Just looking at the PR, I would not really know how to build Keycloak and the App anymore. I think noting down the necessary `mvn` command somewhere in the build instructions would help :rocket:

StefanKock:
> Just looking at the PR, I would not really know how to build Keycloak and the App anymore. I think noting down the necessary `mvn` command somewhere in the build instructions would help 🚀

The Ant target `install` and `install-with-app` are still there as they were since years.

vidi42 (approved):

JonasCir:
@StefanKock Alright thanks for the clarification :+1:

PR #7821, opened 2022-01-31
[#7813] cache current user

### LINKED ISSUE #7813: Investigate usage of `getCurrentUser` and how to attach the current user to the session to avoid BD round trips
### Feature Description
`BaseFacade::getCurrentUser` and `CurrentUserService`.

### Problem Description
Right now, it seems that we have a lot of DB round trips to map the username in the session to the user ID.

### Proposed Change
Resolve the username once per active session.

JonasCir:
Fixes #7813


JonasCir:
@vidi42 
This is still WIP with test failures but I would appreciate a first round of feedback :)

sormas-backend/src/main/java/de/symeda/sormas/backend/user/CurrentUserCache.java
```diff
@@ -0,0 +1,54 @@
...
+		}
+		return instance;
+	}
+
+	public void flush() {
+		cache.clear();
+	}
+
+	public void put(String name, CurrentUser user) {
+		cache.put(name, user);
```
vidi42, line 44: When using Keycloak the username we receive from Keycloak is always with lowercase letters, however the username in the DB is still stored in PascalCase format, which might introduce some conflicts for example when removing a user from the cache.
For example, the callerPrincipal from Keycloak will be "natuser", but when you update the user, you want to remove it from cache, so it will try to remove "NatUser" which it won't find.
My suggestion is to use `AuthProvider.isUsernameCaseSensitive` which if it's true, then convert all usernames to "lowercase" before `putting`, `getting`, `removing` them from the map

vidi42 (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/user/CurrentUserCache.java
JonasCir: Hi Alex I filed https://github.com/hzi-braunschweig/SORMAS-Security/issues/39 to discuss this further. Currently, I don't want to merge the PR as I do not want to introduce a vulnerability here. I'll leave it as draft for now till the concerns I have are gone :smiley:

JonasCir: Resolved by #7907

JonasCir:
@vidi42 Rebased on latest `development`. Can be reviewed now, however, I would like to wait with merging it after I finished #7835 , I really want to be able to measure the changes to see if we can use this approach more often.

JonasCir:
@vidi42 Please just merge directly, I do not have time to get the performance testing up and running

vidi42 (approved):

PR #7823, opened 2022-01-31
Send a notification when reassigning a task #7383

### LINKED ISSUE #7383: Notifications: send a new notification in case of reassignment of a task
### Situation Description
It is possible to reassign a task to another user. However, this reassignment does not cause the system to send a notification to the new user in charge of the task.

### Feature Description
We suggest modifying the code so that the new user in charge of the task is informed by email of the new task to be performed.
At the same time, a notification is sent to the former user in charge of the task to inform him that he is no longer in charge of this task.

### Additional Information

aponcon-atolcd:
Fixes #7383


sormas-api/src/main/resources/strings.properties
```diff
@@ -1194,6 +1194,10 @@ notificationTaskAssociatedCaseLink = Link to the associated case: %s
 notificationTaskAssociatedContactLink = Link to the associated contact: %s
 notificationTaskAssociatedEventLink = Link to the associated event: %s
 notificationTaskAssociatedTravelEntryLink = Link to the associated travel entry: %s
+notificationTaskGeneralUpdatedAssigneeUserSource = Your %s task has been assigned to another user. You are no longer in charge of this task.
+notificationTaskGeneralUpdatedAssigneeUserTarget = The %s task has been assigned to you.
```
MateStrysewske, line 1198:
```suggestion
notificationTaskGeneralUpdatedAssigneeUserTarget = A(n) %s task has been assigned to you.
```

sormas-api/src/main/resources/strings.properties
```diff
@@ -1194,6 +1194,10 @@ notificationTaskAssociatedCaseLink = Link to the associated case: %s
...
+notificationTaskSpecificUpdatedAssigneeUserSource = Your %s task for %s has been assigned to another user. You are no longer in charge of this task.
+notificationTaskSpecificUpdatedAssigneeUserTarget = The %s task for %s has been assigned to you.
```
MateStrysewske, line 1200:
```suggestion
notificationTaskSpecificUpdatedAssigneeUserTarget = A(n) %s task for %s has been assigned to you.
```

MateStrysewske (changes requested):
Right now, the notification is also sent when a new task is created, not only when the assignee of an existing task is changed. That's not consistent with the issue specification, even though it could potentially make sense. I'd suggest to either change this to make sure that this notification is not sent when a new task is created, or to talk back to Carolin and Bernard to check whether the specifications should be updated.

MateStrysewske (approved):

PR #7832, opened 2022-02-01
Feature 6599 person search

### LINKED ISSUE #6599: Add a person search option on creation forms [3]
### Situation Description

Currently, users cannot search for existing persons in the system and simply add another entity (e.g. case) to them. They are forced to go through duplicate detection in order to do so.This is not very user friendly or efficient when users have to handle many entities at once or quickly search for existing persons during the creation process.
### Feature Description

- Add a search button next to the person name in any creation form (cases, contacts, event participants, travel entries, immunizations; also when processing lab messages)
- Upon clicking the button, open a pop-up to search existing persons in the system (cf. screenshot2)
    - When confirming the existing person, link the entity to be created to the existing person when saving.
    - Visually, grey out person related fields and show the persons details that has just been selected (cf. screenshot3)
- When a person has been selected in the creation form, there needs to be the option to remove the person if the linkage or selection was not desired for whatever reason.
- Only show persons that the user has unpseudonymized access to based on the user filter
- Person duplicate detection can be skipped when a person has been selected, but case/contact/event participant duplicate detection still needs to be triggered
### Additional Information

SearchPerson
SearchPerson3
SearchPerson2

BarnaBartha:
Fixes #6599

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventParticipantsController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventParticipantsController.java
@@ -144,6 +144,15 @@ public class EventParticipantsController {
 							},
 							true);
 				} else {
+					EventParticipantCriteria criteria = new EventParticipantCriteria();
+					criteria.withEvent(eventRef);
+					List<EventParticipantIndexDto> currentEventParticipants =
+						FacadeProvider.getEventParticipantFacade().getIndexList(criteria, null, null, null);
+					for (EventParticipantIndexDto participant : currentEventParticipants) {
+						if (dto.getPerson().getUuid().equals(participant.getPersonUuid())) {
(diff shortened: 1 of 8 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventParticipantsController.java
```diff
@@ -144,6 +144,15 @@ public EventParticipantDto createEventParticipant(
 							},
 							true);
 				} else {
+					EventParticipantCriteria criteria = new EventParticipantCriteria();
+					criteria.withEvent(eventRef);
+					List<EventParticipantIndexDto> currentEventParticipants =
+						FacadeProvider.getEventParticipantFacade().getIndexList(criteria, null, null, null);
```
leventegal-she, line 150: would be more performant to create a special backend method with an exists query for checking the person

leventegal-she (changes requested):

leventegal-she (approved):

PR #7846, opened 2022-02-02
(#7811) Dashboard feature config

### LINKED ISSUE #7811: Add a configuration to deactivate dashboard [1]
### Situation Description

Countries having many entities per week fall in  a situation that the dashbard takes a lot of server resources and break  SOMAS
### Feature Description

- Add a new  `DASHBOARD` feature type, enabled by default
- Exclude the dashboard view from the UI when the feature type is disabled (make sure that the user is redirected to another page upon login properly; should already work though)

### Additional Information
@markusmann-vg If the dashboard view is disabled, the user will be redirected to the "cases" directory

vidi42:
Fixes #7811

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/MainScreen.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/MainScreen.java
@@ -17,6 +17,7 @@
  *******************************************************************************/
 package de.symeda.sormas.ui;
 
+import static de.symeda.sormas.ui.UiUtil.enabled;
 import static de.symeda.sormas.ui.UiUtil.permitted;
 
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/MainScreen.java
```diff
@@ -386,19 +390,24 @@ public boolean beforeViewChange(ViewChangeEvent event) {
...
-					defaultView = TasksView.VIEW_NAME;
+				if (permitted(FeatureType.DASHBOARD, null)) {
+					if (permitted(FeatureType.CASE_SURVEILANCE, UserRight.DASHBOARD_SURVEILLANCE_ACCESS)) {
+						defaultView = SurveillanceDashboardView.VIEW_NAME;
+					} else if (permitted(FeatureType.CONTACT_TRACING, UserRight.DASHBOARD_CONTACT_ACCESS)) {
+						defaultView = ContactsDashboardView.VIEW_NAME;
+					} else if (permitted(FeatureType.CAMPAIGNS, UserRight.DASHBOARD_CAMPAIGNS_ACCESS)) {
+						defaultView = CampaignDashboardView.VIEW_NAME;
+					} else if (UserProvider.getCurrent().hasUserRole(UserRole.EXTERNAL_LAB_USER)) {
+						defaultView = SamplesView.VIEW_NAME;
```
MateStrysewske, line 401: Hiding the sample, task or about view behind whether or not dashboards are enabled doesn't make much sense; please move those out of the parent if clause and remove the additional else clause that redirects to the cases view (because the cases view can be disabled for certain users as well, so the about view is the better default).

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #7853, opened 2022-02-02
#7703 - Error request via Rest-API for some cases

### LINKED ISSUE #7703: Error request via Rest-API for some cases
### Bug Description
We are using the SORMAS-REST-API.
We are sending data like cases as request in a json-body via post-method, e.g. “{{baseUrl}}/cases/push”. 
In most cases it works.
In this context, we detected few problems in sending cases via push. 
The error of these problems is “400 Bad Request”; 
“The server detected a syntax error in your request. Check your request and all parameters.”.
Therefore we checked the reason of the error. 
If there is entered an end-date for an exposure (in UNIX), we will receive this error. 

As soon as we remove the property “endDate” for the exposure from the JSON-body, it works and the response would be “OK”. 
Is there a possibility to fix this problem or are there rules for the values of the property “endDate”?

Thank you.

### Screenshots
Reqest in json-body:
request

Response in html:
Unbenannt

### System Details

* Hosting from ITZBund (Germany)
* SORMAS version: 1.66.4
* User Role: ReST-User

cazacmarin:
Fixes #7703


vidi42:
@cazacmarin Since this PR doesn't fix the issue, but adds some logging improvements, please don't close the ticket even if the PR is merged and move the issue to waiting with a commentary asking for the logs.

leventegal-she (approved):

PR #7862, opened 2022-02-02
Feature 7240 display when entity is automatically deleted

### LINKED ISSUE #7240: Display when an entity is going to be automatically deleted [3]
### Situation Description
#7008 adds automatic deletion for entities that is silently happening every night.

### Feature Description
User's should see when an entity is going to be deleted. This is mostly relevant for short-living entities, like travel entries (may have a deletion period as short as 14 days).

Should be displayed in a decent red, when the deletion is within the next 180 days or the user has the right to delete the entity. Otherwise it should be a light grey, so that it's barely visible. Do not show anything, if not automatic deletion is configured.

I suggest to display this at the bottom of the main form of the entity (case, contact, event, event participant, travel entry, immunization).
grafik

The user can click the information to open a popup or tooltip that displays the following information:

> Deletion scheduled for 10/11/2031
> ------------
> End of processing: 09/11/2021
> Deletion period: 10 years

### Implementation

- [x] Extend the CoreEntityFacade (introduced in #7008) to provide deletion information for the entity the user is viewing
  * deletionDate
  * deletionPeriod
  * deletionReference
  * deletionReferenceDate (actual date of the entity depending on the deletionReference)
- [x] Display the information at the bottom of the (first) entity form, as described above.
  * The deletion period should be displayed in the biggest time scale that makes sense (years, months, days)
    * multiple of 30 days = months
    * multiple of 365 days = years
    * something else: always days

### Additional Information

* The "end of processing" is currently simply the change date of the core entity. This will be changed in a follow-up issue
* The "date of origin" for the travel entry will be renamed to the more specific "date of arrival" once we have that field.

alexcaruntu-vita:
Fixes #7240


sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/AbstractCoreEntityFacade.java
```diff
@@ -38,9 +44,15 @@ public void executeAutomaticDeletion(DeletionConfiguration entityConfig) {
...
-		toDeleteEntities.forEach(entity -> {
-			delete(entity);
-		});
+		toDeleteEntities.forEach(this::delete);
+	}
+
+	public AutomaticDeletionInfoDto getAutomaticDeletionInfo(String uuid, CoreEntityType coreEntityType) {
+		DeletionConfiguration deletionConfiguration = deletionConfigurationService.getCoreEntityTypeConfig(coreEntityType);
+		Object[] deletionData = getDeletionData(uuid, deletionConfiguration);
+		Date referenceDate = (Date) deletionData[0];
```
MateStrysewske, line 53: Have you tested this with missing deletion references and periods? Pretty sure that you will run into all sorts of null pointers in this method and in `getDeletionData` below (e.g. in `addDays` when the deletionPeriod is null, or in `getDeleteReferenceField` when deletionReference is null).

sormas-backend/src/main/java/de/symeda/sormas/backend/deletionconfiguration/AbstractCoreEntityFacade.java
```diff
@@ -55,4 +70,16 @@ protected String getDeleteReferenceField(DeletionReference deletionReference) {
 			throw new IllegalArgumentException("deletion reference " + deletionReference + " not supported in " + getClass().getSimpleName());
 		}
 	}
+
+	private Object[] getDeletionData(String uuid, DeletionConfiguration entityConfig) {
```
MateStrysewske, line 74: Please add null checks in this method as well; `getDeleteReferenceField` will throw an exception when entityConfig.deletionReference is null. That should be prevented (i.e. instead this method should return null as well).

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #7863, opened 2022-02-03
Added missing tests for creating samples

jp-sgent:
https://jira.sormas-tools.de/browse/SORQA-14

### DIFF
--- a/sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/events/EventDirectorySteps.java
+++ b/sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/events/EventDirectorySteps.java
@@ -248,6 +248,12 @@ public class EventDirectorySteps implements En {
         });
 
     When(
+        "I click on the created event participant from the list",
+        () -> {
+          webDriverHelpers.clickOnWebElementBySelector(CREATED_PARTICIPANT);
+        });
+
+    When(
         "I click on New Task from event tab",
         () -> webDriverHelpers.clickOnWebElementBySelector(NEW_TASK_BUTTON));
 
--- a/sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/samples/CreateNewSampleSteps.java
+++ b/sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/samples/CreateNewSampleSteps.java
@@ -69,7 +69,7 @@ public class CreateNewSampleSteps implements En {
         });
 
     When(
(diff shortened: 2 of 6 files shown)


sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/samples/CreateNewSampleSteps.java
```diff
@@ -68,13 +68,34 @@ public CreateNewSampleSteps(
...
+        () -> {
+          sample = sampleService.buildAlternateSample();
+          selectPurposeOfSample(sample.getPurposeOfTheSample(), SAMPLE_PURPOSE_OPTIONS);
+          fillDateOfCollection(sample.getDateOfCollection());
+          fillTimeOfCollection(sample.getTimeOfCollection());
+          selectSampleType(sample.getSampleType());
+          selectReasonForSample(sample.getReasonForSample());
+          fillSampleID(sample.getSampleID());
+          fillCommentsOnSample(sample.getCommentsOnSample());
+          webDriverHelpers.clickOnWebElementBySelector(SAVE_SAMPLE_BUTTON);
```
rdutu-vg, line 82: move save action outside this logic to make it reusable

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/events/EventDirectorySteps.java
```diff
@@ -247,6 +247,12 @@ public EventDirectorySteps(
           TimeUnit.SECONDS.sleep(10);
         });
 
+    When(
+        "I click on the created event participant from the list",
+        () -> {
+          webDriverHelpers.clickOnWebElementBySelector(CREATED_PARTICIPANT);
```
rdutu-vg, line 253: this is not always the best approach, because if another test will run in parallel, the first result can be from other test.
Update the locator to indetify the participant based on partial ID, so here the uuid can be collected from created entity and formatted to generate the webelement and click

rdutu-vg (changes requested):

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/events/EventDirectorySteps.java
jp-sgent: We're doing exactly that searching for the event we've just created based on its ID, and only then we create and open the event participant from inside that specific event (not from the general list) - so there's no chance that another test interferes, as it would have to purposefully search for the very same event first in order to create another participant inside it.

Issue #7323, opened 2021-11-18
Add "changeUser" field to entities and history to know which user last changed an entity [5]


MartinWahnschaffe:
### Situation Description
Currently we only have the information on when an entity was created and changed and which user reported it. What is missing is the information who changed an entity. This can currently only be extracted from the audit database. 

In addition there is currently some redundancy between the audit database and the history tables that we want to get rid of. Our idea is to use the history tables to keep track of all changes made to entities (create, update, delete) and to provide a separate log that provides the information that a user has changed or accessed a resource without containing any actual data on the resource itself, so that we don't have any person related data in there.

### Feature Description

- [x] Add a "changeUser" field to the AbstractDomainObject and to all related database tables and history tables
- [x] update the changeUser of the AbstractDomainObject using the PrePersist and PreUpdate annotation. See User.updateJursidictionLevel

**Important**: We have to validate whether this approach actually works and only set's the user when the entity was changed. 
Otherwise we may have to detect this somehow or work with an interceptor. In this case we should have a look at the mechanics of the existing audit log.

### Additional Information

- We recently had the approach to touch all `AbstractDomainObject`s in SQL with a foreach loop: https://github.com/hzi-braunschweig/SORMAS-Project/issues/7303#issuecomment-1015185867
- For now we should not include this in the DTO classes and especially don't have to include it in the mobile app sync process. This is a backend only functionality.

JonasCir:
I think we can use 
```sql
ALTER TABLE cases ADD COLUMN change_user BIGINT,
    ADD CONSTRAINT fk_cases_change_users
    FOREIGN KEY (change_user)
    REFERENCES users (id);
```
Together with the loop [script](https://github.com/hzi-braunschweig/SORMAS-Project/issues/7303#issuecomment-1015185867)

JonasCir:
@MartinWahnschaffe 
I don't think `PrePersist` and `PreUpdate` will work here as `CurrentUserService::getCurrentUser` is an EJB method. 

One idea that I have would be to put this in the save method of `AbstractBaseEjb` where we have access to the current user and push the usage of `BaseFacade` across the code base.

I'm aware this is a big change but:
1. I think we need to do it anyway at some point in time to keep this project maintainable
2. We talked about pushing the `BaseFacade` before in the (related) context of generating the FHIR logs consistently
3. I think this will allow us the really streamline the code base

[...]

Issue #7246, opened 2021-11-12
Refactor CoreAdo to include archiving [3]


MartinWahnschaffe:
### Feature Description
The CoreAdo class is used by several classes that need to be soft-deleted. In addition we have multiple classes that allow archiving. All of those classes implement this on their own. For all classes besides the infrastructure ones, there is currently no generic facade.

### Problem Description
Trigger for this issue is the new field "endOfProcessingDate" that is needed for all core entities (see #7247). The current CoreAdo class does not represent the actual core entity from the SORMAS domain perspective.
Also there is a lot of redundant code in the facades that should be replaced by an abstract implementation.

### Proposed Change
- [x] Rename CoreAdo to DeletableAdo (name is not ideal, but we didn't come up with something better yet)
- [x] Introduce new CoreAdo class that extends DeletableAdo. Should be extended by the following classes: Case, Contact, Event, EventParticipant, Immunization, TravelEntry, Campaign
  * Take the comment from the previous CoreAdo and adjust it as follows: The integral definition of core data is that it is the working representation of a real-world event that is archived once the work is complete.
 * relevant for external systems that share data with or use data from SORMAS.
- [x] Move the existing archived property of those classes to the CoreAdo
- [x] For contact and event Part make the archived property transient in the entities
- [x] Move the archive methods from AbstractBaseEjb to AbstractInfrastructureEjb
- [x] Create an AbstractCoreEjb that does for CoreAdo classes what AbstractInfrastructEjb is doing for InfrastructureAdo classes. Facades for Case, Contact, ... should use this

### Additional Information

* We could introduce an Archivable interface to enable sharing of code for CoreAdo and InfrastructureAdo

JonasCir:
> We could introduce an Archivable interface to enable sharing of code for CoreAdo and InfrastructureAdo
I would really like to see something like this. This is the last missing piece to wire up e.g., Immunization with the abstract classes we have. By this, we can save a lot of code duplication, so I'm all in favor! :)

adinaflorea9:
Today, @roxanamlendea  and I found that on our local testing machines and on the test instances (test-de1, test-de2, test-de3) various entities can not be opened:

- [ ] Cases
- [ ] Contacts
- [ ] Event Participants
- [ ] Samples 
- [ ] Tasks

[...]

PR #7875, opened 2022-02-03
[#7246] remove annotations blocking deployment

JonasCir:
@BarnaBartha 
I fixed some deployment errors caused by annotations on overwritten methods, this is not allowed.
Also removed the `delete` method in subclasses as they were just the same as the parent.

With this PR, there is still the error that `column contact0_.archived does not exist` , so I guess you need add a field to all core ados? :)
If yes then checkout https://github.com/hzi-braunschweig/SORMAS-Project/issues/7323#issuecomment-1023005858 , this will make your life much easier :) Reach out if you have questions to this script.

@MateStrysewske  FYI


JonasCir:
A some already have archived as column of course we need to find the ones which are missing it

JonasCir:
deploys now

JonasCir:
OK I talked to @BarnaBartha and he is taking a couple of things out of this PR and then he will close it

Issue #7674, opened 2022-01-13
[Angular][Backend] Add new mSERS [2] 


laurentiustanvitagroup:
### Situation Description

- In view of the future replacement of the current FE with the Angular component, the Angular team will focus on the required changes.

### Feature Description

- Reference: https://www.sketch.com/s/fd061796-65a3-4991-886a-c8a172667c91/a/l1rRgyx
- We need an endpoint that can receive and handle a disease list 

### Possible Alternatives

- N/A

### Additional Information

- N/A

PR #7897, opened 2022-02-04
#7674 m sers endpoints

marius2301:
Fixes #


sormas-rest/src/main/java/de/symeda/sormas/rest/AggregateReportResource.java
```diff
@@ -45,6 +47,13 @@
 		return result;
 	}
 
+	@POST
+	@Path("/indexList")
+	public List<AggregatedCaseCountDto> getIndexList(AggregateReportCriteria criteria) {
+		List<AggregatedCaseCountDto> result = FacadeProvider.getAggregateReportFacade().getIndexList(criteria);
```
JonasCir, line 53: Nit pick, but return the result directly :) One warning less in the code :smiley:

MateStrysewske (changes requested):
See Jonas' comment. :)

PR #7919, opened 2022-02-08
#5285: Map PCR test specipication from TestReport to PathogenTest

### LINKED ISSUE #5285: [DEMIS2SORMAS] Map variant specific Nucleic acid detection methods [1]
> @FredrikSchaeferVitagroup Can you add instructions on how to extend the sample creation process when processing lab messages to also include these new fields if that's possible?

- [x] The testReport and -Entity need to be extended by a `testPcrTestSpecification` field holding the enum value `PCRTestSpecification`
- [x] The value needs to be transferred from the testReportDto to newly created or selected PathogenTest when processing a labMessage

_Originally posted by @FredrikSchaeferVitagroup in https://github.com/hzi-braunschweig/SORMAS-Project/issues/5029#issuecomment-820413769_

SORMAS-DEMIS-adapter counterpart is [#11](https://github.com/hzi-braunschweig/sormas-demis-adapter/issues/11).

ChristopherRiedel:
Fixes #5285


sormas-api/src/main/java/de/symeda/sormas/api/utils/InfoProvider.java
```diff
@@ -49,7 +49,7 @@ public static synchronized InfoProvider get() {
 	 * @return
 	 */
 	public String getMinimumRequiredVersion() {
-		return "1.67.0";
+		return "1.68.0";
```
leventegal-she, line 52: I don't think this should be part of the mapping ticket

leventegal-she (changes requested):

leventegal-she (approved):

PR #7940, opened 2022-02-09
Change/7907 username handling

### LINKED ISSUE #7907: Align username handling of Keycloak and legacy login
### Feature Description
SORMAS user management

### Problem Description
We want to handle usernames in an case insensitive manner just like Keycloak is doing it. (TBD)

### Proposed Change

- [ ] Finish https://github.com/hzi-braunschweig/S2S-Testbed/pull/11 such that we can test easier (Done but not merged as of now.)
- [ ] remove `AuthProvider.getProvider(configFacade).isUsernameCaseSensitive()`
- [ ] TBD

### Additional info

The following exception can occur if a username is now considered a duplicate and prevents the deployment (database migration fails):
```bash
ERROR: could not create unique index "idx_users_username_lower"
Detail: Key (lower(username::text))=(my_username) is duplicated.]]
```

Duplicates can be found with the following SQL. Rename or delete users to remove duplicates.
```sql
SELECT username, creationdate, id, uuid FROM users WHERE LOWER(username) IN (SELECT LOWER(username) FROM users GROUP BY LOWER(username) HAVING COUNT(*) > 1) ORDER BY username, creationdate;
```

JonasCir:
Fixes #7907


JonasCir:
@vidi42 The other issue I mentioned this morning will be addressed in a separate PR, it need to put more thoughts into it. Does not block us to ship this win, however :)

JonasCir:
@MartinWahnschaffe FYI, this includes a constraint on the DB now, just as we discussd

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -9949,4 +9949,11 @@ ALTER TABLE eventparticipant_history ADD COLUMN archived BOOLEAN;
 
 INSERT INTO schema_version (version_number, comment) VALUES (440, 'Refactor CoreAdo to include archiving #7246');
 
+-- 2022-02-09 Align username handling of Keycloak and legacy login #7907
+DROP INDEX idx_users_username;
```
vidi42, line 9953: Should we worry about systems which are not using Keycloak and might already have users with non unique usernames, like `NatUser` and `natuser` ?

JonasCir:
Currently fixing the tests...

sormas-backend/src/main/resources/sql/sormas_schema.sql
vidi42: I suggest than that we provide a script in a comment to the issue which can help systems identify if they have this kind of issue before upgrading to the newest version

JonasCir: Good idea will do.

JonasCir:
Hm test passes locally, dunno... Just restarted it again

JonasCir:
@vidi42 
All failing test fixed, was related to duplicate users being created what is prohibited now.
I added a script as proposed by you. I also tested it locally: with duplicates deployment fails, after using the script and changing the name everything works as expected.

PR #7961, opened 2022-02-11
#7246 - fill contact and EvP archived flags

### LINKED ISSUE #7246: Refactor CoreAdo to include archiving [3]

BarnaBartha:
…don't have this flag mandatory yet

Fixes #7246

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/CoreAdo.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/CoreAdo.java
@@ -31,14 +31,14 @@ public class CoreAdo extends DeletableAdo {
 
     public static final String ARCHIVED = "archived";
 
-    private boolean archived;
+    private Boolean archived = Boolean.FALSE;
 
     @Column
-    public boolean isArchived() {
+    public Boolean isArchived() {
         return archived;
     }
 
-    public void setArchived(boolean archived) {
+    public void setArchived(Boolean archived) {
         this.archived = archived;
     }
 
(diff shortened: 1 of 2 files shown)


StefanKock (changes requested):
As discussed, don't introduce a tri-state Boolean, but make sure that all data and SQL tables only has two possible states

StefanKock (approved):
Code changes look like I expected them, but did not test it and did not check if all sub classes of CoreAdo are properly covered concerning NOT NULL constraint

MartinWahnschaffe (approved):

MateStrysewske (changes requested):
I would suggest to also add a default (false) to these columns to make sure that they are always automatically populated. We did this for cases and events as well.

MateStrysewske (approved):

PR #7963, opened 2022-02-11
#7792 Make tile server used for map configurable

### LINKED ISSUE #7792: Make tile server used for map configurable [2]
### Feature Description
We are using leaflet js to display maps in SORMAS. The used tile server url is currently hard coded in the leaflet-connector javascript.

### Problem Description
This approach does not allow to use different tile servers or to put a proxy in between.

### Proposed Change

- [ ] Add the configuration to sormas.properties
- [ ] Extend the leaflet map component to use the configured tile server

### Additional Information

Default the one that is currently hardcoded.
If there is no property defined, we don't show the tiles
Validate what happens if there is sth. defined which does not work (404)

Examples:
http://leaflet-extras.github.io/leaflet-providers/preview/index.html

leventegal-she:
Fixes #7792

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/map/LeafletMap.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/map/LeafletMap.java
@@ -96,17 +96,6 @@ public class LeafletMap extends AbstractJavaScriptComponent {
 		// credit where credit's due
 		String attribution = FacadeProvider.getGeoShapeProvider().loadShapefileAttributions();
 		this.addShapefileAttribution(attribution);
-
-		/*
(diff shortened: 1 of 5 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/map/LeafletMap.java
```diff
@@ -88,6 +97,16 @@ public void call(JsonArray arguments) {
 		String attribution = FacadeProvider.getGeoShapeProvider().loadShapefileAttributions();
 		this.addShapefileAttribution(attribution);
 
+		/*
```
vidi42, line 100: This probably makes more sense to be added in the java doc of the constructor or class

leventegal-she: I'l delete it. Just pasted here to copy parts that were needed for testing

vidi42 (approved):

PR #7997, opened 2022-02-15
Change 6879 move health conditions

### LINKED ISSUE #6879: Move health conditions from clinical course to the case [3]
### Feature Description

Cases > Clinical Management > Therapy / Clinical Course / pre-existing conditions
### Problem Description

Currently, the clinical management module is quite dispersed across the case entity and user roles and rights are not clearly separated for this module. E.g. pre-existing conditions, in contrast to the contact entity, are present in the clinical course of the case. This tab is only accessible for clinicians or case officers but other users should also be able to enter pre-existing conditions.
### Proposed Change

- [ ]  Move the pre-existing conditions for cases to the case information tab above the additional medical information ( as seen for contacts)
- [ ] Document the following upon completion of this change:
        - Data migration from existing to new values (if applicable)
        - Changes to database structure/tables (if applicable)
        - Changes in API endpoints (if applicable)

### Additional Information

Frame 63

* Move the health conditions from the clinical course entities and DTO to the case entities and DTO
* In the web database, add a new column to the case table, fill it with the health conditions of the clinical course that belongs to each case, and drop the health conditions column from the clinical course table (not the history table)
* In the mobile database, do the same, but do not drop the health conditions column from the clinical course table (because SQLite doesn't support dropping columns); instead mark the field as deprecated in the entity class
* Update the change date filters accordingly
* In the web app, move the health conditions from the clinical course tab to the case tab (above the reception dates of paper form)
* In the mobile app, health conditions already have their individual tab, but there will likely be some references that need to be updated
* Document the following upon completion in this issue:
        - Data migration from existing to new values (if applicable)
        - Changes to database structure/tables (if applicable)
        - Changes in API endpoints (if applicable)

BarnaBartha:
Fixes #6879


MateStrysewske:
@BarnaBartha There are test failures, can you please have a look at them before I start to review this PR?

MateStrysewske (changes requested):
Unit tests fail

sormas-backend/src/main/java/de/symeda/sormas/backend/util/test.sql
```diff
@@ -0,0 +1,50 @@
...
...0.trimester, (select count(distinct generatedAlias20.id) from EventParticipant as generatedAlias21 inner join generatedAlias21.event as generatedAlias20 inner join generatedAlias21.resultingCase as generatedAlias22 where ( generatedAlias22.id=generatedAlias0.id ) and ( generatedAlias20.deleted = false ) and ( generatedAlias20.archived = false ) and ( generatedAlias21.deleted = false )), generatedAlias0.externalID, generatedAlias0.externalToken, generatedAlias0.internalToken, generatedAlias1.birthName, generatedAlias23.isoCode, generatedAlias23.defaultName, generatedAlias24.isoCode, generatedAlias24.defaultName, generatedAlias0.caseIdentificationSource, generatedAlias0.screeningType, generatedAlias25.name, generatedAlias26.name, generatedAlias27.name, generatedAlias0.clinicianName, generatedAlias0.clinicianPhone, generatedAlias0.clinicianEmail, generatedAlias28.id, generatedAlias29.id, generatedAlias0.previousQuarantineTo, generatedAlias0.quarantineChangeComment, case when ( ( generatedAlias0.reportingUser is not null ) and ( generatedAlias0.reportingUser.id=15L ) ) or ( ( generatedAlias0.responsibleDistrict.id=11L ) or ( generatedAlias0.district.id=11L ) ) then true else false end)
```
MateStrysewske, line 17: I suspect this file shouldn't have been committed? :)

MateStrysewske (changes requested):
Code looks good, will test it tomorrow.

MateStrysewske (changes requested):
The update script in the mobile app does not work properly. The migration is not done. Pretty sure you need to wrap each query in an individual `executeRaw` call.

Please make sure to test the migration before handing it over for review.

Also, please move the health conditions tab next to the symptoms tab.

MateStrysewske (changes requested):
Everything else is fine. You can merge the PR after this last small change. :)

PR #8000, opened 2022-02-15
#6830: Update poi-ooxml

### LINKED ISSUE #6830: Update poi-ooxml from 3.17 to 5.2.0
### Problem Description

#4707 requires an update for poi-ooxml dependency -> this is a major version increase and the changes introduced are non compiling, the methods and how XSSFTable should be used has changed thus a rewrite/refactor of the data dictionary generation is needed.

### Proposed change

1. poi-ooxml 3.17 -> latest 5 or 4 version
2. Fix breaking changes to keep all features running

### Additional information

We have to update to at least version 4.1.2 - ideally we should update to the latest version.

StefanKock:
Fixed breaking changes:
- color definition not with awt, directly with rgb colors (new helper)
- table configuration after filling cells with reference (new helper)

Refactorings along the way:
- Suppressed a warning locally
- Added a private constructor for helper class
- Removed commented code in Generator class

Fixes #6830


StefanKock:
I already compared the generated `SORMAS_Data_Dictionary.xlsx` and `SORMAS_User_Rights.xlsx`: I did not find any differences or flaws.

PR #8039, opened 2022-02-17
#2802: Simplify user rights

### LINKED ISSUE #2802: Simplify user rights for future user role configuration [2]
### Feature Description
We want to allow admins to configure user roles according to their needs #898

### Problem Description
Some user rights don't follow the standard pattern (view, edit, ...) and make the implementation of a user role configuration more complex than necessary.

### Proposed Change

- [x] Remove USER_RIGHTS_MANAGE user right (is not used at all)
- [x] Remove DASHBOARD_VIEW user right. Check for any of the other dashboard user rights instead
- [x] Rename DASHBOARD_SURVEILLANCE_ACCESS, DASHBOARD_CONTACT_ACCESS, DASHBOARD_CAMPAIGNS_ACCESS to DASHBOARD_SURVEILLANCE_VIEW, etc.
- [x] Introduce OUTBREAK_VIEW user right and use in AbstractConfigurationView
- [x] Remove CONFIGURATION_ACCESS user right and replace it with a helper method that checks for any configuration related user right (see AbstractConfigurationView)
- [x] Replace OUTBREAK_CONFIGURE_RESTRICTED and OUTBREAK_CONFIGURE_ALL with OUTBREAK_EDIT and introduce a jurisdiction check
- [x] Replace LINE_LISTING_CONFIGURE_NATION, by using LINE_LISTING_CONFIGURE and a jurisdiction based check (NATIONAL or NONE)
- [x] Replace CASE_MANAGEMENT_ACCESS wit a dedicated CASE_CLINICIAN_VIEW user right that is only used to allow users to see the three clinician details field in case data. See also next point
- [x] Allow the case management export and the "must have case management data" filter when either THERAPY_VIEW or CLINICAL_COURSE_VIEW are given. The export should only include the allowed part (e.g. currently lab users can export the complete case management data, because only CASE_MANAGEMENT_ACCESS is checked). Also the ViewMode.Normal logic should dependent on those two rights.

ChristopherRiedel:
Fixes #2802


cazacmarin (changes requested):
Please check as well android app as some permissions/roles need to be adjusted there as well. Otherwise, app will fail:
 ERROR: Could not find accessor de.symeda.sormas.api.user.UserRight.CASE_MANAGEMENT_ACCESS file://app/src/main/res/layout/fragment_case_read_layout.xml Line:759
  ERROR: Could not find accessor de.symeda.sormas.api.user.UserRight.CASE_MANAGEMENT_ACCESS file://app/src/main/res/layout/fragment_case_edit_layout.xml Line:805

For current implementation which include Sormas UI/API/Backend, all is looking good so far.

cazacmarin (approved):

PR #8041, opened 2022-02-17
#6822 Automatically check history tables

### LINKED ISSUE #6822: Create an automatic test that compares tables with their _history equivalent [3]
### Proposed Change

This is a follow-up to #6121. @StefanKock had a quick try to create an automatic test here: https://github.com/hzi-braunschweig/SORMAS-Project/commit/d562850b9b2dafb7054ea358a949319b357d795f unfortunately with H2 this is very difficult (or maybe not possible at all). We should take the opportunity to set-up unit testing with testcontainers.org instead: https://www.testcontainers.org/modules/databases/postgres/

Todos:

- [ ] create an automatic test that executes the provided script and checks whether new/modified columns are missing in the history tables.
- [ ] set-up postgres instance with testcontainers.org
- [ ] This requires docker, so we have to make sure that this is also automatically provided or that tests are at least skipped on dev systems that don't have it. For the jenkins CI this should not be a problem.
- [ ] Make sure this also runs with github actions (should be possible, see https://stackoverflow.com/questions/66174071/github-actions-for-running-test-with-testcontainers-and-gradle)

vidi42:
Fixes #6822


sormas-backend/src/test/java/de/symeda/sormas/backend/common/StartupShutdownServiceTest.java
```diff
@@ -94,39 +114,130 @@ public void testAuditSchemaVersions() throws IOException {
...
+	public void testHistoryTablesMatch() throws IOException, URISyntaxException {
+
+		SormasPostgresSQLContainer container = new SormasPostgresSQLContainer().withDatabaseName("sormas");
+		container.start();
+
+		Map<String, String> properties = new HashMap<>();
+		properties.put("javax.persistence.jdbc.url", container.getJdbcUrl());
+		properties.put("javax.persistence.jdbc.user", "sormas_user");
+		properties.put("javax.persistence.jdbc.password", "password");
+		properties.put("javax.persistence.jdbc.driver", container.getDriverClassName());
```
leventegal-she, line 127: I would say also `jdbc.user` and `password` should come from the container.

sormas-backend/src/test/resources/testcontainers/setup_sormas.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
leventegal-she, line 1: I would call this file `setup_sormas_db` to state that it sets up only the database not a whole sormas app

leventegal-she (changes requested):

leventegal-she (approved):

PR #8048, opened 2022-02-17
sorqa68

pk-sgent:
Fixes #


sormas-e2e-tests/src/main/java/org/sormas/e2etests/enums/GenderValuesDE.java
```diff
@@ -0,0 +1,23 @@
...
+
+  GenderValuesDE(String humanGender) {
+    gender = humanGender;
+  }
+
+  public static String getRandomGenderDE() {
+    Random random = new Random();
+    return String.valueOf(GenderValuesDE.values()[random.nextInt(values().length)].gender);
+  }
+}
```
rdutu-vg, line 23: help me out a bit to understand what's the need of this new class because UI text is displayed in english, and I can't see in the steps to change the language to german so we can validate the language differences.
My suggestion would be to have this data in the existing GenderValues, and we can refactor the class to store the translation for each locale, and create methods where we pass the desired locale and get the correct translation

sormas-e2e-tests/src/main/java/org/sormas/e2etests/services/CaseService.java
```diff
@@ -72,6 +73,36 @@ public Case buildGeneratedCase() {
...
+    firstName = faker.name().firstName();
+    lastName = faker.name().lastName();
+
+    return Case.builder()
+        .firstName(firstName)
+        .lastName(lastName)
+        .caseOrigin("IM LAND")
+        .dateOfReport(LocalDate.now().minusDays(1))
+        .externalId(UUID.randomUUID().toString())
+        .disease("COVID-19")
```
rdutu-vg, line 86: random disease

sormas-e2e-tests/src/main/java/org/sormas/e2etests/services/CaseService.java
```diff
@@ -72,6 +73,36 @@ public Case buildGeneratedCase() {
...
+        .responsibleRegion(RegionsValues.VoreingestellteBundeslander.getName())
+        .responsibleDistrict(DistrictsValues.VoreingestellterLandkreis.getName())
+        .responsibleCommunity(CommunityValues.VoreingestellteGemeinde.getName())
+        .placeOfStay("ZUHAUSE")
```
rdutu-vg, line 90: can we vary this data?

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CaseDetailedTableViewSteps.java
```diff
@@ -141,6 +143,34 @@ public CaseDetailedTableViewSteps(
               "Reporting user is not correct");
           softly.assertAll();
         });
+
+    When(
+        "I back to Case Directory using case list button",
```
rdutu-vg, line 148: refactor a bit this step and you can also add a wait for a specific element to be present, so we can avoid flaky situations or forcing us to add a wait in all other methods which can be called after this one

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CaseDetailedTableViewSteps.java
```diff
@@ -141,6 +143,34 @@ public CaseDetailedTableViewSteps(
...
+        () -> webDriverHelpers.clickOnWebElementBySelector(BACK_TO_CASES_LIST_BUTTON));
+
+    When(
+        "I check if Case date format is correct",
+        () -> {
+          List<Map<String, String>> tableRowsData = getTableRowsData();
+          Map<String, String> detailedCaseDTableRow = tableRowsData.get(0);
+          softly.assertTrue(
```
rdutu-vg, line 156: add a more specific failure message for both asserts, to understand from results which date is wrong

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CaseDetailedTableViewSteps.java
```diff
@@ -141,6 +143,34 @@ public CaseDetailedTableViewSteps(
               "Reporting user is not correct");
           softly.assertAll();
         });
+
+    When(
+        "I back to Case Directory using case list button",
+        () -> webDriverHelpers.clickOnWebElementBySelector(BACK_TO_CASES_LIST_BUTTON));
+
+    When(
+        "I check if Case date format is correct",
```
rdutu-vg, line 152: Specify from where we take the date to check it, can be in multiple places, and which format (german) to avoid confusions with existing one or what else can come next

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CaseDetailedTableViewSteps.java
```diff
@@ -141,6 +143,34 @@ public CaseDetailedTableViewSteps(
...
+          softly.assertTrue(
+              checkDateFormatDE(detailedCaseDTableRow, "NACHVERFOLGUNG BIS"),
+              "Date format is invalid");
+          softly.assertTrue(
+              checkDateFormatDE(detailedCaseDTableRow, "MELDEDATUM"), "Date format is invalid");
+          softly.assertAll();
+        });
+  }
+
+  boolean checkDateFormatDE(Map<String, String> map, String row) {
```
rdutu-vg, line 165: make it private

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CaseDetailedTableViewSteps.java
```diff
@@ -141,6 +143,34 @@ public CaseDetailedTableViewSteps(
...
+    DateFormat dateFormat = new SimpleDateFormat(DATE_FORMAT_DE);
+    dateFormat.setLenient(false);
+    try {
+      dateFormat.parse(map.get(row));
```
rdutu-vg, line 169: this method can perform also a dummy parse, if the data is empty, or if the parsing passed but is not as expected, then will return true as well.
Add also an assert here, and a logger to print out what was received and what was parsed, so we can also debug if will fail in jenkins.

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CreateNewCaseSteps.java
```diff
@@ -172,6 +199,11 @@ private void fillDateOfReport(LocalDate date) {
     webDriverHelpers.fillInWebElement(DATE_OF_REPORT_INPUT, formatter.format(date));
   }
 
+  private void fillDateOfReportDE(LocalDate date) {
```
rdutu-vg, line 202: we can keep the same method, just refactor it to pass a formatter for the desired market

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/EditCaseSteps.java
```diff
@@ -748,6 +790,11 @@ private LocalDate getDateOfReport() {
     return LocalDate.parse(dateOfReport, DATE_FORMATTER);
   }
 
+  private LocalDate getDateOfReportDE() {
```
rdutu-vg, line 793: same

rdutu-vg (changes requested):

sormas-e2e-tests/src/main/java/org/sormas/e2etests/services/CaseService.java
pk-sgent: I have no translated diseases from EN to DE. I don't wanna make mess, Alwin said to skip it. I don't change it right now

sormas-e2e-tests/src/main/java/org/sormas/e2etests/services/CaseService.java
pk-sgent: In the Cases we never used place of stay different than "Home". It is because "Facility" needs more fields to fill so it can't be random. We also haven't enum for that - never needed

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CaseDetailedTableViewSteps.java
pk-sgent: If parser will have different format than defined in DATE_FORMAT_DE will return false

rdutu-vg: ok, but in this case, we need to assign the parsed value into a variable and check it :D

pk-sgent: Razvan sorry, i don't get why i need to check something twice? parse fucntion is in try catch, if data format received from website is different than i expect then parser will have ParseException - checked it multiple times. You suggest me to create new variable from parser and check if is correct, it have no sense - if it will be incorrect format, parser will throw exception and that checking will never be called ..

rdutu-vg: yes, you're right, my bad, Just checked on my local.
Then we'll proceed as it is, maybe the logger can be added so we can debug it in jenkins if fails :D

pk-sgent: ok :) Logger is good idea, i added it

rdutu-vg (approved):

PR #8055, opened 2022-02-18
Feature 7451 travel entry directory

### LINKED ISSUE #7451: Add bulk deletion and reporting period filter to the travel entries directory [2]
### Situation Description

There is no option to delete travel entries with bulk action. That causes a lot of work.
In many regions, documented entries that have not become a case must be deleted after 14 days. A filter option for a reporting period is therefore additionally relevant.

### Feature Description
- [x] Add a bulk action dropdown to the travel entries directory that includes a delete action (as in other directories). Only show the bulk action dropdown if the user has the TRAVEL_ENTRY_DELETE and the PERFORM_BULK_OPERATIONS user rights.
- [x] Add "Report date from" and "Report date to" filters to the travel entries directory that filters the travel entries by report date according to the selected period.

BarnaBartha:
Fixes #7451


MateStrysewske (changes requested):
The epi week component doesn't contain the most recent week for me. Maybe that will be fixed by tomorrow, not sure what the problem is here. Not directly related to your issue of course, but could you have a look at it and see whether you can fix it?

BarnaBartha:
THe most recent week is there for me in the filter. Can you check again, i updated the branch from development.

MateStrysewske:
This is how it looks for me, even in the case directory; you see Week 9 there?

epiweeks

MateStrysewske (approved):

PR #8115, opened 2022-02-23
 introduce PersonCreateForm #7935

### LINKED ISSUE #7935: Introduce PersonCreateForm [5]
### Feature Description
Forms where a person may be created are currently implemented as extensions of `PersonDependentEditForm<DTO>`. Fields for the person's data are introduced as custom fields rather than property fields for a `PersonDto`, so all annotations in `PersonDto` are ignored.

To avoid this, the creation forms derived from `PersonDependentEditForm<DTO>` should *contain* an embedded `PersonCreateForm` extending `AbstractEditForm<PersonDto>` (which provides the default mechanics to handle annotations in the `Dto` class) rather than *inherit* from a class with no notion of annotations in `PersonDto`

### Problem Description
See  #7743, where the implementation did not follow changes in `PersonDto` annotations.

### Proposed Change
Refactoring as described above

syntakker:
Fixes #7935

To resolve remaining code duplication, see #8117.


MateStrysewske:
@syntakker There have been changes to especially the case create form in the meantime related to the home address (bugfixes, primarily) which probably are the reason for the merge conflicts. Those have to be resolved before I can properly test this PR.

MateStrysewske (changes requested):
(see above)

MateStrysewske (approved):

PR #8140, opened 2022-02-25
Feature 7712 permanent delete

### LINKED ISSUE #7712: Permanent deletion for TravelEntry entity and Immunization [5]
### Situation Description
Currently SORMAS only supports soft-deletion of entities (DeletableAdo). Soft-deletion currently set's the entity's "deleted" flag and removes references.

Is also permanently deletes some of the child entities (e.g. tasks, treatments) while other child entities are kept (e.g. visits), which is quite intransparent. 

### Feature Description
First, it should be possible to permanently delete an entity. We will keep the current process of soft-deleting an entity and extend it with a nightly job that goes through soft-deleted entities and actually deletes them.

Second, we need to clearly define at what point references are removed and children are permanently deleted in both deletion modes.

The scope of this issue is to get started by adding the generic parts and implementing it for the TravelEnry and Immunization.
Deletion of persons, documents and S2S infos will be handled by #7857 and #7854.

Basics:

- [x] split the delete method of BaseAdoService from the delete method of AbstractDeletableAdoService. The one in BaseAdoService should be called deletePermanent and acutally delete the entity. The other one will be called delete and set the deleted flag in the deletable entity. Make sure that for all services that extend AbstractDeletableAdoService deletePermanent is not called (this wil be added starting with this issue). 
- [x] Add DELETE_PERMANENT feature configuration

For travel entries tasks should no longer be deleted immediately when soft-delete is done. There are no other children that need to be taken care of. 

- [x] Move the logic that deletes tasks from the TravelEntryService delete method to the deletePermanent method
- [x] Extend the TaskService.buildActiveTasksFilter method to also check the DELETED flag for case, contact, event, travel entry (I'd use notEqual with true instead of checking both false and null).
- [x] In TravelEntryService.deletePermanent call markAsDeleted for all documents related to the travel entry

For immunizations the Vaccinations belonging to the Immunization will already by automatically deleted by OneToMany cascading.

- [x] In ImmunizationService.deletePermanent remove the Immunization from all SormasToSormasShareInfos that use it.

The permanent deletion should only be done by the backend in the context of nightly jobs:

- [x] Add an executePermanentDeletion method to the AbstractDeletableAdoService that queries all entities that are marked for deletion and calls the deletePermanent method for each of them
- [x] Add an executePermanentDeletion method to the CoreEntityDeletionService that is executed by the automatic deletion cron job, after executeAutomaticDeletion is done. For now only call executePermanentDeletion for the TravelEntryService and ImmunizationService.
- [x] **Discuss** We could run into a transaction timeout here when there are a lot of entities to be deleted. We may have to define a maximum number of entities to be deleted in one transaction and then execute the method calls in batches

BarnaBartha:
Fixes #7712


alexcaruntu-vita (approved):

sormas-backend/src/main/java/de/symeda/sormas/backend/common/CronService.java
```diff
@@ -189,8 +189,13 @@ public void syncInfraWithCentral() {
 		centralInfraSyncFacade.syncAll();
 	}
 
-	@Schedule(hour = "1", minute =  "55", persistent = false)
-	public void deleteExpiredEntities(){
+	@Schedule(hour = "1", minute = "55", persistent = false)
```
MartinWahnschaffe, line 192: I think we should have 10-15 minutes in between those two. Maybe state permanent deletion at 2:10

MartinWahnschaffe (changes requested):
@BarnaBartha Just one thing. Afterwards you can merge.

PR #8179, opened 2022-03-02
Bugfix 8131 a problem has occured when trying to save an event

### LINKED ISSUE #8131: [Events] Error message 'A problem has occurred' appears when trying to save changes made on an event
### Bug Description

An error message 'A problem has occurred. The data could not be saved because it has been changed in the meantime' appears when trying to save the changes made on an event, via the 'Confirm navigation' pop-up.

### Steps to Reproduce

1. Open an existing event;
2. Edit some information form the event page and do not click 'Save' at the bottom of the page;
3. On the 'Event groups' card, click on 'Link new event group';
4. Select an event from the list in the form 'Pick or create event group' (or select 'New event group' and create one);
5. Click 'Save changes' in the pop-up to 'Confirm navigation' that gets triggered when saving the event group.

### Actual Behavior

An error message appears stating that 'A problem has occurred. The data could not be saved because it has been changed in the meantime'; the user is not able to save the changes made at step 2.

### Expected Behavior

No error message appears and the user is able to save the changes made at step 2 via the 'Confirm navigation pop-up'.

### Screenshots

GIF of the steps to reproduce:
Event group

### System Details

* Device: Windows 10
* SORMAS version: 1.69.0-SNAPSHOT
* Server URL: local machine and test-de2.sormas.netzlink.com/
* User Role: Admin, NatUser

### Additional Information

The 'Confirm navigation' pop-up could be triggered when the user clicks 'Link event group' so that the changed information from the event page gets saved before the linking the new event group.

alexcaruntu-vita:
Fixes #8131

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/components/sidecomponent/SideComponent.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/components/sidecomponent/SideComponent.java
@@ -57,9 +57,8 @@ public class SideComponent extends VerticalLayout {
 		}
 	}
 
-	protected void addCreateButton(String caption, UserRight userRight) {
-		UserProvider currentUser = UserProvider.getCurrent();
-		if (currentUser != null && currentUser.hasUserRight(userRight)) {
+	protected void addCreateButton(String caption, UserRight... userRights) {
+		if (userHasRight(userRights)) {
 			Button createButton = ButtonHelper.createButton(caption);
 			createButton.addStyleName(ValoTheme.BUTTON_PRIMARY);
 			createButton.setIcon(VaadinIcons.PLUS_CIRCLE);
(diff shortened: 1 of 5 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/utils/components/sidecomponent/SideComponent.java
```diff
@@ -81,4 +80,16 @@ public Registration addSideComponentEditEventListener(SideComponentEditEventList
 			sideComponentEditEventListener,
 			SideComponentEditEventListener.ON_SIDE_COMPONENT_EDIT_METHOD);
 	}
+
+	private boolean userHasRight(UserRight... userRights) {
+		UserProvider currentUser = UserProvider.getCurrent();
+		if (currentUser != null) {
```
MateStrysewske, line 86: You can basically replace the whole method with one single line of code:

`return Arrays.asList(userRights).stream().anyMatch(r -> currentUser.hasUserRight(r));`

I don't think the null check is needed because there always HAS to be a current user; if that's not the case, we have a problem anyway.

MateStrysewske (changes requested):

MateStrysewske (approved):

PR #8190, opened 2022-03-02
#7820 Remove keycloak roles dependency

### LINKED ISSUE #7820: Update Keycloak to not depend on user roles anymore [3]
### Feature Description
We want to replace the user role usage with permissions. 

### Problem Description
At the moment Keycloak is configured with a subset of user roles which are needed to access the REST API.
Removing the user roles from SORMAS requires that we also remove them from Keycloak, however since we are managing the permissions on the SORMAS side, we would like to be able to just rely on those instead of duplicating them in Keycloak as well.

### Proposed Change
- [x] Remove the user roles (`REST_USER`, `USER`, `REST_EXTERNAL_VISITS_USER`) from Keycloak SORMAS realm definition. [SORMAS.json](https://github.com/hzi-braunschweig/SORMAS-Project/blob/b32d7fc25baf9e12962347e11e969a907e926268/sormas-base/setup/keycloak/SORMAS.json)
- [x] Update `de.symeda.sormas.backend.user.KeycloakService` and remove role related updates
- [x] Optional: find a way to remove the existing roles from users in Keycloak. Probably trough an script which is ran at startup or a DB script which can be handed over to netzlink
- [x] Update any Keycloak documentation which refers to the roles

### Possible Alternatives
If removing the user roles is not working due to EJB RollesAllowed limitations, replace the user roles in Keycloak with permissions.

### Additional Information
This is a prerequisite for https://github.com/hzi-braunschweig/SORMAS-Project/issues/5056

vidi42:
Fixes #7820

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/user/KeycloakService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/user/KeycloakService.java
@@ -27,11 +27,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 
 import javax.annotation.PostConstruct;
 import javax.ejb.EJB;
@@ -44,16 +40,12 @@ import javax.json.JsonObjectBuilder;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 
-import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.microprofile.config.ConfigProvider;
 import org.keycloak.OAuth2Constants;
 import org.keycloak.admin.client.Keycloak;
 import org.keycloak.admin.client.KeycloakBuilder;
-import org.keycloak.admin.client.resource.RealmResource;
-import org.keycloak.admin.client.resource.UserResource;
 import org.keycloak.representations.idm.CredentialRepresentation;
-import org.keycloak.representations.idm.RoleRepresentation;
 import org.keycloak.representations.idm.UserRepresentation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -62,7 +54,6 @@ import com.jayway.jsonpath.JsonPath;
 
 import de.symeda.sormas.api.AuthProvider;
 import de.symeda.sormas.api.Language;
-import de.symeda.sormas.api.user.UserRole;
 import de.symeda.sormas.backend.common.ConfigFacadeEjb.ConfigFacadeEjbLocal;
 import de.symeda.sormas.backend.user.event.PasswordResetEvent;
 import de.symeda.sormas.backend.user.event.UserCreateEvent;
@@ -89,8 +80,6 @@ public class KeycloakService {
 
     private static final String ACTION_UPDATE_PASSWORD = "UPDATE_PASSWORD";
     private static final String ACTION_VERIFY_EMAIL = "VERIFY_EMAIL";
-
-    private static final String DEFAULT_USER_ROLE = UserRole._USER;
 
     private Keycloak keycloak = null;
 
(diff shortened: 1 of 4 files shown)


sormas-rest/src/main/java/de/symeda/sormas/rest/security/KeycloakHttpAuthenticationMechanism.java
```diff
@@ -54,6 +55,10 @@ public AuthenticationStatus validateRequest(HttpServletRequest request, HttpServ
 		Credential credential = new CallerOnlyCredential(keycloakContext.getToken().getPreferredUsername());
 		IdentityStoreHandler identityStoreHandler = CDI.current().select(IdentityStoreHandler.class, new Annotation[0]).get();
 		CredentialValidationResult result = identityStoreHandler.validate(credential);
+		if (result.getStatus() == CredentialValidationResult.Status.VALID) {
```
leventegal-she, line 58: Seems to be a hack, but I know it's hard to find the right way to force the usage of roles returned by the identity store.

leventegal-she (approved):

sormas-rest/src/main/java/de/symeda/sormas/rest/security/KeycloakHttpAuthenticationMechanism.java
vidi42: Unfortunately the whole way we use Keycloak with SORMAS feels like a hack 🙃

sormas-backend/src/main/java/de/symeda/sormas/backend/user/KeycloakService.java
```diff
@@ -269,17 +258,8 @@ private String createUser(Keycloak keycloak, User user) {
...
-        try {
-            ensureRoles(keycloak, userId, user.getUserRoles());
-        } catch (Exception e) {
-            logger.warn("Cannot set the user roles property, will remove the user");
-            keycloak.realm(REALM_NAME).users().delete(userId);
-            throw e;
-        }
 
-        return userId;
+        return pathSegments[pathSegments.length - 1];
```
JonasCir, line 262: Maybe add java doc or a comment what the return value looks like. Might be confusing for people to understand what exactly is returned and why we need the `-1`.

sormas-rest/src/main/java/de/symeda/sormas/rest/security/KeycloakHttpAuthenticationMechanism.java
JonasCir: I'm somewhat concerned that we don not do any authentication at all, at least I cannot see where we check the actual token from keycloak. 

`hasValidAuthorization` does only check the header being present. Then we use `CallerOnlyCredential` which basically seems to ignore the token altogether and then `KeycloakIdentityStore.validate` consequently does not care about the token at all. Can you point me to the code where the Java code checks either a cookie or a token being correctly issued by the registered Keyclaok?

JonasCir (changes requested):

vidi42:
@JonasCir The authentication is handled by the [`org.keycloak.adapters.servlet.KeycloakOIDCFilter`](https://github.com/keycloak/keycloak/blob/16.1.1/adapters/oidc/servlet-filter/src/main/java/org/keycloak/adapters/servlet/KeycloakOIDCFilter.java) which delegates to either the [`org.keycloak.adapters.BasicAuthRequestAuthenticator#authenticate`](https://github.com/keycloak/keycloak/blob/16.1.1/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/BasicAuthRequestAuthenticator.java#L51) or [`org.keycloak.adapters.BearerTokenRequestAuthenticator#authenticate`](https://github.com/keycloak/keycloak/blob/16.1.1/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/BearerTokenRequestAuthenticator.java#L64)

vidi42:
I will merge this PR. @JonasCir if you still have concerns or think we should improve on this part we can open a separate ticket.

JonasCir:
@vidi42 All good, I forgot that this is a chain of filters :)

PR #8234, opened 2022-03-04
Bugfix 7975 person directory chrashes when hospital informant click oncontact

### LINKED ISSUE #7975: [Persons] System crashes when Hospital Informant clicks on the 'Contact' tab in the 'Persons' directory
### Bug Description
The Hospital Informant doesn't have a 'Contacts' directory, so he shouldn't have a 'Contact' filter tab it the 'Persons' directory either. However, this is present and when the user click on 'Contact' the system crashes.

The system crashes and the error message appears also when the hospital informant user clicks on the 'All'
 filter or when the user clicks on the 'Persons' directory.

### Steps to Reproduce

1. Log in with a Hospital Informant user.
2. Open the 'Persons' directory
3. Click on the 'Contact' filter tab.

### Actual Behavior
The system crashes.

### Expected Behavior
The "Contact" filter should not be shown for users without the CONTACTS_VIEW user right.

### Screenshots
image

### System Details

* Device: Windows 10
* SORMAS version: 1.69.0-SNAPSHOT
* Server URL: local server and test-de1.sormas.netzlink.com
* User Role: Hospital Informant

alexcaruntu-vita:
Fixes #7975

### DIFF
--- a/sormas-api/pom.xml
+++ b/sormas-api/pom.xml
@@ -2,7 +2,7 @@
 	<parent>
 		<groupId>de.symeda.sormas</groupId>
 		<artifactId>sormas-base</artifactId>
-		<version>1.69.0-SNAPSHOT</version>
+		<version>1.70.0-SNAPSHOT</version>
 		<relativePath>../sormas-base</relativePath>
 	</parent>
 	<modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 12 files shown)


MateStrysewske (changes requested):
I don't quite get the proposed fix here - wouldn't that break the person directory for e.g. surveillance supervisors? 

Also, for some reason, there are pom changes in this PR that shouldn't be in it. Can you get rid of them?

PR #8241, opened 2022-03-04
bug/8199_case_task_display_wrong_region

### LINKED ISSUE #8199: Case tasks display region and district in directory instead of responsible region and district
### Bug Description

The case tasks in the task management list display the region and district depending on the place of stay. However, this should be responsible region and district as seen for e.g. contacts instead.
### Steps to Reproduce

1. Create a case and set the jurisdiction so that the place of stay falls into the same region and district
2. Create a task for that case
3. Check the task management list for region and district column

### Expected Behavior

The displayed region and district in the task management list should be based on the responsible region and responsible district

### Screenshots

taskEntityJurisdiction
taskEntityJurisdiction2

### System Details

* Device:
* SORMAS version: 1.68.2
* User Role: Surveillance / Contact Supervisor

dinua:
Fixes #8199


sergiupacurariu:
looks good, please format the code before merging

sergiupacurariu (approved):

PR #8242, opened 2022-03-04
#5056 roles allowed with user rights

### LINKED ISSUE #5056: Use RolesAllowed with UserRights instead of UserRoles [5]
### Problem Description
Hard-coded user roles need to be replaced with configurable ones. See #898.

Related to #2804, #2805 and #7820. Should be possible to work on this in parallel.

We need to be able to easily check whether the user has the needed permission to do something. One way to achieve this is to go on using the javax security RolesAllowed annotation, but instead of using it with UserRole, use it with UserRight.

### Proposed Change
The goal is to replace all usages of UserRole in javax security RolesAllowed with user rights.

- [x] Add SORMAS_REST user right as a replacement for the corresponding user role and the manual "USER" role. Give the user right to all user roles that need to access the REST api (users on district level and below for the mobile app, technical user roles that have to access the REST api)
- [x] Don't check the SORMAS_REST user right in every single resource anymore. Instead put it into the two identity stores.
- [x] Add SORMAS_TO_SORMAS_CLIENT user right ...
- [x] For REST_EXTERNAL_VISIT_USER introduce EXTERNAL_VISITS user right. Check for ExternalVisitsResource
- [x] For BAG_USER use the existing BAG_EXPORT user right
- [x] Add static final strings for all user rights to the UserRight enum (see UserRole) for usage with the RolesAllowed annotation
- [x] Move _SYSTEM string constant from UserRole to UserRight (not enum value)
- [ ] Write a code generator to update the mappings in web.xml and glassfish-web.xml for sormas-ui and sormas-rest. This should include all user rights
- [x] Update KeycloakIdentityStore, when it's properly configured and SormasIdentityStore to pass in the effective user rights instead of user roles to the realm
- [x] Replace all usages of UserRole in RolesAllowed annotations with the corresponding UserRight

Out of scope: It's not needed to add RolesAllowed annotations here, where there currently aren't any annotations yet. This will go into a separate issue.

leventegal-she:
Fixes #5056


sormas-backend/src/test/java/de/symeda/sormas/backend/user/UserFacadeEjbTest.java
```diff
@@ -268,13 +266,13 @@ public void testGetValidLoginRoles() {
...
-		Set<UserRole> validLoginRoles = getUserFacade().getValidLoginRoles(user.getUserName(), password);
-		assertThat(validLoginRoles, containsInAnyOrder(SURVEILLANCE_SUPERVISOR));
+		Set<UserRight> validLoginRoles = getUserFacade().getValidLoginUserRights(user.getUserName(), password);
+		assertThat(validLoginRoles, containsInAnyOrder(getUserRoleConfigFacade().getEffectiveUserRights(SURVEILLANCE_SUPERVISOR).toArray(new UserRight[]{})));
 
 		user.setActive(false);
 		getUserFacade().saveUser(user);
 
-		validLoginRoles = getUserFacade().getValidLoginRoles(user.getUserName(), password);
+		validLoginRoles = getUserFacade().getValidLoginUserRights(user.getUserName(), password);
```
vidi42, line 275: Can be renamed to `validLoginRights`

sormas-rest/src/main/java/de/symeda/sormas/rest/security/SormasRestUserRightFilter.java
```diff
@@ -0,0 +1,83 @@
...
+import javax.ws.rs.ForbiddenException;
+import javax.ws.rs.NotAuthorizedException;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.core.Response;
+
+import de.symeda.sormas.api.user.UserRight;
+
+@Priority(AUTHORIZATION)
+public class SormasRestUserRightFilter implements ContainerRequestFilter {
```
MartinWahnschaffe, line 38: @leventegal-she Why is this class needed? Isn't this already done by the Keycloak and SormasIdentityStore classes?

leventegal-she: The identity store classes do not produce the correct status code in the http response.

vidi42 (approved):

MartinWahnschaffe (approved):

PR #8254, opened 2022-03-06
#2804 - Replace direct usage of user roles [2]

### LINKED ISSUE #2084: In Sormas one Gesundheitsamt ID should be linked to one PIA study
### Situation Description
Every Gesundheitsamt can be assigned to one PIA study.

### Feature Description
Config file where one PIA study ID is mapped to one Gesundheitsamt
Start with 1:1 relation

### Additional Information
What happens when more than one Gesundheitsamt use the same SORMAS instance.
Maybe we can work with the owner of a case, so not the Gesundheitsamt itself is linked to a PIA Study, but every case/contact

### LINKED ISSUE #2804: Replace direct usage of user roles [2]
### Feature Description
We want to allow admins to configure user roles according to their needs #898.

Ideally do #2805 first.

### Problem Description
Some user roles are still used directly, which makes it impossible to create or customize user roles. This needs to be replaced by user rights or other mechanics.

### Proposed Change

- [x] Make UserRole deprecated
- [x] Remove AbstractAdoService.isAdmin, isSystem, hasUserRole (no longer used anyway)
- [x] UserService.getLabUsersOfLab: Use jurisdiction instead of user role. Same in SampleGrid and TraskGridComponent (search for usages of LAB_USER)
- [x] ADMIN: Introduce DEV_MODE user right and use it instead of ADMIN for DevModeView and assign to admin user
- [x] ADMIN: Use specific user rights for deletion of entities, e.g. in SampleController.getSampleEditComponent (search for usages of UserRole.ADMIN)
- [x] NATIONAL_* + REST_USER: ActionService.createUserFilter should check Jurisdiction instead.
- [x] DashboardMapComponent: Remove check for national user roles. Just check region and fallback to centerOfAllRegions.
- [x] Add CASE_RESPONSIBLE user right. Users with this right can be assigned as responsible user to a case and be automatically assigned to related tasks (CaseFacadeEjb.assignOfficerOrSupervisorToTask).
- [x] Add CONTACT_RESPONSIBLE user right. Users with this right can be assigned as responsible user to a contact and be automatically assigned to related tasks
- [x] Add EVENT_RESPONSIBLE user right. Users with this right can be assigned as responsible user to a contact and be automatically assigned to related tasks
- [x] Related "Responsible Office"r field strings should be renamed to "Responsible User"

For user role checks related to the weekly reports we have three parts that need to be changed:

- [x] Getting hospital and community informants (call to UserFacade.getUsersByAssociatedOfficer) needs to be replaced by retrieving users that have the right to WEEKLYREPORT_CREATE and have the associatedOfficer
- [x] Getting surveillance officers (UserService.getallByRegionandUserRolesInJurisdiction in WeeklyReportFacade.getSummariesPerOfficer) needs to be replaced by getting all district level users of that region that have the WEEKLYREPORT_CREATE user right
- [x] May already been touched: In UserEditForm:L168 the system gets surveillance officers as values for the associatedOfficer drop down. This should be replaced by users **associated to the district** that have the CASE_RESPONSIBLE or WEEKLYREPORT_CREATE user right. Also add a description for User.associatedOfficer to the description.properties "Superordiante district user responsible for case surveillance or weekly reporting". 

Finally:

- [x] For all user roles check whether they are still directly used in Java methods. Usage in annotions is ok, this will be handled by #5056. Usage related to email/sms is handled by #2805

### Additional Information

#8318 is handling the android app.

cazacmarin:
Fixes #2804


ChristopherRiedel (changes requested):
1. Remove changes in language specific properties files
2. Remove AbstractAdoService.isAdmin, isSystem, hasUserRole (no longer used anyway) as stated in the issue
3. UserService.getLabUsersOfLab: Use jurisdiction instead of user role
4. Adding of "responsible" rights is completely missing

ChristopherRiedel (commented):
Please also check again if every UserRole is not used in JavaClasses anymore besides tests and default user creation

PR #8265, opened 2022-03-07
Feature 7775 automatic archiving config

### LINKED ISSUE #7775: Configuration for automatic archiving [5]
### Feature Description
Currently SORMAS comes with automatic archiving for cases and events. Those are archived when they were not modified for a certain number of days (90 days default). The configuration is currently done in sormas.properties.

### Problem Description
The configuration should be moved to the feature configuration in the database.
It should be extended to cover the configuration for all entites that will need automatic archiving.

### Proposed Change

Note: We will add the configuration for all core entity types, but not implement it's functionality in this issue.

- [x] Add an AUTOMATIC_ARCHIVING feature type
- [x] Add an new entityType field to the FeatureConfiguration class using the CoreEntityType enum. This enum might later become a more generic EntityType enum
- [x] Add a list of entity types to the FeatureType enum that allows to define valid options for AUTOMATIC_ARCHIVING: case, contact, event, event participant, immunization, travel entry. So in the end we would have multiple AUTOMATIC_ARCHIVING entries in the database, each for one entity type
- [x] All of them use a property to define the number of days for automatic archiving. If no property is given, default to 90 days.
- [x] The number of days defined for event participants has to be <= the number of days for events - output validation to log
- [x] The number of days defined for contacts has to be <= the number of days for cases - output validation to log
- ~~Automatic archiving for cases requires automatic archiving for contacts to be active @MartinWahnschaffe decide what to do~~
- [x] Use it to replace the configuration from sormas.properties
- [ ] Extend the feature configuration documentation accordingly

### Possible Alternatives
Use the deletion configuration instead of the feature configuration.

### Additional Information

Out of scope: Not needed for campaigns.

#7774 will be implement archiving for contacts

BarnaBartha:
Fixes #7775


sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -10049,4 +10049,12 @@ ALTER TABLE campaigns_history ADD COLUMN archiveundonereason character varying(5
 
 INSERT INTO schema_version (version_number, comment) VALUES (445, 'CoreAdo: Introduce "end of processing date" #7247');
 
+-- 2022-03-04 Configuration for automatic archiving #7775
+ALTER TABLE featureconfiguration ADD COLUMN entitytype character varying(255);
+ALTER TABLE featureconfiguration_history ADD COLUMN entitytype character varying(255);
+
+CREATE TRIGGER
```
MateStrysewske, line 10056: ?

MateStrysewske (changes requested):

sormas-backend/src/main/resources/sql/sormas_schema.sql
BarnaBartha: wanted to add validations this way but clarified with Martin we don't want them here
so basically forgot to remove it - did it now

MateStrysewske (changes requested):
* Disabling the automatic archiving feature of an entity does not remove the hint from the directory when filtering by archived entities
* The event participant directory hint does not properly reflect the days specified in the database
* I realised that there's no hint in the contact directory; can you add that please?

MateStrysewske (approved):

Issue #8029, opened 2022-02-16
[Audit|Backend] Implement FHIR AuditEvent logging service [3]


JonasCir:
### Situation Description

For the general purpose of audit logging (#7904) we need to be able to log access to the SORMAS backend with details on what exactly has been accessed, without including sensitive data (e.g. names).

This issue is the first step. It defines how the logged data is represented and provides a service that allows to log audit events.

#### AuditEvent

For the logging output format, we are using the [FHIR R4 AuditEvent resource](https://www.hl7.org/fhir/auditevent.html#tabs-json), which is based on the IHE ATNA audit profile. FHIR is a well-known and established standard for exchange of medical related data. 

A `FHIR AuditEvent` resource has a specified [JSON representation](https://www.hl7.org/fhir/auditevent-example.json.html) which is compact and easy to digest by log processing systems like Loki or ELK.

An open source library (`hapifhir`) that provides the needed `AuditEvent` java classes is available here: [Github](https://github.com/hapifhir/org.hl7.fhir.core/tree/master/org.hl7.fhir.r4) [Maven](https://mvnrepository.com/artifact/ca.uhn.hapi.fhir/org.hl7.fhir.r4/5.6.35)

These are the most important fields in the AuditEvent class:

|Content|Description|Field in `AuditEvent` resource|
|-----|------|-------|
|Access type| Create/Read/Update/Delete/Execute| `action`
|timestamp| Timestamp of the event| `recorded`
|actor|Functional instance causing the event (e.g., user). Users should be identified with a human readable name besides using the UUID| `agent`
|Executing instance| Identifaction of the system(component) generating the audit event| `source.site`
|Executing class|Name of the executing class, including method name, parameters, and return values| `source.observer`
|Activity/Event| Description of the activity with unique reference(uuid) w.r.t. the accesses data| `entitiy`

Additional, there is the field `**type**`, which will be populated according to the following table.

| Code | System | Name | Description |
| ---- | ------ | ---- | ----------- |
| 110100 | http://dicom.nema.org/resources/ontology/DCM | Application Activity | Start, Stop |
| 110106 | http://dicom.nema.org/resources/ontology/DCM | Export | Database level export |
| 110112 | http://dicom.nema.org/resources/ontology/DCM | Query | A request for multiple entities |
[...]

PR #8283, opened 2022-03-08
[#8030] add audit logger

### LINKED ISSUE #8030: [Audit|Backend] Implement a suitable log sink for the audit log [5]
### Situation Description

#8029 implements a service for `AuditEvent` based logging. It currently logs into the default log. This needs to be separated so it can be configured individually (chapter 5.2.7) and a fallback mechanism can be used.

Logging mechanism will be based on `Logback`.

### Feature Description
Programmatically configure the `Logger` to [separate it from the existing logging mechanisms](http://logback.qos.ch/manual/loggingSeparation.html), to ideally parse a separate config file. This allows the different SORMAS users to configure the audit logging to their taste and needs.

A possible way (still needs to investigation though) is to make use of the technique described in this [post](https://dennis-xlc.gitbooks.io/the-logback-manual/content/en/chapter-3-configuration/configuration-in-logback/invoking-joranconfigurator-directly.html), where a logger instance is crated from a config file. The idea would be to create on instance of it (aka singleton), which then gets used for every call to the sink.

### Where do we write to
Netzlink requested us to only use a file based [appender](https://logback.qos.ch/manual/appenders.html). It is very likely that we need to provide a tuned version of the config file for production, therefore, it is very important to support configuration via config file.

JonasCir:
Fixes #8030

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/LogSink.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/LogSink.java
@@ -1,0 +1,51 @@
+package de.symeda.sormas.backend.audit;
+
+import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.classic.joran.JoranConfigurator;
+import ch.qos.logback.core.joran.spi.JoranException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.net.MalformedURLException;
+import java.nio.file.InvalidPathException;
+
+public class LogSink {
+    private static final Logger logger = LoggerFactory.getLogger(LogSink.class);
+    private static LogSink instance;
+
+    private Logger auditLogger;
+
+    private LogSink(String fileName) throws MalformedURLException, JoranException {
+
+
+        if (fileName.endsWith("xml")) {
+            File file = new File(fileName);
+            JoranConfigurator configurator = new JoranConfigurator();
+            LoggerContext context = new LoggerContext();
+            configurator.setContext(context);
+            configurator.doConfigure(file);
+            auditLogger = context.getLogger("Audit");
+        } else {
+            throw new InvalidPathException(fileName, "Please provide an XML file!");
+        }
+
+    }
+
+    public static synchronized LogSink getInstance() {
+        if (instance == null) {
+            try {
+                instance = new LogSink("/tmp/audit.xml");
+            } catch (MalformedURLException | JoranException e) {
+                logger.error("Could not create auditLogger: %s", e);
+            }
+        }
+        return instance;
+    }
+
+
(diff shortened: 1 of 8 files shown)


JonasCir:
Ready for review

sormas-backend/src/main/java/de/symeda/sormas/backend/audit/LogSink.java
```diff
@@ -0,0 +1,70 @@
...
+import java.net.MalformedURLException;
+import java.nio.file.InvalidPathException;
+
+/**
+ * Provides a configurable log sink for the SORMAS audit trail.
+ */
+public class LogSink {
+    private static final Logger logger = LoggerFactory.getLogger(LogSink.class);
+    private static LogSink instance;
+    private static final String CONFIG_PATH = "/opt/config/audit-logback.xml";
```
vidi42, line 34: Wouldn't it make sense to provide this value trough a `sormas.properties` property?
That way if the property is not configured we can deactivate the LogSink.
In case the LogSink is deactivated then `getAuditLogger` can probably provide a mock Logger which doesn't log anything.

vidi42 (commented):
Also the new audit-logback.xml file should also be mentioned in `/docs/SERVER_SETUP.md`

sormas-backend/src/main/java/de/symeda/sormas/backend/audit/LogSink.java
JonasCir: I was also thinking about this, maybe we do it in the future. On reason why I went with this for now is that we likely need to use POJOs, as the interceptors in the backend we are going to use are not running in an EJB/CDI context, so using anything which is a bean from there would likely be even more painful.

JonasCir: Also pelase note this is not live for now. The important part is that I can programmatically read the config, which was the ultimate goal. Once I start with #8029, this will probably be absorbed anyway and then we can take care of the configuration aspect.

JonasCir: I might have a solution....

JonasCir: Fixed!

vidi42 (approved):

JonasCir:
@MartinWahnschaffe @vidi42 Small fixes, this is no ready

sormas-base/setup/sormas.properties
```diff
@@ -176,6 +176,11 @@ app.url=
 # Possible Values: true, false
 #auditor.attribute.logging=true
 
+# Config file path of the audit logger
+# Default: ""
```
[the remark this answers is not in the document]
JonasCir: fixed

Issue #8025, opened 2022-02-16
[Audit|Backend] Write calls to backend methods into audit log [8]


JonasCir:
### Situation Description

#8029 provides a service to write audit events into an audit log. We need to make use of this to create log entries for all calls to backend methods.

We will use the interceptor pattern, similar as we are doing it with `PerformanceLoggingInterceptor`.

### Feature Description

Implement the logging of "patient" related calls.

- [x] Add a `kogBackendCall` method to the `AuditLogger` (provided by #8029)
- [x] Parameteres are: called Class, called Method, List of parameters, return value
- [x] The user either also needs to be parameter or provided in another way
- [x] Check if the audit-log is active

In the `logBackendCall` method we need to fill the `AuditEvent` (similar to #8029).

 ~~type, subtype: TODO~~ tracked in #8464 separately
- [x] `action`: See below
- [x] `recorded`: current timestamp
- [x] `agent`: the user (uuid and user name). 
  Coding is "humanuser" from http://terminology.hl7.org/CodeSystem/extra-security-role-type
  who.identified is the uuid
  name is the user name
- [x] `source.site`: URL of the server - see #8029
- [x] `source.type`: see #8029
- [x] `source.observer`: class name + method name, e.g. `de.symeda.sormas.api.caze.CaseFacade.getIndexList`
- [x]    For uuid based DTOs or lists of DTOs log the UUID, e.g. CaseFacade.saveCase(HGT23S-....)
- [x] Result parameters: Same as for call parameters. E.g. CaseFacade.getIndexList would have a list of UUIDs to be shown here
- [x]  May be necessary to exclude methods from being logged (e.g. read access to regions, districts, ...)
- [x] Identify the **action** (Create/Read/Update/Delete/Execute) related to the call. We can't distinguish create and update, so we will only use update.
- [x] For uuid based DTOs or lists of DTOs log the UUID, e.g. CaseFacade.saveCase(HGT23S-....)
- [x] create a helper method and a mapping that takes the called Method object and identifies the action based on the name.
  * Read: count, get, is, has, does, validate
  * Update: save, update, post, set, archive, dearchive
  * Delete: delete, merge
  * Execute: send
[...]

PR #8306, opened 2022-03-09
[#8029] add audit logging for server start

### LINKED ISSUE #8029: [Audit|Backend] Implement FHIR AuditEvent logging service [3]

JonasCir:
Fixes #8029 

```json
{
  "resourceType": "AuditEvent",
  "type": {
    "system": "https://hl7.org/fhir/R4/valueset-audit-event-type.html",
    "code": "110100",
    "display": "Application Activity"
  },
  "subtype": [
    {
      "system": "https://hl7.org/fhir/R4/valueset-audit-event-sub-type.html",
      "code": "110120",
      "display": "Application Start"
    }
  ],
  "action": "E",
  "recorded": "2022-03-10T13:21:46.875+01:00",
  "outcome": "0",
  "outcomeDesc": "Application starting",
  "agent": [
    {
      "type": {
        "coding": [
          {
            "system": "https://www.hl7.org/fhir/valueset-participation-role-type.html",
            "code": "110151",
            "display": "Application Launcher"
          }
        ]
      },
      "name": "SYSTEM"
    }
  ],
  "source": {
    "site": "sormas_b",
    "type": [
      {
        "system": "http://terminology.hl7.org/CodeSystem/security-source-type",
        "code": "4",
        "display": "Application Server"
      }
    ]
  },
  "entity": [
    {
      "what": {
        "reference": "StartupShutdownService"
      },
      "type": {
        "system": "http://terminology.hl7.org/CodeSystem/audit-entity-type",
        "code": "2",
        "display": "System Object"
      }
    }
  ]
}
```

### DIFF
--- a/sormas-base/pom.xml
+++ b/sormas-base/pom.xml
@@ -27,7 +27,6 @@
 		<!-- *** Other dependency versions *** -->
 		<glassfish.jaxb.version>2.3.3</glassfish.jaxb.version>
 		<slf4j.version>1.7.36</slf4j.version>
-		<hapifhir.version>5.7</hapifhir.version>
 		<logback.version>1.2.10</logback.version>
 		<vaadin.version.warning>TODO: Remove bootstrap.js in widgetset</vaadin.version.warning>
 		<vaadin.version>8.14.3</vaadin.version>
--- a/sormas-base/pom.xml
+++ b/sormas-base/pom.xml
@@ -27,6 +27,7 @@
 		<!-- *** Other dependency versions *** -->
 		<glassfish.jaxb.version>2.3.3</glassfish.jaxb.version>
 		<slf4j.version>1.7.36</slf4j.version>
+		<hapifhir.version>5.7</hapifhir.version>
 		<logback.version>1.2.10</logback.version>
 		<vaadin.version.warning>TODO: Remove bootstrap.js in widgetset</vaadin.version.warning>
 		<vaadin.version>8.14.3</vaadin.version>
@@ -584,6 +585,13 @@
 				<version>${slf4j.version}</version>
 				<scope>provided</scope>
 			</dependency>
+			<dependency>
+				<groupId>ca.uhn.hapi.fhir</groupId>
+				<artifactId>hapi-fhir-structures-r4</artifactId>
+				<version>5.7.0</version>
+				<scope>runtime</scope>
+			</dependency>
+
 			<!-- *** Domain libs END *** -->
 
 			<!-- *** Compile dependencies *** -->
--- a/sormas-base/pom.xml
+++ b/sormas-base/pom.xml
@@ -584,10 +584,15 @@
 				<version>${slf4j.version}</version>
 				<scope>provided</scope>
 			</dependency>
-
 			<!-- *** Domain libs END *** -->
 
 			<!-- *** Compile dependencies *** -->
+			<dependency>
+				<groupId>ca.uhn.hapi.fhir</groupId>
+				<artifactId>org.hl7.fhir.r4</artifactId>
+				<version>5.6.36</version>
+				<scope>compile</scope>
+			</dependency>
 
 			<dependency>
 				<groupId>com.googlecode.libphonenumber</groupId>
--- a/sormas-api/src/main/java/de/symeda/sormas/api/ConfigFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/ConfigFacade.java
@@ -143,4 +143,6 @@ public interface ConfigFacade {
 	long getImportFileSizeLimitMb();
 
 	String getAuditLoggerConfig();
+
+	String getAuditSourceSite();
 }
--- a/sormas-backend/pom.xml
+++ b/sormas-backend/pom.xml
@@ -247,6 +247,10 @@
             <artifactId>logback-core</artifactId>
 			<scope>compile</scope>
         </dependency>
+		<dependency>
+			<groupId>ca.uhn.hapi.fhir</groupId>
+			<artifactId>org.hl7.fhir.r4</artifactId>
+		</dependency>
 
     </dependencies>
 
--- a/sormas-backend/pom.xml
+++ b/sormas-backend/pom.xml
@@ -251,8 +251,13 @@
 			<groupId>ca.uhn.hapi.fhir</groupId>
 			<artifactId>org.hl7.fhir.r4</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>ch.qos.logback</groupId>
+			<artifactId>logback-classic</artifactId>
+			<scope>compile</scope>
+		</dependency>
 
-    </dependencies>
+	</dependencies>
 
 	<build>
 		<finalName>${project.artifactId}</finalName>
--- a/sormas-backend/pom.xml
+++ b/sormas-backend/pom.xml
@@ -252,6 +252,12 @@
 			<artifactId>org.hl7.fhir.r4</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>ca.uhn.hapi.fhir</groupId>
+			<artifactId>hapi-fhir-structures-r4</artifactId>
+			<scope>runtime</scope>
+		</dependency>
+
+		<dependency>
 			<groupId>ch.qos.logback</groupId>
 			<artifactId>logback-classic</artifactId>
 			<scope>compile</scope>
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLogger.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLogger.java
@@ -1,3 +1,18 @@
+/*
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2022 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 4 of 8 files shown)


sormas-base/pom.xml
```diff
@@ -584,10 +584,22 @@
 				<version>${slf4j.version}</version>
 				<scope>provided</scope>
 			</dependency>
+			<dependency>
+				<groupId>ca.uhn.hapi.fhir</groupId>
+				<artifactId>hapi-fhir-structures-r4</artifactId>
+				<version>5.7.0</version>
+				<scope>runtime</scope>
```
MartinWahnschaffe, line 591: should be provided and the lib also needs to be added to the serverlibs.pom

sormas-base/pom.xml
```diff
@@ -584,10 +584,22 @@
...
+			</dependency>
 
 			<!-- *** Domain libs END *** -->
 
 			<!-- *** Compile dependencies *** -->
+			<dependency>
```
MartinWahnschaffe, line 597: shouldn't this also be provided as a serverlib in the domain?

MartinWahnschaffe (changes requested):
Already looks quite good!

See my comments on the dependencies.

Why is the source.site such a strange number string in your example?

JonasCir:
@MartinWahnschaffe I don't really know how all this maven dependency magic plays out. With this configuration it just works :tm: .
I tried to provide them as server libs, but then: 1. compilation failed b/c some imports were not available during compile 2. and second I ran into an issue winch already prevents other libs to be deployed as serverlib.

I'm not familiar with this, Is there any particular reason we need to change it? `org.hl7.fhir.r4` must be available at compile time, while `hapi-fhir-structures-r4` is purely a runtime dependency.

MartinWahnschaffe:
Yes, the reason to change is is to make sure the build artifact that needs to be deployed does not become 10MB bigger again. Also it's very likely that the FHIR resource will be needed by the rest, backend and ui archives, so we should not put it into the server three times (included in each archive).

You can have a look at the com.opencsv dependency. If structures is just a runtime depdency you probably don't have to mention it in the backend pom.

On the server URL: We already have someting similar for the app.url property. We can't re-use that one, but it means that we probably already have a mechanism in sormas docker that fills in urls into the property file.

JonasCir:
Hi @MartinWahnschaffe 
thanks for the feedback. I cannot deploy the dependencies as server libs right now due to https://github.com/hzi-braunschweig/SORMAS-Project/issues/8324, which as also affected the dependencies for Keycloak in the past.

My proposed way forward given the time constraints we facing is:
1. Leave the dependencies as they are now
2. Fix #8324 seperately
3. Introduce a new server property for the hostname. (app.url is also set like the rest of the properties, see [here](https://github.com/hzi-braunschweig/SORMAS-Docker/blob/master/sormas/start-server.sh#L236)) 
4. Merge this to unblock https://github.com/hzi-braunschweig/SORMAS-Project/issues/8025

sormas-base/pom.xml
JonasCir: https://github.com/hzi-braunschweig/SORMAS-Project/issues/8324

sormas-base/pom.xml
JonasCir: https://github.com/hzi-braunschweig/SORMAS-Project/issues/8324

MartinWahnschaffe (approved):

PR #8316, opened 2022-03-09
Bugfix 8193 two positive test results of different diseases duplicates the sample

### LINKED ISSUE #8193: Adding 2 positive Pathogen Test Result of different diseases to a Sample, duplicates the Sample
### Bug Description
When the user adds 2 positive Pathogen Test Results (each for a different Disease), the Sample in the Contact's Sample Card gets duplicated.

This is reproducible for:
- Contacts
- Event Participants

### Steps to Reproduce

1. Create a Contact and set the Disease to 'COVID-19';
2. Add a Sample with a 'Positive' Pathogen Test Result (with the Disease set to COVID-19);
3. Click 'Yes' in first pop-up to create a Resulting Case for the Contact;
4. Fill in the 'Create new case' form with the mandatory data;
5. Go back to the Contact and open the Sample;
6. Add a 'New positive test result' to the Sample (with the Disease set to Anthrax);
7. Click 'Yes' in the second pop-up that asks the user to create a case for Anthrax;
8. Open the Contact again and observe the Sample card.

### Actual Behavior
There are 2 Samples. Basically the Sample gets duplicated. If the user opens them, each of the samples show the same 2 pathogen test results (the one for COVID19 and the one for Anthrax).

### Screenshots
samples

### System Details

* Device: Windows 10
* SORMAS version: 1.69.0-SNAPSHOT (19c54c2)
* Server URL: local machine
* User Role: Admin, NatUser

### Additional Information
- Found this issue while testing ticket #7922

alexcaruntu-vita:
Fixes #8193

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -1566,7 +1566,7 @@ public class CaseFacadeEjb extends AbstractCoreFacadeEjb<Case, CaseDataDto, Case
 		if (sourceContact != null) {
 			final Contact contact = contactService.getByUuid(sourceContact.getUuid());
 			final Case caze = service.getByUuid(cazeRef.getUuid());
-			contact.getSamples().forEach(sample -> sample.setAssociatedCase(caze));
+			contact.getSamples().forEach(sample -> sampleFacade.cloneSampleForCase(sample, caze));
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


MateStrysewske (changes requested):
There's a failing test in SampleFacadeEjbTest.testGetIndexListForCaseConvertedFromContact

MateStrysewske (changes requested):
The sample should only be cloned if it already has a case associated with it. If the user does the following, the sample shouldn't be cloned:

- Create a contact with disease X
- Create a sample + positive pathogen test for the contact
- Click "Yes" in the popup that asks the user to create a case for the contact

In this case, we should stick with one sample that is connected to both the contact and the case.

ONLY if the user saves a pathogen test of a different disease to a sample that already has an associated case, and if they decide to create a case for this disease, the sample should be cloned and associated with this new case (and the cloned sample should not be associated with the contact).

MateStrysewske (approved):

PR #8383, opened 2022-03-15
Fixed SQL recursion introduced by #7323

### LINKED ISSUE #7323: Add "changeUser" field to entities and history to know which user last changed an entity [5]

MartinWahnschaffe:
Fixes #7323

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractDomainObject.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractDomainObject.java
@@ -27,6 +27,7 @@ import javax.persistence.Basic;
 import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.EntityListeners;
+import javax.persistence.FetchType;
 import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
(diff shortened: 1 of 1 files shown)


JonasCir (approved):
lgtm, thank you for taking care of this!
I'm also confused this did not show up. I tested it manually and the unit tests also were fine :D

BarnaBartha (approved):

PR #8394, opened 2022-03-15
Feature/8097 show entities name

### LINKED ISSUE #8097: [Data Protection Dictionary] The Excel worksheets of the data protection Dictionary should show the technical entity names [0.5]
### Situation Description
Currently the names of the worksheet are translated, this means that for example the entity is casedata the worksheet is named case. For the data protection officers it would be helpful and less confusing when the worksheets and the technical entity name is equal

### Feature Description
When a data protection dictionary is exported, then all worksheets that contain information about the dto should have the entity name as shown in the attached picture.
dpd

carina29:
Fixes #8097


leventegal-she (changes requested):
There is no need of introducing new captions, the ticket asks to not have any translated sheet names.

The sheet names should be changed to  the classname without the Dto suffix. @see `DataHelper.getHumanClassName`
But this should be done only when generating the data protection data dictionary.

sormas-backend/src/main/java/de/symeda/sormas/backend/info/InfoFacadeEjb.java
```diff
@@ -329,16 +524,21 @@ private String generateDataDictionary(
 		return documentPath.toString();
 	}
 
+	public String getNameForDataDictionaryType(boolean isDataProtectionDictionary, Class<? extends EntityDto> entityClass, String i18nPrefix) {
```
leventegal-she, line 527: I would not introduce a function for this one line called only once

leventegal-she (changes requested):

carina29:
This PR still needs a fix for a failing unit test.

PR #8403, opened 2022-03-16
#8325 Disable beantest and archunit internal debug logs for testing

### LINKED ISSUE #8325: CI build logs are too big and not usable anymore
### Bug Description
I'm not certain which change caused this, but recently I noticed that the output of `mvn verify` got extremely noisy. This has reached a point, where GitHub Actions will not display the log, but only allow us to download it in a raw format.

### Steps to Reproduce
Run `mvn verify` or take a look at a recent GH Action build.

### Expected Behavior
Log is concise and does not repeatedly print unnecessary internal DEBUG messages (like literally a thousand times :)

vidi42:
+ also disable StartupShutdownServiceTest#testHistoryTablesMatch

Fixes #8325

### DIFF
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/AbstractBeanTest.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/AbstractBeanTest.java
@@ -233,6 +233,7 @@ public abstract class AbstractBeanTest extends BaseBeanTest {
 		initH2Functions();
 		// this is used to provide the current user to the ADO Listener taking care of updating the last change user
 		System.setProperty("java.naming.factory.initial", MockProducer.class.getCanonicalName());
+		System.setProperty("logback.configurationFile", "./logback.xml");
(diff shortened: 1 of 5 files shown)


JonasCir:
Hi Alex thank you so much :)
```
 2022-03-16 10:26:08,943 INFO  (de.symeda.sormas.backend.user.KeycloakService.java:90) - Keycloak Auth Provider not active
2022-03-16 10:26:08,943 WARN  (de.symeda.sormas.backend.user.KeycloakService.java:133) - Cannot obtain keycloak instance. Will not create user in keycloak
``` 
I noticed this line to be printed often as well. Is it something we need to worry about or can we remove it as well?

vidi42:
> I noticed this line to be printed often as well. Is it something we need to worry about or can we remove it as well?

We can safely remove those lines as well as they don't provide meaningful information for tests. I changed the root logging level to ERROR

JonasCir (approved):
Ship it!

leventegal-she (approved):

PR #8417, opened 2022-03-17
Reinit the Events view when navigating to it from the Event Group view #7754

### LINKED ISSUE #7754: Access to the event directory filtered on the events of a group
### Feature Description
From the event file, the user can access the directory of events filtered on the events of a group.
This feature requires further development.

### Problem Description
When the user opens the event file from the event directory, the link works very well.
On the other hand, when the user opens the event file from the groups directory (accessible via the "Groups" view), this button returns to the groups view and not to the events view.

### Proposed Change
Force orientation to the "Events" view of the events directory to guarantee the expected operation of this button.

### Additional Information

aponcon-atolcd:
Fixes #7754

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventController.java
@@ -553,6 +553,13 @@ public class EventController {
 	}
 
 	public void navigateTo(EventCriteria eventCriteria) {
+		navigateTo(eventCriteria, false);
+	}
+
+	public void navigateTo(EventCriteria eventCriteria, boolean changeToDefaultViewType) {
+		if (changeToDefaultViewType) {
+			ViewModelProviders.of(EventsView.class).remove(EventsViewConfiguration.class);
+		}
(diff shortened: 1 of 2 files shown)


aponcon-atolcd:
Hi Mate,

The main issue was about this button :
image

This button is inside the Event groups panel inside an Event page and should redirect the user to the Events directory filtered by the related group.
But without the fix, if the user had previously visited the Event Groups directory, the button doesn't redirect the user to the Events directory but to the Event Groups directory.

I tested again the fix I made and it's ok on my side (for this button I mean). 

But I agree with you that the button "Events list" in the header of an Event Group or Event page may redirect to the Event Groups directory (instead of the Events directory) as you described.
Do you think we need to change the behaviour of this button too?

MateStrysewske:
@aponcon-atolcd Aaah, okay, thanks! I misunderstood the specification then. :) Gonna test again today. And no, I don't think that's necessary right now, at least not for the upcoming release (since it wasn't what was specified in the issue in the first place).

MateStrysewske (approved):

PR #8420, opened 2022-03-17
Feature 7845 add date of arrival to travel entries and fill it when importing dea information

### LINKED ISSUE #7845: Add dateOfArrival to travel entries and fill it when importing DEA information [3]
**Please look at the [comment](https://github.com/hzi-braunschweig/SORMAS-Project/issues/7845#issuecomment-1068134559) by @ChristopherRiedel.**

### Situation Description

Travel entries currently don't have a dedicated arrival date field; instead, this date is part of the deaContent column.
The value is needed for automatic deletion (#7736).

### Feature Description

- [x] Add a `dateOfArrival` field to travel entries as a required field. Set it to the creationDate of the TravelEntry (SQL)
- [x] Migrate the existing data (via SQL if possible to avoid long running migration via Java) by setting the date of arrival based on the DeaContentEntry value with the caption (DEA field_id) "Einreisedatum". Keep the DEA value, so that users are able to manually compare the calculated value with the original one
- [x] Automatically fill dateOfArrival when importing CSV data. The DEA field_id for this is "Einreisedatum". Rows that don't have a proper value should be skipped
- [x] When an automatic deletion uses ORIGIN as reference for a travel entry, use the dateOfArrival field as reference.

### Additional Information

Example import file: https://github.com/hzi-braunschweig/SORMAS-Project/files/6747727/export_einreiseanmeldung_210414.2.txt

leventegal-she:
Fixes #7845


sormas-ui/src/main/java/de/symeda/sormas/ui/travelentry/importer/TravelEntryImporter.java
```diff
@@ -123,6 +123,8 @@ protected ImportLineResult importDataFromCsvLine(
 					return ImportLineResult.ERROR;
 				}
 			}
+		} else if(importResult.getResult() == de.symeda.sormas.api.importexport.ImportLineResult.SKIPPED){
```
MateStrysewske, line 126: Code formatting is missing here as well

MateStrysewske (changes requested):
I believe that skipping the entry when the date of arrival is missing is not the right way to go (@MartinWahnschaffe) because it gives no feedback to the user about what's wrong. I would say we should rather label the line as an error and add it to the error template, like we do it in all other cases where the template is not correctly filled.

MateStrysewske (approved):

Issue #7723, opened 2022-01-20
[Angular][Backend] Vaccination getById [2] 


laurentiustanvitagroup:
### Situation Description

- In view of the future replacement of the current FE with the Angular component, the Angular team will focus on the required changes.

### Feature Description

- Reference: https://www.sketch.com/s/fd061796-65a3-4991-886a-c8a172667c91/a/Jnbqq19
- We need to create the getById vaccination endpoint (included on immunisation, or a dedicated route for vaccines). 
- We need to retrieve all of the data for a vaccination.

### Possible Alternatives

- N/A

### Additional Information

- N/A

vidi42:
@marius2301 This seems to break the deployment of `sormas-rest` because in `de.symeda.sormas.rest.ImmunizationResource` there are now two methods adnotated with `@Path("/push")`: `post` and  `postVaccination`

PR #8426, opened 2022-03-17
#7723 get vaccination by uuid and update vaccination based on the uui…

marius2301:
…d using the patch method

Fixes #


sormas-rest/src/main/java/de/symeda/sormas/rest/ImmunizationResource.java
```diff
@@ -143,4 +146,34 @@ public VaccinationDto createVaccination(
...
 
+	@GET
+	@Path("/vaccination/{uuid}")
+	public VaccinationDto getVaccinationByUuid(@PathParam("uuid") String uuid) {
+		return FacadeProvider.getVaccinationFacade().getByUuid(uuid);
+	}
+
+	@POST
+	@Path("/editVaccination")
+	public VaccinationDto createVaccination(@Valid VaccinationDto vaccination) {
```
MateStrysewske, line 157: Please align the naming of the endpoint as well as the method name with POST methods in other resources

MateStrysewske (changes requested):

sormas-rest/src/main/java/de/symeda/sormas/rest/ImmunizationResource.java
marius2301: I made the changes, however there is a difference between the other endpoints.

sormas-rest/src/main/java/de/symeda/sormas/rest/ImmunizationResource.java
```diff
@@ -143,4 +146,34 @@ public VaccinationDto createVaccination(
...
 	}
 
+	@GET
+	@Path("/vaccination/{uuid}")
+	public VaccinationDto getVaccinationByUuid(@PathParam("uuid") String uuid) {
+		return FacadeProvider.getVaccinationFacade().getByUuid(uuid);
+	}
+
+	@POST
+	@Path("/push")
```
vidi42, line 156: Please change the path since it creates a conflict with the already existing method `post`.

sormas-rest/src/main/java/de/symeda/sormas/rest/ImmunizationResource.java
```diff
@@ -143,4 +146,34 @@ public VaccinationDto createVaccination(
...
+	 * @param uuid
+	 * @param vaccinationDataDtoJson
+	 *            - a subset of caseDataDto fields, same structure as vaccinationDataDtoJson
+	 * @return - the updated caseDataDto
+	 * @throws Exception
+	 */
+	@POST
+	@Path("/vaccination/postUpdate/{uuid}")
+	@Experimental
+	public VaccinationDto postUpdate(@PathParam("uuid") String uuid, JsonNode vaccinationDataDtoJson) throws Exception {
```
vidi42, line 175: `the Exception` doesn't seem to be thrown from anywhere, can be removed

PR #8434, opened 2022-03-18
Feature 7713 history delete triggers

### LINKED ISSUE #7713: Delete history data on permanent deletion of entities [5]
### Situation Description
#7712 introduces permanent deletion for core entities (travel entry). When any entity is deleted, data protection also requires us to delete it's history data.

### Feature Description

We will use a database DELETE TRIGGER to propagate the deletion of history data.

- [x] Replace the existing TRIGGER for versioning to only be triggered on CREATE and UPDATE

```
CREATE TRIGGER versioning_trigger BEFORE INSERT OR UPDATE ON <table_name>
    FOR EACH ROW EXECUTE PROCEDURE versioning('sys_period', '<table_name>_history', true);
```

- [x] For each table that has a single id column (in other words for all that aren't join tables like contacts_visits), add a DELETE trigger that delete entries with the same id from the history table
- [x] For each join table add a DELETE trigger to each of the related tables to delete all entries from the history join table that reference the id. E.g. when deleting a contact, delete all entries from contacts_visits_history that have the same contact id.
- [x] Discuss: Can we extend the test provided by #6822 to validate triggers are in place?

### Possible Alternatives

This was the original implementation approach and estimated with [3], but does not provide a solution on how to delete data from join tables.

- Add a DELETE_HISTORY feature config that can only be used together with DELETE_PERMANENT
- Use the PostRemove JPA annotation to trigger the deletion of history data when an entity is deleted. See DefaultAuditListener in backend as an example. Class could be called JpaRemoveListener
- To identify the name of the history table related to the entity, I suggest to use reflection to get the name value from the Entity annotation of the class and append "_history" to it. #6822 will make sure that those tables are correctly setup, but still we have to make sure that this just gives a warning when the table does not exist (and doesn't crash).
- Delete all entries from the history table that have the same id as the deleted entity (native SQL delete)
- Problem: For entities with JoinTable annotation the entries with the id also have to be deleted from the respective history table

BarnaBartha:
Fixes #7713


MateStrysewske (approved):
Looks fine to me, but I shouldn't do the final evaluation here. @MartinWahnschaffe please also have a look over this.

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -10532,5 +10532,703 @@ DROP INDEX IF EXISTS externalshareinfo_caze_id_creationdate_idx;
...
+    FOR EACH ROW EXECUTE PROCEDURE versioning('sys_period', 'task_observer_history', true);
+
+DROP TRIGGER IF EXISTS versioning_trigger ON userroles_userrights;
+CREATE TRIGGER versioning_trigger
+    BEFORE INSERT OR UPDATE ON userroles_userrights
+    FOR EACH ROW EXECUTE PROCEDURE versioning('sys_period', 'userroles_userrights_history', true);
+
+DROP TRIGGER IF EXISTS versioning_trigger ON contacts_visits;
+CREATE TRIGGER versioning_trigger
+    BEFORE INSERT OR UPDATE ON contacts_visits
```
MartinWahnschaffe, line 11215: This may still have to trigger on deletion. Let's discuss

sormas-backend/src/test/resources/checkHistoryTables.sql
```diff
@@ -28,4 +28,50 @@ UNION
...
+  AND t.table_name ~* '_'
+  AND t.table_name NOT LIKE '%_history'
+  AND t.table_name NOT IN ('schema_version', 'systemevent')
+  AND t.table_name NOT like 'pg_%'
+  AND NOT exists(SELECT trigger_name
+                 FROM information_schema.triggers
+                 WHERE trigger_name = 'versioning_trigger'
+                   AND (event_manipulation = 'DELETE')
+                   AND event_object_table = t.table_name)
 ORDER BY remark, table_name , column_name;
```
MartinWahnschaffe, line 76: Let's also include this check:

```suggestion
UNION
SELECT 'invalid versioning trigger' as remark, t.table_name, null as column_name, 'DELETE' as data_type FROM information_schema."tables" t
WHERE t.table_schema = 'public'
  AND NOT (t.table_name ~* '_')
  AND t.table_name NOT LIKE '%_history'
  AND t.table_name NOT IN ('schema_version', 'systemevent')
  AND t.table_name NOT like 'pg_%'
  AND exists(SELECT trigger_name
                 FROM information_schema.triggers
                 WHERE trigger_name = 'versioning_trigger'
                   AND (event_manipulation = 'DELETE')
                   AND event_object_table = t.table_name)
ORDER BY remark, table_name , column_name;
```

MartinWahnschaffe (changes requested):

alexcaruntu-vita (approved):

BarnaBartha:
@MartinWahnschaffe i need some help/input here
the failing test is the testHistoryTablesMatch() - i don't understand why but it behaves differently within the container than on the local instance (both with old and new clean db)
can you please check it?

Issue #7736, opened 2022-01-20
Automatic deletion of personal data


StefanKock:
### Situation Description & Motivation

Data collected for cases, contacts and the other core entities has to be deleted automatically after a given time period.

#### Use cases
- Travel entries have to be deleted 14 days after the date of arrival in country
- Data collected within a case has to be deleted depending on different regulations. Some fields are kept for 10 years, others for only 3 years after the end of processing.

### High-Level Explanation

We need to define and implement the automatic deletion of person-related data within SORMAS. 

- The deletion has to be done privacy-compliant - there is no way to undo the deletion. (at least after a given time period)
- The given time period for the deletion should be given in SORMAS and can be modifyed by the Gesundheitsamt.
- Deletion is happening 
- The Deletion concept can be found here (Important files: Main document; Annex 1(Anlage 01 - Löschregeln) ; Annex 4 (Anlage 04 - Umsetzungsvorgaben für einzelne IT-Systeme)): https://confluence.sormas-tools.de/pages/viewpage.action?pageId=14488475

### Timeline

TBD

#### Concept Phase

- [ ] Specification of deletion concept & implementation work

#### Implementation Phase 1

Automatic-soft deletion of entities.

- [ ] hzi-braunschweig/SORMAS-Project#7008
- [ ] hzi-braunschweig/SORMAS-Project#7240 
- [ ] hzi-braunschweig/SORMAS-Project#7246 
- [ ] hzi-braunschweig/SORMAS-Project#7247
- [ ] hzi-braunschweig/SORMAS-Project#7319    
- [ ] hzi-braunschweig/SORMAS-Project#7249

#### Implementation Phase 2

- [ ] hzi-braunschweig/SORMAS-Project/issues/7712
- [ ] hzi-braunschweig/SORMAS-Project/issues/7713
- [ ] hzi-braunschweig/SORMAS-Project/issues/7716
- [ ] Field-wise deletion in history data
- [ ] Import deletion config from Excel / data protection doc: https://github.com/hzi-braunschweig/SORMAS-Project/issues/6911
- [ ] Deletion of a record after transfer to another health authority -> 30 days!
- [ ] Automatic archiving for all entities

### Additional Information

One idea would be to use VACCUUM_FULL once per month.

Marko-ilmkreis:
I would like to point out that the handling should be different upon the case classification. A "No Case" might be earlier deleted then a confirmed case.

[...]

PR #8442, opened 2022-03-18
Feature 7942 permanent case deletion

### LINKED ISSUE #7942: Permanent deletion for Case [5]
### Situation Description

### Feature Description

**Important:** Currently deleting a case also deletes all of the related contacts. This should no longer be done. Instead:

- [x] Remove the logic that deletes contacts from the CaseService.delete method
- [x] When manually deleting a case ask the user whether all related contacts should also be deleted. Add a deleteWithContacts method to the CaseFacade.

The logic for most of what is needed here is currently executed in CaseService.delete and needs to be moved. The only thing remaining in CaseService.delete should be the soft deletion of related samples.

Related entities to be deleted in deletePermanent method:

- [x] delete tasks (move from the CaseService delete method) 
- [x] delete surveillance reports (move from the CaseService delete method)
- [x] delete clinical visits (move from the CaseService delete method)
- [x] delete treatments (move from the CaseService delete method)
- [x] delete prescriptions (move from the CaseService delete method)
- [x] delete sample's reference and the sample itself if it's not referenced (duplicate the logic from delete method - when a case is permanently deleted all related samples are also permanently deleted)
- [x] call markAsDeleted for all documents related to the case

Entites where the Case reference needs to be removed:

- [x] remove the Case from any SormasToSormasShareInfo that is referencing it and delete it if no longer referenced by another entity
- [x] remove the Case from any ExternalShareInfo that is referencing it
- [x] Case.duplicateOf

Same for the following references, but these **also** need to be removed on soft-delete -> ideally move this unlinking to a separate method called from both soft- and permanent delete.

- [x] Contact.resultingCase
- [x] Contact.sourceCase
- [x] EventParticipant.resultingCase
- [x] TravelEntry.resultingCase
- [x] Immunization.relatedCase

Finally:

- [x] CoreEntityDeletionService: Call executePermanentDeletion for the CaseService

Since we are not deleting clinical visits etc. anymore when the case is soft-deleted we need to do the following:

- [x] Check DELETED flag of related case in ClinicalVisitService.getAllActiveClinicalVisitsAfter and getAllActiveUuids (already done vor ARCHIVED)
- [x] Check DELETED flag of related case in PrescriptionService.getAllActivePrescriptionsAfter and getAllActiveUuids (already done vor ARCHIVED)
- [x] Check DELETED flag of related case in TreatmentService.getAllActiveTreatmentsAfter and getAllActiveUuids (already done vor ARCHIVED)

MateStrysewske:
Fixes #7942


sormas-app/app/src/main/java/de/symeda/sormas/app/rest/SynchronizeDataAsync.java
```diff
@@ -119,10 +119,10 @@ protected Void doInBackground(Void... params) {
 				// Infrastructure always has to be pulled - otherwise referenced data may be lost (e.g. #586)
 				pullInfrastructure();
 				// Pull and remove deleted entities when the last time this has been done is more than 24 hours ago
-				if (ConfigProvider.getLastDeletedSyncDate() == null
-					|| DateHelper.getFullDaysBetween(ConfigProvider.getLastDeletedSyncDate(), new Date()) >= 1) {
-					pullAndRemoveDeletedUuidsSince(ConfigProvider.getLastDeletedSyncDate());
-				}
+//				if (ConfigProvider.getLastDeletedSyncDate() == null
```
BarnaBartha, line 122: i am guessing this is for testing purposes and should be reverted

BarnaBartha (changes requested):

JonasCir:
@MateStrysewske I'm going to reassign to Levi, I'm not very familiar with the share info, since I mainly worked on the underlying infrastructure. Deleting it might have other implications for S2S that I might not be aware of :smile:

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
```diff
@@ -880,62 +888,119 @@ public Predicate createDefaultFilter(CriteriaBuilder cb, From<?, Case> root) {
...
-				.stream()
-				.forEach(c -> clinicalVisitService.deletePermanent(c));
+			List<ClinicalVisit> cvs = clinicalVisitService.findBy(new ClinicalVisitCriteria().clinicalCourse(clinicalCourse));
+			cvs.forEach(c -> clinicalVisitService.deletePermanent(c));
 		}
 
-		// Remove all events linked to case by removing the case_id from event participant
-		caze.getEventParticipants().stream().forEach(eventParticipant -> eventParticipant.setResultingCase(null));
+		// Remove the case from any S2S share info referencing it
+		sormasToSormasShareInfoService.getByAssociatedEntity(SormasToSormasShareInfo.CAZE, caze.getUuid()).forEach(s -> {
```
leventegal-she, line 924: As there is always one share info for one shared object, share info never has multiple shared objects linked,  it would be better to delete the share info. 

Doesn't make sense to keep share infos not linked to any shared object.

Also other systems should be notified that the case has been deleted so the will not try to sync back later changes.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
```diff
@@ -880,62 +888,119 @@ public Predicate createDefaultFilter(CriteriaBuilder cb, From<?, Case> root) {
...
 
-		// Unlink TravelEntries where this case is set as the resulting case
-		List<TravelEntry> travelEntries = travelEntryService.getAllByResultingCase(caze);
-		for (TravelEntry travelEntry : travelEntries) {
-			travelEntry.setResultingCase(null);
-			travelEntryService.ensurePersisted(travelEntry);
-		}
+		// Remove the case from any external share info referencing it
+		externalShareInfoService.getShareInfoByCase(caze.getUuid()).forEach(e -> {
+			e.setCaze(null);
```
leventegal-she, line 931: Same as for share info never happens that an ExternalShareInfo is linked to case and contact too so it can be deleted here.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
```diff
@@ -880,62 +888,119 @@ public Predicate createDefaultFilter(CriteriaBuilder cb, From<?, Case> root) {
...
+			externalShareInfoService.ensurePersisted(e);
+		});
 
-		// Unlink Immunizations where this case is set as the related case
-		immunizationService.unlinkRelatedCase(caze);
+		// Remove the case from its S2S origin info
+		Optional.ofNullable(caze.getSormasToSormasOriginInfo()).ifPresent(o -> {
```
leventegal-she, line 936: there is no join table between origin info so there is no need to do `o.getCases().remove(caze);` the origin server should be notified that the case has been deleted otherwise it would try to sync future changes

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2358,7 +2367,21 @@ public void delete(String caseUuid) throws ExternalSurveillanceToolException {
...
+		deleteCase(caze);
+
+		Optional.of(caze.getContacts()).ifPresent(cl -> cl.forEach(c -> contactService.delete(c)));
+	}
+
 	private void deleteCase(Case caze) throws ExternalSurveillanceToolException {
+
 		externalJournalService.handleExternalJournalPersonUpdateAsync(caze.getPerson().toReference());
 		if (externalSurveillanceToolGatewayFacade.isFeatureEnabled() && caze.getExternalID() != null && !caze.getExternalID().isEmpty()) {
 			List<CaseDataDto> casesWithSameExternalId = getByExternalId(caze.getExternalID());
```
MartinWahnschaffe, line 2387: @MateStrysewske The deletion in external surveillance tool also has to be executed on permanent deletion. I forgot to put this into the issue.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
MartinWahnschaffe: Unsubscribing from the source system is currently out of scope (#7736).

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
MartinWahnschaffe: Unsubscribing from the source system is currently out of scope (#7736).

BarnaBartha (approved):

PR #8462, opened 2022-03-21
Change 8072 hide buried within person present condition for covid 19

### LINKED ISSUE #8072: Hide "buried" within Person present condition for Covid-19 [2]
### Feature Description

Under present condition of person, "buried" is not a relevant value for Covid-19. Cause of data protection and data frugality this value and the associated ones should be hidden.

### Proposed Change

Hide within "present condition of person" the value `buried` for Covid-19.
Following should be not available automatically, please check: 
`burialDate`
`burialPlaceDescription`
`burialConductor`

=> Those fields have to be annotated as well to be able to hide them in the data protection dictionary if needed

Consider Pop-ups, import, export, mobile app, and data protection Dictionary.

### Additional Information

When implementing this we do not consider the countries but only the disease. **EDIT:** This was partly not done. Hiding the burial related fields in export and import was done by generally hiding the fields for germany. All other checks are disease specific. This deviation was done because of time pressure, follow up tickets will be created.
Migration is not needed
Hide them in Mobile App as well.

@Candice-Louw @bernardsilenou @carolinverset @ClaudeMarksteinerCSI

FredrikSchaeferVitagroup:
Fixes #8072 

Many of the changes made here should be much more generic. Due to the high time pressure, it was decided to have an implementation for only the fields described in the ticket. This is faster to get merged and has a lower risk of breaking anything else.

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
@@ -619,7 +619,7 @@ public class PersonEditForm extends AbstractEditForm<PersonDto> {
 		 * It may happen that the person currently has a present condition that usually shall not be shows for the form's disease.
 		 * In that case, the present condition is added as selectable item here.
 		 */
-		if (presentConditionField.getItem(presentCondition) == null) {
+		if (presentCondition != null && presentConditionField.getItem(presentCondition) == null) {
 			Item currentItem = presentConditionField.addItem(presentCondition);
 			currentItem.getItemProperty(SormasFieldGroupFieldFactory.CAPTION_PROPERTY_ID).setValue(presentCondition.toString());
 		}
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
@@ -34,6 +34,8 @@ import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.List;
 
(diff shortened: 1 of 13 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
```diff
@@ -604,6 +611,20 @@ private void updateListOfDays(Integer selectedYear, Integer selectedMonth) {
...
+		PresentCondition presentCondition = getValue().getPresentCondition();
+		ComboBox presentConditionField = getField(PersonDto.PRESENT_CONDITION);
+
+		/*
+		 * It may happen that the person currently has a present condition that usually shall not be shows for the form's disease.
+		 * In that case, the present condition is added as selectable item here.
+		 */
+		if (presentConditionField.getItem(presentCondition) == null) {
+			Item currentItem = presentConditionField.addItem(presentCondition);
+			currentItem.getItemProperty(SormasFieldGroupFieldFactory.CAPTION_PROPERTY_ID).setValue(presentCondition.toString());
```
vidi42, line 624: if the `presentCondition` is null, this will result in a NPE. It just happened on my systems when I tried opening an existing Person

vidi42 (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonEditForm.java
FredrikSchaeferVitagroup: Even though I also noticed that problem, I forgot to add a null check here. Sorry.
There's a fix committed now.

vidi42 (approved):

PR #8519, opened 2022-03-24
#5930 - add back Berlin / Bayern / Bremen to the list of polygons to …

### LINKED ISSUE #5930: Hide other countries hides Bavaria, Bremen and Berlin
### Bug Description

If you select "hide other countries" in the map of the dashboard it also hides Bavaria, Bremen and Berlin:

image

### Steps to Reproduce

1. Open the map
2. choose the layer "Hide other countries"
image

### System Details

* SORMAS version: 1.59.2

BarnaBartha:
…be used for the country coordinates - as they are wrongly removed / not merged correctly with their neighbouring polygon

Fixes #5930

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/geo/GeoShapeProviderEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/geo/GeoShapeProviderEjb.java
@@ -46,9 +46,10 @@ import org.opengis.referencing.operation.MathTransform;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import de.symeda.sormas.api.infrastructure.district.DistrictReferenceDto;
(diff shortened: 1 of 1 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/geo/GeoShapeProviderEjb.java
```diff
@@ -382,6 +383,14 @@ private void buildCountryShape() {
 			}
 		}
 
+		// fixme: this should be removed if the above logic is rewritten/fixed
```
MartinWahnschaffe, line 386: Something like: "fixme: Remove when the shapefiles are no longer part of the resources. In this case the logic to build country shapes will have to be checked again to fix the problem is workarround by this."

PR #8583, opened 2022-03-28
Change 8318 replace user roles android app

### LINKED ISSUE #8318: Replace direct usage of user roles in Android app [2]
### Feature Description
We want to allow admins to configure user roles according to their needs #898.

### Problem Description
Some user roles are still used directly, which makes it impossible to create or customize user roles. This needs to be replaced by user rights or other mechanics.

### Proposed Change

**Depends on the new user rights introduced by #2804.**

For the Android app the following things need to be done:

- [x] Introduce User.hasJurisdictionLevel as a helper method, together with a jursidictionLevel field. When the jurisdictionLevel is null, use the user roles to calculate it (see UserRole.getJurisdictionLevel). When setUserRole of the user is called, set the jurisdictionLevel to null, so it's re-calculated when needed

**Replace calls to User.hasUserRole:**

- [x] CaseDao.build: Use the new CASE_RESPONSIBE user right instead of user role checks
- [x] UserRoleConfigDao: Can stay as-is.
- [x] CaseEditFragement.setUpFieldVisibilities: Check for jurisdictionLevel HEALTH_FACILITY
- [x] CaseNewFragement.onAfterLayoutBinding: Check jursidictionLevels
- [x] LoginActivity.openLandingActivity: CASE_VIEW in combination with UserRight.CASE_RESPONSIBLE or jursidiction levels: HEALTH_FACILITY, COMMUNITY, POINT_OF_ENTRY goes to cases. CONTACT_VIEW in combination with CONTACT_RESPONSIBLE goes to contacts. The other logic stays as-is.
- [x] Weekly reports: Here the hasUserRole check (1x) for community and hospital informant need to be replaced by a check to the WEEKLYREPORT_CREATE user right and the jurisdiction level community or facility. Checks for the surveillance officer (3x) need to be replaced with a check for the WEEKLYREPORT_CREATE user right and the jurisdiction level district.

Other logic to be replaced:

- [x] Replace the logic in CaseDao.assignOfficerOrSupervisorToTask by copying the logic from CaseFacadeEjb.assignOfficerOrSupervisorToTask including getRandomDistrictCaseResponsible and getRandomRegionCaseResponsible
- [x] UserDao.getInformantsByAssociatedOfficer needs to be replaced with getUsersByAssociatedOfficer(User officer, UserRight userRight) similar to the UserService backend method. To identify the userroles for the user right use the UserRight.getDefaultUserRoles method and ignore the UserRoleConfig for now, since this will be changed related to #4461 anyway.
- [x] Replace calls to User.getUserRoles in combination with UserRole.isPortHealthUser by checking whether the user has a pointOfEntry instead of using UserRole.isPortHealthUser. POE users that don't have this can't use the app anyway.

### Additional Information
#2804 is handling the backend part of this.

MateStrysewske:
Fixes #8318


MateStrysewske:
@ChristopherRiedel Since you've worked on the replacement of user roles in the web app, I think it makes sense that you do the review of this PR. :)

sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseNewFragment.java
```diff
@@ -302,12 +299,12 @@ public void onAfterLayoutBinding(final FragmentCaseNewLayoutBinding contentBindi
 			contentBinding.caseDataDifferentPlaceOfStayJurisdiction.setVisibility(GONE);
 		}
 
-		if (user.hasUserRole(UserRole.POE_INFORMANT) && user.getPointOfEntry() != null) {
+		if (user.getPointOfEntry() != null) {
```
ChristopherRiedel, line 302: user.hasJurisdictionLevel(JurisdictionLevel.POINT_OF_ENTRY) && user.getPointOfEntry() != null

ChristopherRiedel (changes requested):

sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseNewFragment.java
ChristopherRiedel: true

ChristopherRiedel (approved):

PR #8720, opened 2022-04-06
performance log analysis #8357

### LINKED ISSUE #8357: Detailed performance analysis of getAllAfter and getIndexList methods [5]
### Problem Description

Retrieving data via ReST during the mobile app synchronisation process can take a lot of time (with as few as 1000 cases taking several minutes to synchronize). We've also received reports (and witnessed) the web app directories taking more than a few seconds to load on systems with a lot of data (e.g. more than 500,000 entries in a directory).

### Proposed Solution

In preparation of potential refactorings and code improvements, analyse the performance of the getAllAfter and getIndexList methods of the following entities:

* Case
* Contact
* Sample
* Person

This should include all EJBs called in the process, and also especially include the pseudonymisation logic because this was already identified as a huge bottleneck at least for the synchronisation process.

Step 1: Activate `PerformanceLoggingInterceptor` on TRACE level to analyse call hierarchy and where time is spent. Document execution times.
Step 2: Dive deeper into hotspot method to find costly transformations (probably in `toDto`/`pseudonymize` methods).

### Additional Information

A potential follow-up issue to improve the pseudonymisation process could include the following action points:
- Skip pseudonymisation completely for entites that we know will not be pseudonymised based on the user's user rights (e.g. an entity inside the user's jurisdiction does not need to be checked for pseudonymisation if the user has both the SEE_PERSONAL_DATA_IN_JURISDICTION and SEE_SENSITIVE_DATA_IN_JURISDICTION user rights)
- Create a cache per entity of all fields that need to be considered during pseudonymisation in order to reduce the usage of reflection

Since the toDto logic also takes some time, another possible change could be to directly select into DTOs, even though that would potentially introduce a significantly increased maintenance effort.

As database where we want to reproduce  the perf. issues and measure the improvements we are using the pseudonymized data from France.

### Deliverables
Confirm that pseudonymization is one cause 
Explore what else causes performance issues
Document KPI (Key performance indicators) and preconditions to be able to measure the improvement
Create follow-up tickets that are obvious

==>
- pseudonymization can be ruled out as a relevant impact on performance
- long runtimes are mainly caused by long running SQL queries with massive joins
- follow-up issues: #8938 for `getIndexList` methods, #8946 for `getAll` methods in the REST API

syntakker:
This is not the final fix for #8357, please make sure not to close the issue when reviewing!

Fixes #8357

### DIFF
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
@@ -257,8 +257,8 @@ public class PerformanceLogAnalysisGenerator {
 			StringBuilder stringBuilder = new StringBuilder();
 			stringBuilder.append("<table>");
 			stringBuilder.append(
-				"<tr>" + header("method") + header("calls") + header("total") + header("max") + header("min") + header("mean") + header("subcalls")
-					+ "</tr>");
+				"<tr>" + header("method") + header("calls") + header("total") + header("max") + header("min") + header("mean")
+					+ (withSubcalls ? header("subcalls") : "") + "</tr>");
 			List<MethodStats> subcallsList = new ArrayList<>(subcalls.values());
 			Collections.sort(subcallsList);
 			for (MethodStats subMethod : subcallsList) {
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
@@ -22,10 +22,13 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.Stack;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -48,7 +51,7 @@ public class PerformanceLogAnalysisGenerator {
             "* {font-family: Verdana, Arial, Helvetica, sans-serif;}\n" +
             "a {font-weight: bold; color: #666; text-decoration: none;}\n" +
             "a:hover {color: #ff9900;}\n" +
-            "th {font-weight: bold;}\n" +
+            "th {font-weight: bold; background: #EEF}\n" +
             "th,td {padding: 0 5px 0 5px; vertical-align: top;}\n" +
             "tr:nth-child(odd) {background: #EEE}\n" +
             "tr:nth-child(even) {background: #FFF}\n" +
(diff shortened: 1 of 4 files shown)


StefanKock:
@MateStrysewske please look into this and help to provide an extension to the troubleshooting guide.

sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
```diff
@@ -0,0 +1,331 @@
...
+            "</html>\n";
+    //@formatter:on
+
+	public static final int TIME_TRESHOLD_YELLOW = 100;
+	public static final int TIME_TRESHOLD_ORANGE = 300;
+	public static final int TIME_TRESHOLD_RED = 1000;
+
+	public static final String OUTPUT_DIRECTORY = "target/performance/";
+
+	private Map<String, Stack<String>> callstacks;
```
StefanKock, line 70: Why instance variable? Better move to method if not needed outside

sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
```diff
@@ -0,0 +1,331 @@
...
+	private Map<String, MethodStats> methodStats;
```
StefanKock, line 71: Why instance variable? Better move to method if not needed outside

sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
```diff
@@ -0,0 +1,331 @@
...
+import java.util.regex.Pattern;
+
+import org.apache.commons.io.FilenameUtils;
+import org.jetbrains.annotations.NotNull;
+
+public class PerformanceLogAnalysisGenerator {
+
+	private static final Pattern startPattern =
+		Pattern.compile("([^ ]* [^ ]*) TRACE *(.*) d.s.s.b.u.PerformanceLoggingInterceptor - Started: ([^ ]*) with parameters .*$");
+	private static final Pattern finishPattern =
```
StefanKock, line 40: Contant -> uppercase

StefanKock (changes requested):

sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
syntakker: Now used outside the method in `getUnfinishedMethodsHtml`.

sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
syntakker: Obsolete and removed.

sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
syntakker: done.

sormas-backend/src/test/java/de/symeda/sormas/backend/performance/PerformanceLogAnalysisGenerator.java
```diff
@@ -0,0 +1,331 @@
...
+import java.util.Stack;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.commons.io.FilenameUtils;
+import org.jetbrains.annotations.NotNull;
+
+public class PerformanceLogAnalysisGenerator {
+
+	private static final Pattern startPattern =
```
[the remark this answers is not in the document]
syntakker: done.

syntakker:
Description added in `TROUBLESHOOTING.md`.

StefanKock (approved):

PR #8754, opened 2022-04-07
Adding additional test for Sample scenario

mp-sgent:


mp-sgent:
Scenario ready to be merged, during development found a bug so this tests will result in failure in execution until fix is provided.

rdutu-vg (approved):

PR #8782, opened 2022-04-11
Feature 8295 automatic manual deletion

### LINKED ISSUE #8295: Automatic permanent deletion for manually deleted entities [3]
### Situation Description
So far automatic deletion just did a soft-delete and there was a separate process that permanently deletes all automatically AND manually soft-deleted entities once a night.
#8293 changes this, because we need to be able to handle manually deleted entities different then automatically deleted ones. Manually deleted entities should stay in the system for another 90 days (example for Germany), so that a user can review and undo the deletion.

### Feature Description
Introduce automatic permanent deletion of manually deleted entities using the existing system for automatic deletion introduced by #7008.

- [x] Extend the DeletionReference with a DELETION value
- [x] Extend the automatic deletion logic for the new deletion reference. The rule applies to all entities that were soft-deleted and should use their change date as reference date.
- [x] Remove the source code for executePermanentDeletion since it is no longer needed.
- [x] Turn lab messages into an ADO (i.e. remove the `deleted` field) and write a migration script that removes all deleted lab messages from the database
- [x] Remove no longer needed DELETE_PERMANENT feature type

MateStrysewske:
Fixes #8295


MateStrysewske:
@BarnaBartha Please have another look at the changes. :)

PR #8784, opened 2022-04-11
#7931 - Permanent deletion for Contact and Visit

### LINKED ISSUE #7931: Permanent deletion for Contact and Visit [3]
### Situation Description

### Feature Description

**Contact**

The soft-delete of all samples related to the contact should stay as is (ContactService .delete calls TaskService.delete)

- [x] Move the logic that deletes tasks from the ContactService delete method to the deletePermanent method
- [x] Duplicate the logic that deletes samples for ContactService.deletePermanent. So when a contact is permanently deleted all related samples are also permanently deleted
- [x] In deletePermanent call markAsDeleted for all documents related to the contact
- [x] In deletePermanent remove the Contact from any SormasToSormasShareInfo that is referencing it and delete it if no longer referenced by another entity
- [x] In deletePermanent **and** delete remove the Contact from any Exposure.contactToCase that is referencing it -> see how unlinking is done for the case (probably a separate method called from both delete and deletePermanent)
- [x] In deletePermanent remove the Contact from any Contact.duplicateOf that is referencing it
- [x] CoreEntityDeletionService: Call executePermanentDeletion for the ContactService

**Visit**

This is similar to what we are doing for persons in #7857: When no contact or case is referencing a visit anymore, it needs to be deleted.

- [x] Add an executePermanentDeletion method to the VisitService. It should query all persons that are no longer referenced by a Case, or Contact and call the VisitService.deletePermanet method for those. Make sure to unit test this.
- [x] Add this call to the end of the executePermanentDeletion method in the CoreEntityDeletionService. Only PersonService should be called afterwards

sergiupacurariu:
Fixes #7931


sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactService.java
```diff
@@ -1373,9 +1383,81 @@ public void delete(Contact contact) {
...
 		externalJournalService.handleExternalJournalPersonUpdateAsync(contact.getPerson().toReference());
+		deleteContactLinks(contact);
+
 		super.delete(contact);
 	}
 
+	@Override
+	public void deletePermanent(Contact contact) {
+		// Delete all tasks associated with this contact
+		List<Task> tasks = taskService.findBy(new TaskCriteria().contact(new ContactReferenceDto(contact.getUuid())), true);
```
MateStrysewske, line 1394: I believe you can remove the task deletion from the `deleteContact` method above because, afaik, it only needs to be done once we permanently delete the contact (since tasks are not shown if they are associated to deleted entities in the first place). We did the same for cases.

MateStrysewske (changes requested):
lgtm apart from the one change request, but this PR should definitely also be reviewed by someone from the Interactions team; I'd suggest @leventegal-she.

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactService.java
sergiupacurariu: updated

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactService.java
```diff
@@ -1373,9 +1383,75 @@ public void delete(Contact contact) {
...
+				sormasToSormasShareInfoService.ensurePersisted(s);
+			} else {
+				sormasToSormasShareInfoService.deletePermanent(s);
+			}
+		});
+
+		deleteContactFromDuplicateOf(contact);
+
+		// Remove the deleted contact from contact_visits
+		visitService.getAllByContact(contact).forEach(visit -> {
```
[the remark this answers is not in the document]
sergiupacurariu: changed

leventegal-she (approved):

MateStrysewske (approved):

Issue #8508, opened 2022-03-23
Add backend checks for contact, visit and missing case-related user rights [5]


MartinWahnschaffe:
### Problem Description
We are missing checks for many of the user rights related to accessing the backend. #5057 and #445 have done this for the CaseFacade and PersonFacade and partly also for the ContactFacade.

Depdends on #8352.

### Proposed Change

Make sure all user rights for the facades listed below are checked. 
For most of the facades there is a corresponding _VIEW user right that should be used as a class anotation, so it's not necessary to annotation every singler getter. 
Most of the user rights can simply be annotation to the methods (RolesAllowed).
For some user rights you may have to do the check within the method instead of using an annotation.

- [x] ContactFacadeEjb
- [x] VisitFacadeEjb
- [x] BAGExportFacadeEjb-> just BAG_EXPORT for the facade
- [x] CaseImportFacadeEjb -> just CASE_IMPORT for the facade
- [x] ExternalJournalFacadeEjb -> just MANAGE_EXTERNAL_SYMPTOM_JOURNAL for the facade
- [x] PrescriptionFacadeEjb
- [x] TreatmentFacadeEjb
- [x] SurveillanceReportFacadeEjb -> use CASE_... rights
- [x] ClinicalVisitFacadeEjb
- [x] Check usage of these facades in the StartupShutdownService and CronService and make sure called facade methods are annotated with RolesAllowed including UserRight._SYSTEM.
- [x] Introduce an ArchUnit test that makes sure that all methods of the Facade interface implemented by the FacadeEjb either are annotated with RolesAllowed or start with "get, count, is, does, has, validate. Execute this test for CaseFacade, ContactFacade, VisitFacade and the other facades mentioned here.

leventegal-she:
@MartinWahnschaffe 

CONTACT_ASSIGN, CONTACT_CLASSIFY are not used, **do we delete them?**

CONTACT_CONVERT is handled in UI code(in CaseController::createFromContact) using a simple case save and changing also other possible contacts and event participants in different transactions, so checking the right in case creation  or contact save could still introduce  corrupt data so it needs to be refactored and moved to the backend so there could use @RolesAllowed to authorise this action, or we could just keep it as a UI only right.

**Refactoring is out of scope for this ticket. A new ticket is should be created if needed.**

CLINICAL_COURSE rights seem to be redundant, could be a combination of TREATMENT and PRESCRIPTION rights.
[...]

PR #8857, opened 2022-04-19
#8807 allow relevant method calls with EXTERNAL_VISITS right

### LINKED ISSUE #8807: Climedo doesn't sync correctly (EJBAccessException)
### Bug Description

There seems to be a user rights or authentication problem when trying to sync climedo with SORMAS. 
The error in logs says: "Client not authorized for this invocation"

### Steps to Reproduce

Register SMS as well as by e-mail, enter the symptoms and submit. 
No error message, but the system does not show the entry, neither under "Follow-up" nor in the "Follow-up view".

Doesn't work with "ExternalVisitDto #personUUID: XXGMVF-AF3BDN-6D2DPY-ECW4KG2I" on https://test1402.sormas.netzlink.com/

### System Details

* Device: 
* SORMAS version: 1.70.2
* Android version/Browser: 
* Server URL: 1402
* User Role:

### Additional Information

Log excerpt [Climedo_User_Not_Authorized.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/8479129/Climedo_User_Not_Authorized.txt)

```bash
Caused by: javax.ejb.EJBAccessException: Client not authorized for this invocation
	at com.sun.ejb.containers.BaseContainer.mapLocal3xException(BaseContainer.java:2397)
	at com.sun.ejb.containers.BaseContainer.postInvoke(BaseContainer.java:2183)
	at com.sun.ejb.containers.BaseContainer.postInvoke(BaseContainer.java:2104)
	at com.sun.ejb.containers.EJBLocalObjectInvocationHandler.invoke(EJBLocalObjectInvocationHandler.java:220)
	at com.sun.ejb.containers.EJBLocalObjectInvocationHandlerDelegate.invoke(EJBLocalObjectInvocationHandlerDelegate.java:90)
	at com.sun.proxy.$Proxy665.toDto(Unknown Source)
	at de.symeda.sormas.backend.caze.__EJB31_Generated__CaseFacadeEjb$CaseFacadeEjbLocal__Intf____Bean__.toDto(Unknown Source)
	at de.symeda.sormas.backend.visit.VisitFacadeEjb.onVisitChanged(VisitFacadeEjb.java:622)
```

FredrikSchaeferVitagroup:
Fixes #8807

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -2610,7 +2610,9 @@ public class CaseFacadeEjb extends AbstractCoreFacadeEjb<Case, CaseDataDto, Case
 		return dto;
 	}
 
-	@RolesAllowed(UserRight._EXTERNAL_VISITS)
+	@RolesAllowed({
+		UserRight._CASE_VIEW,
+		UserRight._EXTERNAL_VISITS })
 	public CaseDataDto toDto(Case source) {
 		return toCaseDto(source);
 	}
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -1378,7 +1378,8 @@ public class CaseFacadeEjb extends AbstractCoreFacadeEjb<Case, CaseDataDto, Case
 	@Override
 	@RolesAllowed({
 		UserRight._CASE_CREATE,
-		UserRight._CASE_EDIT })
+		UserRight._CASE_EDIT,
+		UserRight._EXTERNAL_VISITS })
 	public CaseDataDto save(@Valid @NotNull CaseDataDto dto, boolean systemSave) throws ValidationRuntimeException {
 		return save(dto, true, true, true, systemSave);
 	}
@@ -2609,6 +2610,7 @@ public class CaseFacadeEjb extends AbstractCoreFacadeEjb<Case, CaseDataDto, Case
 		return dto;
 	}
 
+	@RolesAllowed(UserRight._EXTERNAL_VISITS)
 	public CaseDataDto toDto(Case source) {
 		return toCaseDto(source);
 	}
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
@@ -1606,7 +1606,9 @@ public class ContactFacadeEjb
 		return source.toReference();
 	}
 
-	@RolesAllowed(UserRight._EXTERNAL_VISITS)
+	@RolesAllowed({
+		UserRight._CONTACT_VIEW,
+		UserRight._EXTERNAL_VISITS })
 	public ContactDto toDto(Contact source) {
 		return toContactDto(source);
 	}
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
@@ -1606,6 +1606,7 @@ public class ContactFacadeEjb
 		return source.toReference();
 	}
 
+	@RolesAllowed(UserRight._EXTERNAL_VISITS)
 	public ContactDto toDto(Contact source) {
 		return toContactDto(source);
 	}


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2609,6 +2610,7 @@ public CaseReferenceDto convertToReferenceDto(Case source) {
 		return dto;
 	}
 
+	@RolesAllowed(UserRight._EXTERNAL_VISITS)
```
StefanKock, line 2613: This only allows `UserRight._EXTERNAL_VISITS` to use this method.

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -1606,6 +1606,7 @@ public static ContactReferenceDto toReferenceDto(Contact source) {
 		return source.toReference();
 	}
 
+	@RolesAllowed(UserRight._EXTERNAL_VISITS)
```
StefanKock, line 1609: This only allows `UserRight._EXTERNAL_VISITS` to use this method.

StefanKock (changes requested):
I suppose the permissions from the class need to be added for `toDto` methods to not break "regular" usages.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
FredrikSchaeferVitagroup: It should, but it doesn't seem to. I am about to add a comment to the PR listing some of my own open questions. Maybe we can then use this as a basis for discussion.

FredrikSchaeferVitagroup:
The current suggested changes seem to solve the problem. However, there are some things still not clear to me.

1. (Why) is it necessary to add the RolesAllowed-annotation to the toDto of the Case- and ContactFacadeEjb? It seems that other roles can access this method without having a right annotated here.
2. Should we make sure that the External_Visits-right is not forgotten when the RolesAllowed-annotation is introduced to the VisitFacadeEjb and (maybe) the FeatureConfigurationFacadeEjb? Methods there can be triggered by saving an external visit.

StefanKock (approved):
Remarked findings were solved

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1378,7 +1378,8 @@ public CaseDataDto save(@Valid @NotNull CaseDataDto dto) throws ValidationRuntim
 	@Override
 	@RolesAllowed({
 		UserRight._CASE_CREATE,
-		UserRight._CASE_EDIT })
+		UserRight._CASE_EDIT,
+		UserRight._EXTERNAL_VISITS })
```
leventegal-she, line 1382: I would introduce a new method for handling case updates by external visits and should not allow save for external visit right

leventegal-she:
> 1. (Why) is it necessary to add the RolesAllowed-annotation to the toDto of the Case- and ContactFacadeEjb? It seems that other roles can access this method without having a right annotated here.

I would suggest to use the static toDto methods like `toCaseDto` and `toContactDto` and keep the toDto methods not annotated

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
StefanKock: Somehow the method is already for that purpose that @leventegal-she suggested: It's only used by `VisitFacade.saveVisit` but from a variety of origins.

grafik

leventegal-she: But then why the `save` should be allowed for UserRight._EXTERNAL_VISITS?

StefanKock: Looking again on this I think we should maybe exclude the `CaseFacadeEjb.save(CaseDataDto, boolean)` from the `CaseFacade` and change it to private and remove the permissions since it's only internally called by CaseFacadeEjb itself (at least on my mostly updated `development` branch).

StefanKock (changes requested):
We need to discuss the changes and approach.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
StefanKock: Discussed with @FredrikSchaeferVitagroup: The state on 1.70.x is different: saveVisit calls directly `CaseFacade.save(caseDataDto, boolean)` and therefore needs that annotation. In #8508 the new method `CaseFacade.updateSymptomsByVisit` was introduced and makes annotations on `CaseFacade.save(caseDataDto, boolean)` superfluous (the usage of that method could be reduced in that ticket or afterwards).

StefanKock (approved):

PR #8885, opened 2022-04-20
SORQA-161 fix SORDEV-8268

pk-sgent:
fix for "inaktiv". I attach failing jenkins job log because of issue with "Special characters are displayed as question mark " but fix works

image

### DIFF
--- a/sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/entries/CreateNewTravelEntrySteps.java
+++ b/sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/entries/CreateNewTravelEntrySteps.java
@@ -239,7 +239,7 @@ public class CreateNewTravelEntrySteps implements En {
               "Communities are not equal");
           softly.assertEquals(
               aCase.getPointOfEntry(),
-              travelEntry.getPointOfEntry() + " (Inaktiv)",
+              travelEntry.getPointOfEntry(),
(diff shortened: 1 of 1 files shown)


sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/entries/CreateNewTravelEntrySteps.java
```diff
@@ -239,7 +239,7 @@ public CreateNewTravelEntrySteps(
               "Communities are not equal");
           softly.assertEquals(
               aCase.getPointOfEntry(),
-              travelEntry.getPointOfEntry() + " (Inaktiv)",
+              travelEntry.getPointOfEntry(),
```
jp-sgent, line 242: Won't this fail when the point of entry **does** show up as inactive?
Alwin said that "`Within this test both should be acceptable`" so I think we need the assert to alternate between active and inactive version instead of expecting only one of them.

pk-sgent: "The presents of "inactive" is related to the status (active/archived) of the related destrict record. Within this test both should be acceptable. We will test the "inactive" feature separately in [SORQA-199]" - in this test case it will always 'aktiv' which is also correct.

jp-sgent (approved):

Issue #8841, opened 2022-04-15
[User rights] An External Lab Officer cannot open samples


adinaflorea9:
### Bug Description

An External Lab Officer cannot open and view samples.

### Steps to Reproduce

1. Log in with an External Lab Officer user;
2. Navigate to the 'Samples' directory and open an existing sample.

### Actual Behavior

The Sample page displayed is blank.

### Expected Behavior

The Sample page is displayed and it contains the sample's information, fields, cards.

### Screenshots

External Lab Officer's view of the 'Samples' directory:
![image](https://user-images.githubusercontent.com/96909231/163578617-4743ab5b-c313-4805-bb27-840f6ba6da8f.png)

External Lab Officer's view of a sample:
![image](https://user-images.githubusercontent.com/96909231/163578670-e1ffb9da-3dd7-49fa-861f-c6ca6cf927f9.png)

### System Details

* Device: Windows 10
* SORMAS version: 1.71.0-SNAPSHOT(f974e6f)
* Android version/Browser: Chrome
* Server URL: local machine
* User Role: External Lab Officer

### Additional Information

An External Lab Officer has rights for SAMPLE_VIEW and SAMPLE_EDIT.

roxanamlendea:
I tested the ticket on my local machine and I'm reopening the ticket because the user can now open the Sample, but he cannot Edit it, even though it has SAMPLE_EDIT rights.

![image](https://user-images.githubusercontent.com/89522688/164254955-81df4911-85bf-424a-8908-9c6d2776a8b8.png)

abrudanancuta:
there is already a ticket for this case: https://github.com/hzi-braunschweig/SORMAS-Project/issues/8892, I would suggest to close this issue

vidi42:
The sample editing is not blocked by the UserRight, but by the jurisdiction of the user. So I will close this ticket now

roxanamlendea:
Verified ticket on local environment using the latest version of Sormas 1.71.0-SNAPSHOT (d16a724) from the development branch.
Now all users with SAMPLE_VIEW rights can open Samples.

PR #8894, opened 2022-04-20
Bugfix/bug 8797 fix sample creation

### LINKED ISSUE #8797: Only NationalUser and Surveillance Supervisor can create samples
### Bug Description
Only these roles can create samples for a case: NationalUser and Surveillance Supervisor. The roles that should be allowed to created samples for a case should be: NATIONAL_CLINICIAN, ADMIN_SUPERVISOR, SURVEILLANCE_OFFICER, CASE_SUPERVISOR, CASE_OFFICER, EVENT_OFFICER, HOSPITAL_INFORMANT, COMMUNITY_INFORMANT, LAB_USER, COMMUNITY_OFFICER, plus the 2 roles NationalUser and Surveillance Supervisor that are already working

### Steps to Reproduce
1. As one of the roles stated above which should have right to create sample Login
2. Create a case
3. Add to the case created a sample and save it
4. Edit a case
5. Look for the button to add a sample

### Actual Behavior
At step 3 after saving the sample an error message appears and the case has no longer the button to create samples
At step 5 no button to create a sample is in the form

### Expected Behavior
The roles stated above should be allowed to create samples

### System Details

* Device:
* SORMAS version: 1.71.0-SNAPSHOT (52ca13d)
* Android version/Browser:
* Server URL: local env, test.sormas.netzlink.com
* User Role:

carina29:
Fixes #8797


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1375,6 +1375,9 @@ public void updateCompleteness(String caseUuid) {
 	}
 
 	@Override
+	@RolesAllowed({
```
vidi42, line 1378: Why do we need the `CASE_EDIT` right for getting the data from the DB? 
`SAMPLE_VIEW` also should not be needed here

sormas-backend/src/main/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjb.java
```diff
@@ -245,12 +245,13 @@ public void bulkAssignLabMessages(List<String> uuids, UserReferenceDto userRef)
 	}
 
 	@Override
+    @RolesAllowed({
```
vidi42, line 248: I think this method should be restricted to users with the `LAB_MESSAGES` right. Probably adding these roles here and opening it more it's not the right solution

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -308,6 +308,7 @@ public Long getPersonIdByUuid(String uuid) {
 	@Override
 	@RolesAllowed({
 		UserRight._PERSON_VIEW,
+		UserRight._SAMPLE_VIEW,
```
vidi42, line 311: `SAMPLE_VIEW` right should not enable someone to get person details.

vidi42 (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
carina29: CASE_EDIT is an error, locally I had CASE_VIEW annotation (which is also on the entire class but needs to be added again if the method has another annotation) and SAMPLE_VIEW because in case of logging in as an EXTERNAL LAB OFFICER the samples were not visible and in the log messsages appeared that the Client was unauthorized for getCaseDataByUuid method from CaseFacadeEjb and getPersonByUuid method from PersonFacadeEjb to view the samples.

Since yesterday https://github.com/hzi-braunschweig/SORMAS-Project/issues/8841 was fixed, the problem was solved ( I will delete the annotations related to this problem).

sormas-backend/src/main/java/de/symeda/sormas/backend/labmessage/LabMessageFacadeEjb.java
carina29: The bug is not reproducible anymore. The already fixed bugs solved the Sample creation bug too.

carina29:
This bug can not be reproduced anymore.

PR #8905, opened 2022-04-20
#8893-Lab messages adapter should be displayed to the users with LAB_…

### LINKED ISSUE #8893: An error appears when accessing the About page
### Bug Description
An error appears when accessing the About page 

### Steps to Reproduce
1. Have configured the property: Lab_messages to true from the table featureconfiguration
2. Login with an user and access the About page
3. Look in the server logs

### Expected Behavior
No error should appear in the server logs

### System Details

* Device:
* SORMAS version:
* Android version/Browser:
* Server URL:
* User Role:

### Additional Information

[AboutError.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/8520922/AboutError.txt)

carina29:
…MESSAGES right

Fixes #8893

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/AboutView.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/AboutView.java
@@ -143,12 +143,10 @@ public class AboutView extends VerticalLayout implements View {
 		}
 
 		if (FacadeProvider.getFeatureConfigurationFacade().isFeatureEnabled(FeatureType.LAB_MESSAGES)) {
-			if (UserProvider.getCurrent().hasAllUserRights(UserRight.LAB_MESSAGES)) {
-				addExternalServiceVersion(
-					Captions.aboutLabMessageAdapter,
-					() -> FacadeProvider.getLabMessageFacade().getLabMessagesAdapterVersion(),
-					infoLayout);
-			}
+			addExternalServiceVersion(
+				Captions.aboutLabMessageAdapter,
+				() -> FacadeProvider.getLabMessageFacade().getLabMessagesAdapterVersion(),
+				infoLayout);
 		}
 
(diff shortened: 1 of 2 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/AboutView.java
```diff
@@ -135,22 +135,20 @@ private VerticalLayout createInfoSection() {
...
 			CssStyles.style(commitLink, CssStyles.VSPACE_3);
 			infoLayout.replaceComponent(versionLabel, commitLink);
 		}
 
 		if (FacadeProvider.getFeatureConfigurationFacade().isFeatureEnabled(FeatureType.LAB_MESSAGES)) {
-			addExternalServiceVersion(
-				Captions.aboutLabMessageAdapter,
-				() -> FacadeProvider.getLabMessageFacade().getLabMessagesAdapterVersion(),
-				infoLayout);
+			if (UserProvider.getCurrent().hasAllUserRights(UserRight.LAB_MESSAGES)) {
```
vidi42, line 146: this condition can be added in the outside `if` with `&&` instead of a nested `if`

vidi42 (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/AboutView.java
vidi42: On a second thought it probably makes sense to allow all user rights to view the version of the labmessageadapter. So I suggest removing the user right condition check and add the `@PermitAll` to `LabMessageFacadeEjb::getLabMessagesAdapterVersion`

vidi42 (approved):
Looks good. Can be merged once the CI build passes

PR #8914, opened 2022-04-21
Feature 8747 joins reuse

### LINKED ISSUE #8747: Reuse joins for createUserFilter [5]
### Problem Description

Sibling to #8688. Invoking `createUserFilter` from `Task` or `Sample` domain on for example `Contact` creates a lot of redundant joins that cause performance problems.

### Proposed Change

1. Reduce `AdoServiceWithUserFilter.createUserFilter(CriteriaBuilder, CriteriaQuery, From)` to protected and only use in in the same instance where no outside domain context is relevant -> (text replace `protected abstract Predicate createUserFilter(` to lower the visibility in all subclasses).
2. Alter `AbstractCoreAdoService` to provide by default a `createUserFilter(QueryContext)` method, change all subclasses to adapt to this pattern and always use `QueryContext` as starting point to make joins reusable.

### Additional Information

Code snippets from my PoC:
```java

public abstract class AbstractCoreAdoService<ADO extends CoreAdo, J extends QueryJoins<ADO>, Q extends QueryContext<ADO, J>>
	extends AbstractDeletableAdoService<ADO> {

	public abstract Predicate createUserFilter(Q queryContext);

// ..

public class ContactService extends AbstractCoreAdoService<Contact, ContactJoins, ContactQueryContext> {
```

BarnaBartha:
Fixes #8747, blocked as draft for now to not merge it into 1.71.0


StefanKock (changes requested):
The problematic `createUserFilter(CriteriaBuilder, CriteriaQuery, From)` is still used by dozens of undesired invocations, but now also logs a WARN message when it's used. I think the usage needs still to be used drastically and that this WARN message is not acceptable.

StefanKock (approved):
I continued to replace createUserFilter invocations with QueryContext and removed the WARN logging

Issue #8851, opened 2022-04-18
Deleting an Event Group results in endless loading of the page and the server connection is lost


roxanamlendea:
### Bug Description
When the user deletes an Event Group, the action results in endless loading of the page. Even if the user reloads the page, he cannot access the app. If the user reloads the page, the server connection appears lost and the server is trying to reconnect;

~Logs for this issue from my local env~ (deleted, were not related)

### Steps to Reproduce

1. Create an Event and an EventGroup from it.
2. Open the Events Directory and select the 'Groups' tab;
3. Open an Event Group and click on 'Delete';
-> The 'Confirm Deletion' pop-up appears;
4. Click on 'Yes';

### Actual Behavior
The Event Group is not deleted and the page is endlessly loading;

### Expected Behavior
The Event Group is deleted in a reasonable amount of time;

### Screenshots
![image](https://user-images.githubusercontent.com/89522688/163821591-680e79da-969f-42d0-a101-c86e8065c0a2.png)

![image](https://user-images.githubusercontent.com/89522688/163822758-a235f9b3-1322-405e-9dde-e09a3d1ee9bc.png)

### System Details

* Device: Windows 10
* SORMAS version: 1.70.2 and Sormas 1.71.0-SNAPSHOT (s2s-testbed)
* Android version/Browser: Chrome
* Server URL: local env and test1401.sormas-oegd.de
* User Role: Admin NatUser

### Additional Information

- The log seemed to show some successful calls after clicking the delete button, but did not log any exception. Payara and Postgres did not show high CPU or RAM load, it's currently not clear what's happening.
- Cause: #7857

MateStrysewske:
The exception linked in the initial issue is likely not the cause of this problem; it's closely linked to #8847 though and will be taken care of there. This problem here is probably caused by something else.

StefanKock:
Updated the ticket description: Problem still reproducable, but without any StackTrace related to the problem. It seems that the session is idling, no heavy resource load, other session can be started and work with the system.

MateStrysewske:
Moved from the hotfix in favour of #8915 because there's no easy fix for this problem.

StefanKock:
In conjunction with #7931 @MateStrysewske and I found out that `deletePermanent(List<String> uuids)` causes `deletePermanent(ADO ado)` to also start a new JTA tx. We are going to prevent this by not using the same method name for the `List` use case.

roxanamlendea:
[...]

PR #8916, opened 2022-04-21
#8915 - Hide delete button

### LINKED ISSUE #8915: Temporarily disable the deletion of event groups
### Problem Description

Deleting event groups currently leads to the user session idling and the database transaction not executing (#8851). Unfortunately, this is not easily fixable. In order to not let users run into this problem, we need to temporarily disable the deletion of event groups until we have found a proper fix for the issue.

### Proposed Change

Hide the "Delete" button from the event group form.

MateStrysewske:
Fixes #8915

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventGroupController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventGroupController.java
@@ -29,7 +29,6 @@ import com.vaadin.ui.Button;
 import com.vaadin.ui.Label;
 import com.vaadin.ui.Notification;
 import com.vaadin.ui.Notification.Type;
-import com.vaadin.ui.UI;
 import com.vaadin.ui.themes.ValoTheme;
 
 import de.symeda.sormas.api.FacadeProvider;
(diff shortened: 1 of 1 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventGroupController.java
```diff
@@ -203,19 +202,21 @@ public CommitDiscardWrapperComponent<?> getEventGroupEditComponent(String uuid)
 			});
 		}
 
-		if (user.hasUserRight(UserRight.EVENTGROUP_DELETE) && hasRegion) {
-			editView.addDeleteListener(() -> {
-				deleteEventGroup(eventGroup);
-				UI.getCurrent().getNavigator().navigateTo(EventsView.VIEW_NAME);
-			}, I18nProperties.getString(Strings.entityEventGroup));
-		}
+		// TODO Temporarily deleted in #8915, enable again in #8851
```
StefanKock, line 205: Write this marker like this:
"TODO #8851: Enable temporarily deleted button when EventGroup deletion works"

StefanKock (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventGroupController.java
```diff
@@ -203,19 +202,21 @@ public CommitDiscardWrapperComponent<?> getEventGroupEditComponent(String uuid)
...
+//			}, I18nProperties.getString(Strings.entityEventGroup));
+//		}
 
 		// Initialize 'Archive' button
 		if (user.hasUserRight(UserRight.EVENTGROUP_ARCHIVE) && hasRegion) {
 			boolean archived = FacadeProvider.getEventGroupFacade().isArchived(uuid);
-			Button archiveEventButton = ButtonHelper.createButton(archived ? Captions.actionDearchiveInfrastructure : Captions.actionArchiveInfrastructure, e -> {
-				archiveOrDearchiveEventGroup(uuid, !archived);
-			}, ValoTheme.BUTTON_LINK);
+			Button archiveEventButton =
```
StefanKock, line 216: Better not format something else: When you keep it clean, a revert action can work.

MateStrysewske: This has been done by automatic formatting; manually re-formatting it in the wrong way doesn't seem like a good idea to me.

StefanKock: Just disable the automatic formatting for this change to not change something unintended. That's the problem when not everyone formats the code the same way :-(

StefanKock (approved):
Please only one change commit when merging this: amend, fixup or squash

PR #9007, opened 2022-04-29
#8704 rewrite case tests results count

### LINKED ISSUE #8704: Improve slow query for pathogen test result counts in dashboard [2]
### Bug Description
In the presence of a large database, access to the dashboard is very problematic or impossible. 
In the production environment, as soon as a user calls the dashboard, the request(s) are launched as expected but run so long that they cause a time out and an interruption of the server connection. It is therefore mandatory to restart the server. To clarify, the map is disabled by default and only a user action (clicking on a button on the map) displays the points on the map. 
On the test environment with a low volume, no access problem. 
On the test environment hosting the dump of the production database, the display of the dashboard was completed after 26 seconds (precision: only one user on this server)

### Steps to Reproduce

1. Use the provided dump (ask @MartinWahnschaffe)
2. Set PerformanceLoggingInterceptor to TRACE
3. Open the reported function: Log in and open the dashboard
4. Use PerformanceLogAnalysisGenerator to analyse the logfile
5. Deliverable: Point out hotspots where a lot of time is lost

### System Details

* Device:
* SORMAS version: 1.67
* Server URL:
* User Role:admin+national user OR contact supervisor + surveillance supervisor

### Additional Information
@OlivierLeuci @MartinWahnschaffe @MateStrysewske @StefanKock @bernardsilenou

StefanKock:
Fixes #8704, replaces #8973.

Further improvements:

1. Fresh start from `development`, picked needed (and currently missing) method signature ` CaseService.createUserFilter(CaseQueryContext)`.
2. Removed some superfluous change noise on removed signatures.
3. Code comments within the new method `DashboardService.getNewTestResultCountByResultType`, some refactoring for better readable code.
4. Moved the DTO class out of the EJB (in favor for separate java file instead of inner class to have less cluttered EJB class code).
5. Documented SQL queries before/after here: https://github.com/hzi-braunschweig/SORMAS-Project/issues/8704#issuecomment-1112990750

I think we still need to move/integrate `SampleFacadeEjbTest.testGetNewTestResultCountByResultType` into `DashboardFacadeEjbTest`.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/dashboard/DashboardFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/dashboard/DashboardFacade.java
@@ -25,8 +25,6 @@ public interface DashboardFacade {
 
 	String getLastReportedDistrictName(DashboardCriteria dashboardCriteria);
 
-	Map<PathogenTestResultType, Long> getTestResultCountByResultType(List<DashboardCaseDto> cases);
-
 	long countCasesConvertedFromContacts(DashboardCriteria dashboardCriteria);
 
 	Map<PresentCondition, Integer> getCasesCountPerPersonCondition(DashboardCriteria dashboardCriteria);
--- a/sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleFacade.java
@@ -73,8 +73,6 @@ public interface SampleFacade {
 
 	boolean isDeleted(String sampleUuid);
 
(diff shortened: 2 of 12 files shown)


StefanKock:
Compared the SQL queries: This WHERE condition is missing in the new query: `(symptoms4_.onsetDate BETWEEN ? AND ? OR (symptoms4_.onsetDate IS NULL)`

BarnaBartha:
The raised concern regarding the SQL query difference is not valid, the extra conditions come from specifying a NewCaseDateType filter.

BarnaBartha (approved):

PR #9013, opened 2022-04-29
SORQA-210fix

pk-sgent:


pk-sgent:
I removed waiting for element (notification) because sometimes Selenium makes some magic and moving screen so notification disappears. Instead this i put sleep, not good solution but more stable

image

PR #9060, opened 2022-05-03
bug/8992_surv_officer_cannot_generate_case

### LINKED ISSUE #8992: A Surveillance Officer can not generate a new case with a different disease
### Bug Description

A Surveillance Officer can not generate a new case of a different disease if a sample has a pathogen test result of one disease (Covid-19), and it is associated with a case (Covid-19).

The behavior is also reproducible is the user adds a whole new sample with a positive pathogen test result of a different disease (Anthrax) than the disease of the case (Covid-19).

### Steps to Reproduce

1. Log in with a Surveillance Officer;
2. Create a new case (Covid-19) and add a sample with a positive pathogen test result to it (Covid-19);
3. Navigate to the 'Samples' directory and open the sample from step 2;
4. Add a new positive pathogen test result to the sample (use a different disease than Covid-19);
=> Pop-up to generate a new case for the different disease is triggered;
5. Click 'Yes' in the pop-up.

### Actual Behavior

An error message appears: 'A problem has occurred. You have no rights for changing Disease. Please reload the page to see the latest changes'.

The new pathogen test is added to the sample; the new case is not generated.

### Expected Behavior

The user should be able to generate a new case for the different disease of the second pathogen test.

### Screenshots

image

### System Details

* Device: Windows 10
* SORMAS version: 1.71.0-SNAPSHOT(af0201e)
* Server URL: local machine

### Additional Information

A Surveillance officer user does not have rights for CASE_CHANGE_DISEASE; but in the steps described above, the user would not be changing the disease of an existing case, or the disease of a sample or pathogen test.

If the same steps are reproduced but for a sample that is associated with a contact or event participant, the Surveillance Officer can add the new pathogen test of a different disease and generate a new case for that disease successfully.

- This defect was apperently introduced by #445 with backend check of `UserRight.CASE_CHANGE_DISEASE`.
- Follow up for cleaner code: #9066

dinua:
Fixes #8992

### DIFF
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/caze/CaseFacadeEjbTest.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/caze/CaseFacadeEjbTest.java
@@ -50,6 +50,7 @@ import javax.validation.Validation;
 import javax.validation.Validator;
 import javax.validation.ValidatorFactory;
 
+import de.symeda.sormas.api.disease.DiseaseVariant;
 import org.apache.commons.lang3.time.DateUtils;
 import org.hamcrest.MatcherAssert;
 import org.hibernate.internal.SessionImpl;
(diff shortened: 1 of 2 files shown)


StefanKock (changes requested):
I think this is a dirty workaround to manipulate the existing case data before cloning. To have a cleaner approach one could alter the method signature to directly provide the new disease in the clone method and change other details afterwards.

StefanKock:
I agreed with @dinua that:

1. his changes are ok-ish to fix the problem at hand, but a unit test should be provided that shows that the old case was not changed.
2. I will create a ticket to refactor the approach to use the `existingCaseDto` as data container for the data to change

StefanKock (approved):

PR #9062, opened 2022-05-03
edit/view rights in mobile app synchronization #8989

### LINKED ISSUE #8295: Automatic permanent deletion for manually deleted entities [3]

### LINKED ISSUE #8989: Sync to android-app only for allowed data endpoints
### Bug Description

On the mobile app, when logged in with a Surveillance Officer or a Community Officer, the synchronization with the APMIS server does not work. An error message is displayed stating that the user does not have permission 'You do not have needed permissions. Please contact the support'.

### Steps to Reproduce

1. Log in with a Community Informant or a Surveillance Officer
2. Trigger the synchronization with the APMIS server.

### Actual Behavior

An error message appears: 'You do not have needed permissions. Please contact the support'.

### Expected Behavior

No error message should appear; the synchronization should be successful.

### Screenshots

MicrosoftTeams-image (5)

### System Details

* Device: Huawei P20 lite
* SORMAS version: 1.71.0
* Android version/Browser: 9
* Server URL: test.sormas.netzlink.com
* User Role: Community Officer, Surveillance Officer

### Additional Information

This has been reproducible on the mobile app downloaded from:
APK download: https://test.sormas.netzlink.com/downloads/release/
Server URL: https://test.sormas.netzlink.com/sormas-rest/
Username & Password: AD-CommOff / BpDH859yu9Bf and AD-SurvOff / YjziLyHtWCFr

Caused by #8508

syntakker:
Fixes #8989

The simple solution from comment https://github.com/hzi-braunschweig/SORMAS-Project/issues/8989#issuecomment-1112097217 was not workable (lack of separation between real 403 exceptions and intendedly catched ones, unclear interference with the overall synchronization mechanism), so I implemented the second option:
> I'd propose to do the following now:
> 
>     * [ ]  catch this exception for every entity type (can probably be put into a generic method) and ignore it for now. Instead go on with the next entity type
> 
> 
> Maybe in a separate issue that can go into the next sprint:
> 
>     1. Don't pull entity types the user cannot VIEW
> 
>     2. Don't push entity types the user cannot EDIT

Some Facades can be accessed with different UserRights, e.g., `VisitFacadeEjb` with `CONTACT_VIEW` and `CASE_VIEW`, but every UserRole having `CONTACT_VIEW` has also `CASE_VIEW`, so the separation can be avoided here (see comment in `DtoViewAndEditRights` l.90)
This should however refined later, such that the maps in `DtoViewAndEditRights` return lists of possible UserRights instead of only one.

Fixes #8295

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/InfrastructureChangeDatesDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/InfrastructureChangeDatesDto.java
@@ -21,8 +21,6 @@ public class InfrastructureChangeDatesDto implements Serializable {
 	private Date diseaseConfigurationChangeDate;
 	private Date userRoleConfigurationChangeDate;
 	private Date featureConfigurationChangeDate;
-	private Date campaignChangeDate;
-	private Date campaignFormMetaChangeDate;
 
 	public Date getContinentChangeDate() {
 		return continentChangeDate;
@@ -134,21 +132,5 @@ public class InfrastructureChangeDatesDto implements Serializable {
 
 	public void setFeatureConfigurationChangeDate(Date featureConfigurationChangeDate) {
 		this.featureConfigurationChangeDate = featureConfigurationChangeDate;
-	}
-
-	public Date getCampaignChangeDate() {
-		return campaignChangeDate;
-	}
-
-	public void setCampaignChangeDate(Date campaignChangeDate) {
-		this.campaignChangeDate = campaignChangeDate;
-	}
-
-	public Date getCampaignFormMetaChangeDate() {
-		return campaignFormMetaChangeDate;
-	}
-
-	public void setCampaignFormMetaChangeDate(Date campaignFormMetaChangeDate) {
-		this.campaignFormMetaChangeDate = campaignFormMetaChangeDate;
 	}
 }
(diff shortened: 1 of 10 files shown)


leventegal-she:
Couldn't fully test the changed sync of campaign and campaign form meta because of https://github.com/hzi-braunschweig/SORMAS-Project/issues/8989

MartinWahnschaffe:
I have moved the task to check whether campaigns are working to the related bug ticket:
https://github.com/hzi-braunschweig/SORMAS-Project/issues/9076#issuecomment-1118235087

@sergiupacurariu @leventegal-she 
So you can go on to review and merge this.

PR #9063, opened 2022-05-04
9044 sormas returns 403

### LINKED ISSUE #9044: SORMAS returns 403 for requests to endpoints of the ExternalVisitsResource
### Bug Description

This bug breaks external journal functionalities.

### Steps to Reproduce

1. Make any valid call to a related endpoint, e.g. `curl --location --request GET '[server-url]/sormas-rest/visits-external/person/R2I5VY-WQD7GO-XBCJL6-YQ4QCIGI' --header 'Content-Type: application/json; charset=UTF-8' --header 'Authorization: Basic [valid credentials]' --data-raw '`'. Authenticate with a user having the default `REST_EXTERNAL_VISITS_USER` role.
-> 403.

### Expected Behavior

Valid response (e.g. no content when a person with the specified UUID does not exist on the system).

### System Details

* SORMAS version: 1.70.3 and 1.71.0-SNAPSHOT
* Server URL: test and prod systems (it seems that Keycloak authentication needs to be in place). E.g. https://ga1.sormas-oegd.de/.
* User Role: Default `REST_EXTERNAL_VISITS_USER`.

### Additional Information

The bug seems to often appear and disappear when the server gets restarted.

vidi42:
Fixes #9044

Removed `RollesAllowedDynamicFeature`  from glassfish registered 'manually' as it sometimes conflicts with what payara provides with microprofile and causes the 403 result from Keycloak.

Not sure if this is this fixes the issue for good so it needs more testing

### DIFF
--- a/sormas-rest/src/main/java/de/symeda/sormas/rest/RestConfig.java
+++ b/sormas-rest/src/main/java/de/symeda/sormas/rest/RestConfig.java
@@ -60,7 +60,6 @@ public class RestConfig extends ResourceConfig {
 		// Resources.
 		packages(getClass().getPackage().getName());
 
-		// as described in https://jersey.github.io/documentation/latest/security.html
 		register(JacksonFeature.class);
 
 		SwaggerConfig.init();
--- a/sormas-rest/src/main/java/de/symeda/sormas/rest/RestConfig.java
+++ b/sormas-rest/src/main/java/de/symeda/sormas/rest/RestConfig.java
@@ -40,7 +40,6 @@ import io.swagger.v3.oas.models.info.License;
 import org.apache.commons.collections4.SetUtils;
 import org.glassfish.jersey.jackson.JacksonFeature;
 import org.glassfish.jersey.server.ResourceConfig;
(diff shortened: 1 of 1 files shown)


vidi42:
Tried to reproduce this issue again with [S2S-Testbed](https://github.com/hzi-braunschweig/S2S-Testbed) and wasn't able to reproduce it after 5-10 container restarts for sormas_a (with keycloak) and sormas_b (without keycloak).
Before the fix, this was easily reproducible after 2-3 container restarts.

FredrikSchaeferVitagroup:
> Tried to reproduce this issue again with [S2S-Testbed](https://github.com/hzi-braunschweig/S2S-Testbed) and wasn't able to reproduce it after 5-10 container restarts for sormas_a (with keycloak) and sormas_b (without keycloak). Before the fix, this was easily reproducible after 2-3 container restarts.

It's probably necessary to not just check whether the problem is gone now, but also check whether the authorization works apart from that... For example, do users without SORMAS-REST and EXTERNAL_VISITS user right get a 403 now when trying to access a REST endpoint now?

vidi42:
Was done as part of hotfix 1.70.4

PR #9065, opened 2022-05-04
Feature 8609 case export

### LINKED ISSUE #8609: Detailed Export for a large amount of cases does not work [5]
### Feature Description
As we are offering the detailed export of cases and do not have a limit for the number of cases that can be exported, the expectation should be fulfilled that the export works. 

### Problem Description
We have currently at least 2 health departments that complain that the detailed export of cases does not work. 
~~The query logs are available [here](https://wiki.vitagroup.ag/display/SORMAS/Performance+Taskforce+2022)~~ Just a stack trace given that the request takes too long, no insight what's really going wrong.

BarnaBartha:
Fixes #8609. **Only squash merge this** !!!


StefanKock:
On the resulting SQL I have the following results:

1. There are now superfluous methods to be deleted: `prescriptionService.getPrescriptionCountByCases`, `treatmentService.getTreatmentCountByCases`, `clinicalVisitService.getClinicalVisitCountByCases`
2. I think I get the idea with the `CaseExportMapperDto`: It's a workaround to load `Symptoms` and `HealthConditions` entities in the first query and not afterwards. I'm not sure if the new approach helps, since it's still JPA entities that are fetched and now we are missing the `setHint(ModelConstants.HINT_HIBERNATE_READ_ONLY, true)` on the main query (not sure if makes a difference).
3. The are some joins only for the id that can be avoided: ~~`symptoms11_`~~ (or not, see finding 2), `user21_`, `user22_` by this pattern: `joins.getRoot().get(Case.SYMPTOMS).get(Symptoms.ID)` (as in `CaseJurisdictionPredicateValidator` for comparisons)
4. These WHERE conditions reduce the resulting data with INNER JOINs, not sure if was intended.
```sql
  and event24_.deleted = false
  and eventparti23_.deleted = false
  and (event24_.id is not null)
```

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -738,10 +741,40 @@ public List<CaseExportDto> getExportList(
 				cb.isFalse(eventCountRoot.get(EventParticipant.DELETED))));
 		eventCountSq.select(cb.countDistinct(event.get(Event.ID)));
 
+		Subquery<Long> prescriptionCountSq = cq.subquery(Long.class);
+		Root<Prescription> prescriptionCountRoot = prescriptionCountSq.from(Prescription.class);
+		Join<Prescription, Therapy> prescriptionTherapyJoin = prescriptionCountRoot.join(Prescription.THERAPY, JoinType.LEFT);
+		prescriptionCountSq.where(cb.and(cb.equal(prescriptionTherapyJoin.get(Therapy.ID), caseRoot.get(Case.THERAPY).get(Therapy.ID))));
+		prescriptionCountSq.select(cb.countDistinct(prescriptionCountRoot.get(Prescription.ID)));
```
carina29, line 748: getPrescriptionCountByCases from PrescriptionService, getTreatmentCountByCases from TreatmentService, getClinicalVisitCountByCases from ClinicalVisitService should be removed since they will not be used anymore

carina29 (changes requested):

StefanKock:
>     2. I think I get the idea with the `CaseExportMapperDto`: It's a workaround to load `Symptoms` and `HealthConditions` entities in the first query and not afterwards. I'm not sure if the new approach helps, since it's still JPA entities that are fetched and now we are missing the `setHint(ModelConstants.HINT_HIBERNATE_READ_ONLY, true)` on the main query (not sure if makes a difference).
>     3. The are some joins only for the id that can be avoided: ~`symptoms11_`~ (or not, see finding 2), `user21_`, `user22_` by this pattern: `joins.getRoot().get(Case.SYMPTOMS).get(Symptoms.ID)` (as in `CaseJurisdictionPredicateValidator` for comparisons)

Don't fetch JPA entities in the initial query, but keep the follow-up fetch of Entities by ids. Replace joins on `symptoms11_`, `healthcond13_`, `user21_`, `user22_` -> @BarnaBartha 

>     4. These WHERE conditions reduce the resulting data with INNER JOINs, not sure if was intended.
> 
> 
> ```sql
>   and event24_.deleted = false
>   and eventparti23_.deleted = false
>   and (event24_.id is not null)
> ```

Let's check what conditions we have with less data restrictions in the unit tests. @StefanKock 

>     5. The following four conditions seem to be false, so they are not part of the resulting query (only `null` is staticly inserted): `exportHealthConditions`, `exportPrescriptionNumber`, `exportTreatmentNumber`, `exportClinicalVisitNumber`

Extend existing unit test to contain one run with all optional exports activated. @BarnaBartha (@StefanKock to provide usable coed snippet).

StefanKock:
Objection to current state:
`CaseExportMapperDto` is still existing without good reason, and sneakily tried to return over remote interface as its superclass. I That fails because class cannot be deserialized:
```
Caused by: javax.ejb.EJBException: java.rmi.MarshalException: CORBA MARSHAL 1330446393 Maybe; nested exception is: 
	org.omg.CORBA.MARSHAL: WARNUNG: 00810057: Could not load class de.symeda.sormas.backend.caze.CaseExportMapperDto  vmcid: OMG  minor code: 57 completed: Maybe
	at de.symeda.sormas.api.caze._CaseFacade_Wrapper.getExportList(de/symeda/sormas/api/caze/_CaseFacade_Wrapper.java)
	at de.symeda.sormas.ui.utils.CaseDownloadUtil.lambda$0(CaseDownloadUtil.java:51)
	at de.symeda.sormas.api.utils.CsvStreamUtils.writeCsvContentToStream(CsvStreamUtils.java:137)
	... 79 common frames omitted
```

I still think we should get rid of `CaseExportMapperDto`.

StefanKock (changes requested):
Sending `CaseExportMapperDto` over Remote Interface does not work.

StefanKock (approved):
I approve the changes I made with the last commit :-D

carina29 (approved):

PR #9083, opened 2022-05-05
#9081 Avoid fetch of known empty step if current did not exceed stepSize

### LINKED ISSUE #9081: CsvStreamUtils: Avoid last call on empty step
### Problem Description

`CsvStreamUtils` in line 136 runs the while clause until it gets an empty batch. This results always in a follow up fetch of data with 0 returned lines.

### Proposed Change

Stop the while clause if the **processed** result is smaller than the stepSize.

StefanKock:
Fixes #9081


StefanKock:
@MartinWahnschaffe agreed to provide this for 1.71.0, when exports of cases and contacts are at least tested.

BarnaBartha (approved):

StefanKock:
Export of cases and contacts successfully tested (stepSizeForCsvExport: 5000, 2, 3 to check edge cases where the data is more than stepSize), only the needed invocations happened.

PR #9103, opened 2022-05-09
Bugfix 8347 case dependent jurisdiction filters

### LINKED ISSUE #8911: Filters should take into account responsible and place of stay jurisdictions
Verified ticket on the local machine using the latest version of SORMAS from the development branch – 1.71.0-SNAPSHOT(1dc6c47).

As discussed with @MateStrysewske I am reopening the ticket due to the following flows:

I: Contacts directory
1. Create a case whose ‘Responsible jurisdiction’ and ‘Place of stay’ differ (different values for region-district-community);
2. Create a contact with no responsible jurisdiction, but with a source case (the case created at step 1);
3. In the ‘Contacts’ directory, search for the contact from step 2 by using the region-district-community filters and the values of the responsible jurisdiction of the source case.
=> the contact is not listed among the results; the filters only take into account the region-district-community of the ‘place of stay’ of the source case.

Expected behavior: In the ‘Contacts’ directory, the filters for region-district-community should take into account both the responsible jurisdiction and the place of stay of the source case of a contact.

II: Samples directory
1. Create a case whose ‘Responsible jurisdiction’ and ‘Place of stay’ differ (different values for region-district-community);
2. Create a contact with no responsible jurisdiction, but with a source case (the case created at step 1);
3. Add a sample to the contact created at step 2;
4. Navigate to the ‘Samples directory’; 
5. Search for the sample by using the region-district filters and the values of the responsible jurisdiction of the source case.
=> the sample is not listed among the results; the filters for region-district only take into account the values of the ‘place of stay’ of the source case.

Expected behavior: in the ‘Samples’ directory, the filters for region-district should take into account both the responsible jurisdiction and the place of stay of a source case, associated with a contact that has a sample.

III: Samples directory
1. Create a contact that has its own responsible jurisdiction;
2. Add a sample with a positive pathogen test result to the contact created at step 1;
3. Select ‘Yes’ in the pop-up that is triggered after saving the sample;
4. In the ‘Create new case’ form fill in the mandatory information and set different values for the ‘Responsible jurisdiction’ and ‘Place of stay’ of the resulting case;
5. Navigate to the ‘Samples’ directory;
6. Search for the sample from step 2 by using the region-district filters and the values of the responsible jurisdiction of the contact from step 1.
=> the sample is not listed among the results; the filters for region-district only take into account the responsible jurisdiction and place of stay of the resulting case.

Expected behavior: in the ‘Samples’ directory, the filters for region-district should take into account the responsible jurisdiction of the contact associated with the respective sample, the responsible jurisdiction and the place of stay of the resulting case.

_Originally posted by @adinaflorea9 in https://github.com/hzi-braunschweig/SORMAS-Project/issues/8347#issuecomment-1081753418_

Issue started on branch https://github.com/hzi-braunschweig/SORMAS-Project/tree/bugfix-8347-CaseDependentJurisdictionFilters

BarnaBartha:
Fixes #8911


sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
```diff
@@ -725,44 +725,37 @@ public Predicate buildCriteriaFilter(SampleCriteria criteria, SampleQueryContext
...
 				cb,
 				filter,
-				cb.or(
-					cb.and(
-						cb.isNotNull(joins.getCaze()),
-						cb.equal(joins.getCaseResponsibleRegion().get(AbstractDomainObject.UUID), criteria.getRegion().getUuid())),
-					cb.equal(regionExpression, criteria.getRegion().getUuid())));
+				CriteriaBuilderHelper.or(
+					cb,
+					cb.equal(joins.getCaseRegion().get(AbstractDomainObject.UUID), regionUuid),
```
StefanKock, line 735: Could you use `Region.UUID?` instead of `AbstractDomainObject.UUID`, it's makes it more obvious which entity is currently in focus. Same for District and Community where the pattern is the same. See also the other services.

StefanKock (changes requested):
I generelly dislike the AbstractDomainObject.UUID constant to be used (see comment). 

Aside that: Do we have a before/after performance check? I'm not sure how many more joins or conditions we get that could drag the performance down.

BarnaBartha:
@StefanKock We do not have a performance check. I can do that but is it the case that we don't want the feature if it brings a performance degrade?

StefanKock (approved):
I approve the changes concerning architecture, but I don't judge the functional changes if they are correct.

MateStrysewske (changes requested):
Responsible cases of contacts should generally not be considered; for samples and tasks, we should additionally also include the event participant jurisdiction.

MateStrysewske (approved):

PR #9130, opened 2022-05-11
Feature 7304 limit case sync

### LINKED ISSUE #7304: Don't synchronise cases classified as "Not a case" to the mobile apps
### Feature Description

Which cases are synchronised to the mobile apps currently purely depends on the user of the app and whether or not the case is archived or not.

### Problem Description

Especially in the current pandemic situation, some countries create a lot of cases during rapid testing. Most of these cases will be classified as "Not a case" relatively soon because they're not sick. Still, these cases are synchronised to the mobile devices in case they're not also archived manually, resulting in a lot of data to be synchronised that isn't actually of interest for the mobile users.

### Proposed Change

Add a property or feature configuration to turn off the synchronisation of "Not a case" cases that were not reported by the reporting user on a mobile device (i.e. they don't have an app version assigned).

BarnaBartha:
Fixes #7304

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
@@ -53,6 +53,7 @@ import org.apache.commons.lang3.StringUtils;
 import de.symeda.sormas.api.Disease;
 import de.symeda.sormas.api.EditPermissionType;
 import de.symeda.sormas.api.EntityRelevanceStatus;
+import de.symeda.sormas.api.RequestContext;
 import de.symeda.sormas.api.caze.CaseClassification;
 import de.symeda.sormas.api.caze.CaseCriteria;
 import de.symeda.sormas.api.caze.CaseIndexDto;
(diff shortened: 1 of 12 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
```diff
@@ -1201,6 +1202,16 @@ public Predicate createUserFilter(CaseQueryContext caseQueryContext, CaseUserFil
 
 		filter = CriteriaBuilderHelper.or(cb, filter, filterResponsible);
 
+		if (featureConfigurationFacade.isPropertyValueTrue(FeatureType.LIMITED_SYNCHRONIZATION, FeatureTypeProperty.EXCLUDE_NO_CASE_CLASSIFIED_CASES)
+			&& RequestContextHolder.isMobileSync()) {
+			final Predicate limitedCaseSyncPredicate = CriteriaBuilderHelper.and(
```
MateStrysewske, line 1207: I can't imagine that this works as expected. The idea is to exclude cases with the criteria that is specified in the issue description from the list of cases that are returned, but this code doesn't look like it would do that, but instead that it WOULD only synchronise cases that are not classified as not a case, haven't been reported by the user who does the synchronisation, but at the same time have been created in the mobile app. However, since this is then added to the filter with an `or`, I believe that it won't actually have any effect, would it?

I believe that this code needs to be replaced with something like the following:

```suggestion
final Predicate limitedCaseSyncPredicate = cb.not(cb.and(
  cb.equal(casePath.get(Case.CASE_CLASSIFICATION), CaseClassification.NO_CASE),
  cb.or(
    cb.notEqual(casePath.get(Case.REPORTING_USER), currentUser),
    cb.isNull(casePath.get(Case.CREATION_VERSION)))));
filter = CriteriaBuilderHelper.and(cb, filter, limitedCaseSyncPredicate);
```

MateStrysewske (changes requested):

PR #9200, opened 2022-05-16
add instrumentation for performance analysis #8357

### LINKED ISSUE #8357: Detailed performance analysis of getAllAfter and getIndexList methods [5]

### LINKED ISSUE #9219: Add `PerformanceLoggingStopWatch` instrumentation for performance analysis
### Feature Description

In #8357 tooling has been introduced to investigate performance issues. However, this measurement is restricted to calls to EJBs, where instrumentation is available through the `PerformanceLoggingInterceptor`'s logging (with annotation `@AroundInvoke`).

For complex methods within such logged EJB calls (e.g., `PersonService.getAllAfter` with several major SQL queries), more fine grained logging is needed for further investigation. Logging of non-EJB methods and possibly of executions of UI code provides a benefit as well.

The proposed `PerformanceLoggingStopWatch` can replace indivudual performance logging (as introduced, e.g., in PR #9290 in `PersonService.getAllAfter`) with the benefit that results are incorporated in the visual overview generated by the `PerformanceLogAnalysisGenerator`. The possibility to log additional information when logging start or finish of a method (in the example of PR #9290: the number of entries processed) should be implemented.

### Problem Description

The current performance analysis using `PerformanceLoggingInterceptor` and `PerformanceLogAnalysisGenerator` is restricted to EJB calls.

### Proposed Change

The proposal is to introduce a `PerformanceLoggingStopWatch` to be used in arbitrary places in the code. A proof of concept is given in PR  #9200. The logging by `PerformanceLoggingStopWatch` is seamlessly incorporated in the `PerformanceLogAnalysisGenerator`'s analysis, see below for examples.

To be refined:
- is this the right approach
- are there parts of the code where this monitoring should be permanently available, allowing for a quick performance analysis **or**
- should the  `PerformanceLoggingStopWatch`es use rather be demonstrated in an example application outside the productive code.

### Additional Information

Examples for use cases:

Analysis of individual steps within an EJB method, here `PersonService.getAllAfter`:
image

Analysis of non-EJB methods, here `DtoPseudonymizer.pseudonymizeDtoCollection`:
image

syntakker:
~~Fixes #8357~~
~~To clarify: should this instrumentation be added permanently or not?~~

Proof of concept for #9219.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/utils/PerformanceLoggingStopWatch.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/utils/PerformanceLoggingStopWatch.java
@@ -1,0 +1,49 @@
+/*
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2022 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package de.symeda.sormas.api.utils;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PerformanceLoggingStopWatch {
+
+	private static final Logger logger = LoggerFactory.getLogger(PerformanceLoggingStopWatch.class);
+
+	final String methodName;
+	final long startTime;
+
+	public PerformanceLoggingStopWatch(String methodName) {
+		this.methodName = methodName;
+		startTime = DateHelper.startTime();
+		if (logger.isTraceEnabled()) {
+			logStart(methodName, null);
+		}
+	}
+
+	public void stop() {
+		if (logger.isTraceEnabled()) {
+			logFinish(methodName, DateHelper.durationMillies(startTime));
+		}
+	}
+
+	private static void logStart(String invokedMethod, Object[] parameters) {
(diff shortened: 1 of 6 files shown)


MateStrysewske:
LGTM, but @StefanKock should definitely evaluate this.

MateStrysewske (approved):

StefanKock (commented):
I would not like to rush this into 1.72 and on the existing ticket, but as a dedicated ticket to be refined. This could serve as PoC to show the desired use case.

PR #9206, opened 2022-05-16
Bugfix/bug 9080 fix weekly reports page for admins

### LINKED ISSUE #9080: Opening Weekly Reports page with Admin raises NullPointerException
### Bug Description

When opening Weekly Reports with user that has only Admin as user role then a Server error is shown (NullPointerExpection)

### Steps to Reproduce

1. Login with user with role Admin only.
2. Click on Reports in the menu.

### Actual Behavior
Server Error is shown (NPE)

### Expected Behavior

Admin has WEEKLYREPORT_VIEW right [(SORMAS_User_Rights.xlsx)](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-api/src/main/resources/doc/SORMAS_User_Rights.xlsx) and should be able open the Weekly Report page without errors.

### Screenshots

WeeklyReports-Admin-NPE

### System Details

* Device: 
* SORMAS version: 
  * Local: Version: 1.71.0-SNAPSHOT (b900a16)
  * test.sormas.netzlink.com: Version: 1.71.0-SNAPSHOT (ac57983)
* Server URL:
  * Local
  * test.sormas.netzlink.com
* User Role: Admin

### Additional Information

ServerLogs (local):
[WeeklyReports-Admin-NPE.log](https://github.com/hzi-braunschweig/SORMAS-Project/files/8629549/WeeklyReports-Admin-NPE.log)
```
	at de.symeda.sormas.api.report.__WeeklyReportFacade_Remote_DynamicStub.getSummariesPerOfficer(de/symeda/sormas/api/report/__WeeklyReportFacade_Remote_DynamicStub.java)
	at de.symeda.sormas.api.report._WeeklyReportFacade_Wrapper.getSummariesPerOfficer(de/symeda/sormas/api/report/_WeeklyReportFacade_Wrapper.java)
	at de.symeda.sormas.ui.reports.WeeklyReportOfficersGrid.reload(WeeklyReportOfficersGrid.java:173)
	at de.symeda.sormas.ui.reports.ReportsView.reloadGrid(ReportsView.java:160)
	at de.symeda.sormas.ui.reports.ReportsView.enter(ReportsView.java:152)
...
Caused by: java.lang.NullPointerException
	at de.symeda.sormas.backend.report.WeeklyReportFacadeEjb.getSummariesPerOfficer(WeeklyReportFacadeEjb.java:217)
```

`WeeklyReportFacadeEjb.getSummariesPerOfficer` runs into an NPE because of missing `Region`. It seems that `ReportsView.java:63` decides that the user is not on national level and therefore the following backend logic expects a region that the user apperently does not have. I guess the user also another jurisdiction level by roles which results in a lower jurisdiction level role to win `UserRole.getJurisdictionLevel.java:426`

carina29:
Fixes #9080

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportFacadeEjb.java
@@ -157,7 +157,7 @@ public class WeeklyReportFacadeEjb implements WeeklyReportFacade {
 	public List<WeeklyReportRegionSummaryDto> getSummariesPerRegion(EpiWeek epiWeek) {
 		JurisdictionLevel jurisdictionLevel = userService.getCurrentUser().getJurisdictionLevel();
 
-		if (!isNationOrNoneJurisdiction(jurisdictionLevel)) {
+		if (jurisdictionLevel != JurisdictionLevel.NONE & jurisdictionLevel != JurisdictionLevel.NATION) {
 			return new ArrayList<>();
 		}
 
@@ -204,10 +204,6 @@ public class WeeklyReportFacadeEjb implements WeeklyReportFacade {
 		}
 
 		return summaryDtos;
-	}
-
-	public boolean isNationOrNoneJurisdiction(JurisdictionLevel jurisdictionLevel) {
-		return jurisdictionLevel == JurisdictionLevel.NONE || jurisdictionLevel == JurisdictionLevel.NATION;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportFacadeEjb.java
```diff
@@ -205,6 +206,10 @@ public List<WeeklyReportRegionSummaryDto> getSummariesPerRegion(EpiWeek epiWeek)
 		return summaryDtos;
 	}
 
+	public boolean isNationOrNoneJurisdiction(JurisdictionLevel jurisdictionLevel) {
```
MateStrysewske, line 209: I don't see a need to have a separate method for this because it's only used in one place; just move the check to the if clause above please. :) (same for the other class)

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportFacadeEjb.java
carina29: Done :)

MateStrysewske (approved):

PR #9209, opened 2022-05-16
#8933 [DEMIS2SORMAS] Allow users to select an existing sample next to suggested matches

### LINKED ISSUE #8933: [DEMIS2SORMAS] Allow users to select an existing sample next to suggested matches [2]
### Feature Description

Samples > lab messages > processing sample and pathogen test information
### Problem Description

When processing lab messages, users have the option to select existing samples as long as the sample is similar enough to the content of the message. If there is no similar sample, a new sample form opens right away. Unfortunately, users experience insufficient data coming in through in messages which results in the samples not being detected as similar existing entries. Thus, they are forced to create duplicates in order to successfully process messages.

Users should have the option to select other associated samples to the selected entity to counteract insufficient data coming from reporters.
### Proposed Change

- [x] Underneath the "Select sample" radio button and list, provide another option to "Select a different sample"
    - [x] If there are other samples associated to the selected case or contact, list them in a list underneath the new radio button
    - [x] Rename "Select sample" to "Select matching sample"
    - [x] If a similar sample was detected, the "Select matching sample" option should be the default pre-selected radio button
        - [x] The alternative samples associated with the entitiy should still be displayed in the new list but 'greyed out' and only get enabled for selection when "Select a different sample" was pressed.
    - [x] If the entity does not have alternative samples, hide the new list and disable the radio button "Select a different sample".
    - [x] If the entity has associated samples but no matching samples, pre-select the "Select a different sample" radio button to enable the list right away
        - [x] Adjust the info message to the following:

_"The database does not contain samples that seem to be similar to the details of the lab message. However, the selected entity has other associated samples._

_Please look through the list of samples. If you feel certain that one matches the lab message details, select it and click on the Confirm button. Otherwise, click on Create new sample._

_If you are unsure, you can discard this window and cancel the process."_ 

- [x] If the selected entity does not have any associated samples, open a new sample creation form right away as before
- [x] The sample from the first list (select matching sample) should not be shown in the list of different samples

### Additional Information

Select different sample

Select different sample2

Select different sample3

leventegal-she:
Fixes #8933


sormas-api/src/main/resources/strings.properties
```diff
@@ -791,7 +791,7 @@ infoPickOrCreateImmunization = The system already contains at least one immuniza
...
-infoPickOrCreateSample = The database already contains at least one sample that seems to be very similar to the details of the lab message.<br/><br/>Please look through the lists of samples. If you feel certain that one matches the lab message details, select it and click on the <i>Confirm</i> button. Otherwise, click on <i>Create new pathogen test</i> to create a new pathogen test for the sample.<br/><br/>If you are unsure, you can discard this window and cancel the process.
+infoPickOrCreateSample = The database does not contain samples that seem to be similar to the details of the lab message. However, the selected entity has other associated samples.<br/><br/>Please look through the list of samples. If you feel certain that one matches the lab message details, select it and click on the Confirm button. Otherwise, click on Create new sample.<br/><br/>If you are unsure, you can discard this window and cancel the process.
```
FredrikSchaeferVitagroup, line 794: This info does not always fit the situation. It also shows when there is a matching sample:

image

I propose to either show different messages based on what samples there are, or to generalize the message to fit all situations.

FredrikSchaeferVitagroup (changes requested):

sormas-api/src/main/resources/strings.properties
FredrikSchaeferVitagroup: It's generic, but I think it could be a little more instructive. How about that?

```
Please choose one of the options below. 
The list of matching samples contains samples in the database that seem to be similar to the details of the lab message (if any).
The list of different samples contains samples related to the entry you chose, but not similar to the details of the lab message (if any).
You can also create a new sample.
```

leventegal-she: Looks ok to me. But I would keep the also the last sentence: If you are unsure, you can discard this window and cancel the process.

FredrikSchaeferVitagroup: Good point.

FredrikSchaeferVitagroup (approved):

PR #9217, opened 2022-05-17
#7304 - cleanup & fix sonar issues

BarnaBartha:
Fixes #

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/RequestContextHolder.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/RequestContextHolder.java
@@ -27,7 +27,7 @@ public final class RequestContextHolder {
 	}
 
 	public static boolean isMobileSync() {
-		return threadLocalContext.get().getMobileSync();
+		return threadLocalContext.get().isMobileSync();
 	}
 
 	public static void setRequestContext(RequestContextTO requestContext) {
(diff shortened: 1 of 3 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/RequestContextTO.java
```diff
@@ -20,17 +20,17 @@
...
-	private Boolean isMobileSync;
+	private boolean isMobileSync;
 
-	public RequestContextTO(Boolean isMobileSync) {
+	public RequestContextTO(boolean isMobileSync) {
 		this.isMobileSync = isMobileSync;
 	}
 
-	public Boolean getMobileSync() {
+	public boolean getMobileSync() {
```
StefanKock, line 29: Name this also `isMobileSync`, or is there any pattern against that?

sormas-api/src/main/java/de/symeda/sormas/api/RequestContextHolder.java
```diff
@@ -17,21 +17,24 @@
 
 public final class RequestContextHolder {
 
-	private static final ThreadLocal<RequestContextTO> requestContextHolder = new ThreadLocal<>();
+	private static final ThreadLocal<RequestContextTO> threadLocalContext = new ThreadLocal<>();
```
StefanKock, line 20: A constant should be written in uppercase. I'm quite uncertain how this works, I hope it's not shared between invocations or users.

BarnaBartha: Have a look at I18nProperties, it is the same. It is a constant but is set & cleaned up for each request.

StefanKock (approved):

PR #9250, opened 2022-05-19
#9198 - Surveillance Supervisors see the Buttons for deleting Pathoge…

### LINKED ISSUE #9198: Surveillance Supervisors see the Buttons for deleting Pathogen tests while processing Lab Messages, but can not use them.
### Bug Description

Surveillance supervisors by default have the right to process lab messages (currently `LAB_MESSAGES`, [soon ](https://github.com/hzi-braunschweig/SORMAS-Project/issues/8895)`EXTERNAL_MESSAGE_PROCESS`), but not the `PATHOGEN_TEST_DELETE`.

Still, when processing lab messages (in a certain way), they see buttons to remove new and already existing pathogen tests. Removing an existing pathongen test via such button throws an exception `User [user uuid]  is not allowed to delete pathogen tests.`

### Steps to Reproduce

1. As surveillance supervisor with default user rights, start processing a lab message.
2. Refer it to an existing person, case/contact/event-participant, and sample with at least one already saved pathogen test.
-> The `REMOVE THIS PATHOGEN TEST` buttons show.
3. Remove an already saved pathogen test via such button.
4. Try to save the case.
-> Above described exception gets thrown.

### Expected Behavior

`REMOVE THIS PATHOGEN TEST` buttons should be hidden for users without the `PATHOGEN_TEST_DELETE` right, but only for already saved pathogen tests. Unsaved pathogen tests should always be removable.

The default configuration for surveillance supervisors is a little weird then, regarding lab messages processing. But at least theoretically, they should still be able to process most lab messages. @kwa20 Do you agree?

### System Details

* Device: Windows
* SORMAS version: 1.72.0-SHAPSHOT
* Server URL: local dev system
* User Role: surveillance supervisor

cazacmarin:
…n tests while processing Lab Messages, but can not use them.

Fixes #9198


sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -205,19 +207,23 @@ public PathogenTestForm addPathogenTestComponent(
...
-		sampleComponent.addComponent(buttonLayout, sampleComponent.getComponentCount() - 1);
-		discardButton.addClickListener(o -> {
-			sampleComponent.removeComponent(horizontalRule);
-			sampleComponent.removeComponent(buttonLayout);
-			sampleComponent.removeComponent(pathogenTestForm);
-			sampleComponent.removeFieldGroups(pathogenTestForm.getFieldGroup());
-			sampleComponent.removeCommitListener(savePathogenTest);
-			pathogenTestForm.discard();
-		});
+		if ((UserProvider.getCurrent().hasUserRight(UserRight.PATHOGEN_TEST_DELETE) && (nonNull(pathogenTest)))
```
FredrikSchaeferVitagroup, line 210: This does not reliably hide the button for users who may not delete existing pathogen tests. In the [LabMessageProcessingUIHelper.getSampleEditComponent(...)](https://github.com/hzi-braunschweig/SORMAS-Project/blob/c045b5e6009cf28430169be2ce25de7daa054b79/sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/processing/LabMessageProcessingUIHelper.java#L104), existing and new pathogen tests are added as non-null parameters ([here](https://github.com/hzi-braunschweig/SORMAS-Project/blob/c045b5e6009cf28430169be2ce25de7daa054b79/sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/processing/LabMessageProcessingUIHelper.java#L120) and [here ](https://github.com/hzi-braunschweig/SORMAS-Project/blob/c045b5e6009cf28430169be2ce25de7daa054b79/sormas-ui/src/main/java/de/symeda/sormas/ui/labmessage/processing/LabMessageProcessingUIHelper.java#L151)).

FredrikSchaeferVitagroup (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -165,7 +168,8 @@ public PathogenTestForm addPathogenTestComponent(
 		CommitDiscardWrapperComponent<? extends AbstractSampleForm> sampleComponent,
 		PathogenTestDto pathogenTest,
 		int caseSampleCount,
-		Runnable callback) {
+		Runnable callback,
+		boolean isDeleteAllowed) {
```
FredrikSchaeferVitagroup, line 172: I find this parameter name a little misleading. If I set this to false, I would expect the pathogen test to not be deletable.
Yet for a user with the `UserRight.PATHOGEN_TEST_DELETE` right, it is. How about changing the name to isNew?

Also, I think it's a good idea to add some javadoc to this method.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -138,14 +140,15 @@ private void createSample(SampleDto sampleDto, Disease disease, Runnable callbac
 	public PathogenTestForm addPathogenTestComponent(CommitDiscardWrapperComponent<? extends AbstractSampleForm> sampleComponent) {
 
 		int caseSampleCount = caseSampleCountOf(sampleComponent.getWrappedComponent().getValue());
-		return addPathogenTestComponent(sampleComponent, null, caseSampleCount, SormasUI::refreshView);
+		return addPathogenTestComponent(sampleComponent, null, caseSampleCount, SormasUI::refreshView, false);
```
FredrikSchaeferVitagroup, line 143: Why did you set the isDeleteAllowed = false here?
Right now it might work, because the pathogenTest == null, but it think it's a code smell like that.
I my opinion, the isDeleteAllowed should be handed through as a parameter (so it needs to be a parameter of this method as well).

FredrikSchaeferVitagroup (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -159,13 +162,16 @@ public PathogenTestForm addPathogenTestComponent(
 	 *            is valid).
 	 * @param callback
 	 *            use it to define additional actions that need to be taken after the pathogen test is saved (e.g. refresh the UI)
+	 * @param isNew
+	 *            for existing pathogen tests, the 'delete pathogen test' button is hidden for users without UserRight.PATHOGEN_TEST_DELETE permission.
```
FredrikSchaeferVitagroup, line 166:
```suggestion
	 *            for existing pathogen tests, the 'remove this pathogen test' button is hidden for users without UserRight.PATHOGEN_TEST_DELETE permission.
```

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -205,19 +211,23 @@ public PathogenTestForm addPathogenTestComponent(
...
-			sampleComponent.removeComponent(horizontalRule);
-			sampleComponent.removeComponent(buttonLayout);
-			sampleComponent.removeComponent(pathogenTestForm);
-			sampleComponent.removeFieldGroups(pathogenTestForm.getFieldGroup());
-			sampleComponent.removeCommitListener(savePathogenTest);
-			pathogenTestForm.discard();
-		});
+		if ((UserProvider.getCurrent().hasUserRight(UserRight.PATHOGEN_TEST_DELETE) && (nonNull(pathogenTest)))
+				|| isNull(pathogenTest) || isNew) {
+			//pathogenTest is null
```
FredrikSchaeferVitagroup, line 216: I think this comment should be removed.

FredrikSchaeferVitagroup (changes requested):

FredrikSchaeferVitagroup (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/SampleController.java
```diff
@@ -205,19 +211,22 @@ public PathogenTestForm addPathogenTestComponent(
...
-		sampleComponent.addComponent(buttonLayout, sampleComponent.getComponentCount() - 1);
-		discardButton.addClickListener(o -> {
-			sampleComponent.removeComponent(horizontalRule);
-			sampleComponent.removeComponent(buttonLayout);
-			sampleComponent.removeComponent(pathogenTestForm);
-			sampleComponent.removeFieldGroups(pathogenTestForm.getFieldGroup());
-			sampleComponent.removeCommitListener(savePathogenTest);
-			pathogenTestForm.discard();
-		});
+		if ((UserProvider.getCurrent().hasUserRight(UserRight.PATHOGEN_TEST_DELETE) && (nonNull(pathogenTest)))
```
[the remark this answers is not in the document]
FredrikSchaeferVitagroup: Maybe also reverse the order of those conditions to reduce system load.

PR #9325, opened 2022-05-24
#9308 fix bug

### LINKED ISSUE #9308: [Immunizations] 'A problem has occurred' message is thrown when the user links an Immunization (Recovery) to a Case
### Bug Description
When a user wants to link a Case to a Immunization (Recovery) without prior saving the page, a error message is thrown, informing the user that the data has been saved in the meantime.

### Steps to Reproduce

1. Create a Case with the 'Report Date set to May 01;
2. Set the 'Outcome of Case' to 'Recovery' and the 'Date of Outcome' to May 23;
3. Add a Sample with a positive Pathogen Test Result and a Date of Collection and a Date of Result to May 01;
4. Navigate to the Immunizations Directory and create an Immunization (Recovery) for the same person as of the Case from step 1;
5. Set the 'Date of first positive test result' to 01 May and 'Date of Recovery' to May 23; 
6. Do not save the page; Click on 'Link Case' and insert the ID of the Case from Step 1 and click on 'Search Case';
7. In the 'Unsaved Changes' pop-up, click on 'Save Changes'

### Expected Behavior
The user should be able to successfully link the Immunization to the Case; The 'Unsaved Changes' pop-up should probably be triggered right after the user clicks on the 'Link Case';

### Actual Behavior
The system throws 'A problem has occurred' message.

### Screenshots
image

### System Details

* Device: Windows 10
* SORMAS version: Sormas 1.73.0-SNAPSHOT (ec8ae73)
* Server URL: local environment
* User Role: Admin, National User

dinua:
Fixes #9308

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/immunization/ImmunizationController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/immunization/ImmunizationController.java
@@ -167,6 +167,8 @@ public class ImmunizationController {
 				ImmunizationDto immunizationDtoValue = immunizationDataForm.getValue();
 				List<ImmunizationDto> similarImmunizations = findSimilarImmunizations(immunizationDtoValue);
 				if (similarImmunizations.isEmpty()) {
+					//set the change date when the operation is happening
+					immunizationDtoValue.setChangeDate(new Date());
 					FacadeProvider.getImmunizationFacade().save(immunizationDtoValue);
 					if (immunizationDtoValue.getImmunizationStatus() == ImmunizationStatus.ACQUIRED) {
 						NotificationHelper.showNotification(


dinua:
@StefanKock In this scenario the problem was the changeDate for persist changes was after the last value of changeDate in database 
Example:
- the user did some changes in UI and not save them , the changeDate for dto is set to (let's say) 12:45:05
- after that we link the case to immunization and this will persist the entity and set the changeDate to 12:45:06
- in modal view we select to persist the initial changes and we will try to save it with a changeDate lower as last value in DB

MateStrysewske:
@StefanKock is correct, setting the change date in the UI is not appropriate. In general, the change date should not be manipulated here. Roxana already noted the solution to this bug down in the "Expected Behavior" section: The "Unsaved Changes" popup should be triggered directly after clicking the "Link Case" button, BEFORE the link case popup comes up, similar to as we do it with almost all other side components.

dinua:
@StefanKock , @MateStrysewske I did the changes

sormas-ui/src/main/java/de/symeda/sormas/ui/immunization/components/form/ImmunizationDataForm.java
```diff
@@ -264,7 +267,18 @@ protected void addFields() {
 		} else {
 			linkImmunizationToCaseButton = ButtonHelper.createButton(
 				Captions.linkImmunizationToCaseButton,
-				e -> buildAndOpenSearchSpecificCaseWindow(),
+				e -> {
+					if(this.isModified()){
+						actionCallback.accept(() ->{
+							ImmunizationDto immunizationDto = FacadeProvider.getImmunizationFacade().getByUuid(getValue().getUuid());
```
StefanKock, line 273: Code Formatting is not applied here

StefanKock (changes requested):
Coe Formatting is not obeyed, but I cannot judge the functional changes

MateStrysewske (changes requested):
Please ensure that your IDE is properly set up to automatically format code whenever you save. After the code is reformatted, this PR can be merged as the functional part looks good.

PR #9346, opened 2022-05-26
bug/9224_buttons_overlap_for_some_languages

### LINKED ISSUE #9224: Discard and save buttons overlap for point of entry cases in some languages
### Bug Description

On the case form of point of entry cases, the discard save buttons overlap for some user languages (e.g. german).
### Steps to Reproduce

1. Create a case with the origin point of entry but leave the place of stay empty
2. Scroll to the bottom of the page
3. Change the user language to e.g. german
4. Check the button display

### Expected Behavior

Buttons should not overlap
### Screenshots

referPointOFEntryDisplay

### System Details

* Device:
* SORMAS version: 1.71.0

dinua:
Fixes #9224


leventegal-she:
It doesn't seem that the problem is solved
image

I would just move the deletion message out of the toolbar. @MateStrysewske do you know a better place for it?

leventegal-she (changes requested):
Issue not solved

leventegal-she (approved):

PR #9349, opened 2022-05-27
Feature 8967 aggregate reports age groups

### LINKED ISSUE #8967: Addition of age categories to aggregate module (mSERS) [5]
### Situation Description

Aggregated reporting (mSERS) can be used as an alternative to case-based reporting. It currently allows to report number of reported cases, confirmed lab results and deaths for each mSERS-disease per epi week,

For certain disease it's necessary to report these numbers for different age groups (based on the age of the people). 

### Feature Description

* A technical admin should be able to define age groups for each disease using the database. Not all disease have the same age groups and some disease don't need them at all. The age group is not also the same for all instances and can change over time for an instance.
  An example would be 0-28 day, 0-11 months, 12-59 months, 1-4 years, 5-15 years, > 15 years
  This could be limited to a maximum of 8 per disease
* When creating a report using the mobile app and the web app the reporting user needs to be able to enter those numbers for each age group
* The aggregated reports view in the web application needs to display and export these age groups

Examples:

disease | age_group | Total New cases | Total lab confirmed | Total death
-- | -- | -- | -- | --
Covid-19 | 0-28 days |   |   |  
|| 0-11 months |   |  
|| 12-59   months |   |  
|| 1-4 years |   |   |  
|| 5-15 years |   |   |  
|| > 15 years |   |   |  
Meningitis | 0-28 days |   |   |  
|| 0-11 months |   |  
Cholera | 12-59 months |   |  
|| 1-4 years |   |   |  
|| 5-15 years |   |   |  
|| > 15 years |   |   |  
Anthrax | 0-28 days |   |   |  
|| 0-11 months |   |  
|| 12-59   months |   |  
|| 1-4 years |   |   |  
|| 5-15 years |   |   |  
|| > 15 years |   |   |  

#### Date Model

- [x] Add an agegroups field to the disease configuration that allows to define a comma separated list of age groups. Each age group is defined following a simple pattern:
  XA_YB, where X is the "from" age, Y is the "to" age (both inclusive) and A/B is the age type (D: days, M: months, Y: years).
  You can leave the _Y away to define an age group of X+.
  Examples:
  * 0D_28D: "0-28 days"
  * 1M_12M: "1-12 months"
  * 1Y_4Y: "1-4 years"
  * 5Y_15Y: "5-15 years"
  * 16Y: "16+ years"
- [x] Add an age group field (string) to the AggregateReport entities (backend & app) and dtos and the sync process
- [x] Create an utility method that checks whether an age group definition is valid and one that return the caption for it.

#### Report creation in UI

grafik

- [x] Make the disease name bold in the creation form
- [x] When the disease uses age groups, but the disease caption in a dedicated line and add one data entry line for each age group (see mockup)

#### Report creation in app

- [x] Same as in UI

#### Aggregated reports view in UI

- [x] Add a new age group column to the grid
- [x] Group the data by age groups. **Important**: To group don't use the age groups provided by the disease configuration. Instead use the age groups that are used in the actual reported data. Age groups for a disease may be changed over time and thisway we can also display old data without problems.
- [x] Make sure this also reflects in the export

### Additional Information

[aggregate_module (1).ods](https://github.com/hzi-braunschweig/SORMAS-Project/files/8570122/aggregate_module.1.ods)

### Out of scope

This will go into a separate issue:

As a surveillance officer (SO)

- I want to have the feature to document aggregate data on the mobile app and also on the web app. This feature exists already for national users. #9052
- Since SO are limited to districts, the region and districts should be preselected on the creation form and SO should not be able to create reports outside their jurisdiction. SO should be able to create a report for a facility in their district. #9052
- Add a reporting user on the edit form of each report
- There is no possibility to edit a report (either as SO, national user), if a user resubmits another report in the same week, system does not detect them as duplicates but considers them as independent + sum the counts. It would be good to have the option to edit a report. I have add another issue for duplicate detection #9039

BarnaBartha:
Fixes #8967


sormas-api/src/main/java/de/symeda/sormas/api/disease/AgeGroupUtils.java
```diff
@@ -0,0 +1,104 @@
...
+		throw new IllegalArgumentException("Invalid ageGroupTimeUnit definition: " + ageGroupTimeUnit);
+	}
+
+	public static List<String> convertToList(String ageGroupsString) {
+		if (StringUtils.isBlank(ageGroupsString)) {
+			return null;
+		}
+
+		try {
+			final List<String> agrGroupList = Stream.of(ageGroupsString.split(",")).collect(Collectors.toList());
```
sergiupacurariu, line 79: please update the variable name

sergiupacurariu:
- age group calculation does not work: first age group element is always ignored
- group overlap issue, something like this is possible: 0D_28D,1M_36M,1Y (second age group end is greater than starting of the next age group)

- the error should mention something meaningful regarding age groups validation

sergiupacurariu (changes requested):

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -2986,6 +2986,11 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 						+ "hasOptionalHealthFacility boolean, hasAssociatedDistrictUser boolean, portHealthUser boolean, jurisdictionLevel varchar(255));");
 				getDao(UserRole.class).executeRaw("CREATE TABLE users_userRoles(user_id integer, userRole_id integer);");
 
+			case 336:
```
MateStrysewske, line 2989: Increase the DATABASE_VERSION above and test the migration, please

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -2986,6 +2986,11 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
...
+				currentVersion = 336;
+				getDao(AggregateReport.class).executeRaw("ALTER TABLE aggregateReport ADD COLUMN ageGroup varchar(255);");
+				getDao(DiseaseConfiguration.class).executeRaw("ALTER TABLE diseaseConfiguration ADD COLUMN ageGroups varchar(512);");
```
MateStrysewske, line 2992: Use text instead of varchar(512)

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/disease/DiseaseConfigurationDao.java
```diff
@@ -43,6 +50,38 @@ public String getTableName() {
...
+
+		final List<DiseaseConfiguration> diseaseConfigurations = queryForAll();
+
+		final User currentUser = ConfigProvider.getUser();
+
+		for (DiseaseConfiguration configuration : diseaseConfigurations) {
+			final Disease limitedDisease = currentUser.getLimitedDisease();
+			if (limitedDisease == null) {
+				final Disease disease = configuration.getDisease();
+				boolean activeFlagMatches = active.equals(configuration.getActive()) || active.equals(disease.isDefaultActive());
```
MateStrysewske, line 65: Does this make sense? Won't this return a disease as active that is specifically set to inactive on the server, but is active by default?

sormas-app/app/src/main/java/de/symeda/sormas/app/report/aggregate/AggregateReportsFragment.java
```diff
@@ -154,23 +165,88 @@ private void updateByEpiWeek() {
...
+			if (ageGroups == null || ageGroups.isEmpty()) {
+				LayoutInflater inflater = (LayoutInflater) getActivity().getSystemService(Context.LAYOUT_INFLATER_SERVICE);
+				RowReportAggregateLayoutBinding binding =
+					DataBindingUtil.inflate(inflater, R.layout.row_report_aggregate_layout, contentBinding.reportContent, true);
+			} else {
+				LayoutInflater diseaseInflater = (LayoutInflater) getActivity().getSystemService(Context.LAYOUT_INFLATER_SERVICE);
+				RowReportAggregateDiseaseLayoutBinding viewBinding =
+					DataBindingUtil.inflate(diseaseInflater, R.layout.row_report_aggregate_disease_layout, contentBinding.reportContent, true);
+				viewBinding.setDisease(disease);
+				int i = 0;
```
MateStrysewske, line 242: Remove this unused variable

sormas-backend/src/main/java/de/symeda/sormas/backend/disease/DiseaseConfiguration.java
```diff
@@ -126,4 +132,14 @@ public Boolean getExtendedClassificationMulti() {
 	public void setExtendedClassificationMulti(Boolean extendedClassificationMulti) {
 		this.extendedClassificationMulti = extendedClassificationMulti;
 	}
+
+	@Column(length = CHARACTER_LIMIT_DEFAULT)
```
MateStrysewske, line 136: Use CHARACTER_LIMIT_TEXT

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -11525,4 +11525,12 @@ CREATE TRIGGER delete_history_trigger_userroles_smsnotificationtypes
 
 INSERT INTO schema_version (version_number, comment, upgradeNeeded) VALUES (460, 'Replace hard-coded user roles with fully configurable user roles #4461', true);
 
+-- 2022-05-20 Addition of age categories to aggregate module (mSERS) [5] #8967
+ALTER TABLE diseaseconfiguration ADD COLUMN agegroups varchar(512);
```
MateStrysewske, line 11529: Use text here (and in the history table)

MateStrysewske (changes requested):

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
BarnaBartha: tested it, indeed there were issues
but works well now

MateStrysewske (changes requested):
- IF POSSIBLE, age groups should be sorted in the directory (after applying the grouping to diseases, of course), taking into account that days should come before months and months before years.
- Please adjust the bottom padding of the disease headings in the mobile aggregate report view when they have age groups (i.e. the disease should look like it looks when it doesn't have age groups, with similar top and bottom padding).
- At least for me, aggregate reports aren't saved to the mobile database at all; please have a look into that.

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportFacadeEjb.java
```diff
@@ -179,7 +178,11 @@ public List<AggregatedCaseCountDto> getIndexList(AggregateReportCriteria criteri
 		}
 
 		List<AggregatedCaseCountDto> reportList = new ArrayList<>(reportSet);
-		reportList.sort(Comparator.comparing(r -> r.getDisease().toString()));
+		Function<AggregatedCaseCountDto, String> objectObjectFunction = r -> r.getDisease().toString();
+		Comparator<AggregatedCaseCountDto> comparator = Comparator.comparing(objectObjectFunction)
+			.thenComparing(r -> r.getAgeGroup().replaceAll("\\d", ""))
```
MateStrysewske, line 183: This fails with a NullPointerException when there are aggregate reports without age groups in the system (which is very likely)

MateStrysewske, line 183: This will lead to a NullPointerException for aggregate reports without age groups (which are very likely)

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/report/AggregateReportDao.java
```diff
@@ -58,7 +58,13 @@ public List<AggregateReport> getReportsByEpiWeekAndUser(EpiWeek epiWeek, User us
 				public int compare(AggregateReport o1, AggregateReport o2) {
 					return o1.getDisease().toString().compareTo(o2.getDisease().toString());
 				}
-			});
+			}.thenComparing(new Comparator<AggregateReport>() {
+
+				@Override
+				public int compare(AggregateReport o1, AggregateReport o2) {
+					return o1.getAgeGroup().compareTo(o2.getAgeGroup());
```
MateStrysewske, line 65: Will probably lead to a NullPointerException if one of the age groups is null (which is very likely to happen)

MateStrysewske (changes requested):

MateStrysewske (approved):

MateStrysewske (approved):

MateStrysewske (approved):

Issue #4461, opened 2021-02-18
Replace hard-coded user roles with fully configurable user roles and remove REST_USER user role [5]


MartinWahnschaffe:
### Problem Description
The current user roles are hard-coded, but need to become manageable by the admin. There is already a first step made towards this in #830, which allows to override the user rights configuration in the database.

Source: https://github.com/hzi-braunschweig/SORMAS-Project/issues/898#issuecomment-671336289

### Proposed Change

Depends on #2804.

#### Preparation

- [x] Existing UserRole enum should become private DefaultUserRole enum and should be moved to sormas-backend
- [x] Move mapping between DefaultUserRole and UserRight from UserRight enum to DefaultUserRole enum

#### UserRole entity

- [x] The existing UserRoleConfig class needs to become the new UserRole class. When loading user roles only use the UserRole entries from the database (in other words: Don't mix-in user roles created on the fly based on the default user roles or something like that).
- [x] UserRole: Add caption and description fields
- [x] UserRole: Add hasOptionalHealthFacility, hasAssociatedOfficer and portHealthUser boolean fields. Those might be replaced with other mechanics in the future
- [x] UserRole: Add a list of notifications to the user role
- [x] User: Replace Set of old UserRole enum with Set of new UserRole entity. Will require a user:userRoles mapping table.
- [x] Provide a method to create user roles based on all default roles. Should be called similar to the default users when the system is set up
- [x] ~~Should we keep a defaultUserRole field to know that a user roles was created based on a default user role? Could be helpful for migrations later. Otherwise~~ remove the userRole field.
- [x] the getAssignableUserRoles logic can be dropped - it should actually already have been dropped with #659

#### Migration

- [x] When updating from a previous version call the method to create user roles based on the default user roles
- [x] Migration of user roles for existing users

#### Android App

- [x] Replace userRolesJson in User with join table
- [x] UserRoleConfig becomes UserRole
  - Create new table and drop the old one
  - Include all fields that have been added in #4461 to the UserRole. Notification types can be ignored
  - userRights can be stored as json
- [x] Change the UserDao.getUsersByAssociatedOfficer method (introduced by #8318) to retrieve useres based on the actual user rights (join user roles and do a "like" check on the user rights json)
- [x] Same for the logic needed for getRandomRegion/DistrictCaseResponsible introduced by #8318

Out of scope: Add translations for user role names

## Release Notes
[...]

PR #9365, opened 2022-05-31
#4461: fixed problem with users without userroles

ChristopherRiedel:
Fixes


MateStrysewske (changes requested):
Okay, looks like there's not actually an exception thrown there when `getByCaption` fails, instead `importuserUserRole` will just be null. Can you please adjust the code to throw the exception if this is the case, and otherwise execute the rest of the code?

PR #9375, opened 2022-05-31
[#8949] lab external id

### LINKED ISSUE #8949: [DEMIS2SORMAS] Compare the list of Organization Id's with facilities in SORMAS [3]
### Feature Description

When https://github.com/hzi-braunschweig/sormas-demis-adapter/issues/49 gets implemented, not only one but a list of Id's can map to available facilities (specifically laboratories) in SORMAS.
### Problem Description

Thus far, only one Id is considered when mapping the value in `labMessage.labExternalId` to facilities in SORMAS. 
### Proposed Change

- [x] Rename the `LabMessageDto.labExternalId` to `.labExternalIds` [with an 's'] and enable the attribute to store multiple values (via storage as JSON string).
- [x] Consider all available Id's in the list of `labMessage.labExternalId` and compare them with existing facilities
    - [x] If no match was found, select `OTHER` in the facility field and write the respective facility name into the details field as before.
    - [x] If just one Id matches an existing facility, set that facility as reference.
    - [x] If more than one Id matches existing facilities, leave the field empty since the user has to make a decision about which facility is supposed to be referenced.
- [x] Implement the points above as well for test reports.
- [x] Cover the new functionalities with unit tests
- [x] Make sure that the changes don't break sharing lab messages (and test reports) via S2S.
### Additional Information

Related ticket in the SORMAS_DEMIS-adapter: [#49](https://github.com/hzi-braunschweig/sormas-demis-adapter/issues/49).

JonasCir:
Fixes #8949 
@FredrikSchaeferVitagroup ~My tests with regard to S2S are still pending, but this is still ready for review.~

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/InfrastructureDataReferenceDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/InfrastructureDataReferenceDto.java
@@ -32,6 +32,7 @@ public abstract class InfrastructureDataReferenceDto extends ReferenceDto implem
 		this.externalId = externalId;
 	}
 
+	// todo shouldn't this be also a list then?
 	public String getExternalId() {
 		return externalId;
 	}
--- a/sormas-api/src/main/java/de/symeda/sormas/api/InfrastructureDataReferenceDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/InfrastructureDataReferenceDto.java
@@ -32,7 +32,6 @@ public abstract class InfrastructureDataReferenceDto extends ReferenceDto implem
 		this.externalId = externalId;
 	}
 
-	// todo shouldn't this be also a list then?
 	public String getExternalId() {
 		return externalId;
 	}
--- a/sormas-api/src/main/java/de/symeda/sormas/api/externalmessage/ExternalMessageDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/externalmessage/ExternalMessageDto.java
@@ -13,7 +13,6 @@ import javax.validation.constraints.Size;
 
 import de.symeda.sormas.api.Disease;
 import de.symeda.sormas.api.externalmessage.labmessage.TestReportDto;
-import de.symeda.sormas.api.feature.FeatureType;
 import de.symeda.sormas.api.i18n.Validations;
 import de.symeda.sormas.api.person.Sex;
 import de.symeda.sormas.api.sample.PathogenTestResultType;
@@ -23,7 +22,6 @@ import de.symeda.sormas.api.sample.SpecimenCondition;
 import de.symeda.sormas.api.sormastosormas.SormasToSormasShareableDto;
 import de.symeda.sormas.api.user.UserReferenceDto;
 import de.symeda.sormas.api.utils.DataHelper;
-import de.symeda.sormas.api.utils.DependingOnFeatureType;
 import de.symeda.sormas.api.utils.FieldConstraints;
 
 @DependingOnFeatureType(featureType = FeatureType.EXTERNAL_MESSAGES)
@@ -40,7 +38,7 @@ public class ExternalMessageDto extends SormasToSormasShareableDto {
 	public static final String SAMPLE_MATERIAL_TEXT = "sampleMaterialText";
 	public static final String SPECIMEN_CONDITION = "specimenCondition";
 	public static final String REPORTER_NAME = "reporterName";
-	public static final String LAB_EXTERNAL_ID = "labExternalId";
+	public static final String REPORTER_EXTERNAL_ID = "reporterExternalId";
 	public static final String REPORTER_POSTAL_CODE = "reporterPostalCode";
 	public static final String REPORTER_CITY = "reporterCity";
 	public static final String PERSON_FIRST_NAME = "personFirstName";
@@ -77,7 +75,7 @@ public class ExternalMessageDto extends SormasToSormasShareableDto {
 	@Size(max = FieldConstraints.CHARACTER_LIMIT_SMALL, message = Validations.textTooLong)
 	private String reporterName;
 
-	private List<String> labExternalIds;
+	private List<String> reporterExternalIds;
 	@Size(max = FieldConstraints.CHARACTER_LIMIT_SMALL, message = Validations.textTooLong)
 	private String reporterPostalCode;
 	@Size(max = FieldConstraints.CHARACTER_LIMIT_SMALL, message = Validations.textTooLong)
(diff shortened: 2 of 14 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/InfrastructureDataReferenceDto.java
```diff
@@ -32,6 +32,7 @@ protected InfrastructureDataReferenceDto(String uuid, String caption, String ext
 		this.externalId = externalId;
 	}
 
+	// todo shouldn't this be also a list then?
```
vidi42, line 35: Should it? To my understanding a lab message could point to multiple facilities, but a facility is still identified by a single external id

sormas-api/src/test/java/de/symeda/sormas/api/utils/DataHelperTest.java
```diff
@@ -133,6 +134,7 @@ public void testParseDateWithExceptionForDeFormat() throws ParseException {
 	}
 
 	@Test
+	@Ignore
```
vidi42, line 137: We should remove/resolve the `@Ignore` before merging

vidi42 (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/InfrastructureDataReferenceDto.java
JonasCir: Correct. I think we can currently leave it this way. I added a remark for reference DTOs to #9389 in case we want to change something there.

sormas-api/src/test/java/de/symeda/sormas/api/utils/DataHelperTest.java
JonasCir: true

JonasCir:
@vidi42 Is it ok to merge this even if the corresponding DEMIS adapter changes are pending?

vidi42:
> @vidi42 Is it ok to merge this even if the corresponding DEMIS adapter changes are pending?

We should probably wait for the demise issue to be ready to be merged as well.

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -11600,4 +11600,20 @@ ALTER TABLE externalmessage_history ADD COLUMN caze_id bigint;
 
 INSERT INTO schema_version (version_number, comment) VALUES (467, '[DEMIS2SORMAS] Introduce processing for physician reports #8980');
 
+-- 2022-06-13 lab organization ids #8949
+ALTER TABLE labmessage ALTER COLUMN labexternalid TYPE VARCHAR(255)[] USING ARRAY[labexternalid];
+ALTER TABLE labmessage_history ALTER COLUMN labexternalid TYPE VARCHAR(255)[] USING ARRAY[labexternalid];
+
+ALTER TABLE labmessage RENAME COLUMN labexternalid TO labexternalids;
+ALTER TABLE labmessage_history RENAME COLUMN labexternalid TO labexternalids;
```
FredrikSchaeferVitagroup, line 11608: Those commands don't work (anymore), because the labmessage tables got renamed to externalmessage.

sormas-api/src/main/java/de/symeda/sormas/api/InfrastructureDataReferenceDto.java
FredrikSchaeferVitagroup: I also think this should not be a list, especially not for #8949. Shouldn't this comment just be removed?

FredrikSchaeferVitagroup (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/InfrastructureDataReferenceDto.java
JonasCir: removed

sormas-backend/src/main/resources/sql/sormas_schema.sql
JonasCir: fixed

sormas-api/src/test/java/de/symeda/sormas/api/utils/DataHelperTest.java
JonasCir: removed

sormas-ui/src/main/java/de/symeda/sormas/ui/externalmessage/ExternalMessageMapper.java
```diff
@@ -331,14 +331,26 @@ protected ImmutableTriple<String, DiseaseVariant, String> migrateDiseaseVariant(
...
+	public FacilityReferenceDto getLabReference(List<String> labExternalIds) {
+
 		FacilityFacade facilityFacade = FacadeProvider.getFacilityFacade();
-		List<FacilityReferenceDto> labs =
-			labExternalId != null ? facilityFacade.getByExternalIdAndType(labExternalId, FacilityType.LABORATORY, false) : null;
-		if (labs != null && labs.size() == 1) {
-			return labs.get(0);
+		List<FacilityReferenceDto> labs;
+
+		if (labExternalIds != null && !labExternalIds.isEmpty()) {
```
FredrikSchaeferVitagroup, line 339: After the db migration, I got something like `{NULL}` in the labexternalids column of the for one entry.
This seems to be a list with a null element. This if statement yields true in that scenario, which it probably shouldn't... Or?

FredrikSchaeferVitagroup (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/externalmessage/ExternalMessageMapper.java
FredrikSchaeferVitagroup: When trying to process the external message with the `{NULL}` entry, the mapping crashes.

sormas-api/src/main/java/de/symeda/sormas/api/externalmessage/ExternalMessageDto.java
```diff
@@ -76,8 +76,8 @@ public class ExternalMessageDto extends SormasToSormasShareableDto {
 
 	@Size(max = FieldConstraints.CHARACTER_LIMIT_SMALL, message = Validations.textTooLong)
 	private String reporterName;
-	@Size(max = FieldConstraints.CHARACTER_LIMIT_SMALL, message = Validations.textTooLong)
-	private String labExternalId;
+
+	private List<String> labExternalIds;
```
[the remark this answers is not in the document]
JonasCir: fixed

sormas-ui/src/main/java/de/symeda/sormas/ui/externalmessage/ExternalMessageMapper.java
JonasCir: @FredrikSchaeferVitagroup can you try again, I use `.filter(Objects::nonNull)` which should remove the object

FredrikSchaeferVitagroup (approved):

PR #9400, opened 2022-06-02
Bugfix 8888 bulk task assignment

### LINKED ISSUE #8888: Tasks can be assigned to users via bulk edit mode that don't have access to related entities
### Bug Description

When selecting tasks associated to entities in different jurisdictions, users can be assigned via bulk edit mode that don't actually have the right to see the related entity
### Steps to Reproduce

1. Log in as an admin, national user
2. Create e.g. two cases in different jurisdicitons
3. Create tasks for each case
4. Open tasks and enter the bulk edit mode
5. Select the two created tasks and edit them in bulk
6. In the assignee drop-down, select a user (e.g. Supervisor or Officer) that should only have access to one of the tasks and related entities
7. Log in as the assigned user and check the associated entities to the tasks

### Expected Behavior

Only show users in the list that are assignable to all selected tasks:

- When the user clicks on the "Change assignee" checkbox, check the lowest jurisdiction that all of them have in common and filter the list by users with this jurisdiction level or above and this exact jurisdiction (e.g. when the entities have different districts, but the same region, filter the list by all users with jurisdiction level region or nation that either have no region or the same region as the one of the entities).
- When the lowest identical jurisdiction is not HEALTH_FACILITY or POINT_OF_ENTRY, show an info below the checkbox and above the dropdown (only if the checkbox is selected): "You have selected tasks with different [facilities/points of entry/communities/districts/regions]. Only users on [community/district/region/national] level or above will be selectable as assignees."
- Use an identical style as for the hint that appears in the task edit form when a user without an associated email address and phone number is selected (i.e. black text, small info icon to the left).

### Screenshots

taskAssigneeNoAccess

### System Details

* Device:
* SORMAS version: 1.70.2

MateStrysewske:
Fixes #8888

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/InfrastructureHelper.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/InfrastructureHelper.java
@@ -6,6 +6,7 @@ import java.util.Date;
 
 import de.symeda.sormas.api.i18n.I18nProperties;
 import de.symeda.sormas.api.infrastructure.pointofentry.PointOfEntryDto;
+import de.symeda.sormas.api.user.JurisdictionLevel;
 import de.symeda.sormas.api.utils.DataHelper;
 import de.symeda.sormas.api.utils.DateHelper;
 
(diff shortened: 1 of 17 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/infrastructure/InfrastructureHelper.java
```diff
@@ -73,4 +74,22 @@ public static Integer getProjectedPopulation(Integer population, Date collection
 	public static BigDecimal getCaseIncidence(int caseCount, int population, int divisor) {
 		return new BigDecimal(caseCount).divide(new BigDecimal((double) population / divisor), 2, RoundingMode.HALF_UP);
 	}
+
+	public static JurisdictionLevel getSuperordinateJurisdiction(JurisdictionLevel jurisdiction) {
```
BarnaBartha, line 78: this method should already be present in JurisdictionHelper, not sure i understand why moving it here is better

BarnaBartha (approved):

PR #9402, opened 2022-06-02
#4461: only create external system users on startup if userroles are found

ChristopherRiedel:
only update external system users on startup if userroles are found

This is the suggested solution for #4461 finding no. 31 (https://wiki.vitagroup.ag/pages/viewpage.action?pageId=87001413)

Fixes #

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
@@ -546,8 +546,13 @@ public class StartupShutdownService {
 
 		User existingUser = userService.getByUserName(username);
 
+		boolean allUserRolesExist = userRoles != null && !userRoles.isEmpty() && !userRoles.stream().anyMatch(userRole -> userRole == null);
 		if (existingUser == null) {
-			if (!DataHelper.isNullOrEmpty(password)) {
+			if (!allUserRolesExist) {
+				logger.warn(
+					"User '" + firstName + " " + lastName
+						+ "' was not created because at least one of the user roles to be assigned cannot be found in the database.");
+			} else if (!DataHelper.isNullOrEmpty(password)) {
 				User newUser = MockDataGenerator.createUser(userRoles, firstName, lastName, password);
 				newUser.setUserName(username);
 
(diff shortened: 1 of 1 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
```diff
@@ -546,8 +546,13 @@ private void createOrUpdateDefaultUser(Set<UserRole> userRoles, String username,
 
 		User existingUser = userService.getByUserName(username);
 
+		boolean allUserRolesExist = userRoles != null && !userRoles.isEmpty() && !userRoles.stream().anyMatch(userRole -> userRole == null);
```
StefanKock, line 549: Use `collections4.CollectionUtils.isEmpty()` instead of `userRoles != null && !userRoles.isEmpty()`.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
```diff
@@ -546,8 +546,13 @@ private void createOrUpdateDefaultUser(Set<UserRole> userRoles, String username,
...
 		if (existingUser == null) {
-			if (!DataHelper.isNullOrEmpty(password)) {
+			if (!allUserRolesExist) {
+				logger.warn(
+					"User '" + firstName + " " + lastName
```
StefanKock, line 553: Better write `logger.warn("User '{} {}' was not created because at least one of the user roles to be assigned cannot be found in the database.", firstName, lastName)` (here and on following log statements).

StefanKock (changes requested):
Apply the same check in `StartupShutdownService.createDefaultUsers`

ChristopherRiedel:
> Apply the same check in `StartupShutdownService.createDefaultUsers`

done.

vidi42 (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
```diff
@@ -546,8 +556,14 @@ private void createOrUpdateDefaultUser(Set<UserRole> userRoles, String username,
 
 		User existingUser = userService.getByUserName(username);
 
+		boolean allUserRolesExist = !CollectionUtils.isEmpty(userRoles) && !userRoles.stream().noneMatch(Objects::isNull);
```
vidi42, line 559: I think the `!` has to be removed as well as now it will only be true if one of the values is `null`

vidi42 (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
ChristopherRiedel: @vidi42 No, since we want to know if it's not empty so that 'allUserRolesExist'

ChristopherRiedel: my bad

vidi42 (approved):

PR #9414, opened 2022-06-03
#9266 - Properly set classification info

### LINKED ISSUE #9266: 'Case classification' not updated in the mobile app when the case gets classified as 'Not a case'
### Bug Description

The case classification of cases that were created in the vaadin-app and synchronized to the mobile app is not updated if it is set to 'Not a case'.

### Steps to Reproduce

1. In the vaadin-app, log in with an admin+national user and create a case;
2.  Set the 'Case classification' to any other value than 'Not a case';
3. In the android-app, log in with a Surveillance Officer and trigger a synchronization with the server;
=>the case from step 1 appears in the mobile app;
4. In the vaadin-app, update the case's classification to 'Not a case';
7. From the mobile app, do another synchronization;

### Actual Behavior

In the mobile app, the 'Case classification' remains the same as it was after step 3.

### Expected Behavior

- Add a new method to the resources that unifies `getArchivedUuidsSince` and `getDeletedUuidsSince`; call it something like `getObsoleteUuidsSince`; add javadoc that explains that this is supposed to be used by the mobile app in order to remove archived and deleted entities from the device (for cases: and potentially also cases classified as not a case if the respective feature configuration is enabled).
- Write corresponding backend methods that combine `getArchivedUuidsSince` and `getDeletedUuidsSince` (and for cases also add another clause to return the UUIDs of cases that are classified as "Not a case" and have a `classificationDate` bigger than the specified `since` date).
- In the mobile app, call the new unified resource instead of doing separate calls for all entities.

### System Details

* Device: Windows 10 (vaadin-app); Android Studio Emulator - Pixel C (android-app)
* SORMAS version: 1.73.0-SNAPSHOT(92d0095)
* Android version/Browser: Chrome (vaadin-app); Android 11 (android-app) 
* Server URL: local machine
* User Role: Admin+NatUser (vaadin-app); Surveillance Officer (android app)

### Additional Information

#7304

MateStrysewske:
Fixes #9266

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -1584,7 +1584,8 @@ public class CaseFacadeEjb extends AbstractCoreFacadeEjb<Case, CaseDataDto, Case
 		return caseSave(dto, handleChanges, existingCase, toDto(existingCase), checkChangeDate, internal);
 	}
 
-	@RolesAllowed({ UserRight._CASE_EDIT })
+	@RolesAllowed({
+		UserRight._CASE_EDIT })
 	public CaseDataDto updateFollowUpComment(@Valid @NotNull CaseDataDto dto) throws ValidationRuntimeException {
 		Pseudonymizer pseudonymizer = getPseudonymizerForDtoWithClinician("");
 		Case caze = service.getByUuid(dto.getUuid());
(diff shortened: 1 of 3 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2081,13 +2083,13 @@ public void onCaseChanged(CaseDataDto existingCase, Case newCase, boolean syncSh
...
 					}
 				}
-			} else if (newCase.getCaseClassification() != existingCase.getCaseClassification()) {
-				newCase.setClassificationUser(userService.getCurrentUser());
-				newCase.setClassificationDate(new Date());
 			}
-		} else if (newCase.getCaseClassification() != existingCase.getCaseClassification()) {
+		}
+
+		if (setClassificationInfo && existingCase != null && newCase.getCaseClassification() != existingCase.getCaseClassification()) {
```
BarnaBartha, line 2092: when a new case is created (maybe from the rest api) and the classification is already filled shouldn't these fields also be filled?

MateStrysewske:
@BarnaBartha There are test failures, I'll check them tomorrow and request another review from you once they're fixed.

BarnaBartha (approved):

PR #9422, opened 2022-06-04
#9352 - Different errors are thrown when an Admin tries to link or cr…

### LINKED ISSUE #9352: Different errors are thrown when an Admin tries to link or create an Event Group
### Bug Description

Depending on the jurisdiction of the event, when an Admin tries to link or create an event group, there are 2 different error messages that are displayed. For the second scenario, the error is also registered in the local machine logs, as the app seems to crash.

### Steps to Reproduce

I;
1. Logged in with a user that has only 'Admin' as user right, create an event with the location: Country - Deutschland, Region - Baden-Württemberg, District - LK Alb-Donau-Kreis;
2. On the event's page, click on the 'Link event Group' from the 'Event Group Card'.
**Actual Behavior:** An error message is displayed and the user can not link an event group.
image

II:
1. Logged in with a user that has only 'Admin' as user right, create an event with the location: Country - Afghanistan (no region and no district);
2. Open the event's page and click on the 'Link Event Group' from the 'Event Group Card';
3. In the 'Create new event group' pop-up, fill in the mandatory information and click 'Save'.
**Actual behavior:** An error message is displayed and the logs show an error.
image

Logs from the local machine:
[link event group error - Admin.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/8786825/link.event.group.error.-.Admin.txt)

### Expected Behavior

As the Admin has EVENTGROUP_CREATE, EVENTGROUP_EDIT and EVENTGROUP_LINK rights, they should be able to successfully link and create event groups, at least to events that were also created by them.

### System Details

* Device: Windows 10
* SORMAS version: 1.73.0-SNAPSHOT(7bff1b7) and 1.72.1
* Server URL: local machine and release-international.sormas.netzlink.com
* User Role: Admin

cazacmarin:
Fixes #9352

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventGroupFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventGroupFacadeEjb.java
@@ -322,7 +322,7 @@ public class EventGroupFacadeEjb implements EventGroupFacade {
 
 		for (Event event : events) {
 			final JurisdictionLevel jurisdictionLevel = currentUser.getJurisdictionLevel();
-			if (jurisdictionLevel != JurisdictionLevel.NATION) {
+			if ((jurisdictionLevel != JurisdictionLevel.NATION) && !currentUser.isAdmin()) {
 				Region region = event.getEventLocation().getRegion();
 				if (!userService.hasRegion(new RegionReferenceDto(region.getUuid()))) {
 					throw new UnsupportedOperationException(
@@ -366,7 +366,7 @@ public class EventGroupFacadeEjb implements EventGroupFacade {
 		Event event = eventService.getByUuid(eventReference.getUuid());
 
 		final JurisdictionLevel jurisdictionLevel = currentUser.getJurisdictionLevel();
-		if (jurisdictionLevel != JurisdictionLevel.NATION) {
+		if ((jurisdictionLevel != JurisdictionLevel.NATION) && !currentUser.isAdmin()) {
 			Region region = event.getEventLocation().getRegion();
 			if (!userService.hasRegion(new RegionReferenceDto(region.getUuid()))) {
 				throw new UnsupportedOperationException(
@@ -401,7 +401,7 @@ public class EventGroupFacadeEjb implements EventGroupFacade {
 		EventGroup eventGroup = eventGroupService.getByUuid(uuid);
 
 		final JurisdictionLevel jurisdictionLevel = currentUser.getJurisdictionLevel();
-		if (jurisdictionLevel != JurisdictionLevel.NATION) {
+		if ((jurisdictionLevel != JurisdictionLevel.NATION) && !currentUser.isAdmin()) {
 			List<RegionReferenceDto> regions = getEventGroupRelatedRegions(eventGroup.getUuid());
 			for (RegionReferenceDto region : regions) {
 				if (!userService.hasRegion(region)) {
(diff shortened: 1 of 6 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventGroupFacadeEjb.java
```diff
@@ -322,7 +322,7 @@ public void linkEventsToGroups(List<EventReferenceDto> eventReferences, List<Eve
 
 		for (Event event : events) {
 			final JurisdictionLevel jurisdictionLevel = currentUser.getJurisdictionLevel();
-			if (jurisdictionLevel != JurisdictionLevel.NATION) {
+			if ((jurisdictionLevel != JurisdictionLevel.NATION) && !currentUser.isAdmin()) {
```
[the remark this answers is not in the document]
cazacmarin: We  do need admin user. This user needs to have possibility to do anything he wants without  joining any region. Checked with Christopher and Mate.
Mate said:`I guess the problem is that a pure Admin user does not really have access to data, and something is clashing there; so yeah, theoretically the Admin should be allowed to do that, but it kind of depends on what the cause for the error is in the code and how easy it is to fix it in different ways`

sormas-backend/src/main/java/de/symeda/sormas/backend/user/User.java
```diff
@@ -234,6 +235,12 @@ public JurisdictionLevel getJurisdictionLevel() {
 		return jurisdictionLevel;
 	}
 
+	@Transient
```
vidi42, line 238: Could we remove this from the Entity and keep it in the UserProvider to avoid duplication?

sormas-ui/src/main/java/de/symeda/sormas/ui/UserProvider.java
```diff
@@ -97,6 +97,11 @@ public boolean hasNationJurisdictionLevel() {
 		return getJurisdictionLevel() == JurisdictionLevel.NATION;
 	}
 
+	public boolean isAdmin() {
```
vidi42, line 100: We should probably search for `I18nProperties.getEnumCaption(role)` as the caption will be translated to the server's locale during migration

vidi42 (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/user/User.java
cazacmarin: UserProvider is in Sormas-UI project. However, current implementation is on Sormas-Backend side. It means, that current implementation cannot be done in than way.. 
Doing vice-versa -  for accessing it from Sorma-UI and calling Sormas-Backend (but it still can require some more changes). Might be if any other module will require it, at that moment, this code can be refactored.

sormas-ui/src/main/java/de/symeda/sormas/ui/UserProvider.java
cazacmarin: Due to the point that I'm streaming all userRoles, and then looking by caption, this method will not work. `user.getUserRoles().stream().peek(i->i.getCaption().contains(I18nProperties.getEnumCaption(DefaultUserRole.ADMIN))).collect(Collectors.toList())` will return  2 instead of 1, in case if 1 other user still have ADMIN role. And in this case, user which is not authorized to link elements, will be able to do this. 
Due  to that, I'm still thinking that current approach is the best  one :) 
image

vidi42 (approved):

PR #9426, opened 2022-06-06
#8314 replace to string with get caption

### LINKED ISSUE #8314: Move functional information from toString to getCaption methods [5]
### Feature Description

Many `toString` methods contain functional information, e.g. first and last name, disease, classification, etc.

### Problem Description

`toString` methods can be (automatically) used in places where we don't want them to be used. Because of that, it would be better to split the functional information and move them to a different method.

### Proposed Change

- Add `getCaption` methods to all DTOs and entities that have custom `toString` methods (probably makes sense to add them to the superclasses if possible, e.g. EntityDto and AbstractDomainObject, and override them in child classes; default implementation should simply call the `toString` method)
- Fill the `getCaption` methods of all affected classes with the content of the respective `toString` method (if implemented; otherwise no need to override it)
- Replace all custom `toString` methods: Return a string containing the entity type + UUID in parentheses (e.g. "Case (XXXXX-XXXXX-XXXXXX-XXXXX)
- Replace all usages of the `toString` method in the UI with the new `getCaption` method, and also all implicit `toString` usages (e.g. in string format methods)

BarnaBartha:
Fixes #8314

Close action: Definitely squash merge this branch (with meaningful commit message) to remove the tedious commit history.


sormas-api/src/main/java/de/symeda/sormas/api/EntityDto.java
```diff
@@ -128,4 +128,13 @@ public EntityDto clone() throws CloneNotSupportedException {
 
 		return (EntityDto) super.clone();
 	}
+
+	public String getCaption() {
```
StefanKock, line 134: In `getCaption()` I would write `getClass() + StringUtils.SPACE + this.getUuid()` to have a useful default if the subclass does not override it.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractDomainObject.java
```diff
@@ -181,6 +181,15 @@ public int hashCode() {
 		return getUuid().hashCode();
 	}
 
+	public String caption() {
+		return toString();
+	}
+
+	@Override
+	public String toString() {
```
StefanKock, line 189: Superfluous default implementation, but `getClass() + StringUtils.SPACE + getUuid()` would remove a lot of implementation on subclasses. Since this should only appear rarely and in log files, I would not care about the `"Entity"Dto.I18N_PREFIX`.

StefanKock (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/EntityDto.java
BarnaBartha: getCaption defaults to toString if not overwritten
but adding this to the toString makes sense

sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractDomainObject.java
```diff
@@ -181,6 +181,15 @@ public int hashCode() {
 		return getUuid().hashCode();
 	}
 
+	public String caption() {
```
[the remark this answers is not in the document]
BarnaBartha: for me adding the @Transient was not enough, the server wouldn't start up by complaining that there is no such attribute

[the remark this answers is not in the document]
BarnaBartha: it is used in some places when we build some string in the backend or to create ReferenceDtos
not sure how we should move it to DTOs, the backend has the api as dependency not the other way around
if @MateStrysewske agrees with you we can rework this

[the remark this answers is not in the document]
StefanKock: Since the `caption()` method is in entities, it's part of the backend, so it can easily be moved to EJBs where `caption()` is currently invoked, and using existing helper methods there.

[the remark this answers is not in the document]
MateStrysewske: Fine by me to put it into the EJBs

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
```diff
@@ -1723,6 +1724,6 @@ public void setHealthConditions(HealthConditionsDto healthConditions) {
 
 	@Override
 	public String toString() {
-		return this.getUuid() + " - " + this.getExternalID();
+		return CaseDataDto.I18N_PREFIX + StringUtils.SPACE + this.getUuid() + " - " + this.getExternalID();
```
MateStrysewske, line 1727: Just realised that there's a problem with this because cases might not have an external ID; nothing introduced by you, but can you add an `isBlank` check before appending the dash and the external ID?

MateStrysewske (changes requested):
Idea: Add a `getI18nPrefix` method to EntityDto and PseudonymizableIndexDto that returns null; override this method in subclasses and return the respective `I18N_PREFIX` of the class. Use `getI18nPrefix` in `EntityDto.toString` (fall back to `getClass().getSimpleName()` if the prefix is null). This allows us to get rid of most `toString` methods.

sormas-api/src/main/java/de/symeda/sormas/api/person/PersonIndexDto.java
```diff
@@ -4,11 +4,13 @@
 import java.util.Date;
 
 import de.symeda.sormas.api.caze.AgeAndBirthDateDto;
+import de.symeda.sormas.api.caze.CaseDataDto;
```
StefanKock, line 7: Organize Imports ...

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleIndexDto.java
```diff
@@ -26,12 +26,14 @@
 import de.symeda.sormas.api.event.EventParticipantReferenceDto;
 import de.symeda.sormas.api.infrastructure.facility.FacilityHelper;
 import de.symeda.sormas.api.infrastructure.facility.FacilityReferenceDto;
+import de.symeda.sormas.api.person.PersonDto;
```
StefanKock, line 29: Organize Imports

sormas-backend/src/main/java/de/symeda/sormas/backend/action/Action.java
```diff
@@ -29,8 +29,11 @@
 import javax.persistence.Temporal;
 import javax.persistence.TemporalType;
 
+import org.apache.commons.lang3.StringUtils;
+
```
StefanKock, line 33: Organize Imports...

sormas-backend/src/main/java/de/symeda/sormas/backend/activityascase/ActivityAsCase.java
```diff
@@ -35,7 +35,10 @@
 import javax.persistence.Temporal;
 import javax.persistence.TemporalType;
 
+import org.apache.commons.lang3.StringUtils;
```
StefanKock, line 38: Organize Imports

sormas-backend/src/main/java/de/symeda/sormas/backend/campaign/data/CampaignFormData.java
```diff
@@ -30,10 +30,12 @@
 import javax.persistence.Temporal;
 import javax.persistence.TemporalType;
 
+import org.apache.commons.lang3.StringUtils;
 import org.hibernate.annotations.Type;
 
 import de.symeda.auditlog.api.Audited;
 import de.symeda.auditlog.api.AuditedIgnore;
+import de.symeda.sormas.api.campaign.data.CampaignFormDataDto;
```
StefanKock, line 38: Organize Imports

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/Case.java
```diff
@@ -42,12 +42,14 @@
...
 import javax.persistence.Transient;
 
+import org.apache.commons.lang3.StringUtils;
 import org.hibernate.annotations.Type;
 
 import de.symeda.auditlog.api.Audited;
 import de.symeda.auditlog.api.AuditedIgnore;
 import de.symeda.sormas.api.Disease;
 import de.symeda.sormas.api.caze.CaseClassification;
+import de.symeda.sormas.api.caze.CaseDataDto;
```
StefanKock, line 52: Organize Imports

sormas-backend/src/main/java/de/symeda/sormas/backend/clinicalcourse/ClinicalCourse.java
```diff
@@ -4,7 +4,10 @@
 import javax.persistence.Entity;
 import javax.persistence.OneToOne;
 
+import org.apache.commons.lang3.StringUtils;
```
StefanKock, line 7: Organize Imports

StefanKock (changes requested):
I stopped at roughly half the the changes because of too many not cleaned imports on JPA entities.

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
```diff
@@ -1720,8 +1724,13 @@ public void setHealthConditions(HealthConditionsDto healthConditions) {
...
 
+	@JsonIgnore
+	public String getI18nPrefix() {
+		return I18N_PREFIX;
+	}
+
 	@Override
 	public String toString() {
-		return this.getUuid() + " - " + this.getExternalID();
+		return super.toString() + (StringUtils.isNotBlank(this.getExternalID()) ? " - " + this.getExternalID() : StringUtils.EMPTY);
```
StefanKock, line 1734: Why is the `externalID` part of the toString() method. Should only be standard imho.

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseIndexDto.java
```diff
@@ -391,10 +393,15 @@ public CaseReferenceDto toReference() {
...
 
 	@Override
-	public String toString() {
+	public String getCaption() {
 		return CaseReferenceDto.buildCaption(getUuid(), getPersonFirstName(), getPersonLastName());
 	}
 
+	@Override
+	public String toString() {
+		return CaseDataDto.I18N_PREFIX + StringUtils.SPACE + uuid;
```
StefanKock, line 402: Example where default implementation of `toString()` in `PseudonymizableIndexDto` would be helpful

StefanKock (approved):
I checked the recent changes and don't see any objections. 
Only this remark was not addressed (but I would not insist to do this in hope that everyone who creates a subclass will implement the `i18nPrefix` method volantarly): Make `EntityDto.getI18nPrefix()` abstract to force subclasses implement that method.

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
BarnaBartha: was like that before
maybe @MateStrysewske can weigh in on if it is really needed

BarnaBartha:
> The current state looks quite good now. Two reasonable changes that I see:
> 
> 1. Make `EntityDto.getI18nPrefix()` abstract to force subclasses implement that method.
> 2. Maybe rename the method to `i18nPrefix()` to not match the bean pattern and therefore `@JsonIgnore` is then not needed?
> 3. Copy the `getI18nPrefix` pattern into `PseudonymizableIndexDto` to reduce the need for copy-paste pattern in toString() methods of subclasses.
> 4. Add missing `@Override` to methods in  `PseudonymizableIndexDto`.

Did the changes apart from point 1 - and it's corresponding changed for point 3 and 4.
Not all extending classes of entity dto have i18nprefix, it would require a lot more changes and a lot of return null i18nPrefix methods.

StefanKock (approved):
It's okay, then no abstract `i18nPrefix`.

sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
MateStrysewske: Pretty sure it's important for Germany which is why it should stay there, but only if it's filled in (so, as it is right now)

MateStrysewske (approved):

PR #9431, opened 2022-06-07
#8974 - Group aggregated reporting data (mSERS) by jurisdiction and e…

### LINKED ISSUE #8974: Group aggregated reporting data (mSERS) by jurisdiction and epi week [5]
### Situation Description

The aggregated reports view is currently only displaying (and exporting) the four columns disease, new cases, lab confirmations and deaths. The system always aggregates the numbers for the whole jurisdiction that is defined by filters, e.g. when filtering to a district, the numbers of all facilities are aggregated. This is not sufficient.

### Feature Description
The core requirement here is to group the data aggregation by jurisdiction and by epi week.

- [x] Add a grouping field for the jurisdiction. The field only allows to pick a jurisdiction level that is below or equal to the jurisdiction level used in the filters. E.g. when the data is filter to a specific district the available groupings would be district, facility and point of entry (see campaign statistics - there we have already implemented something similar).
- [x] Add a region column to the grid and depending on the grouping also columns for district and facility or point of entry. To do this extend the AggreagtedCaseCountDto and the backend logic that queries the data
- [x] Add an epi week column to the grid and dto. Always group the data by epi week, so when the date filter ranges across multiple epi weeks, each epi week will have entries in the resulting table
- [x] Add a disease filter (just single select)
- [x] The 0-rows, we are currently adding for disease that don't have reported numbers should stay and have to be extended to have one for every epi week. In terms of jurisdiction grouping and age groups don't add multiple 0-rows. Instead use the jurisdiction of the filter and no age group. 
- [x] Add checkbox "show 0-rows for groupings"
- [x] Make sure all of this reflects correctly in the export

grafik

### Additional Information

[export_aggregate_module.ods](https://github.com/hzi-braunschweig/SORMAS-Project/files/8570715/export_aggregate_module.ods)

sergiupacurariu:
…pi week

Fixes #8974


BarnaBartha (approved):

MateStrysewske (changes requested):
- Grouping dropdown content does not have proper captions
- Default selections for epi year and week have been removed, should be added back (i.e. display the current week by default)
- Move the "Apply filters" button next to the time filters
- If I understood the issue description correctly, we need to show one line per epi week per disease; this is currently not the case

MateStrysewske (changes requested):
0-groupings are not working correctly right now. If I display them e.g. for epi week 24, the "epi week" column shows "0" for the diseases that have no reports there, but it should still display 24. Also, we need to show one row per disease per epi week that is displayed (i.e. when filtering from epi week 18 to 24, we need to show at least 7 rows per disease), potentially more if there are age and jurisdiction groupings.

MateStrysewske (approved):

MateStrysewske (changes requested):
There are two failing tests - once they're running through, this PR can be merged. :)

PR #9434, opened 2022-06-07
Change 9299 vaccination dupliate detection

### LINKED ISSUE #9299: Duplicate detection for vaccinations when merging cases [~5]
### Problem Description

While performing case and contact merging (each case/contact has his own person), here an example what happens with the immunizations/vaccinations:
Merging

- non-leading immunization with the person is not discarded 
- vaccinations are added to remaining immunizations without comparing whether they subsequently exist twice. This can complicate evaluations of the number of vaccinations per case/contact due to duplicates.
- Person duplicates are the main issue because immunizations and therefore vaccinations are not duplicated for one person.

### Proposed Change

- Currently, vaccinations are added to remaining immunizations.
    - Merge vaccinations rather than just adding them in the following way:
        - IF entries do not have vaccinationDate AND vaccineName or vaccinationDate OR vaccineName, simply add the vaccination to the remaining immunization
        - IF entries have a matching vaccinationDate AND vaccineName, do not create an additional entry
transfer any additional fields that are empty in the remaining vaccination to the leading vaccination

MateStrysewske:
Fixes #9299

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/immunization/ImmunizationFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/immunization/ImmunizationFacadeEjb.java
@@ -646,7 +646,8 @@ public class ImmunizationFacadeEjb
 	@RolesAllowed({
 		UserRight._IMMUNIZATION_CREATE,
 		UserRight._PERSON_EDIT })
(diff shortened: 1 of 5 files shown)


sergiupacurariu:
- in case a duplicate vaccination is found and the immunization has only that Vaccination, an empty immunization is still added to the leading case. Should this happen?

MateStrysewske:
* Vaccinations with unknown vaccines should not be merged

sergiupacurariu (approved):

PR #9455, opened 2022-06-09
#9286 - Task syn for not a case cases

### LINKED ISSUE #9286: Tasks of unsynchronized cases appear in the mobile app
### Bug Description

If a case is created in the web app and it is classified as 'Not a case', it is not synchronized to the mobile app when the synchronization with the server is triggered. However, tasks that are assigned to the user of the mobile app, and are associated with unsynchronized case, appear in the mobile app, and are not associated to any entity.

### Steps to Reproduce

1. Log in the web app with an admin+national user and create a case and a task associated with the case;
2. Set the assignee of the task a Surveillance Officer that is also logged in the mobile app;
3. Set the case's classification to 'Not a case';
4. In the mobile app, trigger a synchronization with the server;
5. Navigate to the 'Cases' directory and to the 'Tasks' directory.

### Actual Behavior

The case classified as 'Not a case' is not synchronized to the mobile app. However, the task that is associated (in the web app) to this case, is synchronized and appears in the mobile app. 

### Expected Behavior

The case classified as 'Not a case' is not synchronized to the mobile app. The task associated with this case is also not synchronized to the mobile app.

### Screenshots

image

### System Details

* Device: Windows 10 (vaadin-app); Android Studio Emulator - Pixel C (android-app)
* SORMAS version: 1.73.0-SNAPSHOT(631efcd)
* Android version/Browser: Chrome (vaadin-app); Android 11 (android-app)
* Server URL: local machine
* User Role: Admin+NatUser (vaadin-app); Surveillance Officer (android app)

### Additional Information

For context regarding the expected behavior, if a case is archived and the synchronization with the mobile app is triggered, the case and case's associated task are not synced to the mobile app, even if the assignee of the task is the user logged in the mobile app.

MateStrysewske:
Fixes #9286

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactCriteria.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/contact/ContactCriteria.java
@@ -138,7 +138,6 @@ public class ContactCriteria extends BaseCriteria implements Serializable {
 	private Date creationDateTo;
 	private String reportingUserLike;
 	private String personLike;
-
 	private boolean excludeLimitedSyncRestrictions;
 
 	public UserRoleReferenceDto getReportingUserRole() {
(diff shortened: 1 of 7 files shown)


BarnaBartha (approved):
looks ok
i have one worry about the naming of: isExcludeLimitedSyncRestrictions
this is one heavy flag name to grasp
skipLimitedSync or even better limitedSync with the reverse flags would be better for the future debuggers i think

PR #9501, opened 2022-06-14
#4461 - Clear user rights on logout

MateStrysewske:
Connected to #4461


StefanKock:
The change on `ConfigProvider` came with 5ee9cad -> #4461, so this ticket is a duplicate to #4461 and should probably be treated like that.

StefanKock:
Was discussed and approved in a meeting

PR #9573, opened 2022-06-22
#9125, #8576 validations before share or accept

### LINKED ISSUE #8576: [S2S] Avoiding simultaneous work of two health departments - preventing sharing twice to the same target system as long as the target system has not yet accepted or rejected [2]
### Problem Description

Currently a source system can send a case multiple times to a target system with different parameters. This allows two health departments to work together on one case. 

Steps to reproduce:
1. source system: share a case with hand over the ownership (parameters: hand over the ownership = true, comment = "With")
2. source system: share **the same** case **without** hand over the ownership (parameters: hand over the ownership = false, comment = "without")
3. target system: first accept the last request sent (Comment: "Without")
4. target system: then accept the request you sent first (Comment: "With")
5. change the year of birth in the source system and in the target system
6. check the "share-box" (tab: case, on the right, between "vaccination-box" and "reporting-tool-box") in the source system and in the target system
7. check the sync in both directions by changing year of birth in source system and target system

Expected result: 
6.1 the share box in the source system shows the comment "without" (so the source system thinks it shared without hand over the ownership, because this was the last executed)
6.2 the share box in the target system shows the comment "with" (so the target system thinks it accepted with hand over the ownership, because this was the last executed accepted request)
7.1 the sync works in both direction

That is a big problem for example because of automatic reporting system in the background.
(This scenario can occur if one employee share a case by mistake with hand over the ownership, then realises that she/he made a mistake and then shares again without hand over the ownership. Because the last send request is the highest in the target system the employee from target system will accept the last send request first. 

### Proposed Change

A source system should only be able to send a case once as long as the target system has not yet accepted or rejected. If a case  has been sent by mistake, the case must be taken back (revoke), first. Then the case can be shared again. 
Perhaps technically feasible by target system checking whether there is an open request for the GUID already. If so, an error message is displayed in the source system. 
-> must be implemented for contacts to.

=> For Events and external messages this should be done as well

### Additional Information

source system: release-sormas-x
target system: release-de
1.69.1
Edge

### LINKED ISSUE #9125: Reducing the risk that the link between case and associated contact gets lost in source and/or target system [2]
### Situation Description

At the moment the situation is as follows:

1. Send a contact linked to a case to a target system with 'hand over the ownership' = true.
2. Accept the contact in the target system.
3. Change the contact in the target system and save.
4. Send the associated case (for ownership or viewing) to the same target system.
Result:
-After step 3, the link between contact and case in the source system is lost. That's bad.
-After step 4, the contact is not linked to the case in the target system. That is also bad. 

Also the situation is as follows:

1. Send a case associated with a contact (for viewing or ownership) to a target system.
2. Send the contact linked to the case to the same target system for ownership.
3. Accept the contact in the target system first.
4. Change the contact in the target system and save it.
5. Accept the case in the target system.
Result:
-Link between case and contact does not exist in the target system and the link will be deleted in source system if contact is shared with 'hand over the ownership' = true. That's bad. 
-(If the case is accepted first and then the contact, the link is adopted).
### Feature Description

```diff
+First feature (source system): 
```

If a user wants to share a contact, he/she first has to press the button "+share" in the source system. 

1. If this happens, the source system has to check if the contact has a linked case. 

1.a IF NOT
If the contact has not a linked case a popup window appears with the message: "You can not share this contact because you need a source case. Please link the contact to a case". The message can be closed and the user then is back in the contact (tab 'contact'). The pop-up window 'share' does NOT appear. (As long as the contact is not linked to a case). So if the user does not choose a source case and clicks again on the button "+share" the process described in 1.a is repeated. 

1.b IF YES
If the contact is linked to a case in the source system, the pop-up window 'share' opens. 

1.b.1 
As soon as the user in the source system selects an 'organisation' (the target system) via dropdown menu, the source system must check whether the associated case has already been transferred to the _**same**_ target system as specified for the contact.

1.b.1.a IF NOT
If the case associated with the contact has not yet been shared to the _**same**_ target system (status: 'pending' or 'accepted'), the system must inform the user that the associated case has to be sent first. A pop-up message appears: 'If you want to share this contact, you have to sent the associated case first to the _**same**_ target system. Please note that you can sent the case and exclude personal data'. The pop-up message can be closed. After closing the message **the button 'share' is greyed out and is not operable**→ It is **not** possible for the user to sent the contact bevor he sent the associated case to the **same** target system. 

Notice: If the linked case was sent for example to 'target system X' and the user accidentally choose 'target system Y' for the contact the system behave like described in 1.b.1.a (pop-up message and afterwards greyed out and not editable button 'share'). If the user then notice his/her error and selects the 'target system X' for the contact the button 'share' becomes editable and blue. 

1.b.1.b IF YES
If the case associated with the contact has yet been shared to the _**same**_ target system (status: 'pending' or 'accepted') the user can share the contact.

Notice: If the user changes the target system of the contact, the source system must repeat immediately 1.b.1. 

```diff
+Second feature (target system):
```

Bevor the target system accepts a shared contact, the target system must perform the following tests as soon as the "accept" button is clicked on:

Step 1: Is the linked case known in the target system? If yes everything is ok and the contact is accepted. If not step 2 will be executed.

Step 2: If the linked case is not known in the target system but there is an open share request of the matching case, an error message appears that the linked case must first be accepted. The message can be closed. The contact is not accepted. If the user clicks again on the button 'accept' the process described in step 2 is repeated. So it is not possible for the user to accept the contact before the linked case is accepted. 
If there is no open share request of the matching case step 3 will be executed.

Notice: Step 3 happens, if the following happend in the source system: the linked case was sent to 'target system X' first -> then the contact was sent to 'target system X'. After that the shared case was revoked in the source system. 

Step 3: If the linked case is not known in the target system and there is no open share request of the matching case, an error message appears: 'This contact was sent without a case. That's why you can not accept the contact. Please reject the contact and ask the source system to send the linked case first.' This error message appears as soon as the user clicks on "accept". The user has the option to close this error-message-pop-up-window.  If the user clicks again on the button 'accept' the process described in step 3 is repeated.

### Implementation

When sharing a contact do the following:

- [x] Sharing a contact that is not associated to a case should no longer be possible: "You can not share this contact because you need a source case. Please link the contact to a case"
- [x] If the associated case of the contact has not yet been shared to the target system (not 'pending' or 'accepted') display: "If you want to share this contact, you have to send the associated case first to the same target system. Please note that you can send the case and exclude personal data."
  The "share" button in the sharing dialog should be disabled until the user has selected a valid target system.

When receiving a contact from another system:

- [x] When a user tries to accept an incoming contact and the related case does not exist, but a share request exists for it, an error message appears that the linked case must first be accepted.
- [x] When also no share request exists display: 'The contact cannot be accepted, because it was sent without a case. Please reject the contact and ask the source system to send the linked case first.'

leventegal-she:
Fixes #9125
Fixes #8576


vidi42 (approved):

vidi42:
We should not merge this into development as it is required for the hotfix

vidi42:
@leventegal-she can we close this PR now that it's merged into hotfix 1.73.1?

leventegal-she:
> @leventegal-she can we close this PR now that it's merged into hotfix 1.73.1?

I would say yes.

PR #9581, opened 2022-06-22
#9073 - Add a view to list aggregate report data and to highlight dup…

### LINKED ISSUE #9073: Add a view to list aggregate report data and to highlight duplicates [3]
### Feature Description

Currently the mSERS module for aggregated case reporting only provides a view that aggregates the data for a certain jurisdiction. There is currently no way to display, edit and delete the underlying reports and also no way of detecting duplicates in the web app.

### Proposed Change

- [x] Add a new view to the Aggregate Reporting (mSERS) section in SORMAS. The existing view should be called "Aggregated Reporting" and the new view "Report Data"
- [x] The grid will list AggregateReportDto based on the filters and the user filter of the current user (use existing logic). E.g. a surveillance officer will only see aggregate reports of the own district and below.
- [x] Use the same filters available in the existing view, but without the grouping
- [x] The grid should support sorting
- [x] Entries in the grid that are duplicates of others should by displayed with a light red background. Duplicates are aggregate reports that exist multiple times for the exact same jurisdiction, disease and epi week. The goal is to highlight entries that will be ignored to calculate the aggregation totals (#9068), this means the youngest duplicate entry should not be displayed red.
- [x] Add a filter to only display all those duplicates.

### Additional Information

Out of scope:
* Lazy loading is not needed at this point, since the grid will always filter the current epi week by default
* editing & deletion will by handled by #9119

sergiupacurariu:
…licates

Fixes #9073


sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsView.java
```diff
@@ -39,31 +45,55 @@ public class AggregateReportsView extends AbstractView {
 
 	private AggregateReportCriteria criteria;
 
-	private AggregateReportsGrid grid;
+//	private AggregateReportsGrid grid;
```
MateStrysewske, line 48: Remove commented code if no longer needed

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsView.java
```diff
@@ -74,19 +104,40 @@ public AggregateReportsView() {
 
 		addComponent(gridLayout);
 
+		OptionGroup aggregateReportViewSwitcher = new OptionGroup();
```
MateStrysewske, line 107: I believe Martin's idea was to have a submenu as in the configurations or campaigns views, not an option group; @MartinWahnschaffe please confirm

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsView.java
```diff
@@ -159,7 +225,8 @@ private VerticalLayout createFilterBar(UserDto user) {
 
 		aggregateReportsFilterForm.addApplyHandler(e -> {
 			if (epiWeekFilterBarDataValidation()) {
-				grid.reload();
+//				((AggregateReportsGrid) grid).reload();
```
MateStrysewske, line 228: Remove commented code if no longer needed

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportFacadeEjb.java
```diff
@@ -307,6 +310,85 @@ public List<AggregatedCaseCountDto> getIndexList(AggregateReportCriteria criteri
 		return resultList;
 	}
 
+	@Override
+	public List<AggregateReportDto> getDuplicateAggregateReports(AggregateReportCriteria criteria) {
```
MateStrysewske, line 314: Since this method is used to retrieve all reports, not only duplicates, I'd rather call it `getAggregateReports`

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportFacadeEjb.java
```diff
@@ -307,6 +310,85 @@ public List<AggregatedCaseCountDto> getIndexList(AggregateReportCriteria criteri
...
+				.thenComparing(
+					r -> r.getAgeGroup() != null ? Integer.parseInt(r.getAgeGroup().split("_")[0].replaceAll("[^0-9]", StringUtils.EMPTY)) : 0)
+				.thenComparing(AggregateReportDto::getRegion, Comparator.nullsFirst(Comparator.naturalOrder()))
+				.thenComparing(AggregateReportDto::getDistrict, Comparator.nullsFirst(Comparator.naturalOrder()))
+				.thenComparing(AggregateReportDto::getHealthFacility, Comparator.nullsFirst(Comparator.naturalOrder()))
+				.thenComparing(AggregateReportDto::getPointOfEntry, Comparator.nullsFirst(Comparator.naturalOrder())));
+
+		List<AggregateReportDto> onlyDuplicatesReturnList = new ArrayList<>();
+
+		aggregateReportDtoList.forEach(aggregateReportDto -> {
```
MateStrysewske, line 352: I believe this whole part of the code could be simplified by something like this:

```
aggregateReportDtoList.forEach(r -> {
  List<AggregateReportDto> duplicateReports = aggregateReportDtoList.stream().filter(r2 -> !r.equals(r2) && isDuplicate(r, r2));
  if (duplicateReports.stream.anyMatch(r2 -> r.getChangeDate().before(r2.getChangeDate()))) {
    r.setDuplicate(true);
  }
  onlyDuplicatesReturnList.add(r);
  onlyDuplicatesReturnList.addAll(duplicateReports);
}
```

Would of course require to turn onlyDuplicatesReturnList to a Set and convert that set back to a list when returning it, or alternatively doing contains checks. Correct me if I'm wrong, maybe I'm missing something here, or whether you think that my approach doesn't make much sense (absolutely possible of course :D).

MateStrysewske (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsView.java
MartinWahnschaffe: Yes, exactly

sormas-api/src/main/resources/captions.properties
```diff
@@ -229,9 +229,12 @@ aggregateReportEditAggregateReport=Edit aggregate report
 aggregateReportEditReport=Edit report
 aggregateReportReportFound=Aggregate report found
 aggregateReportShowZeroRowsForGrouping = Show 0-rows for groupings
+aggregateReportShowOnlyDuplicateReports = Display only duplicate reports
 aggregateReportNewAggregateReport=New aggregate report
 aggregateReportNewCasesShort=C
 aggregateReportThisWeek=This Week
+aggregateReport_AggregateReportingView=Aggregate reporting
```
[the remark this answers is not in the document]
sergiupacurariu: done

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsView.java
sergiupacurariu: removed

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsView.java
sergiupacurariu: removed

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportFacadeEjb.java
sergiupacurariu: changed

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportFacadeEjb.java
sergiupacurariu: with some additional changes I updated the logic

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsView.java
sergiupacurariu: updated

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/ReportDataView.java
```diff
@@ -0,0 +1,193 @@
...
+					10000);
+			}
+		});
+
+		filterLayout.addComponent(aggregateReportsFilterForm);
+
+		return filterLayout;
+	}
+
+	private boolean epiWeekFilterBarDataValidation() {
```
MateStrysewske, line 173:
```suggestion
	private boolean isEpiWeekFilterValid() {
```

MateStrysewske (changes requested):
Please use an even more transparent colour for the duplicate background (alpha 0.1) in order to increase the contrast to the text colour.

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/ReportDataView.java
sergiupacurariu: changed

sergiupacurariu:
duplicate background changed to alpha = 0.1

MateStrysewske (approved):

PR #9587, opened 2022-06-22
#9579 Replace @RolesAllowed by own @RightsAllowed annotation and chec…

### LINKED ISSUE #9579: Replace @RolesAllowed by own @RightsAllowed annotation (Backport of #9577)
### Proposed Change
#9577 for 1.72

MartinWahnschaffe:
…k with interceptor.

Fixes #9579

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowed.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowed.java
@@ -1,0 +1,25 @@
+package de.symeda.sormas.backend.util;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Mimics behaviour of @RolesAllowed to workaround performance problem.
+ */
+@Documented
+@Retention(RUNTIME)
+@Target({
+	TYPE,
+	METHOD })
+public @interface RightsAllowed {
+
+	/**
+	 * List of rights that are permitted access.
+	 */
+	String[] value();
+}
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowedInterceptor.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowedInterceptor.java
@@ -1,0 +1,113 @@
+/*
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2021 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package de.symeda.sormas.backend.util;
+
+import java.util.HashMap;
+
+import javax.annotation.Resource;
+import javax.annotation.security.PermitAll;
+import javax.ejb.EJB;
+import javax.ejb.SessionContext;
+import javax.interceptor.AroundInvoke;
+import javax.interceptor.InvocationContext;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import de.symeda.sormas.api.i18n.I18nProperties;
+import de.symeda.sormas.api.i18n.Strings;
+import de.symeda.sormas.api.user.UserRight;
+import de.symeda.sormas.api.utils.AccessDeniedException;
+import de.symeda.sormas.backend.user.CurrentUserService;
+
+/**
+ * Implementation for @RightsAllowed annotation to mimic @RolesAllowed
+ */
+public class RightsAllowedInterceptor {
+
+	@Resource
+	private SessionContext sessionContext;
+	@EJB
+	private CurrentUserService currentUserService;
+
+	private final Logger logger = LoggerFactory.getLogger(getClass());
+
+	private final HashMap<String, UserRight> userRightValues = new HashMap<>();
+
+	@AroundInvoke
+	public Object checkRightsAllowed(InvocationContext context) throws Exception {
+
+		if (context.getMethod().getAnnotation(PermitAll.class) != null) {
+			// method permits
+			return context.proceed();
+		}
+
+		RightsAllowed rightsAllowed = context.getMethod().getAnnotation(RightsAllowed.class);
+		if (rightsAllowed == null) {
+
+			if (context.getTarget().getClass().getAnnotation(PermitAll.class) != null) {
+				// class permits & no method annotation
+				return context.proceed();
+			}
+
+			rightsAllowed = context.getTarget().getClass().getAnnotation(RightsAllowed.class);
+		}
+
+		if (rightsAllowed != null) {
+
+			if (userRightValues.isEmpty()) {
+				synchronized (userRightValues) {
+					if (userRightValues.isEmpty()) {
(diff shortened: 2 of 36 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowedInterceptor.java
```diff
@@ -0,0 +1,98 @@
...
+
+	private static final Logger logger = LoggerFactory.getLogger(RightsAllowedInterceptor.class);
+
+	private final HashMap<String, UserRight> userRightValues = new HashMap();
+
+	@AroundInvoke
+	public Object checkRightsAllowed(InvocationContext context) throws Exception {
+
+		PermitAll permitAll = context.getMethod().getAnnotation(PermitAll.class);
+		if (permitAll == null) {
```
StefanKock, line 52: I don't like this deep if nesting, but `@PermitAll` to be handled as prechecked condition.

```java
// if @PermitAll than don't check UserRights
Optional.ofNullable(context.getMethod().getAnnotation(PermitAll.class)).orElseGet(context.getTarget().getClass().getAnnotation(PermitAll.class)).orElse(null);
if (permitAll != null) {
	return context.proceed();
}
```

sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowedInterceptor.java
```diff
@@ -0,0 +1,98 @@
...
+				}
+
+				boolean hasAnyRight = false;
+				for (String right : rightsAllowed.value()) {
+					if (UserRight._SYSTEM.equals(right)) {
+						if (sessionContext.isCallerInRole(right)) {
+							hasAnyRight = true;
+							break;
+						}
+					} else if (currentUserService.hasUserRight(userRightValues.get(right))) { // if (sessionContext.isCallerInRole(right)) does not work - see #9559
```
StefanKock, line 84: Move comment into the line before, no mix of code and comment. And the comment is wrong. Technically it does work, but is to slow.

sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowedInterceptor.java
```diff
@@ -0,0 +1,98 @@
...
+					}
+				}
+
+				if (!hasAnyRight) {
+					logger.error("Missing user rights: " + String.join(", ", rightsAllowed.value()));
+					throw new AccessDeniedException(I18nProperties.getString(Strings.errorForbidden));
+				}
+			}
+		}
+		return context.proceed();
```
StefanKock, line 96: I think the implizit fall-through for @DenyAll and no annotation should mentioned in a code comment here.

sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowed.java
```diff
@@ -0,0 +1,19 @@
...
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Documented
+@Retention(RUNTIME)
+@Target({TYPE, METHOD})
+public @interface RightsAllowed {
```
StefanKock, line 14: Missing comment that this class mimics `@RolesAllowed` to avoid performance problemns

sormas-backend/pom.xml
```diff
@@ -39,6 +39,11 @@
 			<artifactId>commons-validator</artifactId>
 		</dependency>
 
+		<dependency>
```
StefanKock, line 42: Why this dependency import?

StefanKock (changes requested):
Added some comments. Additionally I think the feature implementation should be splitted from replacing the annotation. That makes review easier and is better to cherry-pick (only causes merge conflict on refactoring part but not on feature part).

sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowedInterceptor.java
```diff
@@ -0,0 +1,98 @@
...
+		PermitAll permitAll = context.getMethod().getAnnotation(PermitAll.class);
+		if (permitAll == null) {
+			permitAll = context.getTarget().getClass().getAnnotation(PermitAll.class);
+		}
+
+		if (permitAll == null) {
+
+			// for DenyAll, we continue to use the existing mechanisms
+
+			RightsAllowed rightsAllowed = context.getMethod().getAnnotation(RightsAllowed.class);
```
StefanKock, line 60: Use this:
```java
RightsAllowed rightsAllowed = Optional.ofNullable(context.getMethod().getAnnotation(RightsAllowed.class)).orElseGet(context.getTarget().getClass().getAnnotation(RightsAllowed.class)).orElse(null);
```

sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowedInterceptor.java
leventegal-she: Being a custom implementation I think for @PermitAll would be nice to check that at least there is a current user so we don't allow access without an authenticated user unless we don't have public features.

Or we could introduce our own @PermitAll like annotation with user restriction and use that one instead of the default one.

StefanKock: Rethinking this: The current implementation gives priority of `@PermitAll` on class without checking `@RightsAllowed` on method. That not how Jakarta EE works.

StefanKock: > Being a custom implementation I think for @PermitAll would be nice to check that at least there is a current user so we don't allow access without an authenticated user unless we don't have public features.

I disagree: We also need `@PermitAll` for SYSTEM itself (so no user at all).

leventegal-she: > I disagree: We also need `@PermitAll` for SYSTEM itself (so no user at all).

For SYSTEM we use RightsAllowed(_SYSTEM). That I think will not work for this implementation @MartinWahnschaffe

MartinWahnschaffe: > > I disagree: We also need `@PermitAll` for SYSTEM itself (so no user at all).
> 
> For SYSTEM we use `@RightsAllowed(_SYSTEM)`. That I think will not work for this implementation @MartinWahnschaffe

Yes, it does work. for _SYSTEM the method checks against sessionContext.isCallerInRole

sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowedInterceptor.java
MartinWahnschaffe: The comment is correct. isCallerInRole does not work, because you would have to use @DeclareRoles or @RolesAllowed or any other javax annotation to make sure the roles are correctly initialized here

sormas-backend/pom.xml
MartinWahnschaffe: Forgot to remove it after settling with the right exception class

StefanKock:
I'm currently reviewing the rewritten commits and fixing some flaws. I also want to check if we can add a test that `@RolesAllowed` is not used anymore.

sormas-backend/src/main/java/de/symeda/sormas/backend/util/RightsAllowedInterceptor.java
```diff
@@ -0,0 +1,113 @@
...
+
+		if (context.getMethod().getAnnotation(PermitAll.class) != null) {
+			// method permits
+			return context.proceed();
+		}
+
+		RightsAllowed rightsAllowed = context.getMethod().getAnnotation(RightsAllowed.class);
+		if (rightsAllowed == null) {
+
+			if (context.getMethod().getAnnotation(PermitAll.class) != null) {
```
StefanKock, line 61: Bug: Check on class for `@PermitAll`

StefanKock (approved):
- changed some comments
- adjusted logging level to debug (reduce [log-and-throw](https://phpopendocs.com/learning/java_exception_antipatterns#log-and-throw) pattern)
- fixed bug: check class `@PermitAll`

I did not work on adding an architecture test that `@RolesAllowed` is not used anymore. I checked that it's not used in `sormas-backend`, but still in `sormas-rest`.

MateStrysewske (approved):

StefanKock:
`StartupShutdownServiceTest.testHistoryTablesMatch:121` fails here again. Has to be ignored probably?

PR #9614, opened 2022-06-24
9577 replace roles allowed

### LINKED ISSUE #9577: Replace @RolesAllowed by own @RightsAllowed annotation
### Problem Description

Proven performance problem (#9559) when checking `javax.annotation.security.@RolesAllowed`

### Proposed Change

Replace `@RolesAllowed` by own `@RightsAllowed` annotation and check with interceptor.

### Acceptance Criteria

- [x] `@PermitAll` and `@DenyAll` is also considered.
- [x] UserRight._SYSTEM is also considered (e.g. completeness calculation of case is working)
- [x] Add archUnit test to make sure `@RolesAllowed` is not used in backend (in sormas-rest it's fine)

MartinWahnschaffe:
Fixes #9577

### DIFF
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/ArchitectureTest.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/ArchitectureTest.java
@@ -11,6 +11,7 @@ import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
 import javax.validation.constraints.NotNull;
 
+import de.symeda.sormas.backend.util.RightsAllowed;
 import org.junit.runner.RunWith;
 
 import com.tngtech.archunit.base.DescribedPredicate;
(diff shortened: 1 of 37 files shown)


sormas-backend/src/test/java/de/symeda/sormas/backend/ArchitectureTest.java
```diff
@@ -291,6 +292,14 @@ private void assertFacadeEjbAnnotated(Class<?> facadeEjbClass, AuthMode authMode
 
 			methodChecks.check(classes);
 		}
+
+		// don't use RolesAllowed anymore, until performance problem is fixed
```
StefanKock, line 296: This code block only checks EJBs that are given by a test. I would rather check all classes in the module annotated with `@Singleton`/`@Stateless` that there is no usage of `@RolesAllowed`.

StefanKock (changes requested):
Looks good to me so far, but the check for `@RolesAllowed` should be more general in my opinion.

sormas-backend/src/test/java/de/symeda/sormas/backend/ArchitectureTest.java
MartinWahnschaffe: Guess there is no harm in simply not allowing it for all classes and methods in the package.

StefanKock (approved):
Checking for not using `@RolesAllowed` looks way better and easier than I expected.

PR #9668, opened 2022-06-29
#9623 - Validate age groups gracefully

### LINKED ISSUE #9623: Conversion of age groups should not throw exception [0.5]
### Problem Description
Currently loading the disease configuration during startup the server does not start when there is an invalid age group defined for a disease, because the used converted validates the age group which throws an exception.

### Proposed Change

- [x] The converter should still validate each age group, but should catch the exception, log a warning and ignore the invalid age group

This will also make sure that the mSers module in the UI and app does not throw exceptions, but instead just doesn't show invalid age groups.

MateStrysewske:
Fixes #9623

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/utils/AgeGroupUtils.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/utils/AgeGroupUtils.java
@@ -87,7 +87,7 @@ public class AgeGroupUtils {
 			return null;
 		}
 
-		final List<String> ageGroupList = Stream.of(ageGroupsString.trim().split(",")).collect(Collectors.toList());
+		final List<String> ageGroupList = Stream.of(ageGroupsString.split(",")).map(String::trim).collect(Collectors.toList());
 		ageGroupList.removeIf(s -> {
 			try {
 				validateAgeGroup(s);
--- a/sormas-api/src/main/java/de/symeda/sormas/api/utils/AgeGroupUtils.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/utils/AgeGroupUtils.java
@@ -15,6 +15,7 @@
 
 package de.symeda.sormas.api.utils;
 
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/utils/AgeGroupUtils.java
```diff
@@ -75,38 +83,46 @@ private static String getAgeGroupTimeUnitCaption(String ageGroupTimeUnit) {
...
-			ageGroupList.forEach(s -> validateAgeGroup(s));
-			return ageGroupList;
-		} catch (IllegalArgumentException e) {
-			throw new RuntimeException(
-				"Content of ageGroupsString is not a valid list of ageGroups, or one of the ageGroups in the String does not match a valid ageGroup: "
-					+ ageGroupsString);
-		}
+		final List<String> ageGroupList = Stream.of(ageGroupsString.split(",")).collect(Collectors.toList());
+		final List<String> invalidAgeGroupsList = new ArrayList<>();
+		ageGroupList.forEach(s -> {
```
MartinWahnschaffe, line 93: I'd suggest to user removeIf: https://www.geeksforgeeks.org/arraylist-removeif-method-in-java/

MartinWahnschaffe (approved):

MartinWahnschaffe (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/utils/AgeGroupUtils.java
```diff
@@ -75,38 +82,46 @@ private static String getAgeGroupTimeUnitCaption(String ageGroupTimeUnit) {
...
-		try {
-			final List<String> ageGroupList = Stream.of(ageGroupsString.split(",")).collect(Collectors.toList());
-			ageGroupList.forEach(s -> validateAgeGroup(s));
-			return ageGroupList;
-		} catch (IllegalArgumentException e) {
-			throw new RuntimeException(
-				"Content of ageGroupsString is not a valid list of ageGroups, or one of the ageGroups in the String does not match a valid ageGroup: "
-					+ ageGroupsString);
-		}
+		final List<String> ageGroupList = Stream.of(ageGroupsString.trim().split(",")).collect(Collectors.toList());
```
MartinWahnschaffe, line 90: The trimming has to be done very every single age group.  E.g. "D0_D30, D31_M2" should be a valid age groups string.

MartinWahnschaffe (approved):

PR #9678, opened 2022-06-29
#9677 keep fhircontext instance

### LINKED ISSUE #9677: Cache FhirContext to improve performance of logging
### Problem Description
The FhirContext is currently created for every entry in the audit log (see AuditLoggerEjb.accept).
This is expensive from a performance point of view: https://hapifhir.io/hapi-fhir/apidocs/hapi-fhir-base/ca/uhn/fhir/context/FhirContext.html

> Performance: This class is expensive to create, as it scans every resource class it needs to parse or encode to build up an internal model of those classes. For that reason, you should try to create one FhirContext instance which remains for the life of your application and reuse that instance. Note that it will not cause problems to create multiple instances (ie. resources originating from one FhirContext may be passed to parsers originating from another) but you will incur a performance penalty if a new FhirContext is created for every message you parse/encode. 

### Proposed Change

- [x] Keep the instance within the AudtiLoggerEjb instead of creating it again and again.

We could even use a static instance - the FhirContext is thread-safe. Since we are already using a Singleton here I'd still go with an instance field.

### Acceptance Criteria

* AuditLog is still working as before

MartinWahnschaffe:
Fixes #9677


JonasCir (approved):
Good catch thanks

Issue #8674, opened 2022-04-01
Person can be shared via S2S that cannot be updated afterwards 


JonasCir:
### Bug Description
This case https://release-sormas-x.sormas.netzlink.com/sormas-ui/#!cases/data/SPUZDY-CDXBOO-XDKHIO-2C6V2A3M
was successfully shared, however, updating it (e.g., changing the birthday) won't be synchronized to the other release instance as the server will receive a bad request from the remote.

The reported error in the receiver side (i.e., which returns the 400 status code), we see thanks to #8548 the following log:

```
Formatting the human readable message failed. I18nTag: sormasToSormasSaveException, arguments: [Sie müssen mindestens eine Einrichtungsart für die Adresse angeben, für die Sie die Ansprechpartner Kontaktdaten angegeben haben.] ([de.symeda.sormas.api.sormastosormas.SormasToSormasI18nMessageError.java:50](http://de.symeda.sormas.api.sormastosormas.sormastosormasi18nmessageerror.java:50/))
```

There are two bugs here:
1. `sormasToSormasSaveException` is used in `ValidationHelper.handleValidationError` however, it contains two `%s` format strings in the German version (I did not check the others), however, only the one argument is provided which causes the formatting to fail.
2. The failed validation string is `importPersonContactDetailsWithoutFacilityType` which is used [here](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java#L628) in `PersonFacadeEjb`. Somehow it is possible to share a case successfully with another instance via S2S, however, when we sync changes, the validation kicks in when the remote tries to call back the sharer wants to get back with us.
See this log: [log.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/8398487/log.txt)

Probably this can be fixed by mirroring the validation logic in S2S validation before sending? But duplication is in general a bad idea, so maybe there is away to reuse it. 

### Steps to Reproduce
This is the exported case causing the problem [spudzy_repro.csv](https://github.com/hzi-braunschweig/SORMAS-Project/files/8398422/spudzy_repro.csv). Share it  with excluding personal data and observe that updating it (e.g., changing the birthday) won't be synchronized to the other release instance as the server will receive a bad request from the remote.

### Expected Behavior
Error formatting works correct and synchronization works/invalid cases are prevented from being shared in the first place

### System Details
[...]

PR #9687, opened 2022-06-30
[DO NOT MERGE] Bug/8674 share update validation

JonasCir:
Yet another part of #8674


JonasCir:
something is not working...

PR #9696, opened 2022-06-30
[#8674] validate shares after building them

### LINKED ISSUE #8674: Person can be shared via S2S that cannot be updated afterwards

JonasCir:
Fixes #8674

This PR ensures that the validation logic of the EJBs runs after the shares have been built to ensure that the remote instance can process them correctly (recall that the save method on the remote will call `validate` as well, so we avoid a network round trip other problems as well).

The underlying issue is that our pseudonymizers can return data which does not pass the `validate` method of the EJBs. I made  some adjustments and I would appreciate @MateStrysewske to comment on the changes to
* LocationDto.java (more PersonalData )
* ShareDataBuilderHelper (i.e., should a valid `ContactDto` contain a reporting user?)
* SormasToSormasContactFacadeEjbTest: same w.r.t. `ContactDto`

I'm quite certain that more pseudonymizers show this behaviour and I'm going to file an issue about it tomorrow.

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/ShareDataBuilder.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/ShareDataBuilder.java
@@ -58,8 +58,7 @@ public abstract class ShareDataBuilder<DTO extends SormasToSormasShareableDto, A
 
 	protected abstract PREVIEW doBuildShareDataPreview(ADO data, ShareRequestInfo requestInfo);
 
-	public PREVIEW buildShareDataPreview(ADO data, ShareRequestInfo requestInfo)
-		throws SormasToSormasValidationException, ValidationRuntimeException {
+	public PREVIEW buildShareDataPreview(ADO data, ShareRequestInfo requestInfo) throws SormasToSormasValidationException {
 		PREVIEW shared = doBuildShareDataPreview(data, requestInfo);
 		ValidationErrors errors = validator.validateOutgoingPreview(shared);
 		if (errors.hasError()) {
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/ShareDataBuilder.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/ShareDataBuilder.java
@@ -22,6 +22,7 @@ import de.symeda.sormas.api.sormastosormas.SormasToSormasEntityDto;
 import de.symeda.sormas.api.sormastosormas.SormasToSormasShareableDto;
 import de.symeda.sormas.api.sormastosormas.validation.SormasToSormasValidationException;
 import de.symeda.sormas.api.sormastosormas.validation.ValidationErrors;
+import de.symeda.sormas.api.utils.ValidationRuntimeException;
 import de.symeda.sormas.api.utils.pseudonymization.PseudonymizableDto;
 import de.symeda.sormas.backend.sormastosormas.data.validation.SormasToSormasDtoValidator;
 import de.symeda.sormas.backend.sormastosormas.entities.SormasToSormasShareable;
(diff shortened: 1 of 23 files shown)


MateStrysewske:
@JonasCir Not gonna approve/disapprove this PR, just commenting on what you wanted me to have a look at:

- Adding the annotations to the fields in LocationDto is correct
- Contacts requiring a reporting user is correct as well

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/ShareDataBuilder.java
```diff
@@ -51,6 +64,8 @@ public SHARED buildShareData(ADO data, ShareRequestInfo requestInfo) throws Sorm
 		return shared;
 	}
 
+	protected abstract void validateWithEjbShared(SHARED shared) throws ValidationRuntimeException;
```
leventegal-she, line 67: I think `doBusinessValidation` would be a better name

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/ShareDataBuilderHelper.java
```diff
@@ -98,7 +98,8 @@ public void pseudonymiePerson(PersonDto personDto, boolean pseudonymizedPersonal
 	public ContactDto getContactDto(Contact contact, Pseudonymizer pseudonymizer) {
 		ContactDto contactDto = contactFacade.convertToDto(contact, pseudonymizer);
 
-		contactDto.setReportingUser(null);
+		// reporting user is not set to null here as it would not pass the validation
```
leventegal-she, line 101: Please create a ticket to cleanup the reporting user validation.

It's strange that only contact validation fails if it has no reporting user.

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/ShareDataBuilder.java
JonasCir: updated

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/ShareDataBuilderHelper.java
JonasCir: filed #9701

leventegal-she (approved):

JonasCir:
Hm test fails

leventegal-she (approved):

PR #9704, opened 2022-07-01
sorqa-358

pk-sgent:
image


mp-sgent (approved):

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CreateNewCaseSteps.java
```diff
@@ -937,6 +940,16 @@ public CreateNewCaseSteps(
           webDriverHelpers.waitUntilElementIsVisibleAndClickable(POINT_OF_ENTRY_REGION_BUTTON);
           webDriverHelpers.waitUntilElementIsVisibleAndClickable(POINT_OF_ENTRY_DISTRICT_BUTTON);
         });
+
+    When(
+        "I check if collected case UUID is equal with current",
+        () -> {
+          softly.assertEquals(
+              casesUUID.get(0),
```
rdutu-vg, line 948: why we need a list if we always collect one value?

rdutu-vg (approved):

sormas-e2e-tests/src/test/java/org/sormas/e2etests/steps/web/application/cases/CreateNewCaseSteps.java
pk-sgent: This list is used in some previous test. I reused it without create a new variable to that.

PR #9706, opened 2022-07-01
[#8674] recycle missing refactoring

### LINKED ISSUE #8674: Person can be shared via S2S that cannot be updated afterwards

JonasCir:
Fixes #8674 
@leventegal-she I had some quite nice refactoring laying around, but I decided against adding it to the hotfix PR. But still a win in my opinion, so I decided to recycle my work.

* `validateEntitiesBeforeSend` is now unified in the super class
* `validateEntitiesBeforeShare` contains now the common code path and type specific checks are implemented in sub class

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/sample/SampleShareDataBuilder.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/sample/SampleShareDataBuilder.java
@@ -76,7 +76,7 @@ public class SampleShareDataBuilder
 
 	@Override
 	public void doBusinessValidation(SormasToSormasSampleDto sormasToSormasSampleDto) throws ValidationRuntimeException {
-		sampleFacade.validate(sormasToSormasSampleDto.getEntity());
+		sampleFacade.validate(sormasToSormasSampleDto.getEntity(), true);
 		sormasToSormasSampleDto.getPathogenTests().forEach(pathogenTestFacade::validate);
 		// additional test facade has no validation method
 	}


JonasCir:
@MateStrysewske Just FYI included the fix for failing CI in this aswell :)

leventegal-she (approved):

PR #9707, opened 2022-07-01
Feature 9068 aggregate reports summarize change

### LINKED ISSUE #9068: Don't summarize aggregated report numbers with those of sub jurisdictions in the aggregate report view [3]
### Problem Description
The numbers shown in the aggregate report view always aggregate anything that lies within the filtered/grouped jurisdiction. This means a reported created for a district would be summed up together with reports for facilities in that district.
This is borderline wrong and unintuitive for users. 

### Proposed Change

- [x] No longer combine reports of different jurisdiction levels in the report view of the UI. E.g. when data is aggregated on district level, the system will rely on the reports directly created for the district and only fall-back to aggregating facility-level reports, when there is no report for the district.
- [x] When there are duplicate entries for the exact same jursidiction (e.g. facility) pick the younger one (latest changeDate). A duplicate detection will be added in a separate issue.
- [x] Add a reporting user column to the grid (AggregateCaseCountDto). It should display the user for all rows that were directly reported and are not aggregated. So for facilities and poe it will always show a user, for district, region and whole instance only if there is a direct report for that jurisdiction.

### Implementation Suggestion

1. Query all reports that lie within the filter criteria AND have a jurisidiction below or equal to the grouping level. E.g. when data is filtered to one region and grouped by district, query all reports that are associated to the region AND have at least a district defined. This means reports created directly for the region itself would not be a part of this.
2. The results should be sorted by jurisidiction - starting with those that don't have a sub-jursidiction defined. E.g. A report for the district itself would be in the list before a report of a facility in the district.
3. Go through the reports to aggregate them based on the grouping. 
  * When a district is covered by a direct report for that district ignore all reports for sub-jurisdictions
  * When a district is not covered, summarize all reports for facilities and poes

If the grouping is done by region or even the whole instance, the same needs to happen: When no direct region report exists, summarize all direct district reports. If for one of those districts no direct report exists, summarize all facility and poe reports.

Note: All of this also has to take into account grouping by epic weeks. To achieve this I'd suggest to simply repeat this process for each epi week.

Alternative: Try to directly handle the aggregation with SQL queries or at least to filter our reports that are not needed because we already have a report for the parent jurisdiction.

BarnaBartha:
Fixes #9068


sormas-api/src/main/java/de/symeda/sormas/api/report/AggregatedCaseCountDto.java
```diff
@@ -291,39 +222,181 @@ public void setEpiWeek(Integer epiWeek) {
...
+		if (regionId == null && that.regionId != null) {
+			return true;
+		}
+		if (districtId == null && that.districtId != null) {
+			return true;
+		}
+		if (healthFacilityId == null && that.healthFacilityId != null && pointOfEntryId == null && that.pointOfEntryId == null) {
+			return true;
+		}
+		if (pointOfEntryId == null && that.pointOfEntryId != null && healthFacilityId == null && that.healthFacilityId != null) {
```
MateStrysewske, line 348: I think the last comparison should be a `==` here

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportFacadeEjb.java
```diff
@@ -185,42 +183,147 @@ public List<AggregatedCaseCountDto> getIndexList(AggregateReportCriteria criteri
...
 
-		List<AggregatedCaseCountDto> resultList = em.createQuery(cq).getResultList();
+		List<AggregatedCaseCountDto> queryResult = em.createQuery(cq).getResultList();
 		Map<Disease, AggregatedCaseCountDto> reportSet = new HashMap<>();
 
+		List<AggregatedCaseCountDto> resultList = new ArrayList<>();
+		Map<AggregatedCaseCountDto, List<AggregatedCaseCountDto>> subJurisdictionsToBeMerged = new HashMap<>();
+
+		final AggregateReportGroupingLevel finalGroupingLevel = groupingLevel;
+		for (AggregatedCaseCountDto aggregatedCaseCountDto : queryResult) {
```
MateStrysewske, line 261: I'd like you to add comments to this whole section because it's really tough to read. Also, please consider using stream methods like `removeIf` and potentially even `reduce` if possible in order to shorten the code and make it more readable.

MateStrysewske (changes requested):
Your IDE is not set up correctly. Files are not properly formatted. Please ensure that you have the auto formatter AND save actions set up, with save actions configured to format your files on save.

MateStrysewske (approved):
LGTM. Have requested a review from @MartinWahnschaffe as well to verify that this properly covers everything that was requested in the issue. I believe it does, but maybe I missed something/misinterpreted something.

sormas-api/src/main/java/de/symeda/sormas/api/report/AggregatedCaseCountDto.java
```diff
@@ -291,39 +237,181 @@ public void setEpiWeek(Integer epiWeek) {
...
+			year,
+			epiWeek,
+			disease,
+			newCases,
+			labConfirmations,
+			deaths,
+			ageGroup);
+	}
+
+	public boolean similar(Object o, AggregateReportGroupingLevel groupingLevel) {
```
MartinWahnschaffe, line 300: Why not use AggregateCaseCountDto instead of Object? This is not an equals method anyway.

sormas-api/src/main/java/de/symeda/sormas/api/report/AggregatedCaseCountDto.java
```diff
@@ -291,39 +237,181 @@ public void setEpiWeek(Integer epiWeek) {
...
+		if (o == null || getClass() != o.getClass())
 			return false;
-		if (labConfirmations != other.labConfirmations)
+		AggregatedCaseCountDto that = (AggregatedCaseCountDto) o;
+		boolean similar = year == that.year && epiWeek == that.epiWeek && disease == that.disease && Objects.equals(ageGroup, that.ageGroup);
+		if (groupingLevel != null) {
+			switch (groupingLevel) {
+
+			case REGION:
+				similar = similar && Objects.equals(regionName, that.regionName) && Objects.equals(regionId, that.regionId);
```
MartinWahnschaffe, line 312: I'd only compare the ID, not the name.

sormas-api/src/main/java/de/symeda/sormas/api/report/AggregatedCaseCountDto.java
```diff
@@ -291,39 +237,181 @@ public void setEpiWeek(Integer epiWeek) {
...
+		return similar;
+	}
+
+	public boolean equalJurisdiction(Object o) {
+		if (this == o)
+			return true;
+		if (o == null || getClass() != o.getClass())
 			return false;
-		if (newCases != other.newCases)
+		AggregatedCaseCountDto that = (AggregatedCaseCountDto) o;
```
MartinWahnschaffe, line 335: Same here: Only compare id

sormas-api/src/main/java/de/symeda/sormas/api/report/AggregatedCaseCountDto.java
```diff
@@ -291,39 +237,181 @@ public void setEpiWeek(Integer epiWeek) {
...
-		if (!Objects.equals(ageGroup, other.ageGroup))
+		AggregatedCaseCountDto that = (AggregatedCaseCountDto) o;
+
+		if (regionId == null && that.regionId != null) {
+			return true;
+		}
+		if (districtId == null && that.districtId != null) {
+			return true;
+		}
+		if (healthFacilityId == null && that.healthFacilityId != null && pointOfEntryId == null && that.pointOfEntryId == null) {
```
MartinWahnschaffe, line 360: Probably easier to understand when combining both if statements to this one:

```suggestion
		if (healthFacilityId == null && pointOfEntryId == null && (that.healthFacilityId != null || that.pointOfEntryId != null)) {
```

sormas-api/src/main/java/de/symeda/sormas/api/report/AggregatedCaseCountDto.java
```diff
@@ -291,39 +237,181 @@ public void setEpiWeek(Integer epiWeek) {
...
+	public boolean sameJurisdictionLevel(Object o) {
+
+		if (this == o)
+			return true;
+		if (o == null || getClass() != o.getClass())
 			return false;
-		return true;
+		AggregatedCaseCountDto that = (AggregatedCaseCountDto) o;
+
+		if (regionId == null && that.regionId == null) {
```
MartinWahnschaffe, line 377: What about:
```suggestion
		if (regionId == null || that.regionId == null) {
		return regionId == null && that.regionId == null;
		}
```
Thisway you can completely ignore the region below. Same for the district in the next if.

MartinWahnschaffe: Maybe even easier: Use the getJursidictionLevel method below and compare the results

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportFacadeEjb.java
```diff
@@ -312,6 +353,87 @@ public List<AggregatedCaseCountDto> getIndexList(AggregateReportCriteria criteri
...
+	private List<AggregatedCaseCountDto> summarizeAggregateData(
+		final AggregateReportGroupingLevel finalGroupingLevel,
+		List<AggregatedCaseCountDto> queryResult) {
+		final List<AggregatedCaseCountDto> resultList = new ArrayList<>();
+		final Map<AggregatedCaseCountDto, List<AggregatedCaseCountDto>> reportsToBeSummed = new HashMap<>();
+
+		// extract relevant aggregate data and prepare what needs to be sumed
+		for (AggregatedCaseCountDto dto : queryResult) {
+
+			final Optional<AggregatedCaseCountDto> optionalDto = resultList.stream().filter(a -> dto.similar(a, finalGroupingLevel)).findAny();
```
MartinWahnschaffe, line 365: Would be faster to sort the queryresults by jurisdiction and only check the last resultList entry for similarity instead of always going through the whole list.
We aren't working with too many results here, so it's probably not worth to rewrite this, but maybe worth a comment so that if needed we can easily optimize the performance here.

MartinWahnschaffe (changes requested):

MartinWahnschaffe:
Works for me - just a few cosmetic changes to the code.

MartinWahnschaffe (approved):

MartinWahnschaffe:
@MateStrysewske if all checks pass this can be merged.

PR #9725, opened 2022-07-04
#9119 - Allow editing and deletion of aggregate report data

### LINKED ISSUE #9119: Allow editing and deletion of aggregate report data [3]
### Feature Description

Currently the mSERS module for aggregated case reporting only provides a view that aggregates the data for a certain jurisdiction. There is currently no way to edit and delete the underlying reports.

#9073 introduces a view to list the actual data that was reported and to detect duplicates.

### Proposed Change

- [x] Add a column to edit aggregate reports
- [x] Important to understand: When creating a report, every row in the creation form is an AggregateReportDto (disease, age group, jurisdiction, numbers). So in theory when the user would click edit for one row in the table, it would only show a small popup that allows to edit or delete the one row of the original creation form.
- [x] Instead of doing it like that, when a user tries to edit one aggregate report entry in the grid, we should show the edit form with all aggregate report entries the reporting user of that entry has created for the same epi week and jurisdiction entity.
- [x] Thisway the user has the same experience as when creating a new report
- [x] When there are multiple entries created by the same user for the same epi week, etc. use the youngest one in the edit form
- [x] Within the edit form allow deletion of all aggregate reports shown in the form (not the grid). There is no dedicated user right for this, because editing and setting the values to 0 would also result in a deletion of the entity in the backend.
- [x] Add a delete column that provides the option to delete a row, but only for duplicates

sergiupacurariu:
Fixes #9119


sormas-api/src/main/java/de/symeda/sormas/api/report/AggregateReportFacade.java
```diff
@@ -23,7 +23,11 @@ public interface AggregateReportFacade {
 
 	List<AggregateReportDto> getList(AggregateReportCriteria criteria);
 
+	List<AggregateReportDto> getAllAggregatedReportsFromSameEpiWeekUserAndJurisdiction(AggregateReportDto aggregateReportDto);
+
 	void deleteReport(String reportUuid);
 
+	void deleteAggregatedReports(List<String> uuids);
```
MateStrysewske, line 30:
```suggestion
	void deleteAggregateReports(List<String> uuids);
```

sormas-api/src/main/java/de/symeda/sormas/api/report/AggregateReportFacade.java
```diff
@@ -23,7 +23,11 @@ public interface AggregateReportFacade {
 
 	List<AggregateReportDto> getList(AggregateReportCriteria criteria);
 
+	List<AggregateReportDto> getAllAggregatedReportsFromSameEpiWeekUserAndJurisdiction(AggregateReportDto aggregateReportDto);
```
MateStrysewske, line 26:
```suggestion
	List<AggregateReportDto> getSimilarAggregateReports(AggregateReportDto aggregateReportDto);
```

The current name is too long in my eyes, I'd rather like you to add Javadoc to explain what the method uses to compare the aggregate reports.

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportFacadeEjb.java
```diff
@@ -445,6 +445,54 @@ public List<AggregateReportDto> getList(AggregateReportCriteria criteria) {
...
+		criteria.setRegion(aggregateReportDto.getRegion());
+		criteria.setDistrict(aggregateReportDto.getDistrict());
+		criteria.setHealthFacility(aggregateReportDto.getHealthFacility());
+		criteria.setPointOfEntry(aggregateReportDto.getPointOfEntry());
+		criteria.setEpiWeekFrom(new EpiWeek(aggregateReportDto.getYear(), aggregateReportDto.getEpiWeek()));
+		criteria.setEpiWeekTo(new EpiWeek(aggregateReportDto.getYear(), aggregateReportDto.getEpiWeek()));
+		criteria.setReportingUser(aggregateReportDto.getReportingUser());
+		criteria.setForceJurisdictionCheck(true);
+
+		List<AggregateReportDto> report = getAggregateReports(criteria);
```
MateStrysewske, line 461:
```suggestion
		List<AggregateReportDto> reports = getAggregateReports(criteria);
```

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsEditLayout.java
```diff
@@ -471,6 +490,13 @@ private boolean isValid() {
 		return valid;
 	}
 
+	private void deleteAggregateReportForm() {
```
MateStrysewske, line 493: Method name is not clear to me; we don't delete a form here, do we?

MateStrysewske (changes requested):
- Please use icons instead of buttons as in e.g. the user directory
- Edit window also shows report entries created by other users, not only by the reporting user of the entry that was clicked on

sormas-api/src/main/java/de/symeda/sormas/api/report/DiseaseAgeGroup.java
```diff
@@ -1,43 +1,45 @@
...
 
 import java.io.Serializable;
 import java.util.Objects;
 
+import de.symeda.sormas.api.Disease;
+
 public class DiseaseAgeGroup implements Serializable {
 
-	private String disease;
+	private Disease disease;
```
MartinWahnschaffe, line 25: Thanks for changing this :-)

sormas-api/src/main/java/de/symeda/sormas/api/report/AggregateReportFacade.java
sergiupacurariu: changed

sormas-api/src/main/java/de/symeda/sormas/api/report/AggregateReportFacade.java
sergiupacurariu: changed

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportFacadeEjb.java
sergiupacurariu: changed

sergiupacurariu:
both edit and delete buttons have their icons borders removed
the criteria filter has been updated to filter by Reporting User

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsEditLayout.java
```diff
@@ -251,10 +258,16 @@ public AggregateReportsEditLayout(Window window, AggregateReportCriteria criteri
...
 		legend.addStyleName(CssStyles.VSPACE_TOP_1);
 
-		editForms.sort((e1, e2) -> e1.getDisease().compareTo(e2.getDisease()));
-		for (AggregateReportEditForm editForm : editForms) {
-			addComponent(editForm);
-		}
+		editForms.stream()
+			.sorted(
+				Comparator.comparing(AggregateReportEditForm::getDisease, Comparator.nullsFirst(Comparator.comparing(Disease::toString)))
+					.thenComparing(
```
MartinWahnschaffe, line 264: I think AgeGroupUtils.getComparator can be used here.

sormas-api/src/main/java/de/symeda/sormas/api/report/AggregateReportCriteria.java
```diff
@@ -144,4 +149,21 @@ public Boolean getShowOnlyDuplicates() {
...
+	public UserReferenceDto getReportingUser() {
+		return reportingUser;
+	}
+
+	public void setReportingUser(UserReferenceDto reportingUser) {
+		this.reportingUser = reportingUser;
+	}
+
+	@IgnoreForUrl
+	public Boolean isForceJurisdictionCheck() {
```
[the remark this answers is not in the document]
sergiupacurariu: I renamed the variable, seems more suggestive

sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/AggregateReportsEditLayout.java
sergiupacurariu: changed

PR #9729, opened 2022-07-05
#9726 password hash length

### LINKED ISSUE #9726: Investigate inconsistent generated passwords
### Problem Description

When generating passwords with PostgreSQL it happens that password hashes are generated that are not accepted by SORMAS to be valid.

```sql

-- Mechanism in CI script to reset the password to "QTj]qF90U~-CMLa/"
SELECT encode(digest(CONCAT('QTj]qF90U~-CMLa/', seed), 'sha256'), 'hex') from users where username = 'myuser';

-- Not working password (hash) with seed
UPDATE users set password = '0daf0f39b9f32cc81462f8c4a9f12ac96552474769768aa6811f672692945d98', seed = 'eqQ8HaXBkmjqqBgt' WHERE username = 'myuser';

-- Working password (hash) with seed
UPDATE users set password = '35472f3b451e136556597ec82c16236c85a77dd1718173f2bedd1b1ab3737277', seed = 'xPQCk3J5psSOkLcp' WHERE username = 'myuser';

```

### Proposed Solution

Check what the Java digest methods return as result for that given examples.

StefanKock:
Fixes #9726


sormas-api/src/main/java/de/symeda/sormas/api/utils/PasswordHelper.java
```diff
@@ -93,7 +94,12 @@ public static String encodePassword(String password, String seed) {
 		}
 	}
 
-	public static String hexEncode(byte[] aData) {
-		return new BigInteger(1, aData).toString(16);
+	public static String hexEncode(byte[] data) {
+
+		StringBuilder hex = new StringBuilder(2 * data.length);
```
MartinWahnschaffe, line 99: would suggest to use a char array instead of a string builder here. See https://stackoverflow.com/a/9855338 or implementation of apache commons Hex: https://commons.apache.org/proper/commons-codec/apidocs/src-html/org/apache/commons/codec/binary/Hex.html

MartinWahnschaffe (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/utils/PasswordHelper.java
StefanKock: I replaced the function, but not with the one from commons-codec. ~~We already have it as dependency, but not (yet) in the android-app.~~ I doublechecked, it's already part of the android-app, so I could just use that implementation. @MartinWahnschaffe what's your opinion?

sormas-api/src/test/java/de/symeda/sormas/api/utils/PasswordHelperTest.java
```diff
@@ -50,4 +53,44 @@ public void testCreatePass() {
...
+
+	@Test
+	public void testEncodePasswordWithValidHashLength() {
+
+		String password;
+		String seed;
+
+		password = "QTj]qF90U~-CMLa/";
+
+		for (int i = 1; i <= 1_024; i++) {
```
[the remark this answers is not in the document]
StefanKock: I added a comment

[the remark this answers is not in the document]
MartinWahnschaffe: Ok, thought there is something special about 1024 :-D

sormas-api/src/main/java/de/symeda/sormas/api/utils/PasswordHelper.java
MartinWahnschaffe: If it's already a dependency I'm in favor of using that method.

StefanKock: I switched now to commons-codec and removed the extensive testing (don't retest dependencies), I just keep the test that proves the recent problem at hand to be removed.

MartinWahnschaffe (approved):

PR #9743, opened 2022-07-06
#9073 Don't use primary for non-case-based diseases

### LINKED ISSUE #9073: Add a view to list aggregate report data and to highlight duplicates [3]

MartinWahnschaffe:
Fixes #9073

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/report/aggregate/AggregateReportsFragment.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/report/aggregate/AggregateReportsFragment.java
@@ -212,7 +212,7 @@ public class AggregateReportsFragment extends BaseReportFragment<FragmentReports
 			.collect(Collectors.toList());
 		DatabaseHelper.getAggregateReportDao().sortAggregateReports(userReports);
 
-		List<Disease> diseaseList = DiseaseConfigurationCache.getInstance().getAllDiseases(true, false, false);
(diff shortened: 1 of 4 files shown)


MateStrysewske (approved):

MateStrysewske (changes requested):
Tests are failing

PR #9764, opened 2022-07-07
#9415 Create/Update/Delete of vaccination/immunization should be considered as change of a case for reporting tool

### LINKED ISSUE #9415: Create/Update/Delete of vaccination/immunization should be considered as change of a case for reporting tool [3]
### Problem Description

When vaccination data is updated in a case, this case is not treated as changed afterwards.

immunization data is generally attached to the person, but should always mark the case as changed when a change is made.

otherwise, when using the filter "only cases changed since last shared with reporting tool", this case will not be displayed.

### Steps to Reproduce

1. Open a case (make sure it was send to the reporting tool at least once before)
2. apply changes to vaccinations
3. save
4. open the case directory
5. look for the filter "only cases changed since last shared with reporting tool"
6. the case will not appear

### Proposed Change

Based on discussion below:

- [x] Extend CaseService.createChangeDateFilter method with an optional param to include the change dates of relevant vaccinations AND immunizations. The relevance of a vaccination is defined by it's vaccination date being before the case's onset date or if that is not defined the report date (see VaccinationService.isVaccinationRelevant). Make sure to either let VacciantionService provide this part of the criteria or at least link both aspects of the logic.
- [x] When deleting a vaccination, the change date of the parent immunization has to be incremented. This can probably be best done by added a "incrementChangeDate" method to the BaseAdoService which calls lock(immunization, LockMode.OPTIMISTIC_FORCE_INCREMENT) on the JPA session.
- [x] Call createChangeDateFilter with includeExtendedChangeDateFilters, because person and samples should also be considered for the reporting tool filter.

### Acceptance Criteria

1. A case is shown in the filter when either a related vaccination is created, update or deleted.
2. A case is shown in the filter when it's persons armedForcesRelationType is changed
3. A case is shown in the filter when a sample is created, updated or deleted.

### Additional Information

Technically the behavior is not wrong because something was changed on the person and not on the case. but with other personal data like date of birth this filter works. therefore it is a bug.

leventegal-she:
Fixes #9415

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/BaseAdoService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/BaseAdoService.java
@@ -403,11 +403,9 @@ public class BaseAdoService<ADO extends AbstractDomainObject> implements AdoServ
 		}
 	}
 
-	public void updateChangeDateToNow(ADO ado) {
+	public void incrementChangeDate(ADO ado) {
 		Session session = em.unwrap(Session.class);
 		session.lock(ado, LockMode.OPTIMISTIC_FORCE_INCREMENT);
-
-		ado.setChangeDate(new Timestamp(new Date().getTime()));
 	}
 
 	public interface ExistsPredicateBuilder<ADO extends AbstractDomainObject> {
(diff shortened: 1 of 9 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/common/BaseAdoService.java
```diff
@@ -401,6 +403,13 @@ public boolean exists(ExistsPredicateBuilder<ADO> filterBuilder) {
 		}
 	}
 
+	public void incrementChangeDate(ADO ado) {
```
MateStrysewske, line 406: Not sure whether "increment" is the perfect wording here, why not call it "updateChangeDate" or, even more explicit, "updateChangeDateToNow"?

MateStrysewske (changes requested):
LGTM without having tested it.

carina29 (approved):

PR #9766, opened 2022-07-07
#8314 - Fix location and user role captions

### LINKED ISSUE #8314: Move functional information from toString to getCaption methods [5]

MateStrysewske:
Fixes #8314

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/EntityDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/EntityDto.java
@@ -133,7 +133,7 @@ public abstract class EntityDto implements Serializable, Cloneable, HasUuid {
 	}
 
 	@JsonIgnore
-	public String getCaption() {
+	public String buildCaption() {
 		return toString();
 	}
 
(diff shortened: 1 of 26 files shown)


StefanKock (commented):
It's not explained why `EntityDto.getCaption`  is now named `buildCaption`.

Some problems I see directly:
1. UserRoleDto.getCaption is now decoupled, so a new problem that the default caption method is not returning a meaningful caption string when needed.
2. Maybe the caption is not built internally but precompiled. If there is a reason to move away from `getCaption` then we should use toCaption() for example.

StefanKock (approved):
Discussed with @MateStrysewske the method naming (is consistent to other places) and that Dtos who currently don't have the `buildCaption` method don't need them now.

PR #9792, opened 2022-07-11
#9784 S2S | When sharing back a Contact from target system to the source system the Contact can't be accepted

### LINKED ISSUE #9784: S2S | When sharing back a Contact from target system to the source system the Contact can't be accepted
### Bug Description
When sharing back a Contact from the target system to the source system the Contact can't be accepted because the source system expects a case to be shared as well where the contact is linked.

### Steps to Reproduce
1. Send a Case from system A to system B. Make sure the Case has a Contact linked to it 
2. Share the contact linked to the previous case to a system B with 'hand over the ownership' = true.
3. Accept the Case and afterwards also the contact at system B.
4. Change the contact in system B and save.
5. Share the Contact back to system A

### Expected Behavior
On System A the Contact should be accepted with no problem

### Actual Behaviour
error_sharing_back contact in S2S
The system A does not check the DB for the original shared Case and does not see the link between the shared-back Contact and the Shared Case.

### System Details

* Device: 
* SORMAS version:1.74.0-snapshot
* Android version/Browser:
* Server URL: localhost
* User Role: NatUser

leventegal-she:
Fixes #9784

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/contact/SormasToSormasContactFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/contact/SormasToSormasContactFacadeEjb.java
@@ -23,7 +23,6 @@ import static de.symeda.sormas.backend.sormastosormas.ValidationHelper.buildCont
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


JonasCir:
@leventegal-she 
One question: Is the changed data from the received contact written back to the original shared contact (i.e., is there any merging involved)? Or is the received contact a completely new entity in the DB although originating from instance A in the first place?

JonasCir:
Code looks good

JonasCir (approved):

leventegal-she:
> One question: Is the changed data from the received contact written back to the original shared contact (i.e., is there any merging involved)? Or is the received contact a completely new entity in the DB although originating from instance A in the first place?

If the contact was not yet shared then it creates a new one, otherwise it updates the previously shared one.

leventegal-she:
> One question: Is the changed data from the received contact written back to the original shared contact (i.e., is there any merging involved)? Or is the received contact a completely new entity in the DB although originating from instance A in the first place?

If the contact was not yet shared then it creates a new one, otherwise it updates the previously shared one.

JonasCir:
Thanks :)

PR #9793, opened 2022-07-11
#9785 S2S | Share status does not get updated to the latest modifications for Contacts

### LINKED ISSUE #9785: S2S | Share status does not get updated to the latest modifications for Contacts
### Bug Description
When sharing a contact multiple times to the target system and rejecting it or accepting it on the target system, the status of the Share for the Contact on the Source system does not get updated with the latest changes

### Steps to Reproduce

1. Share a Contact from system A to system B
2. On system B reject the Contact share
3. On system A Share it again 
4. On system A refresh the page and Revoke the share of the Contact
5. Share one more time with system B and check the status of the Share

### Expected Behavior
The status should get updated after each action

### Screenshots
In this screenshot the Contact was Revoked and shared one more time but the status did not change

S2S_status not updated

### System Details

* Device:
* SORMAS version: 1.74.0-snapshot
* Android version/Browser:
* Server URL: Localhost
* User Role: NatUser

leventegal-she:
Fixes #9785

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/SormasToSormasListComponent.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/SormasToSormasListComponent.java
@@ -76,7 +76,7 @@ public class SormasToSormasListComponent extends VerticalLayout {
 		sormasToSormasList = new SormasToSormasList(
 			caze.getSormasToSormasOriginInfo() == null,
 			Captions.sormasToSormasCaseNotShared,
-			(i) -> ControllerProvider.getSormasToSormasController().revokeShare(i, this::reloadList));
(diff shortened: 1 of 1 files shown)


JonasCir:
@leventegal-she Can you briefly explain what caused the bug in the first place and why your changes now fix it?

Code looks good, but I just want to have some background :)

leventegal-she:
> @leventegal-she Can you briefly explain what caused the bug in the first place and why your changes now fix it?
> 
> Code looks good, but I just want to have some background :)

Sometimes after share/reject the share list didn't update depending on how slow/fast the server responded because it was loaded asynchronously using a separate thread.

Now I kept only the initial load to be async so it doesn't block loading the page, but after share/revoke it reloads synchronously.

JonasCir:
thanks understood

JonasCir (approved):
lgtm

PR #9794, opened 2022-07-11
Feature/4462 add user roles view to UI

### LINKED ISSUE #4462: Add user roles view to UI [3]
### Situation Description
Source: https://github.com/hzi-braunschweig/SORMAS-Project/issues/898#issuecomment-671336289

Depends on #4461 

### Feature Description

Mockup: https://www.sketch.com/s/32266e24-8c84-4006-8e1c-27263b30bdb4/a/xrjMW48

- [x] add view. Should be a tab in the users view
- [x] list all user roles as configured in the database
- jurisdiction should be jurisdiction level
- [x] allow filtering by user right (list all roles that have a certain right) and jurisdiction
- [x] allow filtering by enabled: enabled, disabled, all user roles
- [x] allow detailed export. Each user right should go into a separate column -> can be moved from the users section. See #4882 
- [x] Add USER_ROLE_VIEW right. Give it to the admin default user. For migration give those rights to every user role that is allowed to edit users.

carina29:
Fixes #4462


sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserRoleFacadeEjb.java
```diff
@@ -254,6 +297,243 @@ public Map<UserRoleDto, Set<UserRight>> getUserRoleRights() {
...
+					roleRightCell.setCellValue(I18nProperties.getString(Strings.no));
+				}
+				columnIndex++;
+			}
+
+			Cell uuidCell = row.createCell(columnIndex++);
+			uuidCell.setCellValue(userRole.getUuid());
+
+			Cell portHealthUserCell = row.createCell(columnIndex++);
+			portHealthUserCell.setCellValue(userRole.isPortHealthUser());
```
leventegal-she, line 510: boolean values should be translated to yes/no similarly to user rights

do it for this and the next few cells

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserRoleFacadeEjb.java
```diff
@@ -507,16 +513,16 @@ private void generateUserRolesDocument(Map<UserRoleDto, Set<UserRight>> userRole
 			uuidCell.setCellValue(userRole.getUuid());
 
 			Cell portHealthUserCell = row.createCell(columnIndex++);
-			portHealthUserCell.setCellValue(userRole.isPortHealthUser());
+			setCellValueAndStyleForBooleans(portHealthUserCell, userRole.isPortHealthUser(), cellStyles);
```
leventegal-she, line 516: The coloring on the boolean columns is not needed, only the translation

leventegal-she (changes requested):

leventegal-she (approved):

PR #9798, opened 2022-07-11
Feature 9119 allow editing and deletion of aggregate report data

### LINKED ISSUE #9119: Allow editing and deletion of aggregate report data [3]

sergiupacurariu:
Fixes #9119


sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java
```diff
@@ -38,7 +38,7 @@ public Predicate createCriteriaFilter(
 		From<AggregateReport, AggregateReport> from) {
 
 		Predicate filter = null;
-		boolean isForcedJurisdictionCheck = Boolean.TRUE.equals(criteria.isForceJurisdictionCheck());
+		boolean isForcedJurisdictionCheck = Boolean.TRUE.equals(criteria.isConsiderNullJurisdictionCheck());
```
MartinWahnschaffe, line 41: rename

MartinWahnschaffe (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/report/AggregateReportService.java
sergiupacurariu: renamed

MartinWahnschaffe (approved):

PR #9825, opened 2022-07-13
#8920: Added distinction between create and edit user rights

### LINKED ISSUE #8920: Backend user right checks should distinguish between EDIT and CREATE user rights [2]
### Problem Description
Most entity facades have a save method that currently checks whether the user has either the CREATE oder EDIT user right for that entity.
Effectively this means that in the backend a user that does not have the right to create and entity can still do so. Some goes for editing.

### Proposed Change
- [x] Within the save method of each backend facade to the following: After getting the existing entity do a hasUserRight check for either the CREATE (entity does not exist yet) or EDIT (entity exists) user right. 

### Implementation Details

* Use AccessDeniedException - see RightsAllowedInterceptor

ChristopherRiedel:
Fixes #8920


leventegal-she (changes requested):
- `ADDITIONAL_TEST_CREATE` and `_EDIT` is not checked in `AdditionalTestFacadeEjb`
- Would be better to create a *helper method* for the logic so we don't duplicate it in almost all facades
- Also the CI is failing

sormas-backend/src/main/java/de/symeda/sormas/backend/visit/VisitFacadeEjb.java
```diff
@@ -218,6 +220,12 @@ public VisitDto getVisitByUuid(String uuid) {
 	public VisitDto saveVisit(@Valid VisitDto dto) {
 		final String visitUuid = dto.getUuid();
 		final Visit existingVisit = visitUuid != null ? visitService.getByUuid(visitUuid) : null;
+
+		if (existingVisit == null && !userService.hasRight(UserRight.VISIT_CREATE)
```
leventegal-she, line 224: please use `FacadeHelper.checkCreateAndEditRights`

leventegal-she (changes requested):

PR #9867, opened 2022-07-15
#8950 S2S_Processed lab messages should be transferred

### LINKED ISSUE #8950: S2S_Processed lab messages should be transferred [3]
### Feature Description

Currently it is possible to process a lab message. Afterwards the lab message is added to a case. If you share a case with S2S with hand over the ownership and with the sample -> the lab message is not transferred.
Furthermore the lab message can not be opened on source system after the share because the case can only be read (no right anymore to edit or open files) 
### Problem Description

This is a problem, because the target system needs the lab message to start with the investigation of the case. 
If a health department wants to use S2S right know they have to send the lab message by fax afterwards. 
This should be avoided with S2S and means double the work. 
### Proposed Change

**Processed** lab messages will be shared if the parameter 'sample' is set on true.
### Additional Information

source system after share (fields are greyed out because you only have the permission to read) :
image

target system after share (no lab message):
image

leventegal-she:
Fixes #8950

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/sample/SormasToSormasSampleDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/sample/SormasToSormasSampleDto.java
@@ -15,6 +15,7 @@
 
 package de.symeda.sormas.api.sormastosormas.sample;
 
+import java.util.Collections;
 import java.util.List;
 
 import javax.validation.Valid;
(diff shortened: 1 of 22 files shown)


JonasCir (approved):
Hey @leventegal-she,
code changes look good to me and are straight forward, thank you!

Pleas note that I cannot currently test this as my DEMIS setup is currently broken on my machine, so I'm going to leave a comment on the issue for QA to double check this ok?

PR #9895, opened 2022-07-19
#9146 S2S_deactivate share parameter 'share associated contacts' (for cases)

### LINKED ISSUE #9146: S2S_deactivate share parameter 'share associated contacts' (for cases) [2]
### Feature Description

At the moment it is possible to share cases with associated contacts.
### Problem Description

This feature is currently not suitable. If the share was successful and the user in source system adds new contacts to the case then these contacts will not be synchronised. The user in source and target system must be informed about this.
### Proposed Change

We should deactivate feature share parameter 'share associated contacts' for cases and see whether feature is requested by the health departments. If so we have to develop before activate.

For example: It is better if the user in source system has the option to select which contacts he wants to sent (otherwise there is a risk that the function will be abused due to laziness or lack of time). The user in source and target system must be informed about not existing synchronisation and in source system it must be visible in the tab 'contacts' of the case which contacts have been sent. Even better the synchronisation must be developed and then the user in source system has the option to select if added contacts in the future should be synchronised or not and target system must be informed of the decision taken. Also the target system must has the option to accept individually because this prevents that a whole share request has to be sent back because of one or a few contacts. It is also necessary that the user on source system can select to send a case without hand over the ownership with a contact with hand over the ownership.

leventegal-she:
Fixes #9146


JonasCir (changes requested):
When running this with a clean DB, only the following in present in the `featureconfiguration`:

| | |
| :- | :- |
| **id** | 27 |
| **uuid** | D33IID-WF2FFI-VP6PF4-DSVXJL3U |
| **changedate** | 2022-07-21 10:58:40.792 |
| **creationdate** | 2022-07-21 10:58:40.792 |
| **featuretype** | SORMAS\_TO\_SORMAS\_SHARE\_CONTACTS |
| **region\_id** | NULL |
| **district\_id** | NULL |
| **disease** | NULL |
| **enddate** | NULL |
| **sys\_period** | \["2022-07-21 08:58:40.791509+00",\) |
| **enabled** | NULL |
| **properties** | {"SHARE\_SAMPLES" : true, "SHARE\_IMMUNIZATIONS" : true} |
| **entitytype** | NULL |
| **change\_user\_id** | NULL |

JonasCir:
@leventegal-she can you rebase?

JonasCir (changes requested):
Hey,
I'm truly sorry for the delay here :(

I noticed two things:
uebergeben
Please move the note below all check boxes

Secondly: If you create a case with a contact, and then go to the contact page, open the share menu, and select an organization, you will receive the following message which should be adjusted as well:
share_from_contact

Maybe we should include the deactivation note also in the contact view under the share section such that there is no button to share it in the first place?

leventegal-she:
> Hey, I'm truly sorry for the delay here :(
> 
> I noticed two things: uebergeben Please move the note below all check boxes

It is placed as requested, see https://github.com/hzi-braunschweig/SORMAS-Project/issues/9146#issuecomment-1176155432.

> Secondly: If you create a case with a contact, and then go to the contact page, open the share menu, and select an organization, you will receive the following message which should be adjusted as well: share_from_contact
> 
> Maybe we should include the deactivation note also in the contact view under the share section such that there is no button to share it in the first place?

I don't really see what to adjust here. This message has nothing to do with the scope of the this ticket. Also I don't see why to spam the user that he can't send contacts together with cases when he just wants to share a contact. He will be asked to share the case first and there he will see that only the case can be sent at once.

JonasCir (approved):

PR #9936, opened 2022-07-22
Feature 9039 add duplicate warning on creating aggregate reports

### LINKED ISSUE #9039: Add a duplicate warning when creating and editing aggregate reports [3]
### Feature Description

Currently the web app and mobile app allow users to create additional aggregate reports for a jurisdiction entity, without checking if other users have already done so.

### Proposed Change

- [x] When the user edits/creates an aggregate report retrieve all existing reports for the same jurisdiction and epi week, created by other users.
- [x] If any exist display a warning between the filters/jurisdiction selection and the data entry part: "There already exist aggregate reports for 'district/facility name' in this epi week. The corresponding disease names are displayed in red."
- [x] For each disease that already has an entry display the disease label in red.
- [x] Same for the mobile app

### Implementation Details

Blocked by #9119

### Additional Information

1. If multiple reports are created in a week, the reports are considered as independent and sum together. This leads to overcounting -> addressed by #9068
2. Add a duplicate detection method to identify reports with the same jurisdiction and epi week -> #9073
3. It is possible that a duplicate report can be created when using the mobile app offline. During synch, use the most recently modified report as the right one while the others are stored as duplicates. It may not be good to delete them but instead add a filter (on the web app) like merger duplicate to allow users to identify duplicates + decides which to keep as the right record + option to delete the other pairs. -> #9068 and #9073
4. eg if there are two SOs' in district A, if SO1 submit a report for week1 2022, and SO2/ SO1 try to create another report for the same week, then show a pop up that a report exists for that week and ask the user to edit it if needed.  However, a hospital informant can create a report for the same week in district A since this is another user role. A second or the same informant should not also be able to create a new report in the same week.

Caution: A duplicate check is already partly implemented, see `AggregateReportsEditLayout.checkForExistingData` in 2020-03 with #1526.

sergiupacurariu:
Fixes #9039


sormas-app/app/src/main/java/de/symeda/sormas/app/report/aggregate/AggregateReportsFragment.java
```diff
@@ -351,6 +365,24 @@ private void showReportData() {
...
 	}
 
+	private boolean isDuplicateByDiseaseEpiWeekAndInfrastructure(
+		Disease disease,
+		EpiWeek epiWeek,
+		InfrastructureAdo selectedInfrastructure,
+		User selectedUser) {
+
+		List<AggregateReport> duplicates = reports.stream()
+			.filter(
```
MateStrysewske, line 375: I'd rather use `anyMatch` here, hoping that it would perform slightly better because the whole process could then abort as soon as a duplicate is found

MateStrysewske (changes requested):

sormas-app/app/src/main/java/de/symeda/sormas/app/report/aggregate/AggregateReportsFragment.java
sergiupacurariu: that was my initial approach but during testing I realized that this approach does not cover all the situations. In the return statement, I test the duplicates size and actually in the logic there is a difference between one result and more than one results so anyMatch will not work here.

MateStrysewske (changes requested):
Looks good, but can you please add the warning message to the mobile app as well? Otherwise users won't understand why some diseases are written in red :)

sergiupacurariu:
A warning message has been added to Android app as well

MateStrysewske (approved):

PR #9944, opened 2022-07-25
#9847 - Make region and district required for aggregate reports

### LINKED ISSUE #9847: Make region and district required for aggregate reports [0.5]
## Problem description

Aggregate reports created without a district or even a region are useless.

## Proposed Solution

- [x] Make region and district required for aggregate reports.
- [x] Delete existing aggregate reports that don't have region or district. -> point this out in the release notes!
        (Deletion will be done via DB-Update-Script)

## Additional Information

_Originally posted by @bernardsilenou in https://github.com/hzi-braunschweig/SORMAS-Project/issues/9068#issuecomment-1183833896_

Following https://github.com/hzi-braunschweig/SORMAS-Project/issues/9068#issuecomment-1182988376
I have not tested all the steps but one issue that I see is that an aggregate report can be created without a region or district by a national user for example.
1. Just as cases must have a region and a district, an aggregate report must also have both, no matter the user who created it. we can prevent users from creating such data. because they are useless.
2. The issue that arises when filtering by region would not apply since all reports would have a region and a district.
3. Since the main users who create such reports are SO (having no facility) and informant (having a facility), then we can make facility optional for a report. When grouping by facility, then report with Missing facility would not show up

sergiupacurariu:
Fixes #9847


sormas-app/app/src/main/java/de/symeda/sormas/app/report/aggregate/AggregateReportsFragment.java
```diff
@@ -374,7 +372,9 @@ public void doInBackground(TaskResultHolder resultHolder) throws DaoException {
 						if (report.getLocalChangeDate() == null
 							&& (report.getNewCases() == null || report.getNewCases() == 0)
 							&& (report.getLabConfirmations() == null || report.getLabConfirmations() == 0)
-							&& (report.getDeaths() == null || report.getDeaths() == 0)) {
+							&& (report.getDeaths() == null || report.getDeaths() == 0)
+							&& report.getRegion() == null
```
MateStrysewske, line 376: Does it make sense to include region and district here? I think they should be checked before (i.e. submitting the numbers should with no region and district filled in should result in an error); right here, they rather water down the logic

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -11659,4 +11659,13 @@ INSERT INTO userroles_userrights (userrole_id, userright) SELECT userrole_id, 'U
 
 INSERT INTO schema_version (version_number, comment) VALUES (473, 'Add user roles view to UI #4462');
 
+-- 2022-07-25 Make region and district required for aggregate reports
+DELETE FROM aggregatereport
+WHERE region_id IS NULL OR district_id IS NULL;
+
+DELETE FROM aggregatereport_history
```
MateStrysewske, line 11666: I'd keep the history table entries just in case anyone wants to restore data that is deleted with the script above; @MartinWahnschaffe what do you think?

MateStrysewske (changes requested):

sormas-app/app/src/main/java/de/symeda/sormas/app/report/aggregate/AggregateReportsFragment.java
sergiupacurariu: it will be removed

sormas-backend/src/main/resources/sql/sormas_schema.sql
MartinWahnschaffe: Deleting from the history table is redundant, since we already have a trigger that takes care of this.
If we want to create a backup of the deleted data we would have to deactivate this trigger and then create a trigger that creates the history entry on deletion. From my point of view this would not be worth it, but we need a bold comment in the release notes that mentions this. A backup should be created when updating the system anyway.

sergiupacurariu: the code for deletion of aggregatereport_history table was removed

MateStrysewske (approved):

PR #9957, opened 2022-07-26
ensure minimum deletion period 7 days #9471

### LINKED ISSUE #9471: Deletion period can be below 7 days
### Bug Description

https://github.com/hzi-braunschweig/SORMAS-Project/issues/7008 specified that the lowest possible deletionperiod should be 7 days. However, setting a deletion period below 7 does not get validated and can be set in the database.
### Steps to Reproduce

1. Configure the deletionperiod to be 6 days for e.g. entitytype = CASE and deletionreference = END
2. Check the deletionconfiguration table or alternatively the UI display of the entity

### Expected Behavior

The deletionperiod should not be able to be configured below 7 days
### Screenshots

deletionDaysValidate

### System Details

* Device:
* SORMAS version: 1.72.1

syntakker:
Fixes #9471

In class `DeletionConfiguration`, hibernate validation annotations were already present:
```
	@Min(value = 7, message = Validations.numberTooSmall)
	@Max(value = Integer.MAX_VALUE, message = Validations.numberTooBig)
	Integer deletionPeriod;
```

However, as validation is only triggered by hibernate's lifecycle events persist, update and remove, the validation of deletion configurations read from the database is ensured by a database constraint.

@StefanKock The introduction of this constraint is preceded by a silent update of entries violating it. Please consider if this is the way to go, or if we would prefer a hard fail on startup or if we can introduce a warning message. The time of the parameter's update, however, is documented in the `sys_period` column.


StefanKock (approved):
Since this is considered a bug I would say adjusting the deletionPeriod to the minimal allowed value is reasonable.

PR #9976, opened 2022-07-27
#9629 - Allow diseases to be used case-based and aggregated at the sa…

### LINKED ISSUE #9629: Allow diseases to be used case-based and aggregated at the same time [2]
### Problem Description

Diseases have to be usable both for case-based surveillance and aggregate reporting in case that countries are using both approaches simultaneously.

### Proposed Change

Allow diseases to be used for both features by replacing the `caseBased` property of disease configurations.

### Implementation Details

- [x] Rename the `caseBased` property of disease configurations to `caseSurveillanceEnabled`
- [x] Add a new property called `aggregateReportingEnabled`; write an SQL migration script that sets it to the opposite value of `caseSurveillanceEnabled`
- [x] Rename the `caseBasedDiseases` and `aggregateDiseases` lists in DiseaseConfigurationFacadeEjb to `caseSurveillanceDiseases` and `aggregateReportingDiseases` and update the code that builds them based on the new properties
- [x] Update the `getAllDiseases` method by adding a new parameter called `aggregateReporting` (and change the `caseBased` parameter to `caseSurveillance`). Update the code to return all diseases based on the value of these properties; make sure that the final list doesn't contain any diseases twice
- [x] Update usages of the `getAllDiseases` method if necessary; in most places, you won't need to change anything, so I recommend to add an overloaded `getAllDiseases` method without the `aggregateReporting` parameter and call the other `getAllDiseases` method with false
- [x] Make sure that code that retrieves diseases for aggregate reporting calls the `getAllDiseases` method with `caseSurveillance == false`
- [x] Apply the same changes to the mobile app

### Additional Information

Data between case surveillance and aggregate reports are not supposed to match/link, the features stay separate.

sergiupacurariu:
…me time

Fixes #9629


sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -2993,6 +2993,16 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				getDao(DiseaseConfiguration.class).executeRaw("ALTER TABLE diseaseConfiguration ADD COLUMN ageGroupsString text;");
 				getDao(DiseaseConfiguration.class).executeRaw("UPDATE diseaseConfiguration SET changeDate = 0;");
 
+			case 337:
+				currentVersion = 337;
+				getDao(DiseaseConfiguration.class)
```
MateStrysewske, line 2998: This won't work. Column renaming is not supported in the SQLite version we're using. There's two possible workarounds for this; the easier one is to add a new `caseSurveillanceEnabled` column, copy the values from `caseBased` to the new column, and then mark the `caseBased` column as deprecated.

Also, SQLite requires colum names to be written in camel case, otherwise this script won't run through correctly.

Please make sure to test this migration before re-submitting the PR for review :)

sormas-app/app/src/main/java/de/symeda/sormas/app/util/DiseaseConfigurationCache.java
```diff
@@ -76,9 +70,12 @@ private DiseaseConfigurationCache() {
 			} else {
 				nonPrimaryDiseases.add(disease);
 			}
-			if (Boolean.TRUE.equals(configuration.getCaseBased()) || (configuration.getCaseBased() == null && disease.isDefaultCaseBased())) {
+			if (Boolean.TRUE.equals(configuration.getCaseSurveillanceEnabled())
+				|| (configuration.getCaseSurveillanceEnabled() == null && disease.isDefaultCaseBased())) {
```
MateStrysewske, line 74: `isDefaultCaseBased` should be renamed to `isDefaultCaseSurveillanceEnabled`

sormas-app/app/src/main/java/de/symeda/sormas/app/util/DiseaseConfigurationCache.java
```diff
@@ -76,9 +70,12 @@ private DiseaseConfigurationCache() {
...
 				caseBasedDiseases.add(disease);
-			} else {
+			}
+			if (Boolean.TRUE.equals(configuration.getAggregateReportingEnabled())
+				|| (configuration.getAggregateReportingEnabled() == null && !disease.isDefaultCaseBased())) {
```
MateStrysewske, line 78: Diseases should also get a `isDefaultAggregateReportingEnabled` method which should be used here, because this logic is flawed (i.e. it won't work if a disease is enabled for both case surveillance and aggregate reporting by default). All diseases that so far have been neither primary nor case based should be enabled for aggregate reporting by default.

sormas-backend/src/main/java/de/symeda/sormas/backend/disease/DiseaseConfigurationFacadeEjb.java
```diff
@@ -285,10 +288,11 @@ public void loadData() {
...
 				nonPrimaryDiseases.add(disease);
 			}
-			if (enabled(configuration.getCaseBased(), disease.isDefaultCaseBased())) {
-				caseBasedDiseases.add(disease);
-			} else {
-				aggregateDiseases.add(disease);
+			if (enabled(configuration.getCaseSurveillanceEnabled(), disease.isDefaultCaseBased())) {
+				caseSurveillanceDiseases.add(disease);
+			}
+			if (enabled(configuration.getAggregateReportingEnabled(), !disease.isDefaultCaseBased())) {
```
MateStrysewske, line 294: See above, logic is flawed here as well

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -11670,4 +11670,15 @@ ALTER TABLE sormastosormassharerequest_history ADD COLUMN shareassociatedcontact
...
 INSERT INTO schema_version (version_number, comment) VALUES (474, 'S2S_deactivate share parameter ''share associated contacts'' (for cases) #9146');
 
+-- 2022-07-25 Allow diseases to be used case-based and aggregated at the same time
+ALTER TABLE  diseaseconfiguration RENAME COLUMN casebased TO casesurveillanceenabled;
+ALTER TABLE  diseaseconfiguration_history RENAME COLUMN casebased TO casesurveillanceenabled;
+ALTER TABLE diseaseconfiguration ADD COLUMN aggregatereportingenabled boolean;
+ALTER TABLE diseaseconfiguration_history ADD COLUMN aggregatereportingenabled boolean;
+
+UPDATE diseaseconfiguration SET aggregatereportingenabled = NOT casesurveillanceenabled;
+UPDATE diseaseconfiguration_history SET aggregatereportingenabled = NOT casesurveillanceenabled;
```
MateStrysewske, line 11680: Don't manually update the content of history tables

sormas-backend/src/test/java/de/symeda/sormas/backend/TestDataCreator.java
```diff
@@ -1835,12 +1835,13 @@ public PopulationDataDto createPopulationData(RegionReferenceDto region, Distric
 		return populationData;
 	}
 
-	public void updateDiseaseConfiguration(Disease disease, Boolean active, Boolean primary, Boolean caseBased) {
+	public void updateDiseaseConfiguration(Disease disease, Boolean active, Boolean primary, Boolean caseSurveillance, Boolean aggregateDiseases) {
```
MateStrysewske, line 1838:
```suggestion
	public void updateDiseaseConfiguration(Disease disease, Boolean active, Boolean primary, Boolean caseSurveillance, Boolean aggregateReporting) {
```

MateStrysewske (changes requested):

sormas-backend/src/test/java/de/symeda/sormas/backend/TestDataCreator.java
sergiupacurariu: changed

sormas-backend/src/main/resources/sql/sormas_schema.sql
sergiupacurariu: update query for diseaseconfiguration_history removed

sormas-app/app/src/main/java/de/symeda/sormas/app/util/DiseaseConfigurationCache.java
sergiupacurariu: updated

sormas-app/app/src/main/java/de/symeda/sormas/app/util/DiseaseConfigurationCache.java
sergiupacurariu: logic updated

sormas-backend/src/main/java/de/symeda/sormas/backend/disease/DiseaseConfigurationFacadeEjb.java
sergiupacurariu: updated as well

MateStrysewske (approved):

PR #9978, opened 2022-07-28
#9960 Fetch bean-test from within the repo as jar (system scope)

### LINKED ISSUE #9960: Make `bean-test` artifact accessible again
### Problem Description

https://repository.novatec-gmbh.de/content/repositories/novatec is now down (responding with _503 Service Temporarily Unavailable_) since at least 24h. This blocks our tests in Github Actions, because artifact `info.novatec:bean-test` is only hosted in that repository.

### Proposed Solution

1. Contact the hosting company if the repo will come online again.
2. Consider pushing the existing artifact or a fork of it to Maven Central.

### Additional Information

- `info.novatec:bean-test` is the framework we use to tests most of our backend by emulating an environment where EJBs can be tested directly without extensive mocking.
- All build environments that cache maven artifacts still work as long as the local maven repository (or that specific artifact) is not wiped out.
- Related ticket in Source repo: https://github.com/NovatecConsulting/BeanTest/issues/29

StefanKock:
Fixes #9960 (now with system scope)


MartinWahnschaffe:
We should also remove the novatec repository from the pom.
I guess using the local path on the jenkins should also work, but we will find out about this soon enough.

StefanKock:
Unit tests on Gihub Actions and in my DEV environment were successful.

PR #9988, opened 2022-07-29
Change 6748 add task status progress 

### LINKED ISSUE #6748: Add task status "Progress" [0.5]
### Problem Description

Currently the tasks can only have the status pending, done, removed, not executable. It cannot documented if an officer started to work on the task as the is a status "in progress" is missing.

### Proposed Change

Please add this status to the tasks.
image

FredrikSchaeferVitagroup:
Fixes #6748 

Consider possible feature freeze before merging.

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportFacadeEjb.java
@@ -20,6 +20,7 @@ package de.symeda.sormas.backend.report;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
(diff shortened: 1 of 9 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportFacadeEjb.java
```diff
@@ -377,9 +377,8 @@ public void generateSubmitWeeklyReportTasks() {
 				// task
 				continue;
 			} else {
-				TaskCriteria pendingUserTaskCriteria =
-					new TaskCriteria().taskType(TaskType.WEEKLY_REPORT_GENERATION).assigneeUser(user.toReference()).taskStatus(TaskStatus.PENDING);
-				List<Task> existingTasks = taskService.findBy(pendingUserTaskCriteria, true);
+				List<Task> existingTasks = taskService
```
leventegal-she, line 380: I would extend the criteria to accept multiple statuses but if you would like to keep the new method please rename to `findByTypeAndStatus` or similar and maybe use rest params for statuses (eg. TaskStatus ...status )

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/report/WeeklyReportFacadeEjb.java
FredrikSchaeferVitagroup: First thing I did was extending the criteria. The problem with this is that it's used by the task grid, and therefore requires the one status attribute. Having both (one and multiple statuses) seemed ugly to me.
So I will keep the new method, but will adapt the to your requests.

leventegal-she (approved):

PR #10014, opened 2022-08-02
Feature/4462 fixes

carina29:
Fixes #


ChristopherRiedel (changes requested):
1. The translation for USER_ROLE_VIEW is missing
2. Sorting roles by uuid results in an error
3. The list grid does not use the available space and cuts off text even if the space would allow for the full display of the content.
All this was found by @kwa20 (https://github.com/hzi-braunschweig/SORMAS-Project/issues/4462#issuecomment-1202165832). I just checked if these are alrready fixed.

carina29:
1) Was solved
2) A new tichet was created https://github.com/hzi-braunschweig/SORMAS-Project/issues/10032

PR #10030, opened 2022-08-03
Feature 7930 event permanent delete

### LINKED ISSUE #7930: Permanent deletion for Event and Event Participant [3]
### Situation Description
https://github.com/hzi-braunschweig/SORMAS-Project/issues/7712 is a pre-requisite to this.

### Feature Description

**Event Participant**

The soft-delete of all Samples related to the event participant should stay as is (EventParticipantService .delete calls TaskService.delete)

- [x] Duplicate the logic that deletes samples for EventParticipantService.deletePermanent. So when an event participant is permanently deleted all related samples are also permanently deleted
- [x] In deletePermanent remove the EventParticipant from any SormasToSormasShareInfo that is referencing it and delete it if no longer referenced by another entity

**Event**

The soft-delete of all EventParticipant related to the Event should stay as is (EventService.delete calls EventParticipantService.delete)

- [x] Move the logic that permanently deletes tasks and actions from the EventService delete method to the deletePermanent method
- [x] In deletePermanent remove the Event from any SormasToSormasShareInfo that is referencing it and delete it if no longer referenced by another entity
- [x] In deletePermanent remove the Event from any ExternalShareInfo that is referencing it
- [x] In deletePermanent **and** delete remove the Event from all subordinate events that are referencing it > see how this is done for the case (probably a separate method called from both delete and deletePermanent)
- [x] In deletePermanent call markAsDeleted for all documents related to the event
- [x] deleting in event in external surveillance tools also has to be done in deletePermant

**Action**

- [x] In deletePermanent of the ActionService call markAsDeleted for all documents related to the action
- [x] Make sure that facades and services only return "active" actions (actions of events that have not been archived or deleted), except for those methods that return the actions of a specific event. The related methods should be renamed to contain "active" - see other services like ClinicalVisitsService for comparison.

Finally:

- [x] CoreEntityDeletionService: Call executePermanentDeletion for the EventParticipantService and EventService

BarnaBartha:
Fixes #7930

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
@@ -40,7 +40,6 @@ import de.symeda.sormas.api.event.EventActionIndexDto;
 import de.symeda.sormas.api.event.EventCriteria;
 import de.symeda.sormas.api.user.JurisdictionLevel;
 import de.symeda.sormas.api.utils.SortProperty;
-import de.symeda.sormas.backend.caze.Case;
 import de.symeda.sormas.backend.common.AdoServiceWithUserFilter;
 import de.symeda.sormas.backend.common.CriteriaBuilderHelper;
 import de.symeda.sormas.backend.document.DocumentService;
@@ -75,12 +74,7 @@ public class ActionService extends AdoServiceWithUserFilter<Action> {
 		Predicate filter = createActiveFilter(cb, queryContext.getJoins().getEvent());
 		filter = CriteriaBuilderHelper.and(cb, filter, createUserFilter(queryContext));
 		if (since != null) {
-			Predicate dateFilter = createChangeDateFilter(cb, root, since, lastSynchronizedUuid);
-			if (filter != null) {
-				filter = cb.and(filter, dateFilter);
-			} else {
-				filter = dateFilter;
-			}
+			filter = CriteriaBuilderHelper.and(cb, filter, createChangeDateFilter(cb, root, since, lastSynchronizedUuid));
 		}
 		if (filter != null) {
 			cq.where(filter);
@@ -101,37 +95,12 @@ public class ActionService extends AdoServiceWithUserFilter<Action> {
 		Predicate filter = createActiveFilter(cb, queryContext.getJoins().getEvent());
 
 		if (getCurrentUser() != null) {
-			Predicate userFilter = createUserFilter(cb, cq, from);
-			filter = CriteriaBuilderHelper.and(cb, filter, userFilter);
+			filter = CriteriaBuilderHelper.and(cb, filter, createUserFilter(cb, cq, from));
 		}
 
 		cq.where(filter);
-		cq.select(from.get(Case.UUID));
+		cq.select(from.get(Action.UUID));
 
-		return em.createQuery(cq).getResultList();
-	}
-
-	public List<Action> getAllActiveActionsAfter(Date date, User user) {
-
-		CriteriaBuilder cb = em.getCriteriaBuilder();
-		CriteriaQuery<Action> cq = cb.createQuery(getElementClass());
-		Root<Action> from = cq.from(getElementClass());
-		ActionQueryContext queryContext = new ActionQueryContext(cb, cq, from);
-
-		Predicate filter = createActiveFilter(cb, queryContext.getJoins().getEvent());
-		if (user != null) {
-			Predicate userFilter = createUserFilter(queryContext);
-			filter = CriteriaBuilderHelper.and(cb, filter, userFilter);
-		}
-		if (date != null) {
-			Predicate dateFilter = createChangeDateFilter(cb, from, date);
-			filter = CriteriaBuilderHelper.and(cb, filter, dateFilter);
-		}
-		if (filter != null) {
-			cq.where(filter);
-		}
-		cq.orderBy(cb.desc(from.get(Action.CHANGE_DATE)));
-		cq.distinct(true);
 		return em.createQuery(cq).getResultList();
 	}
 
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
@@ -34,13 +34,16 @@ import javax.persistence.criteria.Root;
 import de.symeda.sormas.api.action.ActionCriteria;
 import de.symeda.sormas.api.action.ActionDto;
 import de.symeda.sormas.api.action.ActionStatEntry;
+import de.symeda.sormas.api.document.DocumentRelatedEntityType;
 import de.symeda.sormas.api.event.EventActionExportDto;
 import de.symeda.sormas.api.event.EventActionIndexDto;
 import de.symeda.sormas.api.event.EventCriteria;
 import de.symeda.sormas.api.user.JurisdictionLevel;
 import de.symeda.sormas.api.utils.SortProperty;
+import de.symeda.sormas.backend.caze.Case;
 import de.symeda.sormas.backend.common.AdoServiceWithUserFilter;
 import de.symeda.sormas.backend.common.CriteriaBuilderHelper;
+import de.symeda.sormas.backend.document.DocumentService;
 import de.symeda.sormas.backend.event.Event;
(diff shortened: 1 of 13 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
```diff
@@ -577,8 +597,53 @@ public void delete(Event event, DeletionDetails deletionDetails) {
...
+
+		documentService.getRelatedToEntity(DocumentRelatedEntityType.EVENT, event.getUuid()).forEach(d -> documentService.markAsDeleted(d));
+
+		deleteEventInExternalSurveillanceTool(event);
+		deleteEventLinks(event);
+
+		super.deletePermanent(event);
+	}
+
+	private void deleteEventLinks(Event event) {
```
sergiupacurariu, line 623: I would rename it to deleteSubordinateEvents since this is the only thing this method does

sergiupacurariu (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
```diff
@@ -54,18 +57,68 @@ public class ActionService extends AdoServiceWithUserFilter<Action> {
...
 		CriteriaQuery<Action> cq = cb.createQuery(getElementClass());
+		Root<Action> root = cq.from(getElementClass());
+
+		ActionQueryContext queryContext = new ActionQueryContext(cb, cq, root);
+
+		Predicate filter = createActiveFilter(cb, queryContext.getJoins().getEvent());
+		filter = CriteriaBuilderHelper.and(cb, filter, createUserFilter(queryContext));
+		if (since != null) {
+			Predicate dateFilter = createChangeDateFilter(cb, root, since, lastSynchronizedUuid);
+			if (filter != null) {
```
MateStrysewske, line 79: Use CriteriaBuilderHelper.and here as well instead of the if-else condition; also allows you to get rid of the variable definition for dateFilter

sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
```diff
@@ -54,18 +57,68 @@ public class ActionService extends AdoServiceWithUserFilter<Action> {
...
+		CriteriaBuilder cb = em.getCriteriaBuilder();
+		CriteriaQuery<String> cq = cb.createQuery(String.class);
 		Root<Action> from = cq.from(getElementClass());
 		ActionQueryContext queryContext = new ActionQueryContext(cb, cq, from);
 
-		Predicate filter = null;
+		Predicate filter = createActiveFilter(cb, queryContext.getJoins().getEvent());
+
+		if (getCurrentUser() != null) {
+			Predicate userFilter = createUserFilter(cb, cq, from);
```
MateStrysewske, line 104: Unnecessary variable

sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
```diff
@@ -54,18 +57,68 @@ public class ActionService extends AdoServiceWithUserFilter<Action> {
...
+			filter = CriteriaBuilderHelper.and(cb, filter, userFilter);
+		}
+
+		cq.where(filter);
```
MateStrysewske, line 108: Unique clause would make sense here as well I guess, just to be safe

sergiupacurariu (approved):

sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
```diff
@@ -54,18 +57,68 @@ public class ActionService extends AdoServiceWithUserFilter<Action> {
...
+		cq.select(from.get(Case.UUID));
+
+		return em.createQuery(cq).getResultList();
+	}
+
+	public List<Action> getAllActiveActionsAfter(Date date, User user) {
```
MateStrysewske, line 114: Feels like a redundant method; it basically does exactly the same as getAllAfter minus the batching; could just as well simply call getAllAfter from this method and in getAllAfter check whether batchSize is not null before doing the batching, or simply returning the result list otherwise. Also, I believe that both methods should have the same name then (i.e. both should be called getAllAfter or getAllActiveActionsAfter).

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventFacadeEjb.java
```diff
@@ -340,6 +342,19 @@ public List<String> deleteEvents(List<String> eventUuids, DeletionDetails deleti
 		return deletedEventUuids;
 	}
 
+	@RightsAllowed({
+		UserRight._CASE_DELETE,
+		UserRight._SYSTEM })
+	public void deleteEventInExternalSurveillanceTool(Event event) throws ExternalSurveillanceToolException {
+
+		if (externalSurveillanceToolGatewayFacade.isFeatureEnabled() && event.getExternalId() != null && !event.getExternalId().isEmpty()) {
```
MateStrysewske, line 350: Use StringUtils.isNotBlank instead of combining a null check with an empty check

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
```diff
@@ -577,8 +597,53 @@ public void delete(Event event, DeletionDetails deletionDetails) {
...
+
+		documentService.getRelatedToEntity(DocumentRelatedEntityType.EVENT, event.getUuid()).forEach(d -> documentService.markAsDeleted(d));
+
+		deleteEventInExternalSurveillanceTool(event);
+		deleteSubordinateEvents(event);
+
+		super.deletePermanent(event);
+	}
+
+	private void deleteSubordinateEvents(Event event) {
```
MateStrysewske, line 623:
```suggestion
	private void removeFromSubordinateEvents(Event event) {
```

MateStrysewske: Reason: You're not deleting the subordinate events, you're simply removing the reference to this event

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventService.java
```diff
@@ -577,8 +597,53 @@ public void delete(Event event, DeletionDetails deletionDetails) {
...
+		} catch (ExternalSurveillanceToolException e) {
+			throw new ExternalSurveillanceToolRuntimeException(e.getMessage(), e.getErrorCode());
+		}
+	}
+
+	public List<Event> getByExternalId(String externalId) {
+
+		CriteriaBuilder cb = em.getCriteriaBuilder();
+		CriteriaQuery<Event> cq = cb.createQuery(Event.class);
+		Root<Event> caseRoot = cq.from(Event.class);
```
MateStrysewske, line 642:
```suggestion
		Root<Event> eventRoot = cq.from(Event.class);
```

sormas-backend/src/test/java/de/symeda/sormas/backend/deletionconfiguration/CoreEntityDeletionServiceTest.java
```diff
@@ -232,6 +234,46 @@ public void testImmunizationAutomaticDeletion() {
 		assertEquals(0, getVaccinationService().count());
 	}
 
+	@Test
+	public void testEventAutomaticDeletion() {
+
+		createDeletionConfigurations();
+		DeletionConfiguration coreEntityTypeConfig = getDeletionConfigurationService().getCoreEntityTypeConfig(CoreEntityType.IMMUNIZATION);
```
MateStrysewske, line 241:
```suggestion
		DeletionConfiguration coreEntityTypeConfig = getDeletionConfigurationService().getCoreEntityTypeConfig(CoreEntityType.EVENT);
```

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/action/ActionService.java
BarnaBartha: i don't understand this
can you please elaborate ?

MateStrysewske: Ah sorry, I meant distinct, not unique. :D

MateStrysewske (approved):

PR #10046, opened 2022-08-04
SORQA-431 implementation

pk-sgent:
Fixes #


mp-sgent (approved):

pk-sgent:
it fails because of bug in the Sormas

image

PR #10048, opened 2022-08-04
[#9701] better user validation through annotations

### LINKED ISSUE #9701: Cleanup reporting user validation
### Bug Description
We discovered  in #9696 that reporting user validation is probably inconsistent and is only triggered for Contacts.
Citing @leventegal-she 
> Please create a ticket to cleanup the reporting user validation. It's strange that only contact validation fails if it has no reporting user.

Behavior should be consistent for all entities.

~Please also adjust the tests (e.g., `testShareContact()`) accordingly~ Test works

---

Found:
- `EventFacade`
- `EventParticipateFacade`
- `ImmunizationFacade`
- `CaseFacadeEjb`

JonasCir:
Fixes #9701 
@MateStrysewske This PR adds a validation to the API, however, the reporting user was either already checked in the validate method or it has a not null constraint in the DB. 

I went through the general setup here and I think we should start to phase out `@Required` over time, I'm going to write a dedicated issue for that.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/BaseFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/BaseFacade.java
@@ -31,5 +31,5 @@ public interface BaseFacade<DTO extends EntityDto, INDEX_DTO extends Serializabl
 
 	List<String> getObsoleteUuidsSince(Date since);
 
-	void validate(DTO dto) throws ValidationRuntimeException;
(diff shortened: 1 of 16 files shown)


MateStrysewske (approved):

MateStrysewske:
I'm not that familiar with potential side effects of the `@Valid` annotation. Did a quick test using the contact resource to check the reporting user validation which seems fine, so I've approved the PR.

PR #10062, opened 2022-08-08
Feature 8723 all event part filter

### LINKED ISSUE #8723: Set 'All Event Participants' as the default value of the 'active/archived/all event participants' filter, when an event is archived [0.5]
### Problem Description

When an event is archived (and by default its event participants are archived as well), the 'Event Participants' list remains filtered to 'Active event participants' and the user is not able to see the associated event participants of the archived event.

Event Participants list before archiving the event:
image

Event Participants list after archiving the event:
image

### Proposed Change

When the event is archived, set the value 'All Event Participants' as the default value for the filter 'active/archived/all event participants', on the 'Event Participants' page.

BarnaBartha:
Fixes #8723


sergiupacurariu:
the following flow does not work properly:
- access an archived event
- open the event participants grid
- the relevance status is correctly set to ALL
- change the relevance status to ACTIVE
- switch directly to Cases Directory
- move back to Events Directory and select the archived event
- open the event participants grid
- the relevance status is set to ACTIVE (this is incorrect, it should be set to ALL)

BarnaBartha:
I can't find a solution to this, it is a limitation to vaadin itself, not having access to the target navigation in the view. 
Also i do not think the behaviour is bad, obviously depends on the definition of 'default' value for the filter.
@MartinWahnschaffe @MateStrysewske i need your input here as well. Thanks!

MateStrysewske:
Both approaches (i.e. the value stays as it is after it was changed by the user vs. it is always reset to ALL) are fine by me actually. We should just go with whatever works and is easiest to implement.

sergiupacurariu (approved):

PR #10114, opened 2022-08-11
Change/9701 s2s reporting user

### LINKED ISSUE #9701: Cleanup reporting user validation

JonasCir:
Fixes #9701
The PR does the following:

- Do not set `reportingUser` to null where it is required while building the S2S share
- Add generic method to build the DTO while sharing

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/event/EventShareDataBuilder.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/event/EventShareDataBuilder.java
@@ -52,11 +52,8 @@ public class EventShareDataBuilder
 
 	@Override
 	protected SormasToSormasEventDto doBuildShareData(Event data, ShareRequestInfo requestInfo, boolean ownerShipHandedOver) {
-		Pseudonymizer pseudonymizer =
-			dataBuilderHelper.createPseudonymizer(requestInfo);
-
-		EventDto eventDto = getEventDto(data, pseudonymizer);
-
+		Pseudonymizer pseudonymizer = dataBuilderHelper.createPseudonymizer(requestInfo);
+		EventDto eventDto = getDto(data, pseudonymizer);
 		return new SormasToSormasEventDto(eventDto);
 	}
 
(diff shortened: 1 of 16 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/event/EventShareDataBuilder.java
```diff
@@ -67,20 +64,18 @@ public void doBusinessValidation(SormasToSormasEventDto sormasToSormasEventDto)
...
+		Pseudonymizer pseudonymizer = dataBuilderHelper.createPseudonymizer(requestInfo);
 		return getEventPreview(event, pseudonymizer);
 	}
 
-	private EventDto getEventDto(Event event, Pseudonymizer pseudonymizer) {
-		EventDto eventDto = eventFacade.convertToDto(event, pseudonymizer);
+	@Override
+	protected EventDto getDto(Event event, Pseudonymizer pseudonymizer) {
 
-		eventDto.setReportingUser(null);
```
BarnaBartha, line 79: it is a bit confusing what the overall change wants to be
looking at other parts i would say we need this nulling of the reporting user
some minimal unit tests would shed light on the whole thing as the ticket you are referencing is also a bit lacking in description

JonasCir: Hey, the nulling of the user is exactly what causes the transmission to fail. https://github.com/hzi-braunschweig/SORMAS-Project/pull/9696 already did this for contacts. I can create a unit test.

JonasCir: tests added

JonasCir:
I added test for cases, contacts, events. Event test also includes event participant and samples, cases also covers immunizations.

BarnaBartha (approved):

leventegal-she (approved):

Issue #8541, opened 2022-03-25
Add backend checks for infrastructure & user related user rights [2]


MartinWahnschaffe:
### Problem Description
We are missing checks for many of the user rights related to accessing the backend. 

### Proposed Change

Make sure all user rights for the facades listed below are checked. 
For most of the facades there is a corresponding _VIEW user right that should be used as a class anotation, so it's not necessary to annotation every singler getter. 
Most of the user rights can simply be annotation to the methods (RolesAllowed).
For some user rights you may have to do the check within the method instead of using an annotation.
All infrastructure related facades need to be annotated with the INFRASTRUCTURE... user rights.

- [x] Discuss: Should we rename USER_VIEW? It is ony relevant to decide whether the user should see the users directory in the UI. In general users need to be able to load user data, e.g. to sync all users to the mobile app or to provide users for a drop-down
- [x] UserFacadeEjb -> don't use USER_VIEW
- [x] UserRoleFacadeEjb (introduced by #4461) -> re-use USER_CREATE & USER_EDIT
- [x] Continent/Subcontinent/CountryFacadeEjb 
- [x] Area/Region/District/CommunityFacadeEjb
- [x] FacilityFacadeEjb
- [x] PointOfEntryFacacadeEjb
- [x] Check usage of these facades in the StartupShutdownService and CronService and make sure called facade methods are annotated with RolesAllowed including UserRight._SYSTEM.
- [x] Extend the archunit test implemented by #8508

vidi42:
We should rename `USER_VIEW` to `USER_DIRECTORY_VIEW` to emphasize it just refers to view on the UI

StefanKock:
> We should rename `USER_VIEW` to `USER_DIRECTORY_VIEW` to emphasize it just refers to view on the UI

I discussed with @leventegal-she about the suggested refactoring. I have two objections against it:
1. It breaks the pattern to other `UserRight` names without a real reason. Pointing out that some methods are only for directory view (what many other `ENTITY_VIEW` user rights are also) could be done by an additional permission. And what the use case of the existing `UserRight` is should be documented in code, manual etc. if it is not intuitive.
2. Adding `@PermitAll` to other endpoints concerning users also does not restrict to a logged in user.

vidi42:
@leventegal-she what is needed for this issue to be picked up again?

StefanKock:
Needs to wait for #4461 and resolve discussion points

[...]

PR #10131, opened 2022-08-15
[#9693] refactor isEditAllowed

### LINKED ISSUE #9693: Follow up to 8674 - Basic implementation of `isEditAllowed` is probably not sufficient for some classes
### Bug Description
While working on #8674, I noticed some inconsistencies around `isEditAllowed` specifically the implementation in:

- `CampaignService.java` where a basic implementation was added which needs to be verified if anything else needs to be done
- `CaseService.java` which only contains a basic implementation which is far less sophisticated than the checks for e.g., contacts which looks suspicious.

Investigate the logic and make sure that it is correct and no bugs are introduced by the basic implementation.

JonasCir:
Fixes #9693 
Turns out everything is fine, the checks were just scattered around, I cleaned it up and now everything is just in one place.


MateStrysewske (commented):
I'm a bit doubtful about this PR, honestly. First of all, a method that's called "isEditAllowed" returning anything but a boolean is very counterintuitive to me (I know that this method existed before, but if we're doing a refactoring here, I think renaming the getEditPermissionType methods to this is the wrong way to go). Alternatives could be to only use this method when we're really interested in the ALLOWED enum value and then letting isEditAllowed return a boolean - or to keep the getEditPermissionType name and use it everywhere (and completely get rid of the isEditAllowed name).

Secondly, I don't like replacing the explicit classes with superclasses for the UUID, ARCHIVED and DELETED calls. I know that the IDE suggests those, but @StefanKock has a very good point in saying that using the explicit classes gives more clues as to what exactly we're interested in, and I think that makes a lot of sense. I've added him as another reviewer so that he can say something about it.

Bottom line: I think we should discuss in which way this PR makes sense and whether we might need to change stuff about it.

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactJurisdictionFlagsDto.java
```diff
@@ -20,19 +20,19 @@
 
 public class ContactJurisdictionFlagsDto implements Serializable {
 
-    private Boolean isInJurisdiction;
-    private Boolean isCaseInJurisdiction;
+    private final boolean isInJurisdiction;
```
StefanKock, line 23: Why this refactoring? Is it always decidable without null as third option?

sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractCoreAdoService.java
```diff
@@ -99,7 +98,7 @@ public Map<String, Date> calculateEndOfProcessingDate(List<String> entityuuids)
 
 		Expression aggregatedChangeDateExpression = addChangeDates(new AggregatedChangeDateExpressionBuilder(cb), from, true).build();
 		cq.multiselect(from.get(AbstractDomainObject.UUID), cb.max(aggregatedChangeDateExpression));
-		cq.where(from.get(ADO.UUID).in(entityuuids));
+		cq.where(from.get(AbstractDomainObject.UUID).in(entityuuids));
```
StefanKock, line 101: Use `CoreAdo`

StefanKock (changes requested):
I totally agree with @MateStrysewske:
1. `isEditAllowed` not returning a `boolean` but a currently tri-state enum is a lot against naming conventions. I think it should be the other way around: `getEditPermissionType` is valid and `isEditAllowed` has to be removed or changed to `return getEditPermissionType == EditPermissionType.ALLOWED;` (as convenience, but is still slightly duplicate).
2. Changes of UUID / ARCHIVED constants to superclasses should always reflect the class that you expect to handle. In `AbstractCoreAdoService` changing it all to `CoreAdo` (instead of `Case`) is reasonable and clean code. But don't change it in `CampaignService` since it makes clear that it's the Campaigns attributes.

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactJurisdictionFlagsDto.java
JonasCir: removed from PR.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractCoreAdoService.java
JonasCir: done

JonasCir:
@MateStrysewske @Totally agree.

So I updated the PR quite a bit which gives as good wins w.r.t. readability and LOC.

1. `isEditAllowed` returns a boolean as it should, making the call sites much cleaner.
2. I refactored the edit checks in S2S in such a way that they do not rely on calls into subclasses anymore
3. I introduced the `AbstractEditAllowedDetailView` which removes quite some code duplication from the UI w.r.t. edit checks

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/AbstractSampleView.java
```diff
@@ -118,15 +118,15 @@ protected void setSubComponent(DirtyStateComponent newComponent) {
...
-		if (!isSampleEditAllowed) {
+		Boolean isSampleEditAllowed = isEditAllowed();
+		if (Boolean.FALSE.equals(isSampleEditAllowed)) {
 			component.setEnabled(false);
 		}
 	}
 
-	protected Boolean isSampleEditAllowed() {
-		return FacadeProvider.getSampleFacade().isSampleEditAllowed(getSampleRef().getUuid());
+	protected Boolean isEditAllowed() {
```
StefanKock, line 127: Why keep Object Boolean here?

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
```diff
@@ -1094,7 +1094,8 @@ public Predicate createDefaultFilter(CriteriaBuilder cb, Root<Sample> root) {
 		return cb.isFalse(root.get(Sample.DELETED));
 	}
 
-	public Boolean isSampleEditAllowed(Sample sample) {
+
+	public Boolean isEditAllowed(Sample sample) {
```
StefanKock, line 1098: This should not be an Object Boolean, and the current implementation does not support it. 
`inJurisdictionOrOwned(sample).getInJurisdiction()::Boolean` will run into NullPointerException.

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/AbstractDetailView.java
```diff
@@ -115,4 +115,5 @@ protected R getReference() {
 	 *            The URL parameters String
 	 */
 	protected abstract void initView(String params);
+
```
StefanKock, line 118: Why empty line here? It's against code convention

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/AbstractEditAllowedDetailView.java
```diff
@@ -0,0 +1,34 @@
...
+ * In addition to that, it contains generic code to check if the object is editable.
+ *
+ * @param <R>
+ *            {@link ReferenceDto} with the uuid as parsed from the URL.
+ */
+public abstract class AbstractEditAllowedDetailView<R extends ReferenceDto> extends AbstractDetailView<R> {
+
+	private CoreFacade coreFacade;
+
+	protected AbstractEditAllowedDetailView(String viewName, CoreFacade coreFacade) {
```
StefanKock, line 19: This pattern caches the Facade instance which is discouraged because it does not refresh when the sormas-ear is redeployed.

StefanKock (changes requested):
The approach with usage of `isEditAllowed` and `getEditPermissionType` looks now way cleaner. But some new remarks.

sormas-backend/src/main/java/de/symeda/sormas/backend/common/AbstractCoreFacadeEjb.java
```diff
@@ -259,8 +259,13 @@ public Date calculateEndOfProcessingDate(String entityUuid) {
 		return service.calculateEndOfProcessingDate(Collections.singletonList(entityUuid)).get(entityUuid);
 	}
 
-	public EditPermissionType isEditAllowed(String uuid) {
-		ADO ado = service.getByUuid(uuid);
-		return service.isEditAllowed(ado);
+	@Override
+	public EditPermissionType getEditPermissionType(String uuid) {
```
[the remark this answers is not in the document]
JonasCir: Good point, thanks. I think this is a fundamental utility method which should be callable unconditionally.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/AbstractSampleView.java
JonasCir: changed

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
JonasCir: fixed

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/AbstractEditAllowedDetailView.java
JonasCir: fixed

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/AbstractDetailView.java
JonasCir: fixed

MateStrysewske:
@JonasCir Code looks good to me, can you resolve the merge conflict please?
@StefanKock Can you do another review?

JonasCir:
conflict resolved

MateStrysewske:
I'll have to postpone testing this because it's blocked by https://github.com/hzi-braunschweig/SORMAS-Project/issues/8541.

MateStrysewske (approved):

JonasCir:
@MateStrysewske Thx, can we merge or should we wait for a review from @StefanKock ? :)

MateStrysewske:
Wait for his approval please :)

StefanKock (approved):
Thx for waiting ;-) I have no objections.

JonasCir:
All good, thank you both! :rocket:

PR #10153, opened 2022-08-17
#10091 - Fix sending entities to reporting tool multiple times

### LINKED ISSUE #10091: [ReportingTool] Sending entities to the reporting tool creates two entries in the Reporting tool history
### Bug Description

When events or cases are sent to SurvNet, the reporting tool history contains 2 entries for the same sending process. Furthermore, when multiple events or cases are sent to SurvNet via bulk action, there seems so be two verification messages, that the entity has been sent successfully. 

### Steps to Reproduce

1. Open a case or event 
2. Click on send to reporting tool 
3. Check the reporting tool history 

### Expected Behavior

There should only be one entry in the reporting tool history for each time the entity is sent. 
### Screenshots

Double entry: 
grafik

Double verification message: 
grafik

### System Details

* Device:
* SORMAS version: 1.74.1
* Android version/Browser: Firefox
* Server URL: https://test1402.sormas.netzlink.com
* User Role: Admin

carina29:
Fixes #10091


JonasCir (approved):
lgtm thank you!

PR #10163, opened 2022-08-18
Feature 9645 linked default user role

### LINKED ISSUE #9645: User roles should have optional link to default user role [5]
### Problem Description
Default user roles are changed over time when new user rights are added, existing rights removed or user right configurations changed/fixed.
So far we have tried to migrate these changes by automatically applying new user rights to existing user roles based on their existing user rights. Most often this works fine, but sometimes it doesn't work,. An example for this is #8892: External lab users need to receive the right to see sensitive data, but there is no way to generically decide about this based on other user rights of a user role.

### Proposed Change

- [x] Give the user role entity a new field `linkedDefaultUserRole` that maintains a link to the default user role that was used to create this user role.
  For future migrations, when a change is made to the default user role, all user roles that are linked to it are changed accordingly.
- [x] The default user role + default user logic that is currently relying on the user captions needs to be changed to use this new fields instead, to make it more resilient. See implementation details.
- [x] Edit form: The user can either remove the link to the default user role or even set it for a user role that was manually created.
- [x] "Apply user role template" is visible in both the main and notifications edit view and opens a popup with a drop-down to select the user role, including the description we use in the creation form. When a "linked default user role" is selected in the edit form, it is used as the default value for the drop-down in the popup.
- [x] Creation form: Move the user role template info to the right side of the template selection drop-down.
- [x] Creating a user role based on a template or applying a template should also fill the "linked default user role" field. Either to the used default user role or to the linked default user role of the template.

#### Edit Form
grafik

#### Creation Form
grafik

### Acceptance Criteria
* The automatically created user roles on existing systems and on newly setup systems are correctly linked to the default user roles.

### Implementation Details

- [x] The INSERT statement in the 460 schema update should set the correct linked default user roles
- [x] StartupShutdownService.fillDefaultUserRoles should use the linked default user role instead of the caption
- [x] Same for StartupShutdownService.upgrade to 469 and 471
- [x] Same for the calls to StartupShutdownService.createOrUpdateDefaultUser and createAndPersistDefaultUser - which should probably be unified.
- [x] Add a new schema version update that set's the linkedDefaultUserRole based on the caption for those user roles that don't have it yet. This is only relevant for systems that have already rolled-out schema version 460+

BarnaBartha:
Fixes #9645


sormas-api/src/main/java/de/symeda/sormas/api/i18n/Strings.java
```diff
@@ -14,11 +14,13 @@ public interface Strings {
 	String address = "address";
 	String aggregateReportLegend = "aggregateReportLegend";
 	String and = "and";
+	String apply = "apply";
 	String at = "at";
 	String average = "average";
 	String between = "between";
 	String bpm = "bpm";
 	String by = "by";
+	String cancel = "cancel";
```
leventegal-she, line 23: please use `Captions.actionCancel`

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -11900,4 +11902,25 @@ UPDATE person_history SET birthcountry_id = NULL WHERE birthcountry_id IS NOT NU
...
+    ELSE
+        EXECUTE in_statement;
+        RETURN TRUE;
+    END IF;
+END
+$_$ LANGUAGE plpgsql VOLATILE;
+
+DO $$ BEGIN
+   PERFORM add_column('ALTER TABLE userroles ADD COLUMN linkeddefaultuserrole varchar(255)', 'userroles', 'linkeddefaultuserrole');
+   PERFORM add_column('ALTER TABLE userroles_history ADD COLUMN linkeddefaultuserrole varchar(255)', 'userroles_history', 'linkeddefaultuserrole');
```
leventegal-she, line 11921: the function more generic and generate the query using some variables, it also could handle adding the column on the history table if you want the function to be reusable. Otherwise it would be better to delete the function at the end of the migration.

BarnaBartha: i've changed it so that the parameters are used in the statement generation
i would not include the history table modification in it
we don't have a function for the normal addition either that modifies both tables at once, and there are some rare tables that don't have a history one

leventegal-she: ok

leventegal-she (approved):

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -3027,6 +3027,10 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				currentVersion = 340;
 				getDao(CustomizableEnumValue.class).executeRaw("ALTER TABLE customizableEnumValue ADD COLUMN defaultValue boolean;");
 
+			case 341:
+				currentVersion = 341;
+				getDao(UserRole.class).executeRaw("ALTER TABLE userRoles ADD COLUMN linkedDefaultUserRole varchar(255)");;
```
MateStrysewske, line 3032:
```suggestion
				getDao(UserRole.class).executeRaw("ALTER TABLE userRoles ADD COLUMN linkedDefaultUserRole varchar(255);");
```

sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserRoleService.java
```diff
@@ -77,6 +78,18 @@ public UserRole getByCaption(String caption) {
 		return entity;
 	}
 
+	public UserRole getByLinkedDefaultUserRole(DefaultUserRole linkedDefaultUserRole) {
```
MateStrysewske, line 81: Question: What happens if multiple user roles are linked to the same default user role? Do we prevent this from happening? Do we WANT to prevent this from happening? The logic currently doesn't seem to support something like that, but in practice I can absolutely see admins creating user roles that are very similar to another, with only minor differences, and linking them all to the same default user role.

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/user/UserRoleService.java
BarnaBartha: good point
since the method is only used in the startup service the current implementation findFirst should solve it
but i added an order on the creation date just to be sure and covered it with a test

sormas-api/src/main/resources/descriptions.properties
```diff
@@ -216,7 +216,8 @@ Visit.visitStatus = Select the visit status
 
 #UserRole
 userRoleTemplate=You can choose one of the existing or default SORMAS user roles to pre-configure it
+userRoleVersionUpdate=Version updates of the software may change the user rights ad notification settings of a default user role. When linked, these changes will also be applied to this user role.
```
MateStrysewske, line 219:
```suggestion
userRoleVersionUpdate=Version updates of the software may change the user rights and notification settings of a default user role. When linked, these changes will also be applied to this user role.
```

MateStrysewske (changes requested):
Only a small typo, everything else looks good! You can merge the PR after correcting it.

MateStrysewske:
Additionally, the hint "You can choose one of the existing or default SORMAS user roles to pre-configure it" in the web UI doesn't make much sense anymore since the button has been moved, so I'd say it should be moved to the popup as well.

PR #10181, opened 2022-08-22
Feature/9943 update vaccination status if necessary after delete

### LINKED ISSUE #9943: Inconsistency -When a vaccination is deleted the vaccination status remains VACCINATED [2]
### Problem Description

When a vaccination is deleted, in the description of the case the vaccination status will remain VACCINATED however when a case
is created and has no added vaccination, the status is empty,

New case without vaccinations:
image                                                                
image

Add a vaccination to the same case and delete the vaccination:
image
image

**Important:** This is an inconsistency in Sormas project which created an inconsistency in Survnet project too:
When the above mentioned case is sent to the Reporting Tool, vaccination status will be mapped differently and after a deleted vaccination the case will have VACCINATED status.

### Proposed Change

~~Add the same vaccination status(empty) to the cases without vaccination ( even if the case had a vaccination and it was deleted or the case never had a vaccination)~~

**For the reduced immunization module:**
When a vaccination is deleted (and there is no other vaccination for the corresponding immunization) the user should get asked, wether the vaccination status should be updated (deleted) or should be kept. Depending on what he clicks the vaccination status is updated or not.

carina29:
Fixes #9943


sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseFacade.java
```diff
@@ -220,6 +222,10 @@ void saveBulkEditWithFacilities(
 
 	List<CaseDataDto> getByExternalId(String externalId);
 
+    default List<CaseDataDto> getCasesForWhichVaccinationIsRelevant(VaccinationDto vaccination){
```
MateStrysewske, line 225: Incorrect code formatting; please ensure that you have the correct formatting settings and have enabled the save actions plugin (and set it to reformat your code on save)

MateStrysewske, line 225:
```suggestion
    default List<CaseDataDto> getRelevantCasesForVaccination(VaccinationDto vaccination){
```

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -71,6 +71,8 @@
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 
+import com.vladmihalcea.hibernate.query.SQLExtractor;
```
MateStrysewske: Where is this import used?

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1388,6 +1390,28 @@ public List<CaseSelectionDto> getSimilarCases(CaseSimilarityCriteria criteria) {
 		return entries;
 	}
 
+	@Override
+    public List<CaseDataDto> getCasesForWhichVaccinationIsRelevant(VaccinationDto vaccinationDto) {
```
MateStrysewske: There is a major flaw with this logic: It doesn't take relevance into account at all, but will just return all vaccinations of the person. However, we need to filter them by a) disease (because cases with other diseases should NOT be considered), and also need to take into account the vaccination date in relation to the case date, because cases that have been reported earlier than the vaccination should not be affected (see VaccinationService.isVaccinationRelevant).

sormas-api/src/main/resources/strings.properties
```diff
@@ -220,6 +220,7 @@ confirmationRemoveGridRowTitle=Confirm remove
 confirmationRemoveGridRowMessage=Are you sure you want to remove this row?
 confirmationRemoveGridRowConfirm=Yes
 confirmationRemoveGridRowCancel=No
+confirmationVaccinationUpdate = Do you want to update the vaccination status?
```
MateStrysewske, line 223: This is way too few information for the user. Which vaccination status? Update to what? This needs to be specified so that they can make an informed decision.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1388,6 +1389,28 @@ public List<CaseSelectionDto> getSimilarCases(CaseSimilarityCriteria criteria) {
 		return entries;
 	}
 
+	@Override
+	public List<CaseDataDto> getCasesForWhichVaccinationIsRelevant(VaccinationDto vaccinationDto) {
```
MateStrysewske, line 1393: There are some major flaws with this logic. First of all, it doesn't take into account disease and vaccination date in relation to the case date, i.e. with this logic, cases that aren't relevant for the vaccination at all will be updated. We can only consider cases with the same disease and with a date that makes them relevant for the vaccination (see VaccinationService.isVaccinationRelevant).

Secondly, if I'm not mistaken, this will include cases that, even after the change, will in fact still have relevant vaccinations. I guess this is the part of the logic that you meant is still buggy, but I don't think we can release it in this state because this will mess up data.

MateStrysewske (changes requested):
In addition to the requested changes, we need a unit test for the new backend method in order to ensure that the logic works properly, and still works after potential code changes in the future.

sormas-api/src/main/java/de/symeda/sormas/api/vaccination/VaccinationFacade.java
```diff
@@ -95,4 +98,8 @@ List<VaccinationListEntryDto> getEntriesListWithRelevance(
 	VaccinationDto getByUuid(String uuid);
 
 	VaccinationDto postUpdate(String uuid, JsonNode vaccinationDtoJson);
+
+	default boolean isVaccinationRelevant(CaseDataDto caze, VaccinationDto vaccination) {
```
MateStrysewske, line 102: This method should not have a default return value

sormas-api/src/main/java/de/symeda/sormas/api/vaccination/VaccinationFacade.java
```diff
@@ -64,6 +65,8 @@ VaccinationDto createWithImmunization(
 
 	List<VaccinationListEntryDto> getEntriesList(VaccinationListCriteria criteria, Integer first, Integer max, List<SortProperty> sortProperties);
 
+	List<VaccinationDto> getVaccinationListWithRelevanceForACase(CaseDataDto cazeDto);
```
MateStrysewske, line 68:
```suggestion
	List<VaccinationDto> getRelevantVaccinationsForCase(CaseDataDto cazeDto);
```

sormas-api/src/main/resources/strings.properties
```diff
@@ -236,6 +236,7 @@ confirmationRevokeSormasToSormasShareRequest = Are you sure you want to revoke t
 confirmationSinceExternalMessages = This is the first time messages will be fetched. Do you want to select the start date from which to fetch messages? If you select no, all messages will be retrieved.
 confirmationSeeAllPersons=Are you sure you want to search persons for all association types? This could result in a slow response.
 confirmationExternalMessageCorrection = This message contains corrections to a previous one.<br/>Do you want process  those corrections from this message?
+confirmationVaccinationStatusUpdate = Do you want to update the case's vaccination status field to empty?
```
MateStrysewske, line 239:
```suggestion
confirmationVaccinationStatusUpdate = Deleting this vaccination has lead to at least one case no longer having a valid vaccination. Do you want to remove the vaccination status from all affected cases?
```

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1388,6 +1389,36 @@ public List<CaseSelectionDto> getSimilarCases(CaseSimilarityCriteria criteria) {
 		return entries;
 	}
 
+	@Override
+	public List<CaseDataDto> getRelevantCasesForVaccination(VaccinationDto vaccinationDto) {
```
MateStrysewske, line 1393: The relevance in terms of dates needs to be added to this method so that only relevant cases are returned, i.e. the resulting case list needs to be filtered accordingly.

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -1388,6 +1389,36 @@ public List<CaseSelectionDto> getSimilarCases(CaseSimilarityCriteria criteria) {
...
+		List<Case> cases = em.createQuery(cq).getResultList();
+		List<CaseDataDto> caseDataDtos = cases.stream().map(this::toDto).collect(Collectors.toList());
+		return caseDataDtos;
+	}
+
+	@Override
+	public boolean hasOtherValidVaccination(CaseDataDto caze, String vaccinationUuid) {
+		List<VaccinationDto> relevantVaccinationsForCase = vaccinationFacade.getVaccinationListWithRelevanceForACase(caze);
+		//checking if the vaccination selected for delete is between the relevant vaccinations of the case
+		VaccinationDto vaccination = relevantVaccinationsForCase.stream().filter(v -> v.getUuid().equals(vaccinationUuid)).findFirst().orElse(null);
```
MateStrysewske, line 1418: You should be able to shorten this to: 

`return relevantVaccinationsForCase.stream().filter(v -> !v.getUuid().equals(vaccinationUuid)).count() > 0`

sormas-backend/src/main/java/de/symeda/sormas/backend/vaccination/VaccinationService.java
```diff
@@ -244,6 +246,24 @@ private boolean isVaccinationRelevant(Vaccination vaccination, Date... relevance
 		return false;
 	}
 
+	public List<Vaccination> getVaccinationListForACase(Case caze) {
```
MateStrysewske, line 249:
```suggestion
	public List<Vaccination> getRelevantVaccinationsForCase(Case caze) {
```

This method should already filter the vaccinations by relevance, this should not be outsourced to the method in VaccinationFacadeEjb; I can't think about any scenario in which we want to retrieve vaccinations without relevance here. i.e., the `isVaccinationRelevant` method needs to be called here already.

sormas-ui/src/main/java/de/symeda/sormas/ui/vaccination/VaccinationController.java
```diff
@@ -129,6 +135,10 @@ public void edit(
 			createComponent.addDeleteWithReasonListener((deleteDetails) -> {
 				popupWindow.close();
 				if (doSave) {
+					List<CaseDataDto> cases = getCaseFacade().getRelevantCasesForVaccination(vaccination);
```
MateStrysewske, line 138:
```suggestion
List<CaseDataDto> cases = getCaseFacade().getRelevantCasesForVaccination(vaccination).filter(c -> getCasesFacade().hasOtherValidVaccination(c, vaccination.getUuid())).collect(Collectors.toList());
```

And then use this list for all the other stuff that is done afterwards (which of course also means that you don't need to check for valid vaccinations again in the next line, the `isEmpty` check is sufficient then).

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
carina29: I've filtered the returned list from this methods in VaccinationController ( as you've also suggested in one of your comments)

carina29:
> In addition to the requested changes, we need a unit test for the new backend method in order to ensure that the logic works properly, and still works after potential code changes in the future.

I've finished and pushed the requested changes. I've also tested the changes and everything works fine for relevant and not relevant vaccinations too. I will add some unit tests now.

MateStrysewske (approved):

PR #10195, opened 2022-08-24
#9787 S2S_Revision of the case directory and contact directory_Part 1

### LINKED ISSUE #9787: S2S_Revision of the case directory and contact directory_Part 1
### Problem Description

With the implementation of S2S the case directory and contact directory shows also cases/contacts which no longer belongs to the own health department/source system because they were shared with hand over the ownership to another target system.

The count oft the buttons "All" "Investigation pending" "Investigation done" "Investigation discarded" includes the cases which no longer belongs to the own source system. 
The count oft the buttons "All" "Active contact" "Converted to case" "Dropped" includes the contacts which no longer belongs to the own source system. 

-> That is very confusing for the health departments because if for example the button "all" shows the number 100, they think that they have 100 cases/contacts although it might be that 20 of this cases/contacts yet belong to another system.
So if statistics are made with this counts, the statistics are false. 
### Proposed Change

- [ ] To the left of the dropdown for filtering by active cases, another dropdown is added. This is only visible if the SORMAS 2 SORMAS module is activated in the instance.
It has three selection options:

| Selection | Behaviour |
| :--- | :--- |
| own cases (Eigene Fälle) | Filters the case directory for cases where the current instance is the owner of the data. <br/> (Default selection when opening the case directory) |
| For view (Zur Ansicht) | Filters the case directory for cases that have been transferred "for viewing" (with **no** hand over the ownership) by other instances. |
| All cases | Both (own cases and cases for view) are displayed |

This filter can be freely combined with other filters and applies in addition to any other filter options that may have been set.

- [ ] To the left of the dropdown for filtering by active contacts, another dropdown is added. This is only visible if the SORMAS 2 SORMAS module is activated in the instance.
It has three selection options:

| Selection | Behaviour |
| :--- | :--- |
| own contacts (Eigene Kontakte) | Filters the contact directory for contacts where the current instance is the owner of the data. <br/> (Default selection when opening the contact directory) |
| For view (Zur Ansicht) | Filters the contact directory for contacts that have been transferred "for viewing" (with **no** hand over the ownership) by other instances. |
| All cases | Both (own cases and cases for view) are displayed |

This filter can be freely combined with other filters and applies in addition to any other filter options that may have been set.

### Additional Information

image

leventegal-she:
Fixes #9787


JonasCir (approved):
lgtm

PR #10209, opened 2022-08-26
#9437 - Added synchronization dialog with detailed info

### LINKED ISSUE #9437: Add status indication during the synchronization process in the Mobile app [5]
### Problem Description

When we try to synchronize in the mobile app, there is no indication that how many cases, contacts, tasks, etc need to be synchronized. There is just a loading icon that does not tell any information that how much data or timing is remaining. In this way, the end-user is just blindly waiting and does not know how much time it needs or how much data it needs to be synced yet.  

### Proposed Change

Add some sort of progress bar or a Sync Status tab where the user can see, when it was last synced, and if the synchronization is in progress then how much data is already synced and ~~how much is remaining~~, and ~~how much time is needed more based on current bandwidth~~. This type of information will also help us to improve the mobile app user experience and performance.

#### Details
The main idea is to:

1. Display the steps that are executed during a synchronization (sending new data, receiving infrastructure data, ...).
2. For each step display which entities are being synchronized and how many of them (only for pushed data and for data that has been pulled)

Screenshot of how it looks like after implementation:

syncscreen

- [ ] When a synchronization is started, list the steps for the user, as listed below
- [ ] When data is pushed, also display how many entities are being pushed
- [ ] When data is pulled display how many entities already have been pulled during the current sync process
- [ ] Update the displayed data after every server roundtrip

The steps that are executed are different based on the three synchronization types:

#### sync changes

1. Send new data
2. Pull infrastructure data
3. Delete obsolete data (once per day)
4. Synchronize data
5. Pull modified data

Step 5 is needed, because pushing data may change other data that should be pulled afterwards. Not sure how to name this, so users aren't confused why data is pulled twice.

#### sync complete

1. Pull infrastructure data
2. Pull uuids of infrastructure to delete invalid and pull missing
3. Send new data
4. Pull uuids of all data to delete invalid and pull missing
5. Synchronize data
6. Pull modified data

#### sync complete and re-pull

1. Pull infrastructure data
2. Pull uuids of infrastructure to delete invalid and pull missing
3. Re-pull all data
4. Pull uuids of all data to delete invalid
5. Synchronize data
6. Pull modified data

### Acceptance Criteria

* Users can see what type of data is currently being synchronized
* Users can see the counts for data that has been synchronized
* Synchronization does not leave the user in a situation where nothing is happening for 20 seconds or more.

### Implementation Details

- [ ] AbstractAdoDao needs to provide count methods for queryForNew and queryForModified. Those are meant to provide the counts needed to display what is going to be pushed
- [ ] We need a representation of the steps (send, pull, ...) and sub-steps (pull persons, push persons, ...) that are done during synchronization. This representation is used to update the progress dialog. In a separate issue we may want to temporarily keep this information to allow users to see the progress of the last or current synchronization that was done
- [ ] The representation needs to be updated after every server communication during synchronization

### Additional Information

The screenshot is attached for current implementation.
<details>
Screenshot_20220601-141045_APMIS
</details>

The dto helper classes should ideally be stateless, but currently have lastSyncedEntityDate + Uuid fields. This may have to be changed, because it's possibly negatively affecting the re-pull logic. Discuss with @BarnaBartha and decide whether an issue should be created for this.

### Follow-up issues

1. Allow users to see details on the last or current (if done in background) sync process
2. Display actual numbers of entities to be pulled from the server
3. Calculate an approximate time for the whole sync process

MateStrysewske:
Fixes #9437


MateStrysewske:
@BarnaBartha @MartinWahnschaffe I've worked on this for two weeks now and I'm sure there's some major flaws in it that I'm just not able to see right now. Please make sure to test this after doing a code review if possible.

BarnaBartha (approved):

BarnaBartha:
@MateStrysewske I've tested and sync worked out well for me.
Must say it is the most beautiful feature in Sormas so far, takes the android app to a whole next level.
Nice job!

sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/SynchronizationDialog.java
```diff
@@ -0,0 +1,589 @@
...
+
+	private void showProgressItems(boolean showPulls, boolean showPushes, boolean showDeletions, List<String> captions) {
+
+		getActivity().runOnUiThread(() -> {
+			clearBindings();
+			LayoutInflater inflater = (LayoutInflater) getActivity().getSystemService(Context.LAYOUT_INFLATER_SERVICE);
+
+			captions.forEach(caption -> progressItemBindings.add(createBinding(inflater, caption, showPulls, showPushes, showDeletions)));
+
+			currentBinding = progressItemBindings.get(0);
```
MartinWahnschaffe, line 210: Please move these two calls into the synchronized block. I now it's not relevant in the current situation, but since the wait logic depends on the value of currentBinding, this should be within the block.

sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/SynchronizationDialog.java
```diff
@@ -0,0 +1,589 @@
...
+
+		getActivity().runOnUiThread(() -> {
+			progressItemBindings.clear();
+			progressItemBindingsIndex = 0;
+			contentBinding.syncProgressLayout.removeAllViews();
+			currentBinding = null;
+		});
+	}
+
+	private void waitForUiThread() {
```
MartinWahnschaffe, line 469:
```suggestion
	private void waitForProgressItemsBinding() {
```
Maybe that's better?

MartinWahnschaffe (changes requested):
See comments

sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/SynchronizationDialog.java
```diff
@@ -0,0 +1,585 @@
...
+
+	private void showProgressItems(boolean showPulls, boolean showPushes, List<String> captions) {
+
+		showProgressItems(showPulls, showPushes, false, captions);
+	}
+
+	private void showProgressItems(boolean showPulls, boolean showPushes, boolean showDeletions, List<String> captions) {
+
+		getActivity().runOnUiThread(() -> {
+			assert (currentBinding == null);
```
MartinWahnschaffe, line 207: This lead to an error on my system, which made me aware of the following situation:

When updateSynchronizationStep is called twice, before the UI thread is executed again, it will mean that the logic in showProgressItems is also executed twice, once the UI thread is run. For the second execution the currentBinding will not be null.

Unfortunately simply setting currentBinding to null here will also not solve the problem. The first UI thread execution for showProgressItems will notify the waiting calls that access the currentBinding. If the second exeuction of showProgressItems is running in parallel it will take away the currentBinding and either lead to wrong progress data or worst case a nullpointerexception. @MateStrysewske Let's discuss how we can mitigate this.

MartinWahnschaffe (changes requested):

sormas-app/app/src/main/java/de/symeda/sormas/app/component/dialog/SynchronizationDialog.java
```diff
@@ -0,0 +1,610 @@
...
+		waitForProgressItemBindings();
+		currentBinding.setPushTotal(pushTotal);
+	}
+
+	private void updateDeletions(int deleted) {
+		waitForProgressItemBindings();
+		currentBinding.setDeletions(currentBinding.getDeletions() + deleted);
+	}
+
+	private void showNextCleanupItems() {
```
MartinWahnschaffe, line 216: I'd like to get rid of this method. It splits the updatingProgressItemBindings logic into a second "branch" and I also didn't get why it's needed in the first place.

As far as I understand why could replace it by adding another two entries in SynchronizationStep enum (e.g. CLEAR_UP_INFRASTRUCTURE_MISSING and CLEAR_UP_MISSING) and calling updateSynchronizationStep instead.

MartinWahnschaffe (changes requested):
@MateStrysewske I have fixed the thread sync issue and have identified one last thing I'd like to be changed. Please check my comment and see if it makes sense.

Also please have a look into the original issue and check what has been done.

MartinWahnschaffe (approved):
Thanks! It's much clearer now.

PR #10223, opened 2022-08-29
#9645 - pre-fill popup with linked default user role

### LINKED ISSUE #9645: User roles should have optional link to default user role [5]

BarnaBartha:
- fill default user role on creation of user role

Fixes #9645

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserRoleNotificationsForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserRoleNotificationsForm.java
@@ -17,6 +17,7 @@ package de.symeda.sormas.ui.user;
 
 import static de.symeda.sormas.ui.utils.LayoutUtil.fluidRowLocs;
 
+import de.symeda.sormas.api.user.DefaultUserRole;
 import de.symeda.sormas.api.user.UserRoleDto;
 
 public class UserRoleNotificationsForm extends AbstractUserRoleForm {
@@ -44,5 +45,10 @@ public class UserRoleNotificationsForm extends AbstractUserRoleForm {
 		if (templateRole != null) {
 			notificationTypesCbSet.setInternalValue(templateRole.getNotificationTypes());
 		}
+	}
+
+	@Override
+	DefaultUserRole getDefaultUserRole() {
+		return null;
 	}
 }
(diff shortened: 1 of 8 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserRoleNotificationsForm.java
```diff
@@ -45,4 +46,9 @@ void applyTemplateData(UserRoleDto templateRole) {
 			notificationTypesCbSet.setInternalValue(templateRole.getNotificationTypes());
 		}
 	}
+
+	@Override
+	DefaultUserRole getDefaultUserRole() {
+		return null;
```
leventegal-she, line 52: This should return the linked default user role of the role being edited

leventegal-she (changes requested):

leventegal-she (approved):

PR #10238, opened 2022-08-30
[#10110] call validate for events and move unecessary checks to DTO

### LINKED ISSUE #10110: The missing reporting user validation is not triggered for creating Event
### Bug Description
The missing reporting user validation is not triggered for creating Event

### Steps to Reproduce
1. Make a POST for the endpoint to created event( env_url/sormas-rest/events/push)
2. In the body of the request don't add the attribute : reporting user 
[
    {
        "eventTitle": "JustFromAPI",
        "eventLocation": {
            "uuid": "e8ed5a4e-c5af-4ef9-ac6f-3004f2c4eba8",
            "pseudonymized": false,
            "region": {
                "uuid": "XD5CCG-2K2NRA-E7VK75-ME622CX4",
                "caption": "First Region"
            }
        },
        "eventStatus": "EVENT",
        "eventInvestigationStatus": "PENDING",
        "reportDateTime": 1653385889240
    }
]

### Actual Behavior

The request responds with the status code 200 and response body
[
    "TRANSACTION_ROLLED_BACK_EXCEPTION"
]
### Expected Behavior

The request should responds with the status code 400 and response body
"Invalid data",
    "errors": {
        "reportingUser": "You have to specify a valid user from the SORMAS database"
    }

### System Details

* Device:
* SORMAS version:
* Android version/Browser:
* Server URL:
* User Role:

JonasCir:
Fixes #10110


JonasCir:
HTTP status code is still 200 in case of error...

sormas-api/src/main/java/de/symeda/sormas/api/event/EventDto.java
```diff
@@ -120,11 +122,11 @@ public class EventDto extends SormasToSormasShareableDto {
 	public static final String OTHER_DELETION_REASON = "otherDeletionReason";
 
 	private EventReferenceDto superordinateEvent;
-
-	@Required
+	@NotNull(message = Validations.validEventStatus)
```
MartinWahnschaffe, line 125: The `@Required` annotation is currently used in EntityColumn.getRequired. It's ok to replace it here, but then we also have to extend this method to check for both annotations.

I'd also suggest to add a comment to the Required annotation to use NotNull instead and maybe even deprecate it.

MartinWahnschaffe (changes requested):

sormas-api/src/main/java/de/symeda/sormas/api/event/EventDto.java
JonasCir: I deprecated `@Required`. I wanted to do this for two years now... :smile:

sormas-backend/src/main/java/de/symeda/sormas/backend/info/EntityColumn.java
```diff
@@ -49,7 +51,7 @@ public enum EntityColumn {
 	DATA_PROTECTION(256 * 30, EntityColumn::getDataProtection, false, true, false),
 	CAPTION(256 * 30, EntityColumn::getCaption, false, true, false),
 	DESCRIPTION(256 * 60, EntityColumn::getDescription, true, true, false),
-	REQUIRED(256 * 10, EntityColumn::getRequired, false, true, false),
```
MartinWahnschaffe, line 52: Let's keep the enum name "REQUIRED" otherwise translation will have to be updated.

JonasCir:
fixed

MartinWahnschaffe:
@JonasCir Sorry, two last things:
1. The Android CI didn't run through for 24 and 25. This may just be a CI thing, so I have now trigger a re-run of 24. Please check the results. 
2. Please execute DataDictionaryGenerator to update the data dictionary with the latest changes (e.g. eventInvestigationStatus and eventTitle should now be marked as required in the excel file).

MartinWahnschaffe:
@JonasCir Please double check. Looks like the included dictionary still doesn't list Event.eventInvestigationStatus as required.

JonasCir:
@MartinWahnschaffe fixed now, you can review the last commit :)

MartinWahnschaffe (approved):

PR #10244, opened 2022-08-30
#3845 - fix event participant index list sorting

### LINKED ISSUE #3845: Add lazy loading to event participants grid [1]
### Problem Description

- Opening the ep tab on with 1040 ep takes 1 min. To Reproduce:
1. connect to test1401
2. open the ep tab for event with uuid = WGYIVS-YEGH3C-K5DAWP-6WTICAKY

Expected Behavior:
- Should take about 2 seconds. 
- This issues may also apply to the all sub tabs like contact within a case. Thus we also need to apply the solution (maybe lazy loading) to the contact tab within a case.

### Proposed Change

- [x] Add support for lazy loading to the event participants grid (this includes adding a button to switch between both alternatives like in other grids)
- [x] Make lazy grid the default for event participants and case contacts

### Additional Information

* SORMAS version: 1,52,1
* Server URL: https://test1401.sormas-oegd.de/sormas-ui/#!events/eventparticipants/WGYIVS-YEGH3C-K5DAWP-6WTICAKY
* @StefanKock: Event participants grid is only concerning one event. It's likely that the user filters cause delays, but not the amount of participants per event.

BarnaBartha:
Fixes #3845


sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
```diff
@@ -533,8 +533,6 @@ public List<EventParticipantIndexDto> getIndexList(
 				.and(cb, filter, cb.equal(samples.get(Sample.PATHOGEN_TEST_RESULT), eventParticipantCriteria.getPathogenTestResult()));
 		}
 
-		cq.distinct(true);
```
MateStrysewske, line 536: What's the reason for removing the distinct clause here?

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
```diff
@@ -562,18 +560,19 @@ public List<EventParticipantIndexDto> getIndexList(
...
 					expression = joins.getSamples().get(SampleIndexDto.SAMPLE_DATE_TIME);
-					order.add(sortProperty.ascending ? cb.asc(expression) : cb.desc(expression));
 					break;
 				case EventParticipantIndexDto.FIRST_NAME:
 					expression = person.get(sortProperty.propertyName);
 					break;
 				case EventParticipantIndexDto.CASE_UUID:
 					expression = resultingCase.get(Case.UUID);
 					break;
+				case EventParticipantIndexDto.CONTACT_COUNT:
```
MateStrysewske, line 573: This a) doesn't make sense to me (because you're adding sorting to a property, but then sort by something entirely different), and b) doesn't have any effect because the column is set as not sortable anyway. Should be removed.

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
BarnaBartha: was left there by mistake, saw the column is not sortable after and forgot to remove the code

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
BarnaBartha: it is not needed, was added by Sergiu trying to fix the problem, so i didn't see a reason why it would be needed.

leventegal-she (approved):

PR #10245, opened 2022-08-30
#9807 S2S_Case ID/Contact ID should be clickable and open case/contact afterwards

### LINKED ISSUE #9807: S2S_Case ID/Contact ID should be clickable and open case/contact afterwards [2]
### Problem Description

Currently it is not possible for a user to open a case or contact by clicking on the case ID /contact ID in the 'share request details' pop-up window. That is a problem because the user is accustomed to this behaviour. Furthermore this is the only option for the user to open a case/contact after accepting a share (if he/she does not want to search for it manually in the case/contact directory).
 
### Proposed Change

**_After_** successful acceptance of the share, the case/ contact can be opened by clicking on the case/contact ID.

### Implementation Details

- It can only be clicked and navigate to the page, when the case/contact exists in the system.
- If the case/contact does not exist, you get a popup that says that you have to accept the share request first

### Additional Information

image

leventegal-she:
Fixes #9807


sormas-api/src/main/resources/strings.properties
```diff
@@ -1229,6 +1229,8 @@ messageVaccinationNotRelevantForContact = This vaccination is not relevant for t
 messageVaccinationNoDateNotRelevantForContact = This vaccination is not relevant for this contact because it does not have a vaccination date.
 messageVaccinationNotRelevantForEventParticipant = This vaccination is not relevant for this event participant because the vaccination date is after the event date or event report date.
 messageVaccinationNoDateNotRelevantForEventParticipant = This vaccination is not relevant for this event participant because it does not have a vaccination date.
+messageAcceptRequestToNavigate = The request is not yet accepted. Please accept it to be able to navigate.
```
ChristopherRiedel, line 1232: Please change this to:  The request is not yet accepted. You have to accept it first to be able to navigate there.

ChristopherRiedel (changes requested):
Only a little change requested, everything else is fine.

ChristopherRiedel (approved):

PR #10297, opened 2022-09-07
Provide user rights documentation #10200

### LINKED ISSUE #10200: Provide user rights documentation [1]
### Problem Description
The new user role management (#898) allows admins to define which user rights each user role should have. For this they need to understand what each right means and also how user rights are related (#5058).
[SORMAS_User_Rights.xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-api/src/main/resources/doc/SORMAS_User_Rights.xlsx) exists but does not cover all aspects of this and is not reachable from within the software and not translated to other languages.

### Proposed Change

1. - [x] In the user role edit view extend the tooltips (added by #10084) to also list which other user rights are needed. Example:

>   Able to delete cases.
>   Needed user rights: View cases, Delete tasks from the system, Delete samples from the system, ...

2. - [x] Extend the [SORMAS_User_Roles.xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-api/src/main/resources/doc/SORMAS_User_Roles.xlsx) with another **sheet** that lists all details for the user rights. It should look like the no longer generated [SORMAS_User_Rights.xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/sormas-api/src/main/resources/doc/SORMAS_User_Rights.xlsx), but also contain columns for the group and the needed user rights. The columns for the default user roles should be removed.
  The original source code can be found here: https://github.com/hzi-braunschweig/SORMAS-Project/pull/9794/files#diff-5407cc1f13fe3347221c88ea06f67963075295fad0120fb92717c955756cd711
grafik
Example of the final document: 
  - [x] In the user roles sheet use a pleasant green for the "yes" cells and not background color at all for the "no" cells. Also don't use those cell borders. It should look like this    
grafik
  - [x] Remove SORMAS_User_Rights.xlsx from the repository
  - [x] Update https://github.com/hzi-braunschweig/SORMAS-Project/blob/development/docs/SERVER_CUSTOMIZATION.md#user-role-configuration accordingly

### Already in Place
3. Make a translated version of this Excel document accessible via the user role management

### Out of Scope
4. As an alternative visualization we could generate a markdown text document explaining the user roles, that would look something like this and could be part of the github repository and also be provided as localized information to the user.

> ## Case Surveillance
> ### View existing cases
> Allow users to view cases based on their jurisdiction.
> ### Edit existing cases
> Allow users to edit cases they can view.
>
> Required user right: [View existing cases](View-existing-cases).

- [ ] #10295

### Additional Information

* Could go one step further and also display right hierarchies

leventegal-she:
Fixes #10200


MartinWahnschaffe:
@leventegal-she Please execute UserRolesGenerator.generateUserRoles, so the SORMAS_User_Roles.xlsx in the repository is updated and can easily be reviewed as part of this PR.

FredrikSchaeferVitagroup (approved):

Issue #10128, opened 2022-08-15
Aggregate reports do not disappear immediately from the grid once they are deleted


roxanamlendea:
### Bug Description
Recently deleted aggregated reports do not immediately disappear from the list; they disappear only after a refresh.

### Steps to Reproduce

1. Navigate to the mSERS directory, and select the 'Report Data' tab;
2. Click on the Edit icon of one of the entries of the grid;
3. Click on the delete button and Save;

### Expected Behavior
The aggregate report should disappear immediately from the grid.

### Actual Behavior
The aggregate report disappears only after a refresh.

### Screenshots

![aggregate report does not disappear instantly](https://user-images.githubusercontent.com/89522688/184623583-708be098-780f-43c5-91ef-fc90876ffbfb.gif)

### System Details

* Device: Windows 10
* SORMAS version: Sormas 1.75.0-SNAPSHOT (a6ecbf1)
* Android version/Browser: Chrome
* Server URL: test.sormas.netzlink.com
* User Role: Admin, NatUser

### Additional Information
Similar to #10008

MateStrysewske:
Likely the same issue as #10008, so I'll add a comment there and close this one.

MateStrysewske:
Edited reports are also not updated immediately, but only after a refresh, see #10246.

adinaflorea9:
Verified ticket on the local machine using the latest version of SORMAS from the development branch - 1.76.0-SNAPSHOT(0c95d19).

PR #10344, opened 2022-09-13
#10128 Refresh aggregate reports after edit and deletion

andrei7bal:
#10128 Refresh aggregate reports after edit and deletion


sormas-ui/src/main/java/de/symeda/sormas/ui/reports/aggregate/ReportDataGrid.java
```diff
@@ -99,8 +104,44 @@ private Button createDeleteButton(AggregateReportDto aggregateReport) {
 		return null;
 	}
 
-	public void reload() {
+	public boolean navigateTo(BaseCriteria criteria, boolean force) {
+
```
sergiupacurariu, line 108: please remove this method and pass the navigateTo method in the constructor of ReportDataGrid. The ReportDataGrid is initialized in ReportDataView which has a navigateTo.

sergiupacurariu (approved):

Issue #10377, opened 2022-09-15
Annotation-based auditing


JonasCir:
### Problem Description

For example, if we change the user name, the action gets log through the auditor, however, the name is not included in the log line. This shows that we need a mechanism which allows us two tweak the representation in the audit log in order to customize the result

### Proposed Change

Introduce an annotation which can be put onto fields and methods to build the representation in the log. All annotated fields/result of the method when called will be included in the log line.

### Acceptance Criteria

- Adjustable representation
- easy to use
- future-proof

AndyBakcsy-she:
Validated ticket in a peer testing session with Jonas

StefanKock:
I removed the usage of `reflections` because we suspect it is causing somehow a blocked deployment on the TEST stage.
Sometimes deployment of sormas-ear seems to not finish, the following log message does not appear:
```
Reflections took 460 ms to scan 2 urls, producing 126 keys and 5557 values (org.reflections.Reflections.java:219)
```

StefanKock:
The deployment with the removed reflection usage did work for instances that failed sometimes in the last days: https://jenkins.sormas.netzlink.com/view/SORMAS-Project/job/sormas-DeployToTest/392/console

Issue #9710, opened 2022-07-01
[AUDIT] Important, type-specific information does not show up in the audit log [2]


JonasCir:
### Problem Description
For example, if we change the user name, the action gets log through the auditor, however, the name is not included in the log line. This shows that we need a mechanism which allows us two tweak the representation in the audit log.

### Proposed Change
~Introduce an interface which extends `HasUuid` and let all DTOs implement it. The default implementation should return only the UUID, but in case we need more information a more specific representation should be returned.~

Introduce an annotation which can be put onto fields and methods to build the representation in the log 

### Acceptance Criteria
* No `toString` based representation, for the most important types
* Adjustable representation
* Do it for the user name
* Architecture test

### out of scope
* No `Object@JVM-Address` style strings in log

### Split tickets
- [x] #10335
- [x] #10364
- [x] #10377
- [x] #10403
- [x] #10369 is discarded ~necessary or introduced architecture test fails~

JonasCir:
@MartinWahnschaffe FYI :)

AndyBakcsy-she:
Validated ticket during a peer testing session with Jonas

PR #10386, opened 2022-09-19
Feature/10377 annotation auditing

JonasCir:
Had to close the other PR.
Fixes https://github.com/hzi-braunschweig/SORMAS-Project/issues/10377

I will file another PR which puts the introduced annotations on the DTOs to finish https://github.com/hzi-braunschweig/SORMAS-Project/issues/9710 next week.

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/EntityDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/EntityDto.java
@@ -25,6 +25,8 @@ import org.apache.commons.lang3.StringUtils;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 
+import de.symeda.sormas.api.audit.AuditInclude;
+import de.symeda.sormas.api.audit.AuditedClass;
 import de.symeda.sormas.api.i18n.Validations;
 import de.symeda.sormas.api.utils.FieldConstraints;
 import de.symeda.sormas.api.utils.Outbreaks;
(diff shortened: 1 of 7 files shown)


MartinWahnschaffe:
I really like the implementation! Including the Reflections lib makes it very readable.
The code looks like it should be quite fast, but I'd still like you to do a quick performance check, e.g. by calling this for one or multiple different methods of the CaseFacade 1000x and measuring how much time executing the logging takes in average.
In addition we need to make developers aware of this annotation, so we should add this to the documentation.

JonasCir:
@MartinWahnschaffe I have the results ready:
I measured the insertion of 1000 distinct cases (one after one) via the REST API to measure the impact on the whole stack.
This is the result of baseline (latest `development`) which uses the old logic:
cases_1000_develop
The insertion of a cases takes on average ca 0.025 s while the best-case is around 0.021 s and worst case (ignoring the outliers b/c of cache warming) around 0.035s
Now the new implementation. I did not benchmark this branch but https://github.com/hzi-braunschweig/SORMAS-Project/pull/10407 which is based on this PR and adds the annotation to all DTOs.
cases_1000_new_annotation
The average is again around 0.025. Best case is a little bit better (around 0.02). Turns out our worst case (again ignoring outliers due to cache warming) is now around 0.04. Please note here that #10407 significantly audits more DTOs than the development implementation which in almost all cases straightly goes to `toString` without executing any logic at all. 

Since the average case (pun not intended) is comparable, this implementation is superior, and best case behavior is even better I propose to go ahead with this implementation.

MartinWahnschaffe (approved):

PR #10396, opened 2022-09-19
#10390 create test suite for external visits

### LINKED ISSUE #10390: Create test suite for external visits
### Problem Description

Currently the tests for external visit related backend methods are placed in several test classes and use the generic test user that has all user right and right restriction issues are not caught covered
### Proposed Solution

Create a test class that contains all external visit related backend tests.

leventegal-she:
Fixes #10390

Moved test cases from several test classes into a single test class and using the external visits user for calling backend facade methods related to the external visits resource.

### DIFF
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/visit/ExternalVisitTest.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/visit/ExternalVisitTest.java
@@ -1,0 +1,483 @@
+/*
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2022 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package de.symeda.sormas.backend.visit;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasSize;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.lang.reflect.InvocationTargetException;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Optional;
+
+import javax.persistence.EntityManager;
+import javax.persistence.EntityTransaction;
+
+import org.apache.commons.beanutils.BeanUtils;
+import org.junit.Test;
+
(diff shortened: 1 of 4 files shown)


sormas-backend/src/test/java/de/symeda/sormas/backend/visit/ExternalVisitTest.java
```diff
@@ -0,0 +1,483 @@
...
+import de.symeda.sormas.backend.person.Person;
+import de.symeda.sormas.backend.person.PersonFacadeEjb;
+import de.symeda.sormas.backend.person.PersonService;
+
+/*
+ * If you need to change these tests to make it pass, you probably changed the behaviour of the ExternalVisitsResource.
+ * Please note that other system used alongside with SORMAS are depending on this, so that their developers must be notified of any
+ * relevant API changes some time before they go into any test and productive system. Please inform the SORMAS core development team at
+ * https://gitter.im/SORMAS-Project!
+ */
```
FredrikSchaeferVitagroup, line 73: I'm not certain whether it is a good idea to put this comment only here. It would be more visible when being at the top of each test.

sormas-backend/src/test/java/de/symeda/sormas/backend/visit/ExternalVisitTest.java
```diff
@@ -0,0 +1,483 @@
...
+			setPersonRelevantFields(person);
+			person = entityManager.merge(person);
+		}
+	}
+
+	@Test
+	public void testCreateExternalVisit() {
+		PersonDto cazePerson = creator.createPerson("Case", "Person");
+		CaseDataDto caze = creator.createCase(
+			externalVisitsUser.toReference(),
```
FredrikSchaeferVitagroup, line 243: Maybe it's a more realistic approach to create the entities with a different user (e.g. nationalUser). The externalVisitsUser usually does not create them. This comment applies not only to this line, but to every entity creation in this test class.

FredrikSchaeferVitagroup (changes requested):

sormas-backend/src/test/java/de/symeda/sormas/backend/visit/ExternalVisitTest.java
leventegal-she: those entities are created with the test user, the external visit is used only for calling the method in test e.g. 
```
loginWith(externalVisitsUser);

visitFacade.saveExternalVisit(externalVisitDto);
visitFacade.saveExternalVisit(externalVisitDto2);
```

sormas-backend/src/test/java/de/symeda/sormas/backend/visit/ExternalVisitTest.java
```diff
@@ -0,0 +1,537 @@
...
+	 * If you need to change this test to make it pass, you probably changed the behaviour of the ExternalVisitsResource.
+	 * Please note that other system used alongside with SORMAS are depending on this, so that their developers must be notified of any
+	 * relevant API changes some time before they go into any test and productive system. Please inform the SORMAS core development team at
+	 * https://gitter.im/SORMAS-Project!
+	 */
+	public void testSetSymptomJournalStatus() {
+		PersonDto person = creator.createPerson();
+
+		loginWith(externalVisitsUser);
+		getPersonFacade().setSymptomJournalStatus(person.getUuid(), SymptomJournalStatus.REGISTERED);
```
FredrikSchaeferVitagroup, line 168: This method calls the [PersonFacadeEjb.savePerson(...)](https://github.com/hzi-braunschweig/SORMAS-Project/blob/7a4043b2df3617f24119e4ee541e8c2bf980ac06/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java#L418), which is annotated with `@RightsAllowed(UserRight._PERSON_EDIT)`. 
Isn't this a problem? The External_Visits_User does not have  this right by default. I wonder why this test passes...

FredrikSchaeferVitagroup (approved):

Issue #9852, opened 2022-07-14
Event participants created for archived events are active but listed in the 'Archived event participants' list


adinaflorea9:
### Bug Description

When the feature configuration EDIT_ARCHIVED_ENTITIES is set to 'True', for an archived event, new event participants can be added.

In the UI, these event participants are listed in the 'Archived Event Participants' list of the event. However, they are not archived - the 'Archive' button is visible on the event participant page.

In the DB, in the 'archived' boolean column, the value for these event participants is set to 'false' (i.e. they are active event participants).

### Steps to Reproduce

1. In the DB, set the feature configuration EDIT_ARCHIVED_ENTITIES to 'true';
2. In the web-app, create an event and archive it;
3. Navigate to the 'Event participants' tab and create a new event participant;
4. Observe the new event participant's page in the UI, and it's entry in the DB.

### Actual Behavior
 
The event participant is listed in the 'Archived event participants' list of the event.
On the event participant's page the 'Archive' button is still visible and it can be actioned by the user. 
In the DB, for the event participant's entry,  the value from the column 'archived' boolean is set to 'false' (i.e. they are an active event participant).

### Expected Behavior

Expected behavior to be discussed:
1. Is it desired that event participants created for events that are already archived should also be directly archived when they are created?

2. Is it desired that event participants created for events that are already archived should be active (as the current behavior is) and also listed in the 'Active event participants' list of the event?

### Screenshots

The archived event:
![image](https://user-images.githubusercontent.com/96909231/178943190-517ff3ab-fe5c-479e-994d-35bef81e799e.png)

The event participant:
![image](https://user-images.githubusercontent.com/96909231/178943262-b0ee5914-fe3a-495b-9faa-710b5e0fcea3.png)
![image](https://user-images.githubusercontent.com/96909231/178943288-04aa892e-9650-4cdf-80f2-5ffb140c46ef.png)

### System Details

* Device: Windows 10
* SORMAS version: 1.75.0-SNAPSHOT(7e4e9eb)
* Android version/Browser: Chrome
* Server URL: local machine
* User Role: Admin+NatUser

### Additional Information

@kwa20 perhaps you could please provide some input regarding the expected behavior here?

kwa20:
[...]

PR #10413, opened 2022-09-21
#9852 Add event participants as archived if the event is archived

andrei7bal:
#9852 Add event participants as archived if the event is archived


sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
```diff
@@ -358,6 +358,7 @@ public EventParticipantDto saveEventParticipant(@Valid EventParticipantDto dto,
 		validate(dto);
 
 		EventParticipant entity = fillOrBuildEntity(dto, existingParticipant, checkChangeDate);
+		entity.setArchived(event.isArchived());
```
MateStrysewske, line 361: This will also set archived event participants of active events to active, even if they are supposed to stay active (which could happen quite frequently for events that have never been archived, but still contain archived event participants). Please make sure to move this logic to the logic that archives/dearchives events, i.e. when an event is archived, also archive all of its event participants.

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
StefanKock: > Please make sure to move this logic to the logic that archives/dearchives events, i.e. when an event is archived, also archive all of its event participants.

The remark from @MateStrysewske does not match the expectation of the ticket to automatically archive an event participant when created on an archived event.
I think the `setArchived` should only be called when `existingParticipant == null´, and please add (or hopefully extend an existing unit test) to show that this behaviour is working as expected

StefanKock (changes requested):

andrei7bal:
Created a new pull request (#10498) since this branch contains unwanted commits.

PR #10442, opened 2022-09-26
#10352 hided all, contacts and cases buttons when user does not have …

### LINKED ISSUE #10352: Cases and contacts can be recognized as such in the person directory even if they are not supposed to be recognized
### Bug Description

- persons who have attached cases and contacts can be recognized as cases in the person directory even if they are not supposed to be recognized 
- the quick filter _all_ shows persons who only have one entity as case or one entity as contact
### Steps to Reproduce

1. Log in as an admin
2. Go to the person directory and make sure there are cases and contacts mentioned
3. Create a role _without_ the right to see cases and _without_ the right to see contacts, but with the right to see persons (role described under _additional information_ )
4. Create a user with this role and open the person directory

### Expected Behavior

1. On the left side there is the menu-points _persons_ 
2. On the left side there are _no_ menu-points _cases_ or _contacts_
3. In the person directory the quick filter _case_ and the quick filter _contacts_ are missing
4. Under the quick filter _all_ ONLY the persons should be visible who have at least one entity as _event participant_ or _travel entry_
5. Error message should not appear? 

Additional information: I focus here on cases and contacts but if you see the created user roles below you may recognize that the right VIEW_EVENT PARTICIPANT and VIEW_TRAVEL ENTRY is not given, too. So actually no person or quick filter should be visible.

### Implementation Details

- [ ] In the person directory each quick filter should only be visible when the user has access to the related _VIEW right.
- [ ] As quickfix before implementing #10414: The `PersonAssociation.ALL` option shall only be shown if all other `PersonAssociation` values are allowed.

### Screenshots

Description:
- no quick filter 'contact'
- If i can count i will know there are at least 3 contacts in the system and i also know which persons are the contacts, because i can compare the persons in the quick filter _cases_ with the persons under the quick filter _all_ and the another quick filters are 0
- The error message _Access to the specified resource has been forbidden. Please contact your supervisor or administrator and inform them about it. Please reload the page to see the latest changes_ appears
user role_1

Description: If you click on the quick filter _case_ and if you open a person you cannot see the entity _case_. That is good, but you KNOW that there is the entity because of the shown quick filter
user role_2

Description: This is shown with the user role _admin_:
user role_3

### System Details

* Device: 
* SORMAS version: 1.75
* Android version/Browser: Edge
* Server URL: release-x
* User Role: viewcase / password: the default one

### Additional Information

Maybe the right _PERSON_VIEW_ means, that _all_ persons in the systems are shown by adding this filter. 
We have two problems, if the right is defined like this:
1. Why is the quick filter _contacts_ than not visible?
2. What can a user do, if he wants that a user neither can see the case directory nor the persons who ONLY have a case entity in the person directory but do not want to completely abandon the person directory?

user role:
[sormas_benutzerrollen_2022-09-13_case view.xlsx](https://github.com/hzi-braunschweig/SORMAS-Project/files/9564826/sormas_benutzerrollen_2022-09-13_case.view.xlsx)

image
image
image
image
image
image
image
image

valentinmikleuvg:
…permission

Fixes #10352


StefanKock:
Missing as discussed in a meeting yesterday: 
1. When opening the `PersonsView`, open the first allowed `PersonAssociation` but skip the ALL option (`PersonsView` line 60).
2. Open association by URL still works.
3. Don't show data if no option is allowed.

StefanKock (changes requested):
Please check my last comment.

Additionally: The following test is not to be concerned of, and will not be a problem once this branch is rebased on `hotfix-1.75.3`
```
Error:  Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.072 s <<< FAILURE! - in de.symeda.sormas.api.task.TaskHelperTest
Error:  testGetDefaultDueDate(de.symeda.sormas.api.task.TaskHelperTest)  Time elapsed: 0.017 s  <<< FAILURE!
```

StefanKock:
There will be another PR based on hotfix-1.75.3

PR #10452, opened 2022-09-26
[#9889] log deletion

### LINKED ISSUE #9889: Write permanent deletion of entities to audit log [1]
### Problem Description
The main write mechanism for the audit log is currently based on an interceptor that logs calls to the backend. In addition to that the history tables document the complete history of the creation and manipulation of data in the system.
Both mechanisms don't work for the permanent deletion of entities, because permanent deletion also deletes the history AND permanent deletion is handled within scheduled logic in the backend that is not triggered for every single entity.

### Proposed Change
Add an explicit log call when entities are deleted permanently.

### Acceptance Criteria
* When a case is automatically deleted this leads to log entries for the case itself and all related core entities that are deleted. The log entry contains the uuid of the entity.

### Implementation Details

- [ ] Call the AuditLogger.logBackendCall method from within the deletePermanent method of the BaseAdoService

### Alternative

Use the existing AuditLogInterceptor mechanism and exempt calls to deletePermanent from the "don't log calls to local beans" rule.

JonasCir:
Fixes #9889 

log line looks like this:
```json
{
"resourceType":"AuditEvent",
"action":"D",
"period":{"start":"2022-09-26T15:20:31+02:00","end":"2022-09-26T15:20:48+02:00"},
"recorded":"2022-09-26T15:20:50.144+02:00",
"agent":[{"type":{"coding":[{"system":"https://www.hl7.org/fhir/valueset-participation-role-type.html","code":"110150","display":"Application"}]},"name":"SYSTEM"}],
"source":{"site":"sormas_b","type":[{"system":"http://terminology.hl7.org/CodeSystem/security-source-type","code":"4","display":"Application Server"}]},
"entity":[
    "what":{
        "reference":"public void de.symeda.sormas.backend.caze.CaseService.deletePermanent(de.symeda.sormas.backend.common.AbstractDomainObject)"
    },
    "detail":[{"type":"param","valueString":"R7G5UD-WPLSSS-JKHO56-ZO6JSMSE"}]
]}
```

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
@@ -142,7 +142,7 @@ public class AuditLoggerInterceptor {
 			// ignore certain classes for audit altogether. Statically populated cache.
 			return context.proceed();
 		}
-		Method calledMethod = context.getMethod();
+
 
 		// with this we ignore EJB calls which definitely originate from within the backend
 		// as they can never be called direct from outside (i.e., remote) of the backend
@@ -153,6 +153,8 @@ public class AuditLoggerInterceptor {
 			// ignore local beans
 			return context.proceed();
 		}
+
+		Method calledMethod = context.getMethod();
 
 		if (ignoreAuditMethods.contains(calledMethod) || !allowedLocalAuditMethods.contains(calledMethod)) {
 			// ignore certain methods for audit altogether. Statically populated cache.
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
@@ -162,7 +162,7 @@ public class AuditLoggerInterceptor {
 			return context.proceed();
 		}
 
-		// we have a relevant method and a local bean -> audit local calls directly
+		// we have a relevant method of a remote bean or a local bean that is allowed
 		return backendAuditing(context, calledMethod);
 	}
 
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
@@ -22,11 +22,23 @@ import org.slf4j.LoggerFactory;
 
 import de.symeda.sormas.backend.auditlog.AuditContextProducer;
 import de.symeda.sormas.backend.auditlog.AuditLogServiceBean;
+import de.symeda.sormas.backend.campaign.CampaignService;
+import de.symeda.sormas.backend.caze.CaseService;
+import de.symeda.sormas.backend.common.AbstractDomainObject;
 import de.symeda.sormas.backend.common.ConfigFacadeEjb;
+import de.symeda.sormas.backend.contact.ContactService;
+import de.symeda.sormas.backend.event.EventParticipantService;
+import de.symeda.sormas.backend.event.EventService;
 import de.symeda.sormas.backend.feature.FeatureConfigurationFacadeEjb;
 import de.symeda.sormas.backend.i18n.I18nFacadeEjb;
+import de.symeda.sormas.backend.immunization.DirectoryImmunizationService;
+import de.symeda.sormas.backend.immunization.ImmunizationService;
 import de.symeda.sormas.backend.infrastructure.continent.ContinentFacadeEjb;
 import de.symeda.sormas.backend.infrastructure.subcontinent.SubcontinentFacadeEjb;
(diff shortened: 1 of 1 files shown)


MartinWahnschaffe:
@JonasCir will try to a generic implementation based on the reflections library

JonasCir:
@MartinWahnschaffe Found a quite nice generic way :)

sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
```diff
@@ -101,14 +155,14 @@ public Object logAudit(InvocationContext context) throws Exception {
 		// expression yields true if it IS a local bean => should not be audited and ignored
 		Boolean shouldIgnoreAudit = shouldIgnoreClassCache.computeIfAbsent(target, k -> target.getAnnotationsByType(LocalBean.class).length > 0);
```
MartinWahnschaffe, line 156: Your implementation is excluding the AdoServices from being ignored in general, which goes for all their method calls. To circumvent this you have created allowedLocalAuditMethods to only include the delete calls.

Unfortunately you are now also excluding the method calls for all other remote beans, because their methods are not listed in allowedLocalAuditMethods.

What you need to do instead is to by-pass shouldIgnoreAudit if the method is in allowedLocalAuditMethods

MartinWahnschaffe (changes requested):

JonasCir:
@MartinWahnschaffe Not certain if anything is left to do here? :)

sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
```diff
@@ -96,23 +144,25 @@ public Object logAudit(InvocationContext context) throws Exception {
...
+		Boolean ignoreBeanAudit = shouldIgnoreBeanCache.computeIfAbsent(target, k -> target.getAnnotationsByType(LocalBean.class).length > 0);
 
-		if (ignoreAuditMethods.contains(calledMethod)) {
-			// ignore certain methods for audit altogether. Statically populated cache.
+		if (Boolean.TRUE.equals(ignoreBeanAudit) && !allowedLocalAuditMethods.contains(calledMethod)) {
+			// we have a local bean call which should not be audited -> ignore
 			return context.proceed();
 		}
 
+		// we have a relevant method and a local bean -> audit local calls directly
```
MartinWahnschaffe, line 165:
```suggestion
		// we have a relevant method of a remote bean or a local bean that is allowed
```

sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
```diff
@@ -32,14 +38,23 @@
...
 
+	private static final Reflections reflections;
+	static {
+
+		ConfigurationBuilder configurationBuilder = new ConfigurationBuilder();
+		configurationBuilder.forPackages("de.symeda.sormas.backend").addScanners(SubTypes).setParallel(false);
+		reflections = new Reflections(configurationBuilder);
+	}
+
+	private static final Set<Class<?>> adoServiceClasses = new HashSet<>(reflections.get(SubTypes.of(BaseAdoService.class).asClass()));
```
MartinWahnschaffe, line 52: Please do this within the static initialize of the class

MartinWahnschaffe (changes requested):
Found two small things.
Can be merged, once the 1.76 release is done.

sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
JonasCir: done

sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
```diff
@@ -32,14 +38,27 @@
...
+
+		ConfigurationBuilder configurationBuilder = new ConfigurationBuilder();
+		configurationBuilder.forPackages("de.symeda.sormas.backend").addScanners(SubTypes).setParallel(false);
+		reflections = new Reflections(configurationBuilder);
+	}
+
+	private static final Set<Class<?>> adoServiceClasses;
+
+	static {
+		adoServiceClasses = new HashSet<>(reflections.get(SubTypes.of(BaseAdoService.class).asClass()));
```
MartinWahnschaffe, line 55: move to line 50

MartinWahnschaffe (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/audit/AuditLoggerInterceptor.java
JonasCir: one static initializer now

MartinWahnschaffe (approved):

PR #10456, opened 2022-09-27
[#9091] explicitly remove client role from keycloak user

### LINKED ISSUE #9091: Only allow  sormas users with  "STATISTICS_ACCESS" right to authenticate with sormas-stats keycloak client [3]
### Situation Description
#8927 allows all authenticated SORMAS users to access `sormas-stats` which contains analytics. Thus, it should be accessible only to SORMAS users who are responsible to produce epidemiological reports.

### Feature Description
- only active users with with `STATISTICS_ACCESS` right  should access `sormas-stats`, i.e. `STATISTICS_ACCESS==t` and `ACTIVE ==t`

### Implementation notes

- [x] Check the `STATISTICS_ACCESS` right in SORMAS and see how it can be managed by the admin. Investigate where it is already used in the system.
- [x] Add `sormas-stats` to the testbed
- [x] Presence of the "Navigate to SORMAS Stats button" should be dependent on the access right.
- [x] Add statistics right/role or other suitable primitive to Keycloak.
- [x] Implement synchronization mechanism in SORMAS which adds/removes permission from the user in Keycloak. Make this generic and reusable as we might want to reuse this in the future.
- [x] add check for active user
- [x] write documentation
- [x] open an issue in SORMAS Docker
- ~TODO: Make SORMAS the leading system? What happens if someone edits the user in Keycloak?~
- [x] When saving a user role we also have to make sure to sync all related users to keycloak.

JonasCir:
Fixes #9091 

@MartinWahnschaffe Just like we cannot add a client role to a normal user representation in an update, we cannot remove it.

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/StartupShutdownService.java
@@ -20,7 +20,6 @@ import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.security.SecureRandom;
 import java.sql.Timestamp;
-import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


MartinWahnschaffe:
@JonasCir an me realized that updating a user role does currently not lead to the related users being synced again to keycloak, which is needed for this issue.

JonasCir:
@MartinWahnschaffe Tested and ready for review

MartinWahnschaffe (approved):

PR #10458, opened 2022-09-27
#10305: Added unit test

### LINKED ISSUE #10305: When changing the jurisdiction level at user roles, inconsistent users can be created.
### Bug Description

With the user roles, the jurisdiction levels can be changed. A reduction only becomes effective when the user has also been edited. After a change of the user role, empty mandatory fields appear with the user. This happens without any hint when editing the role.

### Steps to Reproduce

1. Create a new user role using e.g. the "Surveillance Supervisor (Default)" template. 
2. Assign a user to this role.
3. Go back to the user role
4. Set the jurisdiction level to "Community" 
5. Login whist the user and check if the user behaves as configured 
image
6. Login a Admin and update the user
image

* Device: Laptop
* SORMAS version: 1.75.0
* Server URL: test1401
* User Role: Admin and other newly created users

ChristopherRiedel:
Fixes #10305

### DIFF
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/user/UserRoleFacadeEjbTest.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/user/UserRoleFacadeEjbTest.java
@@ -1,0 +1,31 @@
+package de.symeda.sormas.backend.user;
+
+import static org.junit.Assert.assertThrows;
+
+import org.junit.Test;
+
+import de.symeda.sormas.api.i18n.I18nProperties;
+import de.symeda.sormas.api.i18n.Validations;
+import de.symeda.sormas.api.user.JurisdictionLevel;
+import de.symeda.sormas.api.user.UserRoleDto;
+import de.symeda.sormas.api.utils.ValidationRuntimeException;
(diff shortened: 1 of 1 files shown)


sormas-backend/src/test/java/de/symeda/sormas/backend/user/UserRoleFacadeEjbTest.java
```diff
@@ -0,0 +1,31 @@
...
+import de.symeda.sormas.api.i18n.Validations;
+import de.symeda.sormas.api.user.JurisdictionLevel;
+import de.symeda.sormas.api.user.UserRoleDto;
+import de.symeda.sormas.api.utils.ValidationRuntimeException;
+import de.symeda.sormas.backend.AbstractBeanTest;
+
+public class UserRoleFacadeEjbTest extends AbstractBeanTest {
+
+    @Test
+    public void testSaveUserRole() {
```
leventegal-she, line 17: Please rename to `testChangeUserRoleJurisdictionLevel` or something similar so it reflects what is tested by this method

leventegal-she (changes requested):

PR #10465, opened 2022-09-27
#10410 - optional health facility should be false by default

### LINKED ISSUE #10410: Description and default value for user role properties
### Problem Description
When creating a new user role the meaning of the checkboxes "Has optional health facility" and "Port health user" are not clear to users.

grafik

### Proposed Change
- [x] Both values should be "false" by default
- [x] Optional health facility should only be checkable when the user role is not on HEALTH_FACILITY jurisdiction level
- [x] Optional health facility should be set to false when it is disabled based on HEALTH_FACILITY
- [x] Change description (tooltip) of the optional health facility field: "Allows to associate a user to a health facility even though the user is not on health facility jurisdiction level. When the user is entering new data, e.g. a case, the place of stay facility will be pre-filled." The captions should remain unchanged.
- [x] Change description (tooltip) of the port health user field: "User is a port health user and can only access cases and related data that is associated to a point of entry in the user's jurisdiction." The captions should remain unchanged.

carina29:
Fixes #10410


carina29:
@MartinWahnschaffe I've updated the PR. 
The descriptions (tooltips) were changed and the old captions were put back.
HAS_OPTIONAL_HEALTH_FACILITY checkbox will be set back to false default value in case of any Jurisdiction Level change in order to keep the default false values for HAS_OPTIONAL_HEALTH_FACILITY and PORT_HEALTH_USER as it was requested in https://github.com/hzi-braunschweig/SORMAS-Project/issues/10410.

MartinWahnschaffe (approved):
Thanks!

PR #10502, opened 2022-09-30
#10414 limit persons by config and user rights

### LINKED ISSUE #10414: Person directory: Don't include persons based on entity types the user cannot access [3]
### Problem Description
#10352 pointed out problems accessing persons related to entities the user should not be able to access.
For example the person directory currently still lists persons related to a case to a user that does not have the right to view cases.

### Proposed Change
Only show persons that are related to entity types the user can access based on activated features and given user rights.

### Acceptance Criteria

- [x] The person directory only shows persons that are part of an activated feature **and** the current user is permitted to access via user rights.
- [x] All similar backend logic is changed according to this.
- [x] The person directory shows the "All" button, also if any association is not permitted for the current user.
- [x] Probable code redundancy with #10352 concerning permitted `PersonAssociation`s was reduced.

### Implementation Details

We need to apply the same logical changes to several methods in the backend:
1. `PersonService.createUserFilter` should not join entity types the user does not have access to (e.g. exclude travelEntryFilter, when the user is not allowed to view travel entries). For consistency the following methods need to be changed in the same way: 
2. `PersonService.getAllAfter`.
3. `PersonService.getAllUuids` (similar to `PersonService.getAllAfter`)
4. `PersonService.getSimilarPersonDtos`
5. `PersonService.inJurisdictionOrOwned`  (uses `PersonJurisdictionPredicateValidator`)
6. `PersonFacadeEjb.count` needs to match `PersonFacade.getIndexList` (that uses `createUserFilter`).

Change in logic: `PersonService.getAllAfter` and `PersonService.getAllUuids` currently contain logic that additionally takes persons by the users district into account. It was clarified with @MartinWahnschaffe that this must be aligned with all other methods: Getting persons just by users district will be dropped.

StefanKock:
Fixes #10414

Progress:
- [x] Backend logic to calculate which `PersonAssociation`s are allowed.
- [x] `PersonService.createUserFilter`, `PersonFacadeEjb.count` and `PersonService.inJurisdictionOrOwned` changed.
- [x] `PersonService.getAllAfter` changed.
- [x] `PersonService.getAllUuids` changed.
- [x] `PersonService.getSimilarPersonDtos` changed.
- [x] Replaced redundant logic in `PersonsView` by backend logic.

Challenges to solve:
- [x] 1. Two `ExternalVisitTest` cases fail now because `PersonService.inJurisdictionOrOwned` cannot find any allowed association to join.

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
@@ -29,7 +29,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
-import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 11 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
```diff
@@ -150,98 +152,117 @@ public Person createPerson() {
...
+				if (!RequestContextHolder.isMobileSync()) {
+					personUuids.addAll(
+						getAllUuids(
+							TravelEntry.class,
+							TravelEntry.PERSON,
+							(b, q, f) -> travelEntryService.createUserFilter(new TravelEntryQueryContext(b, q, new TravelEntryJoins(f)))));
+				}
+				break;
+			case ALL:
+				// NOOP: check the explicit associations
```
MartinWahnschaffe, line 238: Same here

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
```diff
@@ -251,73 +272,100 @@ public Predicate createUserFilter(CriteriaBuilder cb, CriteriaQuery cq, From<?,
...
 		case TRAVEL_ENTRY:
-			return travelEntryFilter.get();
+			userFilter = createAssociationPredicate(queryContext, personAssociation);
+			break;
 		default:
 			throw new IllegalArgumentException(personAssociation.toString());
 		}
+
+		if (userFilter == null) {
+			throw new IllegalArgumentException("No filter compiled for persons by association " + personAssociation.name());
```
MartinWahnschaffe, line 306: Instead of throwing an exception return a "false" predicate.
This will likely also solve the external visit problems.

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonService.java
```diff
@@ -656,6 +629,82 @@ public List<SimilarPersonDto> getSimilarPersonDtos(PersonSimilarityCriteria crit
...
+			case ALL:
+				// NOOP: check the explicit associations
+				break;
+			default:
+				throw new IllegalArgumentException(personAssociation.toString());
+			}
+		}
+
+		if (associationPredicates.isEmpty()) {
+			throw new IllegalArgumentException("No filter compiled for persons by associations: " + permittedAssociations);
```
MartinWahnschaffe, line 702: Same here.

MartinWahnschaffe (changes requested):

MateStrysewske:
@StefanKock Code looks good to me, would also do some testing after you've removed the Draft status from this PR (I suspect there will be merge conflicts that need to be taken care of first).

MartinWahnschaffe (approved):

MartinWahnschaffe:
@MateStrysewske doesn't seem to have any merge conflicts.

MateStrysewske (approved):

PR #10507, opened 2022-10-03
#10188 - show the configuration page only to users having user right …

### LINKED ISSUE #10188: A user with a configured User Role gets an error message when opening the Configuration Directory
### Bug Description
A user with a configured User Role (Infrastructure = Yes to all and Configuration = No to all) gets an error message when he opens the Configuration Directory. 

### Steps to Reproduce

1. Logged in with an Admin, navigate to the Users Directory;
2. Create a new User Role with the following User Rights: General (Access Sormas UI: Yes); Infrastructure (Yes all); Configuration (No all); the other ones are all No;
3. Create a user with the newly created user role;
4. Log in with that user;

### Actual Behavior
The Configuration Directory is shown and if the user opens it, he receives an error.

### Expected Behavior
The Configuration Directory should show only the tabs that have the corresponding user right set to 'true' (eg.  'View regions/districts/communities/facilities in the system' = true; 'View outbreaks' = false)

### Screenshots
image

### System Details

* Device: Windows 10
* SORMAS version: Sormas 1.75.0-SNAPSHOT (68389ec)
* Server URL: local instance
* User Role:

### Additional Information
The error was triggered because 'View outbreaks' was set to 'false'

BarnaBartha:
…of the group configuration

Fixes #10188

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/UserProvider.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/UserProvider.java
@@ -32,16 +32,10 @@ import de.symeda.sormas.api.user.JurisdictionLevel;
 import de.symeda.sormas.api.user.UserDto;
 import de.symeda.sormas.api.user.UserReferenceDto;
 import de.symeda.sormas.api.user.UserRight;
+import de.symeda.sormas.api.user.UserRightGroup;
 import de.symeda.sormas.api.user.UserRoleDto;
 
 public class UserProvider {
-
-	private static final List configurationUserRoles = Arrays.asList(
-		UserRight.DOCUMENT_TEMPLATE_MANAGEMENT,
-		UserRight.INFRASTRUCTURE_VIEW,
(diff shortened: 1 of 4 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/UserProvider.java
```diff
@@ -86,7 +80,7 @@ public JurisdictionLevel getJurisdictionLevel() {
 
 	public boolean hasConfigurationAccess() {
 		Set<UserRight> currentUserRights = getUserRights();
-		return configurationUserRoles.stream().anyMatch(currentUserRights::contains);
+		return UserRight.getUserRightsOfGroup(UserRightGroup.CONFIGURATION).stream().anyMatch(currentUserRights::contains);
```
sergiupacurariu, line 83: Configuration menu should be shown if Infrastructure rights are active

sergiupacurariu (approved):

Issue #10343, opened 2022-09-12
Missing validations in the backend when data is created via API endpoints


roxanamlendea:
### Bug Description
- Travel Entries can be created via API without Point of Entry Details, even if it's a mandatory field.
- Events can be created via API calls without an Event Title, even if it's a mandatory field.
- Samples can be created without the 'Specify other type' being filled in, even though it's a mandatory field.

### Steps to Reproduce

Scenario 1: Travel Entries can be created via API without Point of Entry Details, even if it's a mandatory field.
1. Create a new Person via POST Person endpoint;
2. Create a Travel Entry via POST Travel Entry endpoint, set 'Other airport' as 'pointOfEntry' and leave the 'pointOfEntryDetails' empty;
3. Check the Travel Entry in the UI.

![image](https://user-images.githubusercontent.com/89522688/189680350-00ce98f0-3876-4452-b73d-16bf366441fe.png)
![image](https://user-images.githubusercontent.com/89522688/189680442-19d09940-97cd-499c-886c-eb483ebea2f8.png)

Scenario 2: Events can be created via API calls without an Event Title, even if it's a mandatory field.
1. Create a new Event via POST Event endpoint and in the body, leave the 'eventTitle' empty;
2. Check the Event in the UI

![image](https://user-images.githubusercontent.com/89522688/189681219-de804ce6-244b-4e20-a7d7-ba193e807c94.png)
![image](https://user-images.githubusercontent.com/89522688/189681276-f2daac06-7082-4dae-aff7-0322d20f6ed6.png)

Scenario 3: Samples can be created without the 'Specify other type' being filled in, even though it's a mandatory field.
1. Create a Sample for a Case via POST Sample endpoint, set 'Other' as 'sampleMaterial' and leave the 'sampleMaterialText' empty.
2. Check the Sample in the UI.

![image](https://user-images.githubusercontent.com/89522688/189681912-f70a1dce-f44f-4717-99ae-0eb643c75628.png)
![image](https://user-images.githubusercontent.com/89522688/189682078-d82c038e-0eb7-49f0-bc53-a6d39a6bffe2.png)

### System Details

* Device: Windows 10
* SORMAS version: Sormas 1.76.0-SNAPSHOT
* Android version/Browser: Chrome
* Server URL: local machine
* User Role: admin, national user

### Additional Information
Edit by @JonasCir :
* This is similar to issue #10110 . You can take a look at the following PR: #10238
[...]

PR #10383, opened 2022-09-17
Bugfix 10343 missing validations via api

andrei7bal:
#10343 Add new validations for required fields when adding data from API
- add new validations for Sample and for TravelEntry
- use @NotEmpty annotation instead of @Requried (it's gonna be deprecated)

PR #10519, opened 2022-10-04
#10343 Add validation new validation for Sample and Travel entries

andrei7bal:
If SampleMaterial is other than SampleMaterialText is required
If PointOfEntry is other than PointOfEntryDetails is required


andrei7bal:
> 

For Event the fix was merged in a different PR: #10383

sergiupacurariu (approved):

sergiupacurariu (approved):

PR #10532, opened 2022-10-04
Automatize "S2S_ Introduce a feature type to hide share action in bulk mode for cases and contacts [1]"

mp-sgent:


mp-sgent:
image

rdutu-vg (approved):

PR #10577, opened 2022-10-07
bug/10538_disable_user_are_showed

### LINKED ISSUE #10538: Disabled user roles are displayed in the list when 'Enabled' appears selected by default
### Bug Description
When an admin user accesses the Users Directory -> User Roles, 'Enabled' appears to be set to default. However, the list of user roles contain both enabled and disabled user roles.
If 'Enabled' is selected manually, the disabled user roles are no longer displayed in the list.

### Steps to Reproduce

1. Log in with an Admin user;
2. Open the Users Directory and select the User Roles tab;
3. Make sure you have at least one or two disabled user roles;
4. Inspect the list of user roles and search if the disabled user roles are displayed;

### Expected Behavior
The disabled user roles should not be displayed; should be displayed only when 'Disabled' is selected;

### Screenshots
image
image

### System Details

* Device: Windows 10
* SORMAS version: 1.76.0 - SNAPSHOT
* Server URL: local server
* User Role: admin

dinua:
Fixes #10538

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserRolesView.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserRolesView.java
@@ -45,6 +45,7 @@ public class UserRolesView extends AbstractUserView {
 
 	public static final String ENABLED_FILTER = I18nProperties.getString(Strings.enabled);
 	public static final String DISABLED_FILTER = I18nProperties.getString(Strings.disabled);
+	public static final String ALL_FILTER = I18nProperties.getString(Strings.all);
 
 	private UserRoleCriteria criteria;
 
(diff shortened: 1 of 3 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserRolesView.java
```diff
@@ -120,7 +121,7 @@ public HorizontalLayout createStatusFilterBar() {
 		enabledFilter.setId(UserRoleDto.ENABLED);
 		enabledFilter.setWidth(200, Unit.PIXELS);
 		enabledFilter.setInputPrompt(I18nProperties.getPrefixCaption(UserRoleDto.I18N_PREFIX, UserRoleDto.ENABLED));
-		enabledFilter.addItems(ENABLED_FILTER, DISABLED_FILTER);
+		enabledFilter.addItems(ALL_FILTER, ENABLED_FILTER, DISABLED_FILTER);
```
sergiupacurariu, line 124: instead of adding a new item could you explore if we could just add a "Boolean.TRUE" instead of null in criteria enabled logic?

sergiupacurariu (approved):

Issue #10515, opened 2022-10-03
'Facility name & description' is doubled in the text of the error message


roxanamlendea:
### Bug Description
When the 'Facility name & description' filed is not filled in on the Event page, an error is thrown and the user cannot save. But, the text 'Facility name & description' is displayed twice in the error message.

### Steps to Reproduce

1. Open an Event;
2. Scroll down to the Location section and make the following selection:
![image](https://user-images.githubusercontent.com/89522688/193614422-0eed07b4-2c76-42b7-88a9-70412aec76b2.png)
3. Try to Savel

### Expected Behavior
'Facility name & description' should not appear twice in the error message

### Screenshots
![image](https://user-images.githubusercontent.com/89522688/193614506-52079ba1-f67a-4893-aa17-4ed15496861b.png)

### System Details

* Device: Windows 10
* SORMAS version :Sormas 1.75.2
* Android version/Browser: Chrome
* Server URL: test1401.sormas-oegd.de
* User Role: admin + national user

### Additional Information
Found while testing ticket #10431

roxanamlendea:
Verified ticket on local setup using the latest version of Sormas 1.77.0-SNAPSHOT (8c31891)

PR #10593, opened 2022-10-11
#10515 Do not add message from cascading cause if they are the same

andrei7bal:
Fixes #
#10515 Do not add same error message from cascading causes if they are the same.


sormas-ui/src/main/java/de/symeda/sormas/ui/utils/CommitDiscardWrapperComponent.java
```diff
@@ -560,9 +560,12 @@ private String findHtmlMessage(InvalidValueException exception) {
 
 	private String findHtmlMessageDetails(InvalidValueException exception) {
 		for (InvalidValueException cause : exception.getCauses()) {
-			String message = findHtmlMessage(cause);
-			if (message != null)
-				return message;
+			if (!cause.getMessage().equalsIgnoreCase(exception.getMessage()))
+			{
+				String message = findHtmlMessage(cause);
```
sergiupacurariu, line 565: please format code

andrei7bal: Done

sergiupacurariu (approved):

PR #10629, opened 2022-10-13
#9893 - hide disease variant field when there are no disease variants…

### LINKED ISSUE #9893: Hide disease variant on mobile app when there is none
### Bug Description

When the disease variant entry is empty in the DB, the UI element for "disease variant" on the web app is not shown. Disease variants appeared on the mobile app even for diseases that do not have a variant.

### Steps to Reproduce

1. Create a measles case on https://release-international.sormas.netzlink.com/ using the mobile app
2. Open the edit form and the system ask for disease variant

### Expected Behavior

Disease variants should appear only for diseases that have categories for variants in the db.
### System Details

* Device:
* SORMAS version: 1.73.2
* Android version/Browser:  A 11
* Server URL:  https://release-international.sormas.netzlink.com/sormas-ui/
* User Role: surveillance + contact officer

### Additional Information

This may be an old issue before 1.73.2.  The web app works as expected, only the mobile app needs adjustment.

BarnaBartha:
… for a disease - make sure empty disease variants works as before and newly synced custom enums are taken into consideration

Fixes #9893

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/util/DataUtils.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/util/DataUtils.java
@@ -76,7 +76,7 @@ public class DataUtils {
 	}
 
 	public static boolean emptyOrWithOneNullItem(List<Item> listIn) {
-		return listIn.isEmpty() || (listIn.size() == 1 && listIn.get(0).getValue() == null);
+		return listIn.isEmpty() || (listIn.size() == 1 && (listIn.get(0) == null || listIn.get(0).getValue() == null));
 	}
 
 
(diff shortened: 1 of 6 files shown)


dinua (approved):

sormas-app/app/src/main/java/de/symeda/sormas/app/util/DataUtils.java
```diff
@@ -75,6 +75,11 @@ public static List<Item> getBooleanItems() {
 		return list;
 	}
 
+	public static boolean emptyOrWithOneNullItem(List<Item> listIn) {
+		return listIn.isEmpty() || (listIn.size() == 1 && listIn.get(0).getValue() == null);
+	}
```
dinua, line 80: i think here would be better to check that the list is empty or all elements in list are null

PR #10650, opened 2022-10-17
#10625 s2 s show warning and remove share requests when case contact is deleted

### LINKED ISSUE #10625: [S2S] Show Warning and Remove share requests on target system when case/contact is deleted on Sending system
### Problem Description

### Proposed Change

### Acceptance Criteria

### Additional Information

leventegal-she:
Fixes #10625


JonasCir:
@carina29 I think the branch you are working on was based on `development`? I think you need to cherry-pick the commits to a correct branch.

carina29:
This PR will be closed since a new feature branch was created from 1.76.1 hotfix branch.
The new PR for this task is: https://github.com/hzi-braunschweig/SORMAS-Project/pull/10718

PR #10652, opened 2022-10-17
#10517 Facility is not pre-filled

### LINKED ISSUE #10517: Facility is not pre-filled for users with an optional health facility
### Bug Description

When the checkbox " Has optional health facility" is checked and the user is related to a certain health faciliy this health facility should be prefilled when the user is for example creating a new case.
### Steps to Reproduce

1.create a new user role
2. check the checkbox "Has optional health facility"
3. create a new user
4. link the user to the created user role and the facility
5.create a case

### Expected Behavior

facility should be prefilled
### Screenshots

image

### System Details

* Device:
* SORMAS version: 1.75.2
* Android version/Browser:
* Server URL: test1402
* User Role: admin

valentinmikleuvg:
Added a certain health facility when a user with optional health facility permission created a new case

Fixes #10517

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
@@ -332,31 +332,46 @@ public class CaseCreateForm extends AbstractEditForm<CaseDataDto> {
 			updateFacility();
 		});
 		facilityOrHome.addValueChangeListener(e -> {
-			FieldHelper.removeItems(facilityCombo);
-			if (TypeOfPlace.FACILITY.equals(facilityOrHome.getValue())
-				|| ((facilityOrHome.getValue() instanceof java.util.Set) && TypeOfPlace.FACILITY.equals(facilityOrHome.getNullableValue()))) {
-				if (facilityTypeGroup.getValue() == null) {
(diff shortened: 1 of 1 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseCreateForm.java
```diff
@@ -332,31 +332,46 @@ protected void addFields() {
...
-				}
 
-				if (facilityType.getValue() != null) {
-					updateFacility();
-				}
+			boolean hasOptionalHealthFacility = UserProvider.getCurrent().hasOptionalHealthFacility();
+			if (hasOptionalHealthFacility) {
+				String facilityId = UserProvider.getCurrent().getUser().getHealthFacility().getUuid();
+				FacilityDto facilityDto = FacadeProvider.getFacilityFacade().getByUuid(facilityId);
+				FacilityType facilityUserType = facilityDto.getType();
```
carina29, line 340: I've also tested another case , changing a Surveillance Officer user role and checking the  "Has optional health facility" checkbox but after logging in with Surveillance Officer user , the same error appeared due to null healthFacility. 
image

image


image

carina29 (changes requested):

carina29 (approved):

PR #10718, opened 2022-10-24
Feature 10625 s2 s show warning and remove share requests when case contact is deleted

### LINKED ISSUE #10625: [S2S] Show Warning and Remove share requests on target system when case/contact is deleted on Sending system

carina29:
Fixes #10625

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasRuntimeException.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasRuntimeException.java
@@ -1,0 +1,25 @@
+/*
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2022 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package de.symeda.sormas.api.sormastosormas;
+
+import de.symeda.sormas.api.i18n.I18nProperties;
+
+public class SormasToSormasRuntimeException extends RuntimeException {
+
+	public SormasToSormasRuntimeException(SormasToSormasException exception) {
+		super(I18nProperties.getString(exception.getI18nTag()));
+	}
+}
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -176,6 +176,7 @@ import de.symeda.sormas.api.sample.PathogenTestType;
 import de.symeda.sormas.api.sample.SampleCriteria;
 import de.symeda.sormas.api.sample.SampleDto;
 import de.symeda.sormas.api.sormastosormas.ShareTreeCriteria;
+import de.symeda.sormas.api.sormastosormas.SormasToSormasException;
 import de.symeda.sormas.api.sormastosormas.SormasToSormasRuntimeException;
 import de.symeda.sormas.api.symptoms.SymptomsDto;
 import de.symeda.sormas.api.symptoms.SymptomsHelper;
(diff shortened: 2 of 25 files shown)


sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasRuntimeException.java
```diff
@@ -0,0 +1,25 @@
+/*
```
JonasCir, line 1: I somehow feel that we already have something like this, let me see

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/rest/SormasToSormasRestClient.java
```diff
@@ -186,6 +188,7 @@ private <T> T handleResponse(Response response, Class<T> responseType) throws So
 			String errorI18nTag = null;
 			List<ValidationErrors> errors = null;
 			Object[] args = null;
+			boolean debug = false;
```
JonasCir, line 191: Is this debug code which can be removed?

JonasCir (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/outgoing/ShareRequestInfoService.java
```diff
@@ -72,9 +74,16 @@ public List<String> getAllNonReferencedShareRequestInfo() {
 		Join<ShareRequestInfo, SormasToSormasShareInfo> shareRequestJoin = from.join(ShareRequestInfo.SHARES, JoinType.LEFT);
 		cq.select(from.get(ShareRequestInfo.UUID));
 		cq.groupBy(from.get(ShareRequestInfo.ID));
-		cq.having(cb.equal(cb.count(shareRequestJoin), 0));
 
-		return em.createQuery(cq).getResultList();
+		Predicate revokedStatus = cb.equal(from.get(ShareRequestInfo.REQUEST_STATUS), ShareRequestStatus.REVOKED);
```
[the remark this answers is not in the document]
carina29: In order to delete the ShareRequestInfos, the getAllNonReferencedShareRequestInfo() will be called before to find the ShareRequestInfos which should be deleted.
image


The ShareRequestInfos till now only retrieved the ShareRequestInfos without joins which meant that they are no longer referenced. In this task the ShareRequestInfos will be still referenced but their status was changed to REVOKED.

Due to above mentioned problem testSormasToSormasShareInfoPermanentDeletion() unit test from CoreEntityDeletionServiceTest failed because the getAllNonReferencedShareRequestInfo did not return the REVOKED ShareRequestInfo as non referenced => it was not deleted

assertEquals(0, getShareRequestInfoService().count());  ( Before the returned count was 1 instead of 0, since the ShareRequestInfo  was not find and deleted).

I'll add a comment in the code too

sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasRuntimeException.java
carina29: I've checked the RuntimeExceptions , we do not have till now a specific exception created for S2S which extends RuntimeException. If you want to use one of the custom RuntimeExceptions( maybe one with a more general naming) let me know.

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/rest/SormasToSormasRestClient.java
carina29: I also think we can delete this flag

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2557,8 +2571,8 @@ public List<String> deleteCases(List<String> caseUuids, DeletionDetails deletion
 					try {
 						deleteCase(caseToBeDeleted, deletionDetails);
 						deletedCasesUuids.add(caseToBeDeleted.getUuid());
-					} catch (ExternalSurveillanceToolRuntimeException e) {
-						logger.error("The case with uuid:" + caseToBeDeleted.getUuid() + "could not be deleted");
+					} catch (ExternalSurveillanceToolRuntimeException | SormasToSormasRuntimeException e) {
+						logger.error("The case with uuid:" + caseToBeDeleted.getUuid() + "could not be deleted", e);
```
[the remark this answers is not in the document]
carina29: Ok

sormas-api/src/main/java/de/symeda/sormas/api/sormastosormas/SormasToSormasRuntimeException.java
JonasCir: done

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/rest/SormasToSormasRestClient.java
```diff
@@ -82,6 +83,7 @@ public SormasToSormasRestClient(
 		mapper = new ObjectMapper();
 		mapper.setVisibility(PropertyAccessor.ALL, JsonAutoDetect.Visibility.NONE);
 		mapper.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+		mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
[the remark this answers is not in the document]
JonasCir: this is ok

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/rest/SormasToSormasRestClient.java
JonasCir: ok

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/share/outgoing/ShareRequestInfoService.java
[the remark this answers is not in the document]
JonasCir: ok thank you!

PR #10781, opened 2022-10-27
#10670 Fix Disease Variant Spinner when Editing Pathogen Test

### LINKED ISSUE #10670: [Mobile app] 'Tested Disease Variant' field appears in the Pathogen Test page (edit mode), for diseases that don't have disease variants configured
### Bug Description

In the mobile app, in the Edit Mode of the Pathogen Test page, the 'Tested Disease Variant' field is visible for diseases without disease variants configured.

The issue is reproducible only in Edit Mode; It is not reproducible in the Creation Form or in Read Mode;
The issue is only reproducible on the mobile app;

### Steps to Reproduce

1. Create a Sample and a Pathogen Test for a disease that doesn't have disease variants configured;
2. Open the Pathogen Test in Edit Mode;

### Expected Behavior
The 'Tested Disease Variant' field should not be visible when a disease with no disease variants is selected;

### Screenshots
image

### System Details

* Device: Pixel C 
* SORMAS version: 1.76
* Android version/Browser: Android 11
* Server URL:

### Additional Information
#9893

andrei7bal:
- in case no Disease Variant is configured for the selected Disease then hide the Spinner when editing the pathogen test

Fixes #10670


MateStrysewske (changes requested):
The disease variants list needs to contain an empty item so that it's possible for the user to clear it; please refer to the logic used in other fragments in the mobile app where a disease variant spinner is used instead (e.g. EventEditFragment).

MateStrysewske (approved):

PR #10785, opened 2022-10-28
Bugfix 10701 s2 s display comment in share directory

### LINKED ISSUE #10701: S2S _ Display the full comment in share directory
### Bug Description

If you open the share directory you first think that you cannot read a long comment.
It is difficult for a user with no technical understanding to find out how he can read the full comment.

We switched the column with the buttons 'accept' and 'reject' with the column 'comment' because in earlier versions the comment was shown in total and it was annoying for the user to scroll to the right before he could press the buttons 'accept' or 'reject'.
We specified: _Text field as already implemented (1.73.1)_ So we did not asked for that change. 

It is important that the user can read the whole comment without effort, because these are important information.
### Steps to Reproduce

1. share a request with a long comment in the comment field
2. go to the share directory of the source system (outgoing) and try to read the comment 
3. go to the share directory of the target system (incoming) and try to read the comment

### Expected Behavior

You should be able to see the whole long comment by scrolling to the right. 
### Screenshots

You have to go step by step to read the whole comment. If you leave the directory the whole process begins again. 
read comment_2

### System Details

* Device:
* SORMAS version: 1.76.0
* Android version/Browser:
* Server URL: release-sormas-x and release-de
* User Role: admin

### Additional Information

This a behavior that can cause a shit storm. We want to celebrate the implementation of S2S.

ChristopherRiedel:
Fixes #10701

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/ShareRequestGrid.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/sormastosormas/ShareRequestGrid.java
@@ -27,12 +27,17 @@ import com.vaadin.data.provider.ListDataProvider;
 import com.vaadin.icons.VaadinIcons;
 import com.vaadin.shared.data.sort.SortDirection;
 import com.vaadin.ui.Component;
+import com.vaadin.ui.DescriptionGenerator;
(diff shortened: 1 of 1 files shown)


JonasCir (approved):
Thanks!

PR #10798, opened 2022-10-31
#10780 - update occupation type when is necessary

### LINKED ISSUE #10780: [S2S] Sync broken because of occupation type
### Bug Description

If a case/contact is shared via S2S and the occupation type is filled in with a value, that does not exist on the other instance this breaks the whole sync of S2S for the person and the case/contact.
There are occupation types present on one system and not present on another because of #10380. 

### Steps to Reproduce

1. share a case/ contact without an occupation type with hand over the ownership
2. accept in target system
3. try if sync works (add/change data (except occupation type) in target system and observe data in source system) 
4. add an occupation type in target system (a value that exists in source system, too) and observe if sync works in source system
5. add an occupation type in target system (a value that do NOT exist in source system) and observe if sync works in source system

### Expected Behavior

Sync should work
### System Details

* Device: Windows 11
* SORMAS version: 1.75.4
* Android version/Browser: Edge
* Server URL: 1402 to release-sormas-x
* User Role: admin

### Additional Information

We have to discuss how to deal with customizable enum values when syncing data:

Here some ideas from @MartinWahnschaffe:

A. Reject the data
B. Accept the data and automatically create an entry in the customizable enums tables - this value would then also be available for other data objects
C. Make the whole customizable enum logic able to deal with values that are not defined in the database.

Besides that I'm wondering whether we should define a full set of occupation types for Germany and reset any different occupation types assigned to persons or migrate those.

carina29:
Fixes #10780


JonasCir (changes requested):
There are two problems:
- `getByOccupationType` throws an exception: : Parameter value [Sonstiges] did not match expected type [de.symeda.sormas.api.person.OccupationType (n/a)]
- `updateIfNecessaryOccupationType` should be based on the value not the translation(i.e., caption)

JonasCir (approved):

PR #10806, opened 2022-11-02
Change 8938 improve get index list

### LINKED ISSUE #8938: Improve performance of PersonFacadeEjb.getIndexList [5]
### Problem Description

This summarizes the problem for getIndexList methods in general, not only for Persons.

For large datasets, some of the `getIndexList` methods are slow, performance should be improved here.

Dataset:

- 1225343 persons
- 85677 cases
- 114782 contacts
- 236583 tasks
- 2600890 immunizations
- 1991 vaccinations
- 15693 samples
- 11846 events
- 534 eventparticipants

Result of a performance analysis:
image

The runtime is mainly due to long running SQL queries, optimization is needed here.

Some preliminary insights:
- in the following query from `PersonFacadeEjb.getIndexList`, runtime is reduced by facto 0.5 when omitting the joins on district, subselect and `case` statement. Since the query is limited to 100 result, a viable solution might be to select Ids /Uuids first and fill the remaining data after for the returned entities. This, however, with respect to the properties possibly being used a sorting criteria.

<details>

```
    select
        distinct person0_.uuid as col_0_0_,
        person0_.firstName as col_1_0_,
        person0_.lastName as col_2_0_,
        person0_.approximateAge as col_3_0_,
        person0_.approximateAgeType as col_4_0_,
        person0_.birthdate_dd as col_5_0_,
        person0_.birthdate_mm as col_6_0_,
        person0_.birthdate_yyyy as col_7_0_,
        person0_.sex as col_8_0_,
        district2_.name as col_9_0_,
        location1_.street as col_10_0_,
        location1_.houseNumber as col_11_0_,
        location1_.postalCode as col_12_0_,
        location1_.city as col_13_0_,
        (select
            personcont12_.contactInformation 
        from
            PersonContactDetail personcont12_ 
        where
            personcont12_.person_id=person0_.id 
            and personcont12_.primaryContact=true 
            and personcont12_.personContactDetailType='PHONE') as col_14_0_,
        (select
            personcont13_.contactInformation 
        from
            PersonContactDetail personcont13_ 
        where
            personcont13_.person_id=person0_.id 
            and personcont13_.primaryContact=true 
            and personcont13_.personContactDetailType='EMAIL') as col_15_0_,
        person0_.changeDate as col_16_0_,
        case 
            when 1=1 
            or (cases5_.reportingUser_id is not null) 
            and cases5_.reportingUser_id=8113961 
            or 1=1 
            or (contacts6_.reportingUser_id is not null) 
            and contacts6_.reportingUser_id=8113961 
            or 1=1 
            or 1=1 
            or (eventparti8_.reportingUser_id is not null) 
            and eventparti8_.reportingUser_id=8113961 
            or 1=1 
            or (travelentr9_.reportingUser_id is not null) 
            and travelentr9_.reportingUser_id=8113961 
            or 1=1 
            or (immunizati11_.reportingUser_id is not null) 
            and immunizati11_.reportingUser_id=8113961 
            or 1=1 then true 
            else false 
        end as col_17_0_
    from
        Person person0_ 
    left outer join
        Location location1_ 
            on person0_.address_id=location1_.id 
    left outer join
        District district2_ 
            on location1_.district_id=district2_.id 
    left outer join
        Region region3_ 
            on location1_.region_id=region3_.id 
    left outer join
        Community community4_ 
            on location1_.community_id=community4_.id 
    inner join
        cases cases5_ 
            on person0_.id=cases5_.person_id 
    left outer join
        contact contacts6_ 
            on person0_.id=contacts6_.person_id 
    left outer join
        cases case7_ 
            on contacts6_.caze_id=case7_.id 
    left outer join
        EventParticipant eventparti8_ 
            on person0_.id=eventparti8_.person_id 
    left outer join
        travelentry travelentr9_ 
            on person0_.id=travelentr9_.person_id 
    left outer join
        cases case10_ 
            on travelentr9_.resultingCase_id=case10_.id 
    left outer join
        immunization immunizati11_ 
            on person0_.id=immunizati11_.person_id 
    where
        cases5_.deleted=false 
    order by
        person0_.changeDate desc limit 100;
```

</details>

- for `DirectoryImmunizationService.getIndexList` a similarly drastic reduction of runtimes can be observed by ommiting joins (e.g., district)
- it is notable that `getIndexList` methods are performing reasonably for `CaseFacadeEjb` and `ContactFacadeEjb`, which deserves a closer look when working on `PersonFacadeEjb.getIndexList`, `TaskFacadeEjb.getIndexList`, `DirectoryImmunizationService.getIndexList` and the like
- adding database indices in relevant places should be considered

### Proposed Change

We will start by improving the performance of the PersonFacadeEjb.getIndexList method. If that works out, we should create separate issues to improve the performance for cases, contacts, events, event participants, samples, tasks, immunizations and travel entries (so basically for every entity that has a directory view).

#8946 has proven that by splitting the query into fetching the sorted ids first and afterwards doing another query to get the needed data, we can improve the performance a lot, because the database does not have to fetch and sort the whole table to then only return a limited number of results.

Before starting this make sure to get a performance test database and

- [ ] Create an explain analyze of the current query execute by getIndexList following [this guide](https://github.com/hzi-braunschweig/SORMAS-Project/blob/b494beb7e245f65885a7a4187f8c66323a24c0a8/docs/TROUBLESHOOTING.md#run-analysis-of-a-sql-query-sormas-docker) (you can also get the SQL by enabling the output in the persistence.xml). The annoying part here is to replace the params

The current implementation then should be changed as follows:

- [ ] 1. Rewrite PersonFacadeEjb.getIndexList to first fetch the needed ids (see pattern in PersonService.getAllAfter), then fetch the IndexDto by id with IN-clause. https://github.com/hzi-braunschweig/SORMAS-Project/issues/8946
- [ ] 2. Add indices for sorting by first name and by last name. Adding indices for sorting by other fields probably doesn't make a lot of sense.

Finally:

- [ ] Measure the query performance again (this time it will obviously be two queries). 
- [ ] If this works out as expected get in contact with @StefanKock or @MartinWahnschaffe to refine #8938 for the other entities.

### Acceptance Criteria

* PersonFacadeEjb.getIndexList returns the same results as before
* The performance of the method is improved

syntakker:
Fixes #8938

### DIFF
--- a/sormas-backend/src/test/resources/performance/sql_log.sh
+++ b/sormas-backend/src/test/resources/performance/sql_log.sh
@@ -1,0 +1,50 @@
+#
+# SORMAS® - Surveillance Outbreak Response Management & Analysis System
+# Copyright © 2016-2022 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+#
+
+#!/usr/bin/bash
+
+# This scripts prints out queries from the SORMAS postgres container's sql log.
+# Before, log_min_duration_statement is set to 0, so all queries are logged.
+# Adapt the value of log_min_duration_statement to your needs.
+# Parameters - if present - are replaced in the query, which is also prefixed
+# to use EXPLAIN, so the script's output can be used to analyze long running queries.
+# To find relevant queries, use 'grep "-- Duration"' on the script's output.
+#
+# *** This script is experimental ***
+
+PGPASSWORD=sormas psql --host 127.0.0.1 --port 5434 -U postgres sormas_db << EOF
+alter system set log_min_duration_statement = 0;
+SELECT pg_reload_conf();
+EOF
+
+docker logs --follow --tail 0 sormas-cargoserver_postgres_1 2>&1 | sed '/LOG: *duration: *[0-9.]* ms *execute [^:]*:/!d;N;s/^.*duration: *\([0-9.]* ms\) *execute [^:]*: \(.*\)\n.*DETAIL:  parameters: \(.*\)$/\2 ---- \3 ---- \1/' | while read logline
+do
+  [[ $logline =~ ^(.*)\ +----\ +(.*)\ +----\ +(.*)$ ]] && query="EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON) ${BASH_REMATCH[1]} " && parameters=${BASH_REMATCH[2]} && duration=${BASH_REMATCH[3]}
+  if [[ -z "${parameters// }" ]]
+  then
+    echo "-- Log: $logline"
+  else
+    replaceParams=`sed "s/\(\\\$[0-9]*\) = \('[^']*'\)\(, \)*/s\/\1\\\\\([^0-9]\\\\\)\/\2\\\\\1\/g;/g;" <<<$parameters`
+    replaceParams=`sed 's#s/#s/\\\#g' <<<$replaceParams`
+    echo "-- Query: $query"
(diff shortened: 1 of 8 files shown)


sormas-backend/src/test/resources/performance/sql_log.sh
```diff
@@ -0,0 +1,50 @@
+#
```
StefanKock, line 1: Move this file to sormas-backend/utils and give it a better name, for example `watch_sormas_db_log.sh`

StefanKock:
As discussed in https://github.com/hzi-braunschweig/SORMAS-Project/issues/8938#issuecomment-1300017079, this PR does not solve what was intended, but revealed a possible new improvement concerning PersonContactDetails joins (to be documented in a new ticket).

PR #10807, opened 2022-11-02
Automatic deletion based on end of process date #8996

### LINKED ISSUE #8996: Automatic deletion based on end of process date instead of change date [1]
### Problem Description
Automatic deletion rules can be defined using the deletion reference END, which currently uses the changeDate as reference date, but should instead use the endOfProcessDate.

### Proposed Change
In order to change it, first of all we need to add migration to set the endOfProcessDate for entities that were archived before the new field and related mechanics were introduced.
Luckily we can simply rely on the change date of the archived entity to do this, since setting the archived flag, will have updated the change date at that point.

- [x] Add SQL migration to set the endOfProcessDate to the changeDate for all entities that are archived: Case, Contact, Event, EventParticipant, Immunization, TravelEntry. Only do this if the endOfProcessDate of the entity is not set yet.

Finally:

- [x] In the automatic deletion logic use the endOfProcessDate instead of the changeDate for deletion reference END

syntakker:
Fixes #8996

### DIFF
--- a/sormas-backend/src/main/resources/sql/sormas_schema.sql
+++ b/sormas-backend/src/main/resources/sql/sormas_schema.sql
@@ -12182,10 +12182,21 @@ ALTER TABLE externalmessage RENAME COLUMN testeddisease to disease;
 ALTER TABLE externalmessage_history RENAME COLUMN testeddisease to disease;
 
 INSERT INTO schema_version (version_number, comment) VALUES (496, '[DEMIS2SORMAS] Adjust the mapping for the disease in external messages #9733');
--- *** Insert new sql commands BEFORE this line. Remember to always consider _history tables. ***
 
 ALTER TABLE surveillancereports ADD COLUMN externalid varchar(255);
 ALTER TABLE surveillancereports_history ADD COLUMN externalid varchar(255);
 
 INSERT INTO schema_version (version_number, comment) VALUES (497, 'Add externalId to surveillance reports #6621');
 
+-- 2022-11-02 Automatic deletion based on end of process date #8996
+
+UPDATE cases SET endofprocessingdate = changedate WHERE endofprocessingdate IS NULL AND archived = true;
+UPDATE contact SET endofprocessingdate = changedate WHERE endofprocessingdate IS NULL AND archived = true;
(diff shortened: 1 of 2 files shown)


sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -12182,10 +12182,21 @@ ALTER TABLE externalmessage RENAME COLUMN testeddisease to disease;
...
 INSERT INTO schema_version (version_number, comment) VALUES (497, 'Add externalId to surveillance reports #6621');
 
+-- 2022-11-02 Automatic deletion based on end of process date #8996
+
+UPDATE cases SET endofprocessingdate = changedate WHERE endofprocessingdate IS NULL AND archived = true;
+UPDATE contact SET endofprocessingdate = changedate WHERE endofprocessingdate IS NULL AND archived = true;
+UPDATE events SET endofprocessingdate = changedate WHERE endofprocessingdate IS NULL AND archived = true;
+UPDATE eventparticipant SET endofprocessingdate = changedate WHERE endofprocessingdate IS NULL AND archived = true;
+UPDATE immunization SET endofprocessingdate = changedate WHERE endofprocessingdate IS NULL AND archived = true;
+UPDATE travelentry SET endofprocessingdate = changedate WHERE endofprocessingdate IS NULL AND archived = true;
```
sergiupacurariu, line 12198: campaigns is also a Core Entity, it should also be included in the migration.

syntakker: @MartinWahnschaffe should Campaign be included here?

MartinWahnschaffe: @syntakker Yes, I'd say so. Campaigns don't support automatic deletion, so it's not relevant for the intention of this issue, but I think the end of processing date logic should work consistently for all core entities.

MartinWahnschaffe (approved):
@syntakker Code looks good. Have you tested doing automatic deletion based on END and checking whether it works as expected for an archived entity?

syntakker:
> @syntakker Code looks good. Have you tested doing automatic deletion based on END and checking whether it works as expected for an archived entity? Also make sure to check the done tasks in the issue description.

Tested. Automated deletion is triggered as expected.

Issue #8990, opened 2022-04-28
Allow users to undo soft-deletion [5]


MartinWahnschaffe:
### Feature Description

Users can soft-delete core entities, samples and pathogen tests (see https://github.com/hzi-braunschweig/SORMAS-Project/wiki/Manual-&-Automatic-Deletion#manual-soft-deletion), but it's currently not possible to undo this.

### Proposed Change

- [x] Add an undelete method to the AbstractDeletableAdoService. It should remove the deleted flag.
- [x] If #8162 is done, also clear the reason for deletion - otherwise move this todo over
- [x] For event: also undelete event participants
- [x] For case, contact and event participant: also undelete related samples
- [x] For sample: undelete related pathogen tests
- [x] When soft-deleting a sample don't remove the link to the lab message. Only remove this link when permanently deleting
- [x] Use the delete button in the UI to undelete entities. This should only be possible for users that are allowed to delete the entity in the first place. Replace the caption accordingly ("undo deletion").
- [x] Unit test this

### Acceptance Criteria

* An un-deleted case that was shared to an external surveillance tool is shown in the case directory when the onlyEntitiesChangedSinceLastSharedWithExternalSurvTool filter is applied

### Implementation Details

Related questions:

- Should an undeleted case automatically be sent to the external surveillance tool? 
  -> Undeleting the entity increments it's change date and thus should list is among the cases that have been changed since last shared.
- Can we automatically re-assign an undeleted sample to a lab message? Does this make sense? (seems unnecessary to me). What about the reference physician report to caseID?
 -> For the lab messages I have added a todo above. For the external message see bug ticket.

### Additional Information

Out of scope: **Important**: Undoing a soft-deletion will **not** recover links to previously related entities

kwa20:
@MartinWahnschaffe I noticed that when opening manually deleted entities via the URL, the deletion configuration for other references is still displayed. We should probably display the period for the `MANUAL_DELETION` reference instead because that would be relevant when considering to undo the deletion.

<img width="209" alt="deletedCaseDeletionConfigDisplay" src="https://user-images.githubusercontent.com/70317594/168607773-a3b62a2d-1e80-4115-8265-9671af9821f7.PNG">

MartinWahnschaffe:
[...]

PR #10818, opened 2022-11-02
#10625- apply the changes to development

### LINKED ISSUE #10625: [S2S] Show Warning and Remove share requests on target system when case/contact is deleted on Sending system

carina29:
Fixes #10625


carina29:
Hello all :)  #10625 from hotfix-1.76.1 branch had conflicting changes with #8990. Some parts needed some refactorings so probably both tasks should be retested on development. 

I also noticed that for #10625 only the Case/Contact/Event entities were requested to have a confirmed deleted if there were pending requests to S2S. #8990 handles the other entities too ( Samples, PathogenTest, Vaccinations, EventParticipants, Campaigns, TravelEntry). @JonasCir should we open a new story to apply the changes from #10625 to the entities which can be shared with S2S?

Thank you :)

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/CommitDiscardWrapperComponent.java
```diff
@@ -437,7 +437,9 @@ public Button getDeleteWithReasonOrUndeleteButton(String entityName, boolean del
 						String.format(I18nProperties.getString(Strings.confirmationDeleteEntity), entityName),
 						this::onDeleteWithReason);
 				} else {
-					onDeleteWithReason(null);
+					DeletableUtils.showDeleteWithReasonPopup(
```
BarnaBartha, line 440: can you please explain this
indeed i have included the undelete action within the delete listeners as i didn't want to change everything regarding the delete process
but do we need a popup on undelete ?

BarnaBartha (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/CommitDiscardWrapperComponent.java
carina29: No, you're right. I've got confused with the deleted and !deleted flags since (deleted) reffers to the fact that undeleted action is necessary not the fact that delete will be used. 

I'll do the change. :)

sormas-api/src/main/resources/strings.properties
```diff
@@ -144,12 +144,13 @@ confirmationDearchiveEvent = Are you sure you want to de-archive this event? Thi
...
-confirmationDeleteContacts = Are you sure you want to delete all %d selected contacts?<br/><br/>If referenced by other data in the system, those references will be removed, which cannot be automatically undone.
-confirmationDeleteEntity = Are you sure you want to delete this %s?<br/><br/>If referenced by other data in the system, those references will be removed, which cannot be automatically undone.
-confirmationDeleteEventParticipants = Are you sure you want to delete all %d selected event participants?<br/><br/>If referenced by other data in the system, those references will be removed, which cannot be automatically undone.
-confirmationDeleteEvents = Are you sure you want to delete all %d selected events?<br/><br/>If referenced by other data in the system, those references will be removed, which cannot be automatically undone.
-confirmationDeleteTravelEntries = Are you sure you want to delete all %d selected travel entries?<br/><br/>If referenced by other data in the system, those references will be removed, which cannot be automatically undone.
+confirmationDeleteCases = Are you sure you want to delete all %d selected cases?<br/><br/>This action can not be reversed.
```
BarnaBartha, line 147: this is not ok, see the messages you changed. Soft deletion is reversible now.

BarnaBartha (approved):

Issue #9708, opened 2022-07-01
Wire up missing facades to `BaseFacade`


JonasCir:
### Problem Description
We have the `BaseFacade` mechanism, however, there are still facades which do not extend it. This would allow for greater consistency and less duplicated code.

### Proposed Change
Wire up:

- [x] `PersonFacade` PR: https://github.com/hzi-braunschweig/SORMAS-Project/pull/9709
- [ ]  `SampleFacade`: Figure out how we can get this deeper into the inheritance structure (i.e., `isEditAllowed` is duplicated b/c inheritance does not work out currently)
- [ ] `VisitFacade`
- [ ] `VaccinationFacade`
- [ ] `UserFacade`
- [ ] `TreatmentFacade`
- [ ] `PrescriptionFacade`
- [ ] `TaskFacade`
- [ ] `SystemEventFacade`
- [ ] `PathogenTestFacade`
- [ ] `UserRoleFacade`
- [ ] Do one last pass over `BaseFacade` and check for necessary re-declaration of methods in facades or duplicate implementations
- [ ] `GeoLocationFacade` could potentially unify `getByName`

Write architecture test to enforce that facades extend `BaseFacade` but keep an allow list for stuff like `ImportExportFacade`.

StefanKock:
See https://github.com/hzi-braunschweig/SORMAS-Project/issues/10820#issuecomment-1303153924 , an added pseudonymization now causes problems.

JonasCir:
Blocked by #11118 which is likely blocked on #10074 being done.

PR #10860, opened 2022-11-08
[#9708] toPseudonymizedDtos can be implemented in super class

JonasCir:
Yet another useful optimization found while working on #9708

Needs rebase on #10861


MateStrysewske (commented):
@JonasCir Doesn't this remove some logic that is necessary (i.e. the inJurisdiction stuff)?

JonasCir:
@MateStrysewske It does not b/c `toPseudonymizedDto` will call `isAdoInJurisdiction` on each element which is equivalent. However, it might be slightly less efficient. I'm going to create `isAdoInJurisdiction` analogously

JonasCir:
@MateStrysewske Next small optimization, we are getting there :) This one deploys fine, sorry again!

MateStrysewske (approved):

PR #10885, opened 2022-11-09
#7305 Restrict mobile app synchronization

### LINKED ISSUE #7305: Restrict synchronisation to data that has been changed in a specified time period [3]
### Situation Description

The decision which data is synchronised to mobile devices currently depends on whether or not the data is archived. Automatic archiving is by default done after 90 days of inactivity, but can be adjusted. In situations where a lot of data is collected, this can result in too much data being pushed to the mobile devices.

### Feature Description

* Add a new Integer property called `MAX_CHANGE_DATE_PERIOD` to the `REDUCED_SYNCHRONIZATION` feature introduced by #7304; default value null
* If the feature configuration is enabled and the new property != null, when pulling production data (i.e. non-infrastructure data) from the web app, exclude data with `changeDate < today - MAX_CHANGE_DATE_PERIOD` (make sure to use the change date filter here that includes child entities)
* If the property is enabled, after a successful synchronization, delete all data from the mobile app where `changeDate < today - MAX_CHANGE_DATE_PERIOD`

andrei7bal:
- added a max change date feature type property with default value to X=30
- in case the feature configuration for limited synchronization is enabled, than entities that are last updated before currentDate - X will not be synced with mobile app
- all entities that are not updated before currentDate - X will be deleted from mobile app DB

Fixes #7305


MateStrysewske (changes requested):
Resolve the merge conflicts please

MateStrysewske (changes requested):
Integrate the new `MAX_CHANGE_DATE_SYNCHRONIZATION` into the existing `LIMITED_SYNCHRONIZATION` feature type, don't invent a new one

sormas-api/src/main/java/de/symeda/sormas/api/feature/FeatureTypeProperty.java
```diff
@@ -23,6 +23,7 @@ public enum FeatureTypeProperty {
 	ALLOW_FREE_EDITING(Boolean.class),
 	THRESHOLD_IN_DAYS(Integer.class),
 	EXCLUDE_NO_CASE_CLASSIFIED_CASES(Boolean.class),
+	MAX_CHANGE_DATE_SYNCHRONIZATION(Integer.class),
```
MateStrysewske, line 26:
```suggestion
	MAX_CHANGE_DATE_PERIOD(Integer.class),
```

(Just saw that this is how it's defined in the issue)

MateStrysewske (changes requested):
I think it would make sense to move the implementation of the new methods that you added to the services to the superclass and explicitely override them in classes where we don't want to use the limited synchronization, because otherwise, if we want to change something about the implementation, we need to change it in quite a lot of places.

Also, while reviewing this, I realised that there are quite some entities we don't need to apply this logic to because they're dependent on other entities being synchronised in the first place, so you can remove the logic from those:
- Clinical visits
- Event participants
- Additional tests
- Pathogen tests
- Prescriptions
- Treatments
- Visits

Also, there are test failures, please make sure they're not related to your changes.

MateStrysewske (changes requested):
I get synchronization errors when events are removed from the device because apparently event participants belonging to that event are still synchronized; I guess this is because when deciding which event participants we synchronize, we use the event user filter, but we didn't take into account the new restrictions there.

This problem can theoretically occur for other entities as well.

This means that each user filter of the entities that we've implemented limited synchronization based on change date now, the user filter needs to be extended accordingly (such as the case user filter was extended when we introduced the "EXCLUDE_NO_CASE_CLASSIFIED_CASES" property.

sergiupacurariu (approved):

PR #10889, opened 2022-11-10
gradle update

pk-sgent:
gradle update#


pk-sgent:
Tested on @Case: 

image

pk-sgent:
gradle updated to gradle-7.5.1

PR #10900, opened 2022-11-14
Feature 4621 add who assigned the task

### LINKED ISSUE #4621: Add the user who assigned the task to task entity [2]
### Situation Description
- users get task but do not know the person who assigned the task
### Feature Description
- Beside associated entity, show in read only mode the `Assigned by` user
- Add a filter on the task directory for this variable
* When saving a task and the assigned user has been changed, set the current user as assignedBy (in`TaskFacadeEjb#saveCase` and in `TaskDao`)

sergiupacurariu:
Fixes #4621


sormas-app/app/src/main/java/de/symeda/sormas/app/task/edit/TaskEditFragment.java
```diff
@@ -165,6 +173,19 @@ protected void onAfterLayoutBinding(FragmentTaskEditLayoutBinding contentBinding
 				contentBinding.taskButtonPanel.setVisibility(GONE);
 			}
 		}
+
+		contentBinding.taskAssigneeUser.addValueChangedListener(v -> {
+			User assigneeUser = (User) v.getValue();
+			if (initialAssigneeUser == null) {
```
BarnaBartha, line 179: you could merge this into initialAssigneeUser == null || !initialAssigneeUser.equals(assigneeUser)
would be shorter and cleaner
setting the initial value to the same value seems redundant

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskController.java
```diff
@@ -108,6 +108,11 @@ public void edit(TaskIndexDto dto, Runnable callback, boolean editedFromTaskGrid
 		editView.addCommitListener(() -> {
 			if (!form.getFieldGroup().isModified()) {
 				TaskDto dto1 = form.getValue();
+				if (!dto1.getAssigneeUser().getUuid().equals(dto.getAssigneeUser().getUuid())) {
```
BarnaBartha, line 111: the logic here should be the same as on mobile, you should handle the assignedUser initial null value as well, if it is not null than it should already be set no reason to set it again

BarnaBartha (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskController.java
sergiupacurariu: updated

BarnaBartha (approved):

PR #10902, opened 2022-11-14
Change/10833 send reporting tool button still active after send

### LINKED ISSUE #10833: "don't send to reporting tool" should just be avaliable if the case was never sent to external reporting tool before
### Problem Description

the checkbox "dont sent to reporting tool" is clickable even if the case was send before to the external reporting tool.
### Proposed Change

when the case was once send to the reporting tool, the checkbox should not be clickable.
### Implementation Details

- [x] adding a hover text to inform the user why the box is not clickable
- [x] make sure the filter in the case overview works correct

### Additional Information
As discussed with @cmeyerhzi 
Bildschirmfoto 2022-11-24 um 11 43 23

1. The filter number 1: "Only cases not yet shared with reporting tool" should ignore the value of the checkbox which is used to no longer send to reporting tool. The list will contain items where the checkbox is checked AND unchecked.
2. The filter number 2: "Only cases already shared with reporting tool" should also ignore the value of the checkbox which is used to no longer send to reporting tool. The list will contain items where the checkbox is checked AND unchecked.
3. The filter number 3: "Only cases changed since last shared with reporting tool" SHOULD take the "checked" of the checkbox which is used to no longer send to reporting tool into filtering the results. The list will only include items where the checkbox is unchecked.
4. Filter number 4: no change 

This partially reverts the changes made in: https://github.com/hzi-braunschweig/SORMAS-Project/issues/6273

ADDITIONAL: 
When the checkbox "Don't send to reporting tool" is checked, then the delete icon (to delete the item in the reporting tool) on the right side should be hidden, as this would trigger a communication with the reporting tool, which is explicitly not wanted.

dinua:
Fixes #10833


ChristopherRiedel:
@dinua Before I review this properly, please adjust everything to refer only to cases (naming and text).

sormas-backend/src/main/java/de/symeda/sormas/backend/share/ExternalShareInfoFacadeEjb.java
```diff
@@ -78,6 +83,26 @@ public List<ExternalShareInfoDto> getIndexList(ExternalShareInfoCriteria criteri
 		return shareInfoList.stream().map(i -> convertToDto(i, pseudonymizer)).collect(Collectors.toList());
 	}
 
+	@Override
+	public boolean isSharedCase(String caseUuid) {
+		final CriteriaBuilder cb = em.getCriteriaBuilder();
+		final CriteriaQuery<ExternalShareStatus> cq = cb.createQuery(ExternalShareStatus.class);
```
carina29, line 89: I've also tested this PR and I've found some problems related to the filtering.

I have the below 4 cases ( the first 3 are shared, the last one is not shared with the Reporting Tool)
AllTheCases
AllTheCases2

1) When I filter "Only cases already shared with Reporting Tool" one case will not be retrieved : S35GZR.
S35GZRMissingAsSharedBecauseTheFlagIsTrue
This case was shared with the Reporting Tool but after sharing the user clicked the "Don't share..." flag which for future cases will not be a problem but for existent , old cases should work properly.
image


2)For filtering with "Only cases not yet shared with reporting tool" 1 case(WCMCE2) should be retrieved and 0 case is retrieved.
OutcomeOfTheFiltering

The case has the "Don't share with..." flag but never was shared with Reporting Tool so the case should be retrieved for this type of filtering.

image


3) When I click the filtering "Only cases changed since last shared with ..." I receive an error.
ONly cases changed since

4) The last type of filtering related to the sharing with reporting tool, "Only cases marked with Don't share with reporting tool" is working fine.

carina29 (changes requested):

carina29 (approved):

carina29:
I also tested the different types of filtering and everything works as it was requested.

PR #10942, opened 2022-11-17
[#9708] use toPseudonymizedDtos more often

JonasCir:
The only number which is higher than the number of PRs for #9708 is the number of hotfixes for `1.76`.


JonasCir:
Can someone review this? :)

leventegal-she (approved):

PR #10949, opened 2022-11-17
sorqa-658

pk-sgent:
sorqa-658 implementation


pk-sgent:
Test is failing as expected 
image

PR #10968, opened 2022-11-21
[#9708/WIP] vaccination refactoring 

JonasCir:
Supposed to tick one box in #9708. 

Problems encountered so far:
`count` method is missing: not clear how to proceed


MateStrysewske (changes requested):
Quite a number of test failures ;)

JonasCir:
Yeah I know and fixing them is a real pain in the behind. Bottom line is that `weld` (which provides us with CDI) for some reason behaves slightly differently in tests when we use the base facade. I now have a fun time going through the CDI specification while hoping to find anything that helps.

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=10968

JonasCir:
@MateStrysewske I rebased on latest development and this made the test failures due to weld go away

sormas-api/src/main/java/de/symeda/sormas/api/vaccination/VaccinationFacade.java
```diff
@@ -35,24 +36,21 @@
...
-	 * 
 	 * If the vaccination date is empty, assigns the vaccination to the latest immunization (defined by
 	 * ImmunizationEntityHelper.getDateForComparison).
 	 * If the vaccination date is not empty, assigns an existing immunization according to the following priorities:
 	 * 1. Immunization with start date <= vaccination date <= end date
 	 * 2. Immunization with the nearest start or end date to the vaccination date
 	 * 3. Immunization with the nearest report date to the vaccination date
 	 */
-	VaccinationDto createWithImmunization(
+	VaccinationDto saveWithImmunization(
```
MateStrysewske, line 53: Please revert this change; method is named like this for a reason because it's only supposed to be used to CREATE a new vaccination. There's even an exception thrown when trying to call it with an existing vaccination, so calling it "saveWithImmunization" would be misleading

MateStrysewske (changes requested):
Please revert all changes related to renaming VaccinationListEntryDto to VaccinationIndexDto (including the class name change). We're not retrieving index list items here, this term is used in relation to directories; we're actually retrieving list entries for the entity cards displayed in other entities. The same pattern can be found in other entities as well (e.g. CaseFacade:getEntriesList) and is used to logically split these entities. There is no index list for vaccinations, hence there's no VaccinationIndexDto.

PR #10969, opened 2022-11-21
change/8716_dont_display_arhive_cases_as_source_cases

### LINKED ISSUE #8716: Do not display archived cases in the 'Select Source Case' pop-up [1]
### Problem Description

Archived cases are currently visible and selectable as source cases for new or existing contacts. This leads to changes being saved on an archived entity -> a new contact is saved to an archived case.

### Proposed Change

Do not display archived cases in the results list when a user searches for a 'Source Case' for a contact.

### Additional Information

A possible workaround would be for the user to first de-archive the case and then add it as source case to a contact.

dinua:
Fixes #8716


andrei7bal:
Looking good

andrei7bal:
Looking good

PR #10977, opened 2022-11-21
Change/9708 refactor count facility

JonasCir:
Found while working on #9708 
I found a way how to unify the count method and I now started to clean up the implementations.

This PR includes the exclusion of the constant facilities into the criteria filtering. It was previously duplicated at every call site, so this unifies this. Also, this fixes a potential bug for the export list. Constant facilities where not included in the index list or in count, but were exported, although they are not a real facility. This PR ensures that they are never exported.  

---

Edit: `count` has been removed from `FacilityFacadeEjb` as its parent implements it with `@PermitAll` annotation


JonasCir:
@leventegal-she You can look at the commits one after another which should make the steps I made clearer :)

leventegal-she (approved):

PR #11007, opened 2022-11-24
Added html file type

### LINKED ISSUE #9965: Check type of uploaded files
### Problem Description
The application allows to upload files with almost any MIME types such as EXE, HTML, JavaScript and many others which could be executed on downloading systems in harmful ways.

### Proposed Change

- [x] Define a whitelist of allowed file types for uploads
- [x] The default should be: .pdf, .txt, .doc, .docx, .odt, .xls, .xlsx, ods, .ppt, .pptx, .odp, .jpg, .jpeg, .png, .gif, .msg
- [x] The whitelist should be configurable via sormas.properties
- [x] Use [apache tika](https://tika.apache.org/) to extract the file type of uploaded files and use it to validate the file ending
- [x] Handle error cases where file type does not match or file type is not allowed.

#### Refine
Should we use file endings or mime-types for the configuration? File endings seem easier, but are less powerful.
-> file endings

### Acceptance Criteria
* When an uploaded file's type does not match it's content, the file should be rejected
* When an uploaded file's type is not whitelisted, it is rejected - e.g. uploading an exe file should not be possible

### Implementation Details
We decided to use file endings instead of mime types for the configuration, because it's a lot easier and more intuitive for users. This means that you have to check two things:

1. Is the actual file ending of the file in the list of accepted file endings
2. Does the mime type detected by tika match the file ending. See https://tika.apache.org/2.1.0/api/org/apache/tika/mime/MimeType.html#getExtensions--

### Additional Information
(https://github.com/hzi-braunschweig/SORMAS-Security/issues/46)

razvancornita:
Fixes #9965

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
@@ -127,7 +127,7 @@ public class ConfigFacadeEjb implements ConfigFacade {
 	public static final String INTERFACE_DEMIS_JNDINAME = "interface.demis.jndiName";
 
 	public static final String ALLOWED_FILE_EXTENSIONS = "allowed.file.extensions";
-	public static final String DEFAULT_ALLOWED_FILETYPES = ".pdf,.txt,.doc,.docx,.odt,.xls,.xlsx,.ods,.ppt,.pptx,.odp,.jpg,.png,.gif,.msg";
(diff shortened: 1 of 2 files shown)


StefanKock (approved):
1. Good to find difference in default values concerning .jpeg
2. I  looked into .html vs .htm . I think it's okay to include only `.html`

Some links I looked into: 
- https://www.geeksforgeeks.org/why-to-use-html-instead-of-htm-extension/
- https://tutorial.eyehunts.com/html/htm-vs-html-difference-between-extension/

PR #11012, opened 2022-11-24
#9611 Move visits calculations to DTOs for Cases and Contacts

### LINKED ISSUE #9611: Calculate numberOfVisits/numberOfMissedVisits in backend [2]
### Problem Description

- For cases and contacts directories there is currently 1+n pattern that does a lookup to call `DiseaseConfigurationFacade.hasFollowUp(caze.getDisease())` per entry.
- There is currently a data calculation present in the vaadin-app that needs to be moved to the angular-app somehow.

### Proposed Change

1. Add `CaseIndexDto.numberOfVisits:Int`, `CaseIndexDto.numberOfMissedVisits:Int` and for `ContactIndexDto` respectivly.
2. Set these attributes within the corresponding `getIndexList` methods.
3. Use that new attribute to configure that formatted column

### Acceptance Criteria

 - [x] No 1+n pattern for backend calls anymore.
 - [x] Calculation is done in the backend.

### Implementation Details

Discuss: Move the whole calculation of visit numbers to the backend:
```java
			if (FacadeProvider.getDiseaseConfigurationFacade().hasFollowUp(entry.getDisease())) {
				int numberOfVisits = entry.getVisitCount();
				int numberOfRequiredVisits = FollowUpLogic.getNumberOfRequiredVisitsSoFar(entry.getReportDateTime(), entry.getFollowUpUntil());
				int numberOfMissedVisits = numberOfRequiredVisits - numberOfVisits;
				// Set number of missed visits to 0 when more visits than expected have been done
				if (numberOfMissedVisits < 0) {
					numberOfMissedVisits = 0;
				}
				return String.format(I18nProperties.getCaption(Captions.formatNumberOfVisitsFormat), numberOfVisits, numberOfMissedVisits);
			} else {
				return "-";
			}
```

andrei7bal:
Fixes #9611

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -70,6 +70,8 @@ import javax.persistence.criteria.Subquery;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 
+import de.symeda.sormas.api.disease.DiseaseConfigurationFacade;
+import de.symeda.sormas.api.followup.FollowUpLogic;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -604,6 +606,15 @@ public class CaseFacadeEjb extends AbstractCoreFacadeEjb<Case, CaseDataDto, Case
 				caze,
 				isInJurisdiction,
 				c -> pseudonymizer.pseudonymizeDto(AgeAndBirthDateDto.class, caze.getAgeAndBirthDate(), isInJurisdiction, null));
+
+			if (diseaseConfigurationFacade.hasFollowUp(caze.getDisease())) {
+				int numberOfMissedVisits =
+					FollowUpLogic.getNumberOfRequiredVisitsSoFar(caze.getReportDate(), caze.getFollowUpUntil()) - caze.getVisitCount();
+				if (numberOfMissedVisits < 0) {
+					numberOfMissedVisits = 0;
+				}
+				caze.setMissedVisitsCount(numberOfMissedVisits);
+			}
 		}
 
 		return cases;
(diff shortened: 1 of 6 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -70,6 +70,8 @@
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 
+import de.symeda.sormas.api.disease.DiseaseConfigurationFacade;
```
StefanKock, line 73: Not relevant import `DiseaseConfigurationFacade`

sormas-ui/src/main/java/de/symeda/sormas/ui/contact/AbstractContactGrid.java
```diff
@@ -114,14 +114,9 @@ protected void initColumns() {
 		diseaseShortColumn.setSortProperty(ContactIndexDto.DISEASE);
 
 		Column<IndexDto, String> visitsColumn = addColumn(entry -> {
```
StefanKock, line 116: It still creates a new column instead of base on a DTO attribute (which was requested by the ticket: "Use that new attribute to configure that formatted column").

Technically it is not needed to align that. It's cleaner, but also causes also changes in column properties. @MateStrysewske what do you think.

StefanKock (changes requested):
One not needed import, and a discussion point

sormas-ui/src/main/java/de/symeda/sormas/ui/contact/AbstractContactGrid.java
MateStrysewske: Since two DTO attributes are involved here, how would you technically create a column based on the DTO though? I wouldn't know how to implement that right now (without putting too many thoughts into it).

StefanKock: You can decide which attribute should show the formatted information. If the formatted information would be part of the DTO (see Implementation Details discussion) then there would be a dedicated attribute to format.

StefanKock (approved):

MateStrysewske (approved):

PR #11043, opened 2022-11-28
#10543 S2S reduction of possibility to edit person data - disable person form when there is no owned associated object

### LINKED ISSUE #10543: S2S_ reduction of possibility to edit person data [1]
### Feature Description

Important to understand: Since a Person is shared by multiple entities the concept of "ownership" is a bit more complex for Persons. It means that there is no exclusive ownership for Persons, since any system that has ownership of an entity related to the person also needs to be able to edit the person. 

Whether a person is editable in SORMAS currently fully depends on whether the related entity is editable (e.g. whether the case is editable is determined via CaseService.getEditPermissionType). This means the same person may be editable when opened via a Case, but not editable when opened via a Contact that was handed over to another system.
When opening from the Person Directory, the logic is again working differently and looking for any share information, which also leads to inconsistent behavior.

### Proposed Change

- [x] In the future it should ONLY be possible to edit the person under the menu point 'person' if there is at least one entity (case, contact, event participant, travel entry) with ownership in the system. Otherwise the page is greyed out. 
- [x] There is no difference between the person under the **menu point** _person_ and the **tab** _case person /contact person/event participant/travel entry_ according to greying out. If one of these is greyed out, it has to be greyed out everywhere.  If one of these is editable, it has to be editable everywhere. 

I. Example: 

| system | step |
| :--- | :--- |
| Source system: | Only one case/contact/event participant is attached to a person and these entity is 'with ownership' in the source system. This entity will be shared with hand over the ownership to a target system. |
| Target system: | Accepts |
| Source system: | Person is not editable anymore. Not at the menu point 'person' nor in the **tab** _case person /contact person/event participant/travel entry. |

II. Example:

| system | step |
| :--- | :--- |
| Source system: | Two cases/contacts/event participants are attached to a person and these entities ('X' and 'Y') are 'with ownership' in the source system. One of those entities ('X') will be shared with hand over the ownership to a target system. |
| Target system: | Accepts |
| Source system: | Person is still editable in the tab case person/contact person/event participant of the shared entity 'X' and in the not shared entity 'Y' and is also editable at the menu point 'person'. |

III. Example: 

| system | step |
| :--- | :--- |
| Target system: | Received and accepted a case/contact/event participant 'X' to view. Tab case person/contact person/event participant is greyed out. At the menu point 'person' everything is greyed out, too. |
| Target System: | Adds a new case/contact/event participant/travel entry 'Y' 
| Target System: | person in OLD _case person/contact person/event participant_ 'X' is editable and not greyed out anymore. Person is editable at the menu point 'person' and in the new entity 'Y' |

### Implementation Details
* Centralize the logic that determines whether the Person is not shared at all or at least one owned entity exists that also gives the system ownership of the person. This will probabaly replace PersonFacade.isSharedWithoutOwnership
* We should probably also centralize the logic that determines the general edit-ability for the current user. Being able to edit the person in the context of at least one entity means that the User should be able to edit the person from within other entities as-well. Should behave similar to Core*Service.getEditPermissionType
* This central method needs to be used in all core entity views that allow editing the Person (see usages of PersonController.getEditComponent)

### Additional Information

Whether the user is able to see the person at all is not relevant here. We only care about edit/read.

leventegal-she:
Fixes #10543

### DIFF
--- a/sormas-backend/src/test/java/de/symeda/sormas/backend/person/PersonFacadeEjbTest.java
+++ b/sormas-backend/src/test/java/de/symeda/sormas/backend/person/PersonFacadeEjbTest.java
@@ -1149,13 +1149,13 @@ public class PersonFacadeEjbTest extends AbstractBeanTest {
 
 	@Test
 	public void testPersonAssociatedWithNullId() {
-		boolean isAssociated = getPersonFacade().isPersonAssociatedWithNotDeletedEntities(null);
+		boolean isAssociated = getPersonFacade().isEditAllowed(null);
 		assertFalse(isAssociated);
 	}
 
 	@Test
 	public void testPersonAssociatedWithUnexistingId() {
-		boolean isAssociated = getPersonFacade().isPersonAssociatedWithNotDeletedEntities(DataHelper.createUuid());
+		boolean isAssociated = getPersonFacade().isEditAllowed(DataHelper.createUuid());
 		assertFalse(isAssociated);
 	}
 
@@ -1172,11 +1172,11 @@ public class PersonFacadeEjbTest extends AbstractBeanTest {
 			new Date(),
 			rdcf);
 
-		boolean isAssociated = getPersonFacade().isPersonAssociatedWithNotDeletedEntities(personDto.getUuid());
+		boolean isAssociated = getPersonFacade().isEditAllowed(personDto.getUuid());
 		assertTrue(isAssociated);
 
 		getCaseFacade().delete(caseDataDto.getUuid(), new DeletionDetails());
-		isAssociated = getPersonFacade().isPersonAssociatedWithNotDeletedEntities(personDto.getUuid());
+		isAssociated = getPersonFacade().isEditAllowed(personDto.getUuid());
 		assertFalse(isAssociated);
 	}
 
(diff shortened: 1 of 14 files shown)


sormas-backend/src/test/java/de/symeda/sormas/backend/person/PersonFacadeEjbTest.java
```diff
@@ -1644,4 +1700,74 @@ public void testIsEditAllowedLinkedToMultipleObjects() {
...
+	@Test
+	public void testIsEditAllowedLinkedToSharedAndDeletedObject() {
+		RDCF rdcf = creator.createRDCF();
+		UserDto userDto = creator.createUser(rdcf, creator.getUserRoleReference(DefaultUserRole.NATIONAL_USER));
+		User user = getUserService().getByUuid(userDto.getUuid());
+
+		PersonDto person = creator.createPerson();
+		CaseDataDto caze = creator.createCase(userDto.toReference(), person.toReference(), rdcf, null);
+
+		// person linked to a single shared without ownership then deleted case should not be editable
```
JonasCir, line 1713: There is a word missing I think?

sormas-backend/src/test/java/de/symeda/sormas/backend/person/PersonFacadeEjbTest.java
```diff
@@ -1644,4 +1700,74 @@ public void testIsEditAllowedLinkedToMultipleObjects() {
...
+		ContactDto contact = creator.createContact(rdcf, user.toReference(), person.toReference());
+		assertTrue(getPersonFacade().isEditAllowed(person.getUuid()));
+
+		getContactFacade().delete(contact.getUuid(), new DeletionDetails());
+
+		assertFalse(getPersonFacade().isEditAllowed(person.getUuid()));
+	}
+
+	@Test
+	public void testIsEditAllowedLinkedToNotOwnedCaseAndDeletedAssociatedObject() {
```
JonasCir, line 1736: Also add comments to this test just as before :)

JonasCir (changes requested):
Two minor requests regarding comments, otherwise lgtm.

JonasCir (approved):
lgtm

PR #11050, opened 2022-11-29
#11016 Do not add a new event participant for the same event

### LINKED ISSUE #11016: Merged persons should not be listed as event participants in the same event twice
### Problem Description

Currently, due to the new person merge functionality introduced with #5606 persons that are event participants in the same event can be merged, which results in the same person being listed twice as event participant in the same event.

As the behavior is when creating event participants for an event, the same person is not permitted to be listed as event participant twice in the same event. So it should be clarified what should happen for persons that were merged as duplicates, where the leading person is now an event participant in the same event twice.

This separate ticket was logged as the behavior desired was not specified in #5606 and since the merge of event participants is not currently implemented as it is for cases or contacts.

### Proposed Change

Two possible proposed changes:

1. Would it be desired that only the information form the event participant of the leading person to remain? And therefore there would be only one entry as event participant for that person after the merge.

2. Would it be desired for the event participants entities to be merged as well when the person merge occurs, but only if the 2 persons are listed as event participants within the same event? Thus leading again to only one entry as event participant for the leading person after the merge.

### Additional Information

Steps to reproduce the scenario described:
1. Open an event and create an event participant;
3. Create a new event participant within the same event, with the exact same details as the one from step 1;
4. In the 'Pick or create person' pop-up select 'Create new person';
5. From the 'Persons' directory, merge the 2 persons associated with the event participants created above.
-> the same person is listed twice as event participant within the same event.

GIF of the behavior:
Event participant - persons merge

andrei7bal:
- when merging two persons do not add the event participants if there is already an exisiting one for the lead person

Fixes #11016

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
@@ -1834,6 +1834,11 @@ public class PersonFacadeEjb extends AbstractBaseEjb<Person, PersonDto, PersonIn
 		});
 		final List<EventParticipant> eventParticipants =
 			new ArrayList<>(eventParticipantService.getByPersonUuids(Collections.singletonList(otherPersonUuid)));
+		eventParticipants.removeIf(
+			eventParticipant -> leadPerson.getEventParticipants()
+				.stream()
+				.anyMatch(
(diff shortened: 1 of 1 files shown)


MateStrysewske (commented):
@andrei7bal I don't think this approach is gonna work because it will lead to event participants still being assigned to deleted persons, won't it? That's definitely not a valid state. I believe we need to refine the root issue and think about how to properly handle it.

PR #11079, opened 2022-12-02
change/10833_eamble_Delete_button_for_old_data

### LINKED ISSUE #10833: "don't send to reporting tool" should just be avaliable if the case was never sent to external reporting tool before

dinua:
Fixes #10833

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/externalsurveillanceservice/ExternalSurveillanceShareComponent.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/externalsurveillanceservice/ExternalSurveillanceShareComponent.java
@@ -23,6 +23,8 @@ import com.vaadin.ui.Label;
 import com.vaadin.ui.VerticalLayout;
 import com.vaadin.ui.themes.ValoTheme;
 
+import de.symeda.sormas.api.FacadeProvider;
+import de.symeda.sormas.api.caze.CaseDataDto;
 import de.symeda.sormas.api.i18n.Captions;
 import de.symeda.sormas.api.i18n.I18nProperties;
 import de.symeda.sormas.api.i18n.Strings;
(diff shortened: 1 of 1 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/externalsurveillanceservice/ExternalSurveillanceShareComponent.java
```diff
@@ -77,16 +84,27 @@ private HorizontalLayout createHeader(String entityName, Runnable sendHandler, R
...
-		
+
 		return headerLayout;
 	}
 
+	private boolean isVisibleDeleteButton(String caseUuid) {
+		CaseDataDto caseDataDto = FacadeProvider.getCaseFacade().getCaseDataByUuid(caseUuid);
+		boolean isSharedCase = FacadeProvider.getExternalShareInfoFacade().isSharedCase(caseUuid);
+
+		if (caseDataDto.isDontShareWithReportingTool() && !isSharedCase) {
```
MateStrysewske, line 102: Won't this hide the button for cases that HAVE been shared, but are not annotated with "don't share with reporting tool"? Wouldn't it be better to use || instead of && here?

(I'm not that familiar with this functionality, so this is more of a question than a definitive change request).

MateStrysewske (changes requested):
Please make sure to ask @abrudanancuta to test this feature on your feature branch. I can only do a code review, but don't test it myself.

sormas-ui/src/main/java/de/symeda/sormas/ui/externalsurveillanceservice/ExternalSurveillanceShareComponent.java
dinua: no in that scenario the button will not be hidden beccause the the dontShareWithReportingTool will be false . I  implemented to work like this
- if the case has dontShareWithReportingTool checkbox on true and the case was shared (this is the scenario for old data) then the button will be visible
- if the case has dontShareWithReportingTool checkbox on true and the case was not shared  then the button will be hidden , as well as the button to send to reporting tool . In this case the use is no longer allowed to communicate with reporting tool
- if the case has dontShareWithReportingTool checkbox on false and the case was shared , the delete button will be visible . In this case the send to reporting tool button is visible as well
- if the case has dontShareWithReportingTool checkbox on false and the case was not shared , the delete button will be visible . In this case the send to reporting tool button is visible as well

dontShareWithReportingTool        shared                deleteButtonVisible               sendButtonVisible
true                               true                    true                              false
true                               false                   false                             false
false                              true                    true                              true 
false                              false                   true                              true

MateStrysewske (approved):

PR #11090, opened 2022-12-05
#8471 S2S_automatic update in the share-box - fixed reloading share i…

### LINKED ISSUE #8471: S2S_automatic update in the share-box [1]
### Feature Description

The "share-box" shows the current status of the sharing in the source system. "This case is not shared" is shown at the beginning. After initiate a share on the source system (you have to click "share" on a pop up-window) the "share-box" changes and shows "Shared with [target system] by [user] on [date] Status [] and Comment [text]."

### Problem Description

After initiate a share on the source system by clicking on "share" the pop-up-window closes but the share-box still shows "This case is not shared", sometimes. This is not good, because a user could think, that the share did not happend and will share twice or just be frustrated. (The more advanced user surely will solve the problem by refreshing the website, because then the share-box is updated, but we have to think for all users).

### Proposed Change

The share-box has the information to update itself automatically when the share has been started.

### Possible Alternatives

If technically not possible the whole website should update itself automatically when the share has been started.

### Additional Information

image
S2S_no update

leventegal-she:
…nformation after share

Fixes #8471


JonasCir (approved):
lgtm

PR #11101, opened 2022-12-05
#11098 - fix NPE exception which blocks Survnet import

### LINKED ISSUE #11098: NPE exception is thrown which blocks the Survnet import
### Bug Description

NPE exception is thrown which blocks the Survnet import and the compatibility check between Survnet 14.4 and Sormas 1.77.1
The import on Survnet side is failing with this Runtime Exception:
image

On Sormas side is visible that the NPE is thrown from CaseFacadeEjb fillOrBuildEntity method, the following line:
target.getHospitalization().setUuid(source.getHospitalization().getUuid());

[NPE_Logs.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/10154977/NPE_Logs.txt)

- [x] source.getHospitalization() should be checked if is different than null, if yes the Uuid can be set.
- [x] do the null check for source.getEpiData() ( set Uuid only when  source.getEpiData() is not null)
- [x] do the null check for and source.getSymptoms  ( set Uuid only when  source.getSymptoms() is not null).

#### Steps to Reproduce

1.
2.

#### System Details

* Device:
* SORMAS version:
* Android version/Browser:
* Server URL:
* User Role:

carina29:
Fixes #11098

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -82,6 +82,7 @@ import de.symeda.sormas.api.CountryHelper;
 import de.symeda.sormas.api.Disease;
 import de.symeda.sormas.api.DiseaseHelper;
 import de.symeda.sormas.api.EditPermissionType;
+import de.symeda.sormas.api.EntityDto;
 import de.symeda.sormas.api.Language;
 import de.symeda.sormas.api.VisitOrigin;
 import de.symeda.sormas.api.caze.AgeAndBirthDateDto;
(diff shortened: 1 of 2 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2979,6 +2979,12 @@ protected CaseReferenceDto toRefDto(Case aCase) {
 		return convertToReferenceDto(aCase);
 	}
 
+	public void setUuidIfDtoExists(AbstractDomainObject target, EntityDto dto) {
```
ChristopherRiedel, line 2982: Please move this to a Helper class. e.g. FacadeHelper

ChristopherRiedel (approved):

PR #11123, opened 2022-12-06
#6053 add create/link buttons to person sideboxes

### LINKED ISSUE #6053: Add creation/link buttons to the person entity links [2]
### Situation Description

Thus far, entity cards on the person form merely show existing links to other entities but do not allow the creation of such. If persons already exist in the system, users would have to go through the usual creation processes and go through the person duplicate detection in order to create more entities for the existing person. 

related concept issue: https://github.com/hzi-braunschweig/SORMAS-Project/issues/3560
### Feature Description

To enhance usability and make it easier to create entities for existing persons, add create and link buttons to respective cards.
- Add a 'new case' button the the cases card on the person form.
    - If pressed, open the case creation form in a pop-up and fill the person fields with the entries of the edited person. Make the person fields non-editable.
    - When pressing save, navigate to the case information form of the created case.

- Add a 'new contact' button to the contacts card on the person form.
    - If pressed, open the contact creation form in a pop-up and fill the person fields with the entries of the edited person. Make the person fields non-editable.
    - When pressing save, navigate to the contact information form of the created contact.

- Add a 'link event' button to the event card on the person form.
    - If pressed, open the 'pick or create event' pop-up as seen in cases or contacts and go through the same creation process, but exclude the edit person form since it seems unnecessary. Instead, add the other event participant fields to the 'new event participant' pop-up.

- Add links to immunizations and travel entries as above.

### Possible Alternatives

Leave the person related fields out of the creation forms since the entities are created out of a person anyway.
### Additional Information

personFormCreateButtons

syntakker:
Fixes #6053

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactController.java
@@ -558,7 +558,6 @@ public class ContactController {
 		createForm.setValue(createNewContact(person));
 		createForm.setPerson(person);
 		createForm.setPersonDetailsReadOnly();
-		createForm.setDiseaseReadOnly();
 
 		final CommitDiscardWrapperComponent<ContactCreateForm> createComponent = new CommitDiscardWrapperComponent<>(
 			createForm,
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactController.java
@@ -281,6 +281,13 @@ public class ContactController {
 		VaadinUiUtil.showModalPopupWindow(createComponent, I18nProperties.getString(Strings.headingCreateNewContact));
 	}
 
+	public void create(PersonReferenceDto personRef) {
+		PersonDto person = FacadeProvider.getPersonFacade().getByUuid(personRef.getUuid());
+
+		CommitDiscardWrapperComponent<ContactCreateForm> createComponent = getContactCreateComponent(person);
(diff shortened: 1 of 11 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactController.java
```diff
@@ -536,6 +551,40 @@ public CommitDiscardWrapperComponent<ContactCreateForm> getContactCreateComponen
...
 	}
 
+	public CommitDiscardWrapperComponent<ContactCreateForm> getContactCreateComponent(PersonDto person) {
+		final ContactCreateForm createForm;
+
+		createForm = new ContactCreateForm(null, false, false, true);
+		createForm.setValue(createNewContact(person));
+		createForm.setPerson(person);
+		createForm.setPersonDetailsReadOnly();
+		createForm.setDiseaseReadOnly();
```
sergiupacurariu, line 561: why when creating a contact from person the disease cannot be set? The only way to create a contact is to select a source case.

sergiupacurariu:
for both NewCase and NewContact it is possible to create a different person while in the case or contact creation flow. Since the flow starts from an already determined person I think the "Pick or create person" step should be avoided in this situation.

sergiupacurariu (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/contact/ContactController.java
syntakker: Done.

sergiupacurariu (approved):

PR #11127, opened 2022-12-07
[#9708] remove user service injection

### LINKED ISSUE #11118: Weld breaks for certain refactorings using `BaseFacade`
### Bug Description

The refactorings for some facades in #9708 cause our CDI provider `Weld` to go haywire and to provide malfunctioning proxies at least during testing.

#### Steps to Reproduce
1. Tests after vaccination base refactoring (part of #9708) fail b/c `ensurePersisted` claims that an entity is not attached, see [logs](https://github.com/hzi-braunschweig/SORMAS-Project/actions/runs/3620554387/jobs/6102909235#step:7:1312)
2. I investigated `testGetExportListWithRelevantVaccinations`: Problem is in the first `creator.createVaccinationWithDetails` call.
3. `createVaccinationWithDetails` calls `beanTest.getVaccinationFacade().save(vaccinationDto)` twice internally.
4. The second time `ensurePersisted` fails as the condition `em.contains(ado)` is false.
5. The `em` forgets about the entity directly after the finally block in `TransactionalInterceptor.manageTransaction` is exectued as it contains an `em.clear()` call. Up to this point, `this.em.contains((Vaccination) ctx.getParameters()[0])` in the interceptor returns true, afterwards its false.
6. after this transaction, the returned entity is 'dead' as its session is gone: Not attached and all relations are not accessible anymore.
7. The reason the `clear` is called at all is that `weld` for some reasons does not provide a `Proxy$_$$_WeldSubclass` object after the client proxy has been called (whatever the heck that is...)
8. See the following screenshot for a good example where we see the `invoke` call in the stack trace triggering the complete interceptor chain (incl. the correct transaction) after `beantest.getVaccinationFacade.save()` has been called in `createVaccination`
transaction-present. You simply have to :heart: JavaEE stack traces for their brevity.
9. However, after the refactoring, no more JavaEE magic happening between our  `beantest.getVaccinationFacade.save()` call and the backing EJB. Sad.
transaction-missing
10. Please note in above screenshot that this *only* affects the EJB. The service call is still done through EJB runtime magic. Also note that while the `WeldClientProxy` stack frame is present in both, the `Proxy$_$$_WeldSubclass` frame is only present in the working example from `development`.
11. we get bonus points for weirdness as I checked the test behavior for events which already use the base mechanism through core: They work perfectly fine which shows that the problem is not caused by using base _per se_.

#### System Details

* Device:
* SORMAS version:
* Android version/Browser:
* Server URL:
* User Role:

JonasCir:
Found while working on #9708 

Instead of cascading the service injection, we just inject the `UserService` in the abstract parent class.

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/campaign/CampaignFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/campaign/CampaignFacadeEjb.java
@@ -50,7 +50,6 @@ import de.symeda.sormas.backend.common.AbstractCoreFacadeEjb;
 import de.symeda.sormas.backend.common.CriteriaBuilderHelper;
 import de.symeda.sormas.backend.user.UserFacadeEjb;
 import de.symeda.sormas.backend.user.UserRoleFacadeEjb.UserRoleFacadeEjbLocal;
-import de.symeda.sormas.backend.user.UserService;
 import de.symeda.sormas.backend.util.DtoHelper;
 import de.symeda.sormas.backend.util.Pseudonymizer;
 import de.symeda.sormas.backend.util.QueryHelper;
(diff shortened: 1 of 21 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11127

MateStrysewske (approved):

PR #11132, opened 2022-12-07
#10958 listener pattern

### LINKED ISSUE #10958: Avoid directories calling count method several times [3]
### Problem Description

When opening a directory the corresponding count method is called several times apperently without any good reason. This is a possible performance problem if count queries take long. Seen count calls:
- Persons: 2
- Cases: 2
- Samples: 3

The implementation pattern prevalent in the directories is that the active status button shows the number of total elements which triggers the first call before the DataProvider does the querying.

### Steps to Reproduce

1. Activate `PerformanceLoggingInterceptor`, watch log
2. Log in
3. Open Persons directoy

### Screenshots

<details>
<summary>Invocation of PersonFacade.count called twice (method call stack)</summary>

grafik

</details>

### Proposed Change

On every grid directory: Update the active status button after the grid gets updated with the same value that the `DataProvider` just found.

### Acceptance Criteria

- [x] When opening a directory, the linked count method should only be called once.

### Implementation Details

- Implement a `Listener` pattern where the `DataProvider` notifies the view that the size changed (with the count number) to update the active status button. 
- It's probably a good idea to wrap the currently used `DataProvider` implementations to implement the Listener pattern there.
- Consider weak references to avoid memory leaks.

### System Details

* SORMAS version: 1.77.0-SNAPSHOT, but certainly happening since a long time
* Server URL: local
* User Role: admin

### Additional Information

- The used `DataProvider` implementations neither cache the count value nor provide a Listener pattern for count update.
- `com.vaadin.data.provider.DataProviderWrapper` does other stuff with filtering, can probably not be used for what we need.

StefanKock:
Fixes #10958. Current status:

- [x] Listener pattern implemented on top of existing Event/Listener API
- [x] Made needed changes for `PersonsView`
- [ ] All other directories are missing

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseContactsView.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseContactsView.java
@@ -434,7 +434,7 @@ public class CaseContactsView extends AbstractCaseView {
 		CssStyles.removeStyles(activeStatusButton, CssStyles.BUTTON_FILTER_LIGHT);
 		if (activeStatusButton != null) {
 			activeStatusButton
-				.setCaption(statusButtons.get(activeStatusButton) + LayoutUtil.spanCss(CssStyles.BADGE, String.valueOf(grid.getItemCount())));
+				.setCaption(statusButtons.get(activeStatusButton) + LayoutUtil.spanCss(CssStyles.BADGE, String.valueOf(grid.getDataSize())));
 		}
 	}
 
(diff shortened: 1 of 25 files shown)


StefanKock:
I found a better approach to centralize fetching the `dataSize`, but I noticed that the apply filter button does not trigger a data refresh anymore. With debugging I found out that it is not related to how the status buttons are updated, but how I implemented the `DataProviderWrapper` is problematic (apparently breaking how events are handled).

I will reconsider my approach. The basic problem with the wrapper is that I cannot delegate and trigger events to the inner `DataProvider` because `AbstractDataProvider.fireEvent` is protected.

StefanKock:
Dismissing this approach because #11154 works better.

PR #11139, opened 2022-12-08
#4060 - Manual archiving of tasks

### LINKED ISSUE #4060: Manual archiving of tasks [2]
### Feature Description

We as a health department (Gesundheitsamt Friedrichshain-Kreuzberg) implemented a workflow that is completely based on Sormas.
We are using "case investigation"-tasks that are marked as done when finished.
After the investigation a supervisor double-checks the investigated case and then at the end of the shift an automated script send a letter/mail to every person that has been investigated on that day.

The view of tasks marked as "done" is becoming more cluttered by the day and we would like to remove this clutter by archiving the tasks that have been double-checked and received mail.

Removing the tasks does not work (interferes with "case investigation" status), deleting the tasks is not wanted as the tasks contain important information - most importantly, the name of the person who did the investigation.

### Proposed Change

Add a functionality to manually archive tasks - analogues to the "archive" button in case view.

### Implementation Details

- Batch processing for de-/archiving in the backend needed, not one-by-one on `Task` entity.

### Additional Information

Possible alternatives:
- A bulk edit archive function in the task management view - anologue to the bulk adit funcitonality in the case directory - would work as well.
- If another, clearer, workflow is possible using existing methods in Sormas we might also switch to that if it does not add extra overhead to the investigating persons - some of them did not work on PCs before so we wantto keep their tasks as simple as possible.

MateStrysewske:
Fixes #4060


sergiupacurariu:
1) Is there a logic based on which we added TASK_ARCHIVE right to certain user Roles? I haven't seen any logic in the ticket description about this
2) Let's say the user wants to see the archived tasks from the Task grid. The Task form is not consistent on each task opened from the archived section. Those which have been manually archived have the De-Archive button active while those tasks which inherited the archived status from the core entity have the Archive button active.
3) I understood the situation mentioned in the ticket for which we implemented this but once implemented it could be used in other situations as well. Shouldn't we have a reason for archiving? Do we want to keep the implementation light since is not a core entity?

MateStrysewske:
@sergiupacurariu 
1. The logic I used is basically the same user roles that can archive other entities, but also supervisors because those are usually the users who will hand out tasks to others, so they should also be in control of them. Not specified in the issue though.
2. That's interesting, we probably have different kinds of logic there, i.e. the archived flag is likely not set to true for these tasks. We might think about changing that, would be part of a different issue though.
3. I guess so, yes. I figured that since we didn't implement this for bulk archiving, we probably also don't need it for single archiving, and I don't believe we want to turn tasks into core entities that can be soft-deleted.

PR #11154, opened 2022-12-09
#10958 listener pattern V2

### LINKED ISSUE #10958: Avoid directories calling count method several times [3]

StefanKock:
Fixes #10958. Current status:

- [x] Listener pattern implemented on top of existing Event/Listener API for `FilteredGrid`
- [x] Made needed changes for 
  - `PersonsView`
  - `CasesView` 
  - `SamplesView`
  - `CampaignsView`, `CampaignDataView`, `CampaignStatisticsView` (no dataSize displayed)
  - `ContactsView`, `CaseContactsView`
  - `CaseVisitsView`, `ContactVisitsView` (no dataSize displayed)
  - `UsersView`
  - `UserRolesView` (no dataSize displayed)
  - `ImmunizationsView`, `TravelEntriesView` (no dataSize displayed)
  - `AggregateReportsView`, `ReportDataView` (no dataSize displayed)
  - `ShareRequestsView` (no dataSize displayed)
  - `EventsView`, `EventParticipantsView`
  - `EventSelectionGrid`, `EventGroupSelectionGrid` (no dataSize displayed), new `GridHeightResizer` (pattern before was creating triple count query)
  - `TasksView`
  - `ExternalMessagesView`
  - Infrastructure: `AreasView`, `CommunitiesView`, `ContinentsView`, `CountriesView`, `DistrictsView`, `FacilitiesView`, `PointsOfEntryView`, `RegionsView`, `SubcontinentsView`
- [x] All other directories are missing: Replace usage of existing `FilteredGrid.setDataProvider(DataProvider)` method and `FilteredGrid.getItemCount`
- [x] Remove `FilteredGrid.getItemCount` when all usages are replaced.


StefanKock (commented):
One thing to change.

To discuss:

1. Why are there so many changes like 
`searchField.clear(); criteria.nameLike(StringUtils.EMPTY);` ?
2. I don't know why `grid.addDataSizeChangeListener(e -> rowCount.update(grid.getDataSize()));` was neede to be done in this ticket, but I would not argue against it.

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/FilteredGrid.java
```diff
@@ -256,10 +256,10 @@ public void bulkActionHandler(Consumer<Set> callback, boolean allowAdminOverride
 	/**
 	 * @deprecated Use {@link #getDataSize()}, because this call triggers another count query for lazy {@link DataProvider}s.
 	 */
-	@Deprecated
-	public int getItemCount() {
-		return getDataProvider().size(new Query<>());
-	}
+//	@Deprecated
```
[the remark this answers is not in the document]
sergiupacurariu: The changes you mentioned to points 1 and 2 are meant to keep the existing behavior in the Infrastructure menu and fix some bugs also. There was an issue regarding bulkEdit mode which has not been noticed so far. If you look at how the bulkEdit is implemented you can see that in bulkEdit mode the search field should be disabled. Well, this works only exactly when you press the "EnterBulkEditMode" button. The field does not stay disabled once another dropdown is changed or you go in and out of the form (cache-related issue). Since I was working here and while during tests I felt they are annoying I also looked to fix them. The part that the searchField is cleared once you enter the bulkEdit mode comes from me. I felt it is the right behavior. If you don't do this the other dropdowns which remain available in the bulkEdit mode are almost useless. This can be removed but I would not recommend it. 
The method getItemCount() has been removed.

sormas-ui/src/main/java/de/symeda/sormas/ui/configuration/infrastructure/CountriesGrid.java
```diff
@@ -19,11 +19,11 @@
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
-import de.symeda.sormas.api.feature.FeatureType;
 import org.apache.commons.lang3.StringUtils;
 
 import de.symeda.sormas.api.EntityRelevanceStatus;
 import de.symeda.sormas.api.FacadeProvider;
+import de.symeda.sormas.api.feature.FeatureType;
```
StefanKock, line 26: FeatureType is apperently a superfluous import

sormas-ui/src/main/java/de/symeda/sormas/ui/task/TaskGrid.java
```diff
@@ -206,31 +206,11 @@ private void navigateToData(TaskIndexDto task) {
...
-					query.getLimit(),
-					query.getSortOrders()
-						.stream()
-						.map(sortOrder -> new SortProperty(sortOrder.getSorted(), sortOrder.getDirection() == SortDirection.ASCENDING))
-						.collect(Collectors.toList()))
-				.stream(),
-			query -> (int) FacadeProvider.getTaskFacade().count(query.getFilter().orElse(null)));
-		setDataProvider(dataProvider);
-		setSelectionMode(SelectionMode.NONE);
+		setLazyDataProvider(FacadeProvider.getTaskFacade()::getIndexList, FacadeProvider.getTaskFacade()::count);
```
StefanKock, line 209: Probably unused imports left, please check

StefanKock (commented):
I found 2 smaller flaws concerning code quality that should be checked. Apart from that there are two changes in `ContinentService` and `SubcontinentService` that are not directly related to the ticket (case insensitive search) but could stay.

If the code flaws are fixed this ticket could be merged from my perspective.

PR #11157, opened 2022-12-12
Bug/9360 port health tab missing

### LINKED ISSUE #9360: Admin, National Clinician - Port Health Tab missing - The tab is missing when admin creates a case with origin as point of entry
### Bug Description
The tab Port Health is missing when Admin, National Clinician creates a Case with origin as point of entry

### Steps to Reproduce
1. As user having the assigned role only Admin or National Clinician, create a Case with origin: Point of entry
2. Complete only the mandatory fields
3. Save the new Case
4. Look for the Port Health tab

### Actual Behavior
Admin cannot see the Port health tab in the Case edit form

### Expected Behavior
Admin, National Clinician has the right to View and Edit Port health and should be able to see the Port Health tab in the Case edit form
PORT_HEALTH_INFO_VIEW = Yes
PORT_HEALTH_INFO_EDIT = Yes

### System Details

* Device:
* SORMAS version: 1.72.1
* Android version/Browser:
* Server URL:
* User Role:

dinua:
Fixes #9360


sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseController.java
```diff
@@ -1848,4 +1847,19 @@ private static class JurisdictionValues {
 		protected String facilityDetails;
 	}
 
+	public boolean hasPointOfEntry(CaseDataDto caze) {
```
MateStrysewske, line 1850: Since the aim of this method is only to check whether the case has a point of entry or not, please change the backend method to return a boolean instead of the whole point of entry; don't query for the point of entry, but only check whether the field is not null.

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseController.java
```diff
@@ -1341,15 +1341,14 @@ public CommitDiscardWrapperComponent<MaternalHistoryForm> getMaternalHistoryComp
...
-		if (casePointOfEntry == null) {
+		if (!hasPointOfEntry(caze)) {
 			return null;
 		}
 
-		PointOfEntryDto pointOfEntry = FacadeProvider.getPointOfEntryFacade().getByUuid(casePointOfEntry.getUuid());
+		PointOfEntryDto pointOfEntry = FacadeProvider.getPointOfEntryFacade().getByCaseUuid(caseUuid);// getByUuid(casePointOfEntry.getUuid());
 		PortHealthInfoForm form = new PortHealthInfoForm(pointOfEntry, caze.getPointOfEntryDetails());
-		form.setValue(caze.getPortHealthInfo());
+		form.setValue(getPortHealthInfo(caze));
```
MateStrysewske, line 1351: Is it really necessary to do an additional query for the port health info here? Since that's an embedded entity, shouldn't it be present irrespective of the user rights?

MateStrysewske (changes requested):

MateStrysewske (changes requested):
Sorry, this was a mistake :)

MateStrysewske (changes requested):
The tab is properly showing now, but the port health tab is not properly pseudonymized, which is a data privacy problem that we need to tackle before we can merge this ticket.

- Make sure that the point of entry and point of entry details fields are pseudonymized in the same way as on the case details page
- Add the `@SensitiveData` annotation to ALL fields of `PortHealthInfoDto` that allow the user to freely enter text (probably all String fields) and verify that they are marked as Confidential for the user

MateStrysewske:
@sergiupacurariu Can you please do a quick review of this? I've done a bit of refactoring and some missing logic. :)

sergiupacurariu (approved):

PR #11159, opened 2022-12-12
Change/10251 clean non atomic endpoints

### LINKED ISSUE #10251: Improve exception handling and documentation for not atomic POST-based REST methods [5]
### Problem Description

Currently, `EntityDtoResource::savePushedDto` is a non-atomic bulk method which means that it will not abort the request in case one entity of the passed collection is invalid or an error occurs while processing it.

There are a couple of other downsides with the current approach:
- `ExceptionMapper` do not work as `savePushedDto` tries to handle exceptions on its own, however, in a less cleaner way than mapper could do it.
- HTTP status codes/responses are confusing. E.g., sending one invalid element will return `200 OK` and `[ERROR]` as response
- We do not support atomic bulk operations although this is a very valid use case (not handled as part of this issue)

### Proposed Change

Changes in **sormas-rest**:

- [x] Create a `ValidationRuntimeExceptionMapper`, decide which http status code to use
- [x] Instead of returning a list of PushResult, return a list of custom Response objects that have the field "status" (http status code) and an optional string field "body".
- [x] Instead of doing a custom error handling in `EntityDtoResource` re-use the existing exception mappers via javax.ws.rs.ext.Providers.getExceptionMapper. You can inject the interface using the `@Context` annotation.
- [x] This needs to be replaced in ~27 Resources each having one post method. Just change the return type, everything else is handled by savePushedDto generically.
- [x] CaseResource.postCasesDetailed is no longer needed
- [x] Extend the swagger documentation to reflect how this works.
- [x] Update related unit tests

Changes in **sormas-app**:

In theory these changes can be made in a second branch that is based on the sormas-rest branch and implemented by a second developer.

- [x] The matching method return types in the *Retro classes needed to be adjusted
- [x] The error handling in AdoDtoHelper.pushEntities needs to interpret the html codes instead of the enum values.

### Example of a response:
```
[
  {
    "code": 200
  },
  {
    "code": 409 // what used to be TOO_OLD
  },
  {
    "code": 400,
    "body": "Field validation XY error details"
  },
]
```

### QA
This issue changes to response of post requests that save a list of entities. Before it was a list of enum values (in JSON strings), now it's a list where each entry has an HTML based code and optionally a body that may contain a string with more details.
If the list has only one element the overall result code of the response will be the same as the code of the element.
If the list has multiple entries the result will be 207 (multi-status).
It would be good to test all post methods that allow saving a list of entities - maybe we already have automatic tests for this. 

### Acceptance Criteria
- Generated HTTP responses in case of an error are reasonably structured
- OpenAPI doc clearly describes how non atomic batch methods work and how to interpret the result

### Additional Information

Pick one response scheme. For this, go through the links below: There is no standard for how the response should look like. Facebook has a pure JSON based response scheme while Google relies on multipart responses.

=> we go with a facebook like approach based on JSON

* 207 Multi-Status: 	[RFC4918](https://www.iana.org/go/rfc4918)
* https://stackoverflow.com/questions/28596688/rest-api-bulk-create-or-update-in-single-request
* https://developers.facebook.com/docs/graph-api/batch-requests?_fb_noscript=1
* https://stackoverflow.com/questions/52111826/rest-api-append-bulk-to-api-to-create-bulk-create-on-same-resource
* https://www.mscharhag.com/api-design/bulk-and-batch-operations
* https://www.codementor.io/blog/batch-endpoints-6olbjay1hd
* https://adidas.gitbook.io/api-guidelines/rest-api-guidelines/execution/batch-operations
* https://docs.microsoft.com/en-us/power-apps/developer/data-platform/webapi/execute-batch-operations-using-web-api

### Follow-Up

- [x] @ChristopherRiedel Create an issue to update the survnet converter accordingly
* Discuss: Should we go through the resulting codes and update the overall result code accordingly? E.g. when all result codes in the list are 409 should the overall code then also be 409?
* Make the detailed error messages now returned by the server available to users of the Android app
* Provide atomic batch methods. Should we provide two methods or one with a prameter `atomic`.

JonasCir:
Fixes #10251 
@MartinWahnschaffe 
There are the following open questions, but I wanted to get things rolling:

1. The tickets talks about changes in `sormas-app`. I have never worked on the App before, so I think it makes sense if someone else picks this todo.
1. There is one refinement item about external visits as this changes the return type. Not certain what to do here
1. https://github.com/hzi-braunschweig/sormas-survnet-interface/issues/343 needs to be done after this is merged.
1. `LineListingResource.postFeatureConfigurations` still uses the old one as `saveFeatureConfigurations` does return void. Should we change this?


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11159

MartinWahnschaffe:
**DON'T MERGE THIS, AS LONG AS THE NEEDED CHANGES IN SORMAS-APP AREN'T DONE.**

sormas-rest/src/main/java/de/symeda/sormas/rest/exception/mappers/ConstraintViolationExceptionMapper.java
```diff
@@ -1,6 +1,6 @@
 /*
  * SORMAS® - Surveillance Outbreak Response Management & Analysis System
- * Copyright © 2016-2021 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
+ * Copyright © 2016-2022 Helmholtz-Zentrum für Infektionsforschung GmbH (HZI)
```
ChristopherRiedel, line 3: If the year is already adjusted it should be consistently 2023. This is only an example, please change completely for all touched classes.

ChristopherRiedel (changes requested):
Looks very good apart from the year numbers.

ChristopherRiedel (approved):

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11159

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11159

ChristopherRiedel (approved):
I have now tested a lot and could not find any problems.

MartinWahnschaffe:
Thanks. I need to discuss the issue with QA before merging this. Realised it may affect our automatic tests.

PR #11162, opened 2022-12-12
#6052 - Display side component in all person views

### LINKED ISSUE #6052: Add the info cards to the person view of each respective entities [3]
### Feature Description

Thus far, it is not easy to navigate to the person page of cases, contacts or event participants that also shows all connected entities. Users currently would have to navigate to the person directory and then search the person who's connected entity they have just been working on. This makes it quite pesky to for example check if a person had a previous infection.

### Proposed Change

To enhance usability and make it easier to see the persons related entities, add the related entities to the respective person tab of every entity person tab.
- [ ] Display the content of each related entity as seen on the person form also on each respective person tab and allow navigation through the edit icons
- [ ] In the case of the respective entity that the person is currently editing, hide the edit icon and instead highlight the entity that the user has opened. 

Person tab entity associations

### Additional Information

The cards should be extended as well when https://github.com/hzi-braunschweig/SORMAS-Project/issues/6054 gets implemented.

MateStrysewske:
Fixes #6052


MateStrysewske:
@BarnaBartha A review would still be appreciated, but isn't as urgent because we're not supposed to merge this issue for the upcoming release.

BarnaBartha (approved):

PR #11171, opened 2022-12-13
#8465 - Display soft-deleted entities to users that have the right to…

### LINKED ISSUE #8465: Display soft-deleted entities to users that have the right to delete [8]
### Situation Description
It must be possible to review which entities were manually deleted and to revert the deletion.

This belongs to epic #7736 which describes the automatic deletion and re-works how (manual) soft-deletion works. 

### Change Proposal
The following goes for the directory views of cases, contacts, events, event participants, samples, travel entries and immunizations.

- [x] Besides being able to view active and archived entities it should also be able to view deleted entities
- [x] "all ..." should be renamed to "active and archived ..."
- [x] Add another column to display the reason for deletion (combine enum + free text)
- [x] But only when the user has the corresponding ..._DELETE user right
- [x] For event participants this filter is currently missing completely and needs to be added

The user should be able to open the entity and undo it's deletion (see #8990).

### Additional Information
Out of scope: Soft-deleted entities will sometimes be removed from related entities (e.g. a contact from it's source case). When reverting the deletion, this information will not be re-created. If needed the user can manually re-assign the entity.

sergiupacurariu:
… delete

Fixes #8465


BarnaBartha (changes requested):
The changes look good
I have found two problems or missing implementation:
- campaigns - no way to see the the deleted ones - this we need for sure
- pathogen tests - also no way to see the deleted ones (this one needs to be clarified how it is wanted - but form the ticket description we definitely need it as well)

sergiupacurariu:
- campains related requests were fixed
 - pathogen tests - new ticket was created #11223

BarnaBartha (approved):

PR #11182, opened 2022-12-15
Change 11013 two step get index list

### LINKED ISSUE #11013: Two step `getIndexList` for CoreAdos [8]
### Problem Description

`getIndexList` uses `sort distinct` on all attributes that are fetched for directory DTOs, which takes more time with more rows to process. The problem gets bigger with more left joins that multiply rows.

### Proposed Change

Improve performance of `getIndexList` methods (by avoiding extensive sorting on all data columns) that fetch data for the following directories of CoreAdos that tend to grow heavily over time:
1. `Case`
2. `Contact`
3. `Event`
4. `EventParticipant`
5. `TravelEntry`
6. `Immunization`

Apply the developed pattern for the following CoreAdos to first fetch ids, and then fetch data afterwards.

### Acceptance Criteria

- [ ] Performance analysis before/after is documented (with info about given dataset size).
- [ ] At least one unit test exists at least for each `getIndexList` method.

### Implementation Details

1. Apply this pattern (already used in `PersonFacadeEjb`):
```java
getIndexList {

  // 1. Fetch ids of entries in the window given by limit & offset
  getIndexListIds {
    sortBy
  }

  // 2. Fetch data for the fetched ids
  executeBatched {
    sortBy
  }

  // 3. Post-processing: calculated values, references, collections
  // 4. Pseudonymization
  // 5. Log execution time (DEBUG)
}
```

2. Extract generic pattern to fetch Ids, sortBy or something else that is repeated code (if possible).

### Additional Information

- Follow-up for #8938.
- Other CoreAdos `Campaign` and `DirectoryImmunization` are not considered to be needed to rework.

syntakker:
Fixes #11013

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -54,6 +54,7 @@ import javax.ejb.TransactionAttribute;
 import javax.ejb.TransactionAttributeType;
 import javax.enterprise.concurrent.ManagedScheduledExecutorService;
 import javax.inject.Inject;
+import javax.persistence.Tuple;
 import javax.persistence.TypedQuery;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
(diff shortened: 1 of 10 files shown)


syntakker:
An abstract implementation of the two-step-getIndexList is possible, but was not feasible in time for the feature freeze.

The model for this abstraction could be `CaseListCriteriaBuilder`, where the specifications relevant for an individual entity are already encapsuled in a way to be general enough for an abstract two step process for getIndexList (`type`, `selectionProvider`, `orderExpressionProvider`). Additionaly a method to create the respective `where` filter has to be provided.

In the current implementation, however, only the query is built, while its execution is performed outside the method. An abstract implementation of a two step getIndexList method should also include the two executions of the constructed query (for ids first, and then batched for Dtos).

```java
	private <T> CriteriaQuery<T> buildIndexCriteria(
		Class<T> type,
		BiFunction<Root<Case>, CaseQueryContext, List<Selection<?>>> selectionProvider,
		CaseCriteria caseCriteria,
		OrderExpressionProvider orderExpressionProvider,
		List<SortProperty> sortProperties,
		boolean detailed)
```

BarnaBartha (approved):

Issue #3733, opened 2020-12-09
Rename serverlibs to domainlibs module, change packaging [1]


MartinWahnschaffe:
### Problem Description

Dependabot requires a pom.xml to evaluate your Java dependencies. It had expected to find one at the path: /sormas-base/dependencies/serverlibs.pom/pom.xml.

![grafik](https://user-images.githubusercontent.com/23217632/101635237-f6959380-3a29-11eb-8d53-58f4ee35dcf3.png)

Additionally, also IntelliJ does not like a POM that it not named `pom.xml`

### Proposed Solution

- We probably have to change our serverlibs.pom to `sormas-domainlibs/pom.xml` (pom.xml file as a module beside the others), rename the module to `sormas-domainlibs`.
- Move also the assemblyLibJar.xml and assemblyZip.xml into that subdirectory.
- Refactor it also in Ant scripts to still be able to deploy on local DEV system (rename target to deploy-domainlibs).
- Check that nothing breaks in Shell scripts.

### Possible Alternatives

- Keep naming `serverlibs` to not have too much refactoring with all the packaging and setup.

StefanKock:
One remark here: If Dependabot only wants to find dependencies, this change is not needed since all dependencies are listed `sormas-base/pom.xml`. Considering this I do not see a benefit doing this.

JonasCir:
@tbroyer FYI regarding `serverlibs.pom`

tbroyer:
Thanks for the ping @JonasCir.

IntelliJ IDEA doesn't quite like the `serverlibs.pom` either. In #3728 I also renamed it to `pom.xml` for that reason; and I see no reason for not naming it `pom.xml`. Might be even better to maybe move it around too, say, to a `sormas-serverlibs/pom.xml` at the root of the project next to the other `sormas-base`, `sormas-api`, etc. modules.

JonasCir:
Also, fixing this will improve maven package caching in the CI see https://github.com/hzi-braunschweig/SORMAS-Project/pull/3732

Issue #10074, opened 2022-08-08
Replace CDI unit testing framework


MartinWahnschaffe:
### Problem Description
Unit testing of the SORMAS backend is currently facing multiple problems:

* Bean-test no longer maintained (e.g. no support for jakarta-ee, using very old versions of weld + deltaspike) -> we would have to fork and maintain on our own
* Combination of bean-test and mocking is not working very good. E.g. it's tricky to change the logged-in user during test run and hard to mock configuration properties, feature configurations
* Transaction logic is partly not testable (e.g. annotation TransactionAttribute)
* Validation annotations are not testable (e.g. #7611)
* Parts of logic data stored in JSON format can't be tested (e.g. properties of feature configurations). -> This will not be addressed here. Maybe an update of H2 can also fix this problem.
* Is overall a bit slow, because we are setting up the complete backend CDI container for each test class

#### Possible Test Approaches

Generally speaking there are three approaches to testing backend functionality:

1. Basic unit tests + mocking
* Mocking all the dependencies of tested services and facades would be a lot of work and make the tests a lot less helpful
2. Integration tests (e.g. based on Arquillian)
* Every test run requires an archive to be built and deployed which makes the testing slower. Too slow if you need to run thousands of test cases.
* Creating the deployments, especially resolving the required artifacts and classes, is not trivial.
* Setup tends to be complicated.
* Still helpful **for functionality that can't be tested otherwise** (e.g. transaction based stuff or database specific stuff)
3. CDI-based testing (provides bean container that automatically injects dependencies)
* This is the approach we have followed so far and **should go on using**.
* It's currently facing the problems mentioned above, though

### Proposed Change

- [ ] Find a replacement for the novatec bean-test framework.

* cdi-test:  Small open source project, but maintained, using a current version of weld and is not relying on deltaspike (-> faster)
  It's supposed to make mocking of parts of the system a lot easier
* cdi-unit: http://cdi-unit.github.io/cdi-unit/
* weld-testing: https://dzone.com/articles/weld-junit-easy-testing-of-cdi-beans

- [ ] Check whether both can be used in parallel to simplify migration
- [ ] Make transaction logic testable, e.g. provide an interceptor for TransactionAttribute annotation
- [ ] Make validation testable
- [ ] Create implementation/migration issue(s)

### Additional Information

Alternative approach:Use Spring instead of JavaEE for the backend. This would probably be a very big effort. If we are considering this it would have to be evaluated. 

[...]

PR #11184, opened 2022-12-15
[#3733] rename serverlibs pom to make intellij and dependabot happier

JonasCir:
Fixes https://github.com/hzi-braunschweig/SORMAS-Project/issues/3733

All my assigned task are currently blocked, so I decided to go through my old backlog, also b/c my commitment will be reduced next year.

Title says it all.


StefanKock (changes requested):
Renaming `dependencies` to `sormas-serverlibs` is not a good way, because other files in that folder are not related. 
Proposal: Move serverlibs.pom to `../sormas-serverlibs/pom.xml` (aside all other modules, not nested) and with it `assemblyZip.xml` and `assemblyLibJar.xml`.
The directory `dependencies` will be removed later when #10074 and #11230 are done.

Other problem: When we refactor a maven module, our hotfix process with heavily crush. We should schedule this really good to not do more harm than needed.

StefanKock:
I'll provide a new PR.

PR #11188, opened 2022-12-15
Bugfix 6053 test findings

### LINKED ISSUE #6053: Add creation/link buttons to the person entity links [2]

syntakker:
Fixes #6053

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/i18n/Strings.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/i18n/Strings.java
@@ -1183,6 +1183,7 @@ public interface Strings {
 	String messagePersonAddedAsEventParticipant = "messagePersonAddedAsEventParticipant";
 	String messagePersonAlreadyCaseInEvent = "messagePersonAlreadyCaseInEvent";
 	String messagePersonAlreadyEventParticipant = "messagePersonAlreadyEventParticipant";
+	String messageThisPersonAlreadyEventParticipant = "messageThisPersonAlreadyEventParticipant";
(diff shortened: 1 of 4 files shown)


sergiupacurariu:
The part regarding case creation from person page still works like that: the "Pick or create case" is triggered right after "Create new case" is initiated. It should be triggered only after the "Create new case" form is saved.

sergiupacurariu (approved):

PR #11189, opened 2022-12-15
#10034-User right Sormas UI and Rest should be included automatically when selecting a right

### LINKED ISSUE #10034: User Role - User right Sormas UI and Rest should be included automatically when selecting a right [0.5]
### Problem Description
 A new Role is created with status Enable and some rights for Case management as example. Admin creates a new User with the new Role. At this point the new User created is not able to login because it has not selected the right dor Sormas UI

### Proposed Change
There can be 3 solutions:
1. To automatically select the rights Sormas UI and Rest when at least one of the other rights is selected
2. To not be able to save the Role without having Sormas UI or Rest selected
3. To have by default selected this right Sormas UI or Rest when a new Role is created

### Implementation Details

When a user is saved without one of those rights selected there should be a warning message displayed

valentinmikleuvg:
when you create a new User Role, the Access Sormas UI and Access Sormas Rest rights are selected by default, and you cannot save an User Role without one of those rights

Fixes #10034

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserRoleController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserRoleController.java
@@ -15,7 +15,9 @@
 
 package de.symeda.sormas.ui.user;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


leventegal-she (changes requested):
It is not implemented as discussed in the planning
See implementation details for the requirement.

> When a user is saved without one of those rights selected there should be a warning message displayed

This means no default selection and let the user save the role, just display a custom warning message before save to notify the user that the roll will not be usable for login.

sormas-api/src/main/resources/strings.properties
```diff
@@ -1284,6 +1284,7 @@ messageSormasToSormasSimilarPersonFound = There is at least one similar person i
 messageDeleteWithPendingShareRequest = There is a pending share request. With a deletion, the share request will be revoked and deleted.
 messageCannotMergeMoreThanTwoPersons = You need to select two persons for merge!
 messageAutomaticDeletionStarted = Automatic deletion has been started and will be executed in the background. Please note that, depending on the amount of data that is deleted, this process can take some time.
+messageUserUnusableForLogin = The user will be unusable for login
```
leventegal-she, line 1287: please rename to `messageUserRoleUnusableForLogin` because it refers to user role
and update the message to *Users with only this role will not be able to login because the role does not have "Access Sormas UI" nor "Access Sormas REST" right*

leventegal-she (changes requested):

sormas-api/src/main/resources/strings.properties
valentinmikleuvg: ok, I did the requested changes

leventegal-she (approved):

PR #11228, opened 2022-12-20
#10033-User Role-When a role is being enable to have at least one right selected

### LINKED ISSUE #10033: User Role - When a role is being enable to have at least one right selected
### Problem Description
When creating a new Role and making it enable, it can be saved without having a right selected. Because of this behaviour we can assign the new role at the users but having no right on it

### Proposed Change
When a Role is changing the status to enable, to request in a warning message that at least one right should be selected

valentinmikleuvg:
after the user role was created, a warning message saying that the user role was created without rights is displayed
Fixes #10033


carina29 (approved):

carina29:
@valentinmikleuvg Since this change should go in 1.79 release, this PR can be merged after 1.78 was released.

PR #11233, opened 2022-12-21
#11017 - Persons with incomplete home addresses from different jurisd…

### LINKED ISSUE #11017: Persons with incomplete home addresses from different jurisdictions can not be merged
### Problem Description

When using the person merge #5606 to try and merge persons that are from different jurisdictions (in their home address section) a warning appears if one of the persons has an incomplete home address.

A separate ticket has been created as in #5606 it was specified that information from the fields that are not filled in for the leading person should be filled with information from the discarded person. But it was not specified what should be done when the information is conflicting.

The scenario proposed considers 2 persons that are duplicates, one entry could be from a while ago when the person had an incomplete home address from jurisdiction 1, and the second entry for that person in their latest one, and has a complete home address from jurisdiction 2.

Steps:
1. Create a person that has the following home address:
image

2. Create a second person that would be considered a duplicate of the one from step1, but use the following home address:
image

3. Try to merge the 2 persons and have the person from step 1 as the leading person;
-> the merge is not possible, as the information from the community field from the discarded person can not be migrated into the leading person.
image

### Proposed Change

Two options for the proposed change:

1. Would it be desired for the whole home address to be updated for the leading person when the home address information is from different jurisdictions (different districts, and/or different communities)?

2. Would it be desired for the whole home address to be transferred into the leading person in the form of a 'New entry' in the 'Addresses' section of the leading person?

sergiupacurariu:
…ictions can not be merged

Fixes #11017

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
@@ -31,6 +31,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -120,6 +121,7 @@ import de.symeda.sormas.api.utils.DataHelper.Pair;
 import de.symeda.sormas.api.utils.DateHelper;
 import de.symeda.sormas.api.utils.SortProperty;
 import de.symeda.sormas.api.utils.ValidationRuntimeException;
+import de.symeda.sormas.api.uuid.HasUuid;
 import de.symeda.sormas.api.vaccination.VaccinationDto;
 import de.symeda.sormas.backend.FacadeHelper;
 import de.symeda.sormas.backend.caze.Case;
@@ -1869,26 +1871,34 @@ public class PersonFacadeEjb extends AbstractBaseEjb<Person, PersonDto, PersonIn
 			LocationDto otherAddress = otherPersonDto.getAddress();
 			otherAddress.setAddressType(PersonAddressType.OTHER_ADDRESS);
 			otherAddress.setAddressTypeDetails(I18nProperties.getString(Strings.messagePersonMergedAddressDescription));
-			leadPersonDto.addAddress(otherPersonDto.getAddress());
+			leadPersonDto.addAddress(otherAddress);
 		} else {
 			DtoHelper.copyDtoValues(leadPersonDto.getAddress(), otherPersonDto.getAddress(), false);
 		}
 	}
 
 	private boolean differentLocation(LocationDto firstAddress, LocationDto secondAddress) {
-		if (firstAddress.getCountry() != null && !firstAddress.getCountry().getUuid().equals(secondAddress.getCountry().getUuid())) {
+
+		java.util.function.Predicate<Function<LocationDto, HasUuid>> differentInfrastructurePredicate = (Function<LocationDto, HasUuid> function) -> {
+			return function.apply(firstAddress) != null
+				&& function.apply(secondAddress) != null
+				&& !DataHelper.equal(function.apply(firstAddress).getUuid(), function.apply(secondAddress).getUuid());
+
+		};
+
+		if (Boolean.TRUE.equals(differentInfrastructurePredicate.test(LocationDto::getCountry))) {
 			return true;
 		}
 
-		if (firstAddress.getRegion() != null && !firstAddress.getRegion().getUuid().equals(secondAddress.getRegion().getUuid())) {
+		if (Boolean.TRUE.equals(differentInfrastructurePredicate.test(LocationDto::getRegion))) {
 			return true;
 		}
 
-		if (firstAddress.getDistrict() != null && !firstAddress.getDistrict().getUuid().equals(secondAddress.getDistrict().getUuid())) {
+		if (Boolean.TRUE.equals(differentInfrastructurePredicate.test(LocationDto::getDistrict))) {
 			return true;
 		}
 
-		if (firstAddress.getCommunity() != null && !firstAddress.getCommunity().getUuid().equals(secondAddress.getCommunity().getUuid())) {
+		if (Boolean.TRUE.equals(differentInfrastructurePredicate.test(LocationDto::getCommunity))) {
 			return true;
 		}
 
@@ -1900,21 +1910,37 @@ public class PersonFacadeEjb extends AbstractBaseEjb<Person, PersonDto, PersonIn
 			}
 		}
 
-		if (firstAddress.getCity() != null && !firstAddress.getCity().equals(secondAddress.getCity())) {
+		boolean oneMatch = Boolean.FALSE;
+		boolean notNullValues = Boolean.FALSE;
+		boolean different = Boolean.FALSE;
+		List<Pair<String, String>> addressValues = new ArrayList<>();
+		addressValues.add(new Pair<>(firstAddress.getCity(), secondAddress.getCity()));
+		addressValues.add(new Pair<>(firstAddress.getPostalCode(), secondAddress.getPostalCode()));
+		addressValues.add(new Pair<>(firstAddress.getStreet(), secondAddress.getStreet()));
+		addressValues.add(new Pair<>(firstAddress.getHouseNumber(), secondAddress.getHouseNumber()));
+
+		for (Pair<String, String> addressTypePair : addressValues) {
+			if (addressTypePair.getElement0() != null && addressTypePair.getElement1() != null) {
+				if (!DataHelper.equal(addressTypePair.getElement0(), addressTypePair.getElement1())) {
+					different = Boolean.TRUE;
+				} else {
+					oneMatch = Boolean.TRUE;
+				}
+			} else {
+				if (addressTypePair.getElement1() != null) {
+					notNullValues = Boolean.TRUE;
+				}
+			}
+		}
+
+		if (different) {
 			return true;
 		}
 
-		if (firstAddress.getPostalCode() != null && !firstAddress.getPostalCode().equals(secondAddress.getPostalCode())) {
-			return true;
+		if (notNullValues) {
+			return !oneMatch;
 		}
 
-		if (firstAddress.getStreet() != null && !firstAddress.getStreet().equals(secondAddress.getStreet())) {
-			return true;
-		}
-
-		if (firstAddress.getHouseNumber() != null && !firstAddress.getHouseNumber().equals(secondAddress.getHouseNumber())) {
-			return true;
-		}
 		return false;
 	}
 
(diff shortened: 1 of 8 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -1793,7 +1793,7 @@ public void mergePerson(PersonDto leadPerson, PersonDto otherPerson) {
 
 	@Override
 	@RightsAllowed(UserRight._PERSON_EDIT)
-	public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean mergeProperties) {
+	public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean mergeProperties) throws CloneNotSupportedException {
```
BarnaBartha, line 1796: who throws this exception? it does not make sense

sormas-backend/src/test/java/de/symeda/sormas/backend/person/PersonFacadeEjbTest.java
```diff
@@ -926,8 +928,22 @@ public void testMergePersonsAndRemoveDuplication() {
...
+		leadPersonAddress.setRegion(new RegionReferenceDto(rdcf1.region.getUuid()));
+		leadPersonAddress.setDistrict(new DistrictReferenceDto(rdcf1.district.getUuid()));
+
+		LocationDto otherPersonAddress = otherPerson.getAddress();
+		otherPersonAddress.setRegion(new RegionReferenceDto(rdcf2.region.getUuid()));
+		otherPersonAddress.setDistrict(new DistrictReferenceDto(rdcf2.district.getUuid()));
+		otherPersonAddress.setCommunity(new CommunityReferenceDto(rdcf2.community.getUuid()));
+
 		leadPerson.setAddresses(Collections.singletonList(LocationDto.build()));
-		otherPerson.setAddresses(Collections.singletonList(LocationDto.build()));
```
BarnaBartha, line 930: the test was modified in such a way that the old asserts cover it - but i would extend the asserts as now you have more information in the addresses

BarnaBartha (changes requested):

sormas-backend/src/test/java/de/symeda/sormas/backend/person/PersonFacadeEjbTest.java
sergiupacurariu: asserts extended and added more tests for different situations

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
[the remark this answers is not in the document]
sergiupacurariu: updated both merge methods

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -1815,8 +1815,16 @@ public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean m
 					contactDetailDto.setPrimaryContact(false);
 				}
 			}
+
+			LocationDto leadPersonAddress = (LocationDto) leadPersonDto.getAddress().clone();
+			LocationDto otherPersonAddress = otherPersonDto.getAddress();
 			DtoHelper.copyDtoValues(leadPersonDto, otherPersonDto, false);
 
+			if (differentLocation(leadPersonAddress, otherPersonAddress)) {
```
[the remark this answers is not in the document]
sergiupacurariu: copyDtoValue method was overloaded

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -1859,6 +1864,60 @@ public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean m
 		service.ensurePersisted(leadPerson);
 	}
 
+	private void processPersonAddressMerge(PersonDto leadPersonDto, PersonDto otherPersonDto) {
+		if (differentLocation(leadPersonDto.getAddress(), otherPersonDto.getAddress())) {
+			LocationDto otherAddress = otherPersonDto.getAddress();
+			otherAddress.setAddressType(PersonAddressType.OTHER_ADDRESS);
+			otherAddress.setAddressTypeDetails(I18nProperties.getString(Strings.messagePersonMergedAddressDescription));
+			leadPersonDto.addAddress(otherPersonDto.getAddress());
```
MateStrysewske, line 1872: I know that this works because of call-by-reference, but it still looks odd to not use `otherAddress` directly here, but access it through otherPersonDto instead

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -1859,6 +1864,60 @@ public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean m
...
+		} else {
+			DtoHelper.copyDtoValues(leadPersonDto.getAddress(), otherPersonDto.getAddress(), false);
+		}
+	}
+
+	private boolean differentLocation(LocationDto firstAddress, LocationDto secondAddress) {
```
MateStrysewske, line 1878: Won't this method inevitably throw a null pointer if the second address does not have a country, region, district, or community because we call `getUuid()` on it without a null check?

Also, won't this method return true (i.e. different addresses) if e.g. the first address has a city, but the second has none even though that shouldn't lead to the addresses being considered different?

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
sergiupacurariu: rewrote the logic for determining the duplicate address

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -1859,6 +1866,84 @@ public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean m
...
+		}
+
+		if (Boolean.TRUE.equals(differentInfrastructurePredicate.test(LocationDto::getCommunity))) {
+			return true;
+		}
+
+		if (firstAddress.getFacility() != null) {
+			FacilityDto firstAddressFacilityDto = facilityFacade.getByUuid(firstAddress.getFacility().getUuid());
+			if (secondAddress.getCommunity() != null
+				&& !firstAddressFacilityDto.getCommunity().getUuid().equals(secondAddress.getCommunity().getUuid())) {
```
MateStrysewske, line 1908: Communities are optional for facilities, so this can still result in a NPE (because you're not checking whether the facility of the first address has one)

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -1859,6 +1866,84 @@ public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean m
...
+		boolean notNullValues = Boolean.FALSE;
+		boolean different = Boolean.FALSE;
+		List<Pair<String, String>> addressValues = new ArrayList<>();
+		addressValues.add(new Pair<>(firstAddress.getCity(), secondAddress.getCity()));
+		addressValues.add(new Pair<>(firstAddress.getPostalCode(), secondAddress.getPostalCode()));
+		addressValues.add(new Pair<>(firstAddress.getStreet(), secondAddress.getStreet()));
+		addressValues.add(new Pair<>(firstAddress.getHouseNumber(), secondAddress.getHouseNumber()));
+
+		for (Pair<String, String> addressTypePair : addressValues) {
+			if (addressTypePair.getElement0() != null && addressTypePair.getElement1() != null) {
```
MateStrysewske, line 1922: Use `isNotBlank` instead of a null check so that we consider null and a String of whitespaces to be equal

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -1859,6 +1866,84 @@ public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean m
...
+				if (!DataHelper.equal(addressTypePair.getElement0(), addressTypePair.getElement1())) {
+					different = Boolean.TRUE;
+				} else {
+					oneMatch = Boolean.TRUE;
+				}
+			} else {
+				if (addressTypePair.getElement1() != null) {
```
MateStrysewske, line 1929: Write this as `else if` (since we only have one condition in this block anyway)

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -1859,6 +1866,84 @@ public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean m
...
+			LocationDto otherAddress = otherPersonDto.getAddress();
+			otherAddress.setAddressType(PersonAddressType.OTHER_ADDRESS);
+			otherAddress.setAddressTypeDetails(I18nProperties.getString(Strings.messagePersonMergedAddressDescription));
+			leadPersonDto.addAddress(otherAddress);
+		} else {
+			DtoHelper.copyDtoValues(leadPersonDto.getAddress(), otherPersonDto.getAddress(), false);
+		}
+	}
+
+	private boolean differentLocation(LocationDto firstAddress, LocationDto secondAddress) {
```
MateStrysewske, line 1880:
```suggestion
	private boolean areDifferentLocations(LocationDto firstAddress, LocationDto secondAddress) {
```

Also, I'd move this method to LocationFacadeEjb

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
```diff
@@ -1859,6 +1866,84 @@ public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean m
...
+		if (firstAddress.getFacility() != null) {
+			FacilityDto firstAddressFacilityDto = facilityFacade.getByUuid(firstAddress.getFacility().getUuid());
+			if (secondAddress.getCommunity() != null
+				&& !firstAddressFacilityDto.getCommunity().getUuid().equals(secondAddress.getCommunity().getUuid())) {
+				return true;
+			}
+		}
+
+		boolean oneMatch = Boolean.FALSE;
+		boolean notNullValues = Boolean.FALSE;
```
MateStrysewske, line 1914: The name is kind of confusing imo, maybe something like `secondLocationHasAddressValues` makes it clearer what this variable is supposed to contain?

MateStrysewske (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
sergiupacurariu: changed

MateStrysewske (approved):

PR #11255, opened 2022-12-23
#9742 Display errors as list

### LINKED ISSUE #9742: Bad fromatting of error message for missing information aroundVesiculopustular rash
### Bug Description

When setting the Vesiculopustular rash = yes, not setting any value for the other subsequent mandatory fields, and then trying to save the 'Create new visit' form in the 'Cases' directory and in the 'Contacts' directory is displayed differently than the same message from the 'Symptoms' tab of cases.

Message in the 'Symptoms' tab of cases:
image

Message in the 'Create new visit' form of cases and contacts:
image

Expected behavior: the warning message triggered in the 'Create new visit' form should resemble the one from the 'Symptoms' tab.
### Steps to Reproduce

1. For a case, creaet a new visit
2. Vesiculopustular rash = yes
3. Try to save without filling any value of the now displayed fields

### System Details

* Device:
* SORMAS version: 1.74.0-SNAPSHOT
* Android version/Browser:
* Server URL:
* User Role:

andrei7bal:
Fixes #9742

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/CommitDiscardWrapperComponent.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/CommitDiscardWrapperComponent.java
@@ -619,10 +619,12 @@ public class CommitDiscardWrapperComponent<C extends Component> extends Vertical
 						}
 						htmlMsg.append("</ul>");
 					} else if (firstCause != null) {
+						htmlMsg.append("<ul>");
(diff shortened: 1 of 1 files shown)


andrei7bal:
I was not able to test with Vesiculopustular in Case and Contact Visits. This is an improvement anyway as it will display errors as list even though it's just one element. As I saw errors from Visits form are thrown as multiple causes, so maybe this will fix it for Vesiculopustular too.

PR #11281, opened 2023-01-05
 #11076 - delete icon should appear if the case/event was shared with…

### LINKED ISSUE #11076: [Reporting Tool] Button for deletion should not be displayed if the Case was not sent to the Reporting Tool yet
### Problem Description
1) The Reporting Tool card of Cases that were not sent to the Reporting Tool has the option for deletion (the delete button). If the button is clicked, the user is able to delete the Case from the Reporting Tool. However, this shouldn't be possible if the Case was not sent to the Reporting Tool in the first place.

image

2) Added by @carina29: The same functionality ( displaying/hiding delete button in case of Events was also totally missing so this functionality will be also added with this change). => When an Event was already shared the Delete icon will always appear, when not the Delete icon will not appear.

### Proposed Change
1) The delete button should not be displayed for Cases that were not sent to the Reporting Tool.
2) The same behaviour described above should be in place for Events too

Added

carina29:
… Reporting Tool

Fixes #11076

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/externalsurveillanceservice/ExternalSurveillanceShareComponent.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/externalsurveillanceservice/ExternalSurveillanceShareComponent.java
@@ -117,17 +117,11 @@ public class ExternalSurveillanceShareComponent extends VerticalLayout {
 	}
 
 	private boolean isVisibleDeleteButtonForCase(String caseUuid) {
-		if (caseUuid != null) {
 			return FacadeProvider.getExternalShareInfoFacade().isSharedCase(caseUuid);
-		}
-		return true;
 	}
 
 	private boolean isVisibleDeleteButtonForEvent(String eventUuid) {
-		if (eventUuid != null) {
 			return FacadeProvider.getExternalShareInfoFacade().isSharedEvent(eventUuid);
-		}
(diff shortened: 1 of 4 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/externalsurveillanceservice/ExternalSurveillanceShareComponent.java
```diff
@@ -89,24 +93,39 @@ private HorizontalLayout createHeader(
...
+		headerLayout.setWidth(100, Unit.PERCENTAGE);
+	}
+
+	private boolean isVisibleDeleteButtonForCase(String caseUuid) {
 		if (caseUuid != null) {
-			CaseDataDto caseDataDto = FacadeProvider.getCaseFacade().getCaseDataByUuid(caseUuid);
-			boolean isSharedCase = FacadeProvider.getExternalShareInfoFacade().isSharedCase(caseUuid);
+			return FacadeProvider.getExternalShareInfoFacade().isSharedCase(caseUuid);
+		}
+		return true;
```
sergiupacurariu, line 123: caseUuid is checked for null twice this return will not be reached in the current implementation

sergiupacurariu (changes requested):

sergiupacurariu (approved):

PR #11354, opened 2023-01-17
fixed #10852

### LINKED ISSUE #10852: [S2S] Sharing a Contact or Case individually should not be possible if the Sample is linked to both entities
### Bug Description
The Contact and the Case share cannot be accepted by the receiving instance if they have a Sample which is linked to both entities (was used for the resulting Case).
The issue is not reproducible if a Case is created for the Contact by 'Create a Case for this Contact Person' without a Sample.

### Steps to Reproduce

1. On Sormas A, Create a Contact (and select a Source Case);
2. Add a Sample with a positive Pathogen Test Result and create a resulting Case;
3. Share the Case with 'Hand over the ownership';
4. On Sormas B, try to accept the Case 
-> the Case cannot be accepted because the database does not contain the UUID of the Contact;
image
5. Go back to Sormas A also share the Contact with 'Hand over the ownership' (in order to be able to share the Contact, you need to share the Source Case first: 'Exclude personal data');
6. Go back on Sormas B and accept the Source Case;
7. Then, try to accept the Contact;
-> the Contact cannot be accepted because the database does not contain the UUID of the Case;
image
8. Try to accept the Case again;
-> the Case cannot be accepted because the database does not contain the UUID of the Contact;
image

### Actual Behavior
The user is in a loop and is unable to accept the share for the Contact and for the Case;

### Expected Behavior
Don't be able to share a Contact or Case individually if the Sample is linked to both entities

### System Details

* Device: Windows 10
* SORMAS version: 1.76.2
* Server URL: release-sormas-x.sormas.netzlink.com and release-de.sormas.netzlink.com
* User Role: Admin + National User

ChristopherRiedel:
Fixes #10852


leventegal-she (changes requested):
the validations should be done in the `validateEntitiesBeforeShareInner` methods

sormas-backend/src/main/java/de/symeda/sormas/backend/sormastosormas/entities/contact/SormasToSormasContactFacadeEjb.java
```diff
@@ -175,6 +179,17 @@ protected void validateEntitiesBeforeShareInner(
 				}
 			}
 		}
+
+		if (isWithSamples) {
+			Set<Sample> samples = contact.getSamples();
+			if (samples != null && samples.stream().anyMatch(s -> s.getAssociatedContact() != null)) {
```
leventegal-she, line 185: I think it should be
 anyMatch(s -> s.**getAssociatedCase**() != null)

leventegal-she (changes requested):
Please create unit tests

PR #11373, opened 2023-01-19
#11319 - Entities read-only can be soft deleted from API

### LINKED ISSUE #11319: Entities read-only can be soft deleted from API [1]
### Problem Description
The case described refers to the environments that have S2S.

As NatUser+Admin having a Case, Contact, that is read-only(to view) can make an API request to delete that entity.
In the UI the entity is not possible to be deleted because the delete button is not enable.

The entity is in read-only mode because:

- was handed over to another health department
- was received with only to view from another health department

### Proposed Change
- Allow users who have the respective delete user right to delete read-only entities ONLY if they are in their jurisdiction

### Additional Information
The changed appeared while testing this ticket https://github.com/hzi-braunschweig/SORMAS-Project/issues/8465

sergiupacurariu:
Fixes #11319


sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2588,6 +2588,9 @@ public void deleteWithContacts(String caseUuid, DeletionDetails deletionDetails)
 	private void deleteCase(Case caze, DeletionDetails deletionDetails)
 		throws ExternalSurveillanceToolRuntimeException, SormasToSormasRuntimeException {
 
+		if (!caseService.inJurisdictionOrOwned(caze)) {
```
carina29, line 2591: If you have time, it would be great to have some unit tests for deleteCase, deleteCases in CaseFacadeEjbTest
and deleteContact, deleteContacts in ContactFacadeEjbTest with this scenario ( when the jurisdiction is different and when is the same and these methods are called).

sergiupacurariu: tests added

carina29 (approved):

PR #11381, opened 2023-01-20
#11326 - [Samples] District of samples associated with cases only is …

### LINKED ISSUE #11326: [Samples] District of samples associated with cases only is not displayed on the grid
### Bug Description

In the 'Samples' directory, for the column 'District' the associated district is not displayed for samples that have only cases associated.

When filtering based on the values from the responsible region and district of the case, the sample is displayed as a result, but the district is still not displayed in the grid.

#### Steps to Reproduce

1. Create a case and add a sample to it;
2. Navigate to the 'Samples' directory and observe the 'District' column values.

### Actual Behavior

The 'District' of samples that have only cases associated is not displayed. 

### Expected Behavior

The district should be displayed for samples that have only cases associated. 

#### Screenshots

GIF of the behavior:
Sample district

#### System Details

* Device: Windows 10
* SORMAS version: 1.79.0; 1.78.0
* Server URL: local machine; https://release-international.sormas.netzlink.com/
* User Role: Admin+NatUser

sergiupacurariu:
…not displayed on the grid

Fixes #11326

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
@@ -185,7 +185,7 @@ public class SampleService extends AbstractDeletableAdoService<Sample>
 
 		final Join<Sample, Contact> contact = joins.getContact();
 		final Join<Contact, District> contactDistrict = joins.getContactDistrict();
-		final Join<Case, District> contactCaseDistrict = joins.getContactCaseDistrict();
+		final Join<Case, District> contactCaseDistrict = joins.getContactCaseResponsibleDistrict();
 
 		final Join<EventParticipant, Event> event = joins.getEvent();
 		final Join<Location, District> eventDistrict = joins.getEventDistrict();
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
@@ -181,7 +181,7 @@ public class SampleService extends AbstractDeletableAdoService<Sample>
 		SampleJoins joins = sampleQueryContext.getJoins();
 
 		final Join<Sample, Case> caze = joins.getCaze();
-		final Join<Case, District> caseDistrict = joins.getCaseDistrict();
+		final Join<Case, District> caseDistrict = joins.getCaseResponsibleDistrict();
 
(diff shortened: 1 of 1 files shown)


sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
```diff
@@ -181,11 +181,11 @@ public List<SampleIndexDto> getIndexList(SampleCriteria sampleCriteria, Integer
 		SampleJoins joins = sampleQueryContext.getJoins();
 
 		final Join<Sample, Case> caze = joins.getCaze();
-		final Join<Case, District> caseDistrict = joins.getCaseDistrict();
+		final Join<Case, District> caseDistrict = joins.getCaseResponsibleDistrict();
```
leventegal-she, line 184: should be a logic here:

district if it's set, responsible district otherwise

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
sergiupacurariu: updated

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
```diff
@@ -181,11 +181,11 @@ public List<SampleIndexDto> getIndexList(SampleCriteria sampleCriteria, Integer
 		SampleJoins joins = sampleQueryContext.getJoins();
 
 		final Join<Sample, Case> caze = joins.getCaze();
-		final Join<Case, District> caseDistrict = joins.getCaseDistrict();
+		Join<Case, District> caseDistrict = caze.get(Case.DISTRICT) != null ? joins.getCaseDistrict() : joins.getCaseResponsibleDistrict();
```
leventegal-she, line 184: caze.get(Case.DISTRICT) != null  is always true.

BTW I re-checked the logic in this method and it would be ok to use the responsible district only

leventegal-she (changes requested):

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleService.java
sergiupacurariu: changed back to initial approach

leventegal-she (approved):

PR #11385, opened 2023-01-23
#8395 fix adding multiple tests on sample creation

### LINKED ISSUE #8395: Adding 2 positive Pathogen Test Results with the same Disease to a Sample, duplicates the Sample, and the pop-up message for creation of a resulting case appears twice
### Bug Description
When the user adds 2 positive Pathogen Test Results (for the same Disease), the Sample in the Contact's Sample Card gets duplicated.
Also, both pop-up messages are the same, and they both inform the user that the case will be set as a resulting Case of this Contact.

### Steps to Reproduce

1. Create a new Contact and set the Disease to COVID-19;
2. Add a Sample with a 'positive' Pathogen Test Result, with the Disease set to COVID-19; do not save yet;
3. In the Same form add another 'positive' Pathogen Test Result with the Disease set to COVID-19; and Save;
4. In the following pop-up select 'Yes':
image
5. In the Create New Case form, fill in the mandatory data and Save;
6. In the following pop-up select 'Yes':
image
7. In the Pick or Create Case pop-up, select 'Pick an existing Case' and then 'Confirm;
8. Inspect the Samples Card of the Case;

### Actual Behavior
- There are 2 Samples in the Sample Card. Basically the Sample gets duplicated. If the user opens them, each of the Samples show the same 2 Pathogen Test Results that were created on steps 2 and 3;
- Also, both pop-up messages are the same, and they both inform the user that the case will be set as a resulting Case of this Contact;

### Screenshots
image
image
image

### System Details

* Device: Windows 10
* SORMAS version: 1.70.0-SNAPSHOT (d874e30(
* Server URL: local machine
* User Role: Admin, NatUser

### Additional Information
Looks like the same issue as in ticket #8193, but, as discussed with MateStrysewske , I opened a different ticket because it's a different workflow.

leventegal-she:
Fixes #8395

Till now the pathogen tests where saved one by one through the PathogenTestForm components.
I've changed the logic to save them in the controller and handling result types for all of the created and changed tests at once. See `savePathogenTests`  and `handleAssociated` methods in [PathogenTestController.java](https://github.com/hzi-braunschweig/SORMAS-Project/pull/11385/files#diff-a56687c090792a362f26e52348a6ffd3738bcd55b8404daded1a287450401b5f)
The logic is:
1. For tests with same disease as the one on case/contact/event participant
   * if there is a positive test and the case then handle it: create case from contact, change sample result, etc
   * else if there are only negative tests handle it:  change sample result, etc. 
2. For tests with other disease
  * Handle them grouped by disease: create case for the new desease

For case 1. shouldn't we just check the latest test instead? @MateStrysewske @MartinWahnschaffe


MateStrysewske (changes requested):
I have a feeling that the samples are wrongly associated under specific circumstances:

- Create a contact with disease COVID-19
- Create a sample and add three pathogen tests before saving the sample: 2 positive tests for COVID-19 and 1 positive test for Cholera
- Choose "Yes" when asked about creating a case for Cholera, and also "Yes" when asked about creating a resulting case for COVID-19
- Check the sample directory - there are two samples, which is correct, since we needed to clone the sample for Cholera, but the contact is actually associated to the Cholera sample, not the Covid sample, which doesn't seem to make a lot of sense; I guess the reason is that in my case, the prompt to create the Cholera case came before the prompt to create a resulting case for the contact, and thus the Cholera sample was created first, and the Covid sample is the one that got cloned.

I'm honestly not quite sure how to fix this issue; my idea would be to always process the pathogen tests first that match the disease of the entity associated with the sample, and only afterwards process the others; the outcome should be that the prompt to create a resulting case for the contact is triggered before the prompt to create a new case for the differing disease.

MateStrysewske (changes requested):
The contact workflow seems to work fine now, but the event participant workflow is still buggy (as documented by Roxana in the issue) - can you check that again please?

1) Create an event participant for an event with Covid-19
2) Create a sample with two pathogen tests: One for Covid-19, the other for Cholera (Cholera needs to be the second one)
3) Save the sample; Click "Yes" when asked to create a Cholera case; click "Yes" when asked to create a Covid-19 case -> warning message appears and pathogen test can't be saved.

sormas-vitagroup:
SonarCloud analysis: Please find the results at https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11385

MateStrysewske (changes requested):
We now have the same issue as with contacts before, i.e. the event participant is associated with the wrong sample (the Cholera one instead of the Covid one).

Also, please check the workflow for travel entries, I suspect that we might run into similar problems there.

leventegal-she:
> We now have the same issue as with contacts before, i.e. the event participant is associated with the wrong sample (the Cholera one instead of the Covid one).
> 
> Also, please check the workflow for travel entries, I suspect that we might run into similar problems there.

Fixed for event participant. 
Travel entries don't have samples so there is no similar flow and issue

sormas-vitagroup:
SonarCloud analysis: Please find the results at https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11385

MateStrysewske (approved):

PR #11409, opened 2023-01-27
qu-auto/SORQA-639 Delete a case in source system with handing ownership before acceptance

hms-sgent:
Fixes #


hms-sgent:
image

hms-sgent:
image

PR #11420, opened 2023-02-01
qu-auto/SORQA-641 Stabilize "Delete a contact in source system with handing ownership before acceptance"

hms-sgent:
…ted shared contact

Fixes #


hms-sgent:
image

PR #11423, opened 2023-02-01
qu-auto/SORQA-639 Delete a case in source system with handing ownership before acceptance

hms-sgent:
…ted shared case

Fixes #


hms-sgent:
image

PR #11431, opened 2023-02-02
#10876 [Facility fields] Make the display of facility related fields …

### LINKED ISSUE #10876: [Facility fields] Make the display of facility related fields consistent across the app [2]
### Problem Description

Currently, for a user that has access only to sensitive data and no access to personal data, or vice versa, there are inconsistencies in the way that facility related fields are displayed in the app.

The field 'Facility category' is not classified as either personal or sensitive - the behavior for it is inconsistent therefore in the app. 

### Proposed Change

For a user that has access to only sensitive or only personal data:

- [x] Person page

Facility category field - field is editable (but the changes made can not be saved as the data from this field influences the data in the other facility related fields, which in turn are not editable) - proposed change -> become not editable, display 'Confidential';

Facility type field - is not annotated as either personal or sensitive - for a user that can see only sensitive data the field is empty and not editable / for a user that can see only personal data the field is visible and editable - proposed change -> the field should be not editable, and should display 'Confidential';

Facility field - not editable, displays 'Confidential' (expected behavior);

Facility name and description field - is annotated as both sensitive and personal - not displayed at all for either the user that has access to only sensitive or only personal data - proposed change -> filed should be displayed, should be not editable, should display 'Confidential';

- [x] Addresses - new entry - proposed change -> same as for 'Person page';

- [x] Event participant page - proposed change -> same as on 'Person page';

- [x] Immunization page - proposed change -> same as on 'Person page';

- [x] Case page - the fields related to 'Place of stay' and respectively facility are not displayed at all - proposed change -> they should be displayed, not editable, display 'Confidential'.

### Additional Information

Please consider applying the changes to the mobile app as well.

Inconsistencies observed while testing for #10560.

leventegal-she:
…consistent across the app

Fixes #10876

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseExportDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseExportDto.java
@@ -147,6 +147,8 @@ public class CaseExportDto extends AbstractUuidDto {
 	@PersonalData
 	@SensitiveData
 	private String community;
+	@PersonalData
+	@SensitiveData
 	private FacilityType facilityType;
 	@PersonalData
 	@SensitiveData
--- a/sormas-api/src/main/java/de/symeda/sormas/api/location/LocationDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/location/LocationDto.java
@@ -142,6 +142,7 @@ public class LocationDto extends PseudonymizableDto {
 	@Size(max = FieldConstraints.CHARACTER_LIMIT_SMALL, message = Validations.textTooLong)
 	private String addressTypeDetails;
 	@PersonalData
+	@SensitiveData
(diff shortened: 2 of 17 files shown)


sergiupacurariu:
The PersonView shows data and addresses to a user which is outside of any of the entities connected to it. The PersonView from Cases, Contacts, TravelEntry, and Immunization are still viewable. On the other hand the data on the Event Participant which belongs to the same person everything is Confidential. It seems it is not consistent among all the PersonViews.

sergiupacurariu:
The facility fields are confidential when a Person is accessed by a user who does not have any of the entities related to that person in their's jurisdiction. However, some of the fields are editable. The logic based on which they remain editable is verified in a different ticket.

sergiupacurariu (approved):

PR #11460, opened 2023-02-08
#10569 [S2S] Filter directories when navigating from duplicate detect…

### LINKED ISSUE #10569: [S2S] Filter directories when navigating from duplicate detection warning dialog [2]
### Feature Description

A duplicate detection is done when accepting a case or contact. 
When a duplicate is found a warning message is displayed with option to navigate to case merge, contact merge, contacts directory or person directory after accepting the share request as described in https://github.com/hzi-braunschweig/SORMAS-Project/issues/9527#issuecomment-1265442536.

### Proposed Change

When the user choses to "accept and navigate" the grids on those pages should be filtered to show only the similar cases, contacts or persons

### Acceptance Criteria

**Accepting case**

1. There is at least one similar case: Case merge overview should show only the similar cases to the ones being accepted
2. From the similar cases at least one similar case that was converted from a contact: Case merge overview should show only the similar cases to the ones being accepted
3. There is at least one similar contact not converted to case: show only those contacts

**Accepting contact**

1. There is at least one similar contact: Contact merge overview should show only the similar contacts to the ones being accepted
2. From similar contacts at least one is converted to case: Contact merge overview should show only the similar contacts to the ones being accepted
3. There is at least one similar case: If there are multiple contacts to be accepted and with multiple similar cases, navigate to contacts directory and show only the contacts with similar cases, otherwise navigate to the contact details so the user can convert it from there

**Accepting case or contact**

There are only similar persons (no similar case and/or contact): Persons directory should show only the similar persons

leventegal-she:
…ion warning dialog - block automatic loading of duplicate cases until first manual load on merge cases view

Fixes #10569

also fixed duplicate loading of directory pages because of the url update on the ManScreen.java


sormas-vitagroup:
SonarCloud analysis: Please find the results at https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11460

MateStrysewske (changes requested):
The hint that opens up when you enter the merge directory is supposed to be displayed every time you enter it, not only once - basically, the current behaviour on the master branch is that the filters are preserved, but the actual data retrievel is postponed until the user has clicked on "Confirm". This should be reproduced here again.

leventegal-she:
> The hint that opens up when you enter the merge directory is supposed to be displayed every time you enter it, not only once - basically, the current behaviour on the master branch is that the filters are preserved, but the actual data retrievel is postponed until the user has clicked on "Confirm". This should be reproduced here again.

Blocking the load for the first time it makes sense, but forcing the user to always close a dialog and click on confirm filters is a bad UX and annoying for them knowing that once the filters are changed and confirmed they get stored and when they come back to the merge view it will bi filtered as before. So I think if a user confirmed to load all even if it takes long or it already filtered so the list is smaller and loads faster than SORMAS should ask the same thing again and again.

MateStrysewske:
@leventegal-she I disagree partly, because I see a potential situation where loading the list takes quite a long time, and you don't want to be forced to load the same list that you filtered by before the next time you enter the screen - you might actually want to use different filters, but you have to wait for the list to load first.

However, given that Martin has also worked on improving the performance of the grid significantly by limiting the amount of results, I agree that we can leave it like it is right now because my point from above is probably no longer that relevant. i.e. you can resolve the merge conflict and then merge the issue :)

MateStrysewske (approved):

PR #11495, opened 2023-02-13
#11482 s2s enable delete button for read only entities

### LINKED ISSUE #11482: S2S -  Enable delete button for read-only entities
### Problem Description
The case described refers to the environments that have S2S.

**Scenario**:

1. As NatUser+Admin having a Case/Contact
2. Send the Case/Contact with handover to another health department
3. The target health department receives and accepts the Case/Contact
4. Go back to source health department of the Case/Contact
5. Try to delete the entity => At this point this action is not possible from UI because the button to delete is disable, but from the API is possible as it was fixed in this ticket https://github.com/hzi-braunschweig/SORMAS-Project/issues/11319

An entity is in read-only mode because:

- was handed over to another health department(the above explained scenario)
- was received with only to view from another health department

### Proposed Change
As health department should be able to delete the entity that is read-only(to view) from UI, the delete button to be enable. The reason to have the button enable is for the situation when the health department doesn't need anymore that entity in their system. At this moment there is no possibility from UI to delete an entity that has to-view.

The users that will be allowed to do this action of deletion has to have:

- delete user right
- if they are in their jurisdiction

The BE part was implemented in this ticket https://github.com/hzi-braunschweig/SORMAS-Project/issues/11319

The remaining part to have consistency is to implement the FE part of enabling the delete button from the explained cases having an entity in read-only mode

leventegal-she:
Fixes #11482

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
@@ -1625,15 +1625,12 @@ public class CaseService extends AbstractCoreAdoService<Case, CaseJoins> {
 	@Override
 	public EditPermissionType getEditPermissionType(Case caze) {
 
-		if (caze.getSormasToSormasOriginInfo() != null && !caze.getSormasToSormasOriginInfo().isOwnershipHandedOver()) {
-			return EditPermissionType.REFUSED;
-		}
-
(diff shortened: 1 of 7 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11495

ChristopherRiedel (approved):

PR #11556, opened 2023-02-27
#11527 - Reintroduce proper similarity name search

### LINKED ISSUE #11527: Unrelated persons suggested as duplicates during linelisting contact within a case
### Bug Description

-Persons that are not similar are suggested during the creation of contacts within a case.

#### Steps to Reproduce

1. Open a case and line list about three contacts, use names such as 111, 222, etc
2. System sometimes suggests persons that do not share similar names or tell you that there are duplicate but the list is empty (See attached images below)

#### System Details

* Device: web
* SORMAS version: 1.81.0-snapshot (7e6f577)
* Android version/Browser:
* Server URL: https://test.sormas.netzlink.com/sormas-ui/#!about
* User Role: national user + admin

### Expected Behavior

- Only similar persons should be suggested
- Duplicate empty lists should not be shown  
- If this is caused by a low value for the name similarity threshold on the server, then it may be good to restrict its value within a range that will not result in the system suggesting persons that are not likely duplicates.
### Additional Information

1. Person's name is 222 222 and Aurelien is suggested as duplicate
image

2. Person's name is 333 333 but random names suggested
image

3. Person's name is 555 but duplicate detection is triggered with an empty list
image

MateStrysewske:
Fixes #11527

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/person/PersonSimilarityCriteria.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/person/PersonSimilarityCriteria.java
@@ -11,6 +11,8 @@ public class PersonSimilarityCriteria extends BaseCriteria implements Cloneable 
 	public static final String BIRTHDATE_MM = "birthdateMM";
 	public static final String BIRTHDATE_DD = "birthdateDD";
 
+	private String firstName;
+	private String lastName;
 	private Sex sex;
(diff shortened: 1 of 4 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11556

leventegal-she (approved):

PR #11563, opened 2023-03-01
#10721 create basic samples dashboard

### LINKED ISSUE #10721: Create basic samples dashboard [8]
### Feature Description

For the creation of a sample dashboard, a new button to switch between the surveillance, contact and sample dashboard is needed. 
The creation of sample dashboard is tracked in https://github.com/hzi-braunschweig/SORMAS-Project/issues/10723. 

### Proposed Change

- [x] Add a `DASHBOARD_SAMPLES_VIEW` user right and give it to all default user roles with the `DASHBOARD_SURVEILLANCE_VIEW` and `SAMPLES_VIEW` user rights.
- [x] Extend the dashboard switcher option group with a "Sample" option.
Enum boxes fields
- [x] Add a feature type property called `DASHBOARD_SAMPLES` to the `DASHBOARD` feature type, default to true. If set to false, remove the "Sample" option from the dashboard switcher
  - [x] Ideally (TBD whether this is out of scope of this issue, but for consistency reasons I @MateStrysewske think it makes sense), also add `DASHBOARD_SURVEILLANCE`, `DASHBOARD_CONTACTS` and `DASHBOARD_CAMPAIGNS` feature type properties and apply the same logic to the dashboard switcher; if all three feature types are set to false, don't display the dashboard at all
- [x] Add filters: In general look at all samples the user can access. Exclude deleted samples, include archived ones.
  - region of the associated entity. Note that a sample may have multiple associated entities, so duplicates need to be excluded. Add a tooltip to the filter
  - district. Same as region
  - type of sample (sampleMaterial)
  - date filter allowing to filter by date or epi week, as on surveillance dashboard, with the following options:
    1. Sample collection date (sampleDateTime)
    2. Report date of associated entity (e.g. case, contact, event of event participant)
    3. Most Relevant date: priority is pathogen test date (earliest if multiple exist for the sample) + fall back to sample collection date. Associated entity is not relevant
  - disease (of the associated entity); including the option to choose "no disease" (NOT similar to an empty filter!), which is necessary for event participants of events without an assigned disease -> use a dropdown filter as on the contact dashboard
- [x] Add an element that displays the overall samples count as in the following picture.  
           - All samples == total number of samples,
           - the counts and proportions correspond to the variable  "final lab results": Add counts and proportions for `Indeterminate, pending, negative, positive, not done`.  This is a bit different from the widget on the surveillance dashboard, because it considers all samples (not only the latest sample of cases) and also shows the "not done" status
image

**OPEN QUESTIONS:**
- What should the region and district filters filter by? The jurisdiction of their parent entity (e.g. case, contact) or the address of the associated person (because that's what is supposed to be used on the map according to #10745)? --> Proposal by @MartinWahnschaffe Let's go with the jurisdiction of the parent entity. The person's address may be empty and from the responsibility point of view the jurisdiction is the right choice, I think.
- ~~How do we deal with "copied" samples? E.g. when we create a resulting case from a contact, the sample of the contact is duplicated, so in SORMAS, we now have two samples (one for the contact, one for the case) referring to the same physical sample. If we count those, we end up with a higher sample number than the number of samples that have been physically taken; if we don't count those, how do we determine which sample to pull the final laboratory result from?~~ Normally case and contact share the same sample.

### Acceptance Criteria

* The button Sample is implemented as described above.
* sum of count for all enums == All samples
* sum of proportion ==100%
* The needed database query may be slow. Should be tested on a system with at least 10k samples and the resulting query (with region filter active) should be documented with https://explain.dalibo.com/.

leventegal-she:
Fixes #10721

https://explain.dalibo.com/plan/71fe5dd39460c7b7#plan


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11563

sergiupacurariu:
the Disease combo box from the Sample Dashboard filter does not contain all diseases

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11563

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleJoins.java
```diff
@@ -152,6 +152,10 @@ public void setContactJoins(ContactJoins contactJoins) {
 		this.contactJoins = contactJoins;
 	}
 
+	public CaseJoins getContactCaseJoins() {
+		return getContactJoins().getCaseJoins();
```
sergiupacurariu, line 156: this added method is not used

sergiupacurariu (changes requested):

sergiupacurariu (approved):

sormas-backend/src/main/java/de/symeda/sormas/backend/sample/SampleJoins.java
leventegal-she: I will remove in the next sample dashboard related ticket I'm working on right now

PR #11565, opened 2023-03-01
#11360 - [Person merge] Introduce merging of event participants in th…

### LINKED ISSUE #11360: [Person merge] Introduce merging of event participants in the person merge flow [5]
### Feature Description

With #5606 the person merge was introduced, which led to the same person being listed twice or more as event participant in the same event. As a follow-up it was discussed and decided to introduce an extra step to the person merge - merge of event participants.

The merge of event participants should be added as an extra step to the person merge flow currently implemented. 

**User rights necessary:**
- [x] PERSON_MERGE is required (as without it the user could not access the person merge flow);
- [x] If the user does NOT have the view event participants right, we check whether an event participant merge would be necessary when merging the person; if yes, we prevent the user from merging the persons (with a generic message about missing user rights); if no, we do the merge.
- [x] The user should not be required to have edit and delete rights for event participants; we should have a backend method that is only called when merging event participants that circumvents those checks.

**Merge event participants:**
- [x] should be triggered in the person merge flow if the user selects either 'Pick' or 'Merge' person;
- [x] should be triggered if the persons selected for the person merge are both active or archived event participants within the same event;
~~- [x] add the 'Merge' and 'Discard' buttons at the bottom right of the pop-up;~~
- [x] add the 'Pick', 'Merge' and 'Discard' buttons at the bottom right of the pop-up;
~~- [x] 'Merge' should be disabled until the user selects one of the entries;~~
~~- [ ] 'Merge' and 'Pick' should be disabled until the user selects one of the entries;~~
- [x] when clicking 'Merge':
* keep the selected event participant as the remaining entry/leading event participant;
* link all entities of the discarded event participant to the remaining one;
* introduce a confirmation pop-up after the user clicks merge event participants (similar to the one we already have for persons) to inform the user that the action is not reversible; provide options to continue or discard; if discard is selected, redirect them back to the 'Merge event participant' pop-up view;
* make sure that after the merge the discarded event participant is no longer linked to any entities;
* the whole merging should only be done after the user has gone through the whole process; i.e. don't merge the persons until the user has decided what to do with the event participants -> decide whether to merge persons or event participants first based on what makes more sense in terms of related entities;
- [x] #11585

- [x] when clicking 'Pick':
* keep the selected event participant as the remaining entry/leading event participant;
* the entities of the discarded event participant should not be linked with the remaining event participant (as to be consistent with the behavior of case and contact merge);
* the confirmation pop-up that is already implemented and appears when merge is selected should also appear after the user clicks 'Pick';
* the fields 'Responsible region', 'Responsible district', 'Involvement description' and 'Vaccination status' should not be filled with values from the discarded event participant (in line with the behavior for pick persons, cases and contacts).

- [x] Note - the person information from the event participant tab should behave just as it was implemented and it should be merged or picked based on the choice done by the user in the first pop-up of the 'Merge persons' view.

- [x] when clicking 'Discard' in the 'Merge event participants' pop-up, the user should be taken back to the 'Pick or merge person' pop-up view; 
- [x]  only one event participant should be selectable at one time; if the persons selected are event participants in more than one event, then 1 person per event should be selectable at a time;
- [x] if the persons selected are both event participants in more than one event, then in the merge event participants window the event participant entries should be displayed grouped by event, one underneath the other; it should be clearly visible to which event each of them is linked to (the event title should appear in the grid, as listed below);
- [x] if both event participants have resulting cases, keep the case of the remaining event participant, and the case of the discarded event participant should still remain in the system (it will be associated with the remaining person) but it will not be associated with the event anymore.

**Merge event participants pop-up display:**
- [x]  the grid in the 'Merge event participants' widow should be similar to what we already have for the other merge entities window: EVENT PARTICIPANT ID; PERSON ID; FIRST NAME; LAST NAME; AGE AND BIRTHDATE; SEX; DISTRICT(of the event/event participant); EVENT ID; EVENT TITLE; INVOLVEMENT DESCRIPTION; RESULTING CASE ID; CONTACT COUNT.

- [x] Message that should be displayed in the pop-up beneath the heading and above the grid that contains the event participants:
_The selected persons both have one event participant in at least one event. Merging the persons would lead to duplicate event participants.
For each event, please select one of the event participants below as the leading event participant. 
'Merge': The discarded event participant will be removed from the event. Its associated entities and information will be linked and/or added to the remaining event participant, but not overwritten. You will not be able to undo this action.
'Pick': The discarded event participant will be removed from the event. Its associated entities and information will not be linked and/or added to the remaining event participant. You will not be able to undo this action._

Additional information that should be taken into account:
When checking for event participants, we only check for person ID and don't consider any of the other person-related fields.

### Additional Information

This was created as a follow-up to #5606  and #11016 (which will be discarded once this current ticket is refined) after the discussion from the refinement session from the 12th of January.

sergiupacurariu:
…e person merge flow - additional change requests

Fixes #11360

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/person/PersonFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/person/PersonFacade.java
@@ -82,7 +82,12 @@ public interface PersonFacade extends BaseFacade<PersonDto, PersonIndexDto, Pers
 
 	void mergePerson(PersonDto leadPerson, PersonDto otherPerson);
 
-	void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean mergeProperties, List<String> selectedEventParticipantUuids);
+	void mergePerson(
+		String leadPersonUuid,
+		String otherPersonUuid,
+		boolean mergeProperties,
+		List<String> selectedEventParticipantUuids,
+		boolean mergeEventparticipantProperties);
 
 	boolean isPersonSimilar(PersonSimilarityCriteria criteria, String personUuid);
 
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
@@ -1803,13 +1803,18 @@ public class PersonFacadeEjb extends AbstractBaseEjb<Person, PersonDto, PersonIn
 
 	@Override
 	@RightsAllowed(UserRight._PERSON_EDIT)
-	public void mergePerson(String leadPersonUuid, String otherPersonUuid, boolean mergeProperties, List<String> selectedEventParticipantUuids) {
+	public void mergePerson(
+		String leadPersonUuid,
+		String otherPersonUuid,
+		boolean mergePersonProperties,
+		List<String> selectedEventParticipantUuids,
+		boolean mergeEventParticipantProperties) {
 
 		if (leadPersonUuid.equals(otherPersonUuid)) {
 			throw new UnsupportedOperationException("Two different persons need to be selected for merge!");
 		}
 
-		if (mergeProperties) {
+		if (mergePersonProperties) {
 			final PersonDto leadPersonDto = getByUuid(leadPersonUuid);
 			final PersonDto otherPersonDto = getByUuid(otherPersonUuid);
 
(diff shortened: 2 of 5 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11565

leventegal-she (approved):

PR #11580, opened 2023-03-02
qu-auto/SORQA-856 Stabilize "Validate newly created User Role cannot be deleted if assigned towards an user"

hms-sgent:
Fixes #


hms-sgent:
image

PR #11620, opened 2023-03-10
#11449: Added diseaseVariant to ExternalMessage and prefilling cases …

### LINKED ISSUE #11449: [DEMIS - External messages] Disease variant for External messages [2]
### Feature Description

Disease variants need to be extended in the 'External messages' - lab messages. This is part of epic #11447.

Add disease variant field to ExternalMessage and prefill the according field in SORMAS (currently this seems to be retrieved only in the 'Disease variant' field of pathogen tests).

### Proposed Change

The value set from the ExternalMessage.testedDiseaseVariant should be taken from the disease variants which can be configured in the DB - (https://github.com/hzi-braunschweig/SORMAS-Project/wiki/Adding-Disease-Variants-to-a-SORMAS-Server).

- [x] Add diseaseVariant to dto and entity.

- [x] Fields for disease variant need to be added to the UI:
- new column 'Disease variant' on the grid similar to the one in other directories; it should be filled with the disease variant value from the external message;
image

- introduce filter fields based on 'Disease' and 'Disease Variant' (as the ones we already have in the other directories);
image image

- extend it in the pop-up forms that appear while processing the external messages ('Create new case' only as it is the only creation form that contains the field) the disease variant should be prefilled with the variant from the message, alongside with the disease. (This  already happens when reaching the step of processing the associated sample, where the disease variant is already filled with the variant from the message);

### Acceptance Criteria

The user should still be able to edit the disease variant that was prefilled, just as they can currently edit the disease that is also prefilled 
### Additional Information

[#80](https://github.com/hzi-braunschweig/sormas-demis-adapter/issues/80) will provide the variant.

ChristopherRiedel:
…when processing

Fixes #11449


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11620

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11620

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11620

leventegal-she (changes requested):
If you put an unknown disease variant than the external message page crashes.

You should not use @Convert(converter = DiseaseVariantConverter.class) on the field and you should convert from string to DiseasesVariant when you create the DTOs

ChristopherRiedel:
> If you put an unknown disease variant than the external message page crashes.
> 
> You should not use @convert(converter = DiseaseVariantConverter.class) on the field and you should convert from string to DiseasesVariant when you create the DTOs

The csv mapping in the adapter prevents incorrect variants from entering the system. We can therefore assume that we will only receive correct variants.

leventegal-she (approved):

PR #11652, opened 2023-03-16
#11465 - Limit case duplicate merging comparison based on creation da…

### LINKED ISSUE #11465: Limit case duplicate merging comparison based on creation date and archived status [3]
### Problem Description

In #9054 we have improved the performance of the case duplicate merging query. This doesn't mean that the duplicate detection is fast in any case. The whole process still executes a lot of comparison logic. In the example the query had to compare 372 of the 1500 cases with each of the ~85k cases, resulting in 31 mio. comparisons. You can easily see where this grows when any of both sides grows.

### Proposed Change
Decrease the comparison amount by doing the following:

- [x] Backend: Only compare cases with cases that were created in the same period or before (currently we are comparing with all other cases). You will find the duplicates that would fall out of the result as soon as you looked at the future period. In addition it has the benefit that redoing a check for the same week will not give you new results, because new cases have been created in later weeks.
- [x] UI: Introduce a filter for the case status (active, archived, all). By default only check active cases. The CaseCriteria already has the option EntityRelevanceStatus for this, so only the UI needs to be extended for this
- [x] UI: Set the default filtered period for the creation date to the last week. I'd suggest to use DateHelper.getPreviousEpiWeek in combination with getEpiWeekStart and getEpiWeekEnd for this.

### Implementation Details

- [ ] The join from the second case to the person needs to use a sub query for the person that orders the persons by id. This will lead to a much better query planning, because postgres realizes that it's beneficial to materialize the joined data.

The overall resulting query should look like this (when called with the admin), with the modified sections being highlighted. Note that this is the variant when only not-archived cases are querried.

> select case0_.id as col_0_0_, case1_.id as col_1_0_, case0_.creationDate as col_2_0_ 
> from cases case0_ cross join cases case1_ 
> left outer join Person person2_ on case0_.person_id=person2_.id 
> left outer join Symptoms symptoms3_ on case0_.symptoms_id=symptoms3_.id 
> left outer join **(SELECT * FROM Person ORDER by id)** person4_ on case1_.person_id=person4_.id
> left outer join Symptoms symptoms5_ on case1_.symptoms_id=symptoms5_.id 
> 
> where case0_.deleted=FALSE and case1_.deleted=FALSE 
> **and case0_.archived=FALSE and case1_.archived=FALSE** 
> and case0_.creationDate>'2021-12-30' and case0_.creationDate<'2021-12-31' 
> and case0_.deleted=FALSE and similarity(((person2_.firstName||' ')||person2_.lastName),((person4_.firstName||' ')||person4_.lastName))>0.65
> and case0_.disease=case1_.disease 
> and abs(date_part('EPOCH',case0_.reportDate)-date_part('EPOCH',case1_.reportDate))<=2592000 
> and (person2_.sex is null or person4_.sex is null or person2_.sex='UNKNOWN' or person4_.sex='UNKNOWN' or person2_.sex=person4_.sex) 
> and (person2_.birthdate_dd is null or person2_.birthdate_mm is null or person2_.birthdate_yyyy is null or person4_.birthdate_dd is null or person4_.birthdate_mm is null or person4_.birthdate_yyyy is null or person2_.birthdate_dd=person4_.birthdate_dd and person2_.birthdate_mm=person4_.birthdate_mm and person2_.birthdate_yyyy=person4_.birthdate_yyyy) 
> and (symptoms3_.onsetDate is null or symptoms5_.onsetDate is null or abs(date_part('EPOCH',symptoms3_.onsetDate)-date_part('EPOCH',symptoms5_.onsetDate))<=2592000) 
> **and case1_.creationDate<case0_.creationDate** 
> **and case1_.creationDate<'2021-12-31'**
> 
> order by case0_.creationDate desc
> 
> limit 100;

### Additional Information

* It's not clear yet how the users manual will be updated.

sergiupacurariu:
…te and archived status

Fixes #11465

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
@@ -1944,6 +1944,7 @@ public class CaseService extends AbstractCoreAdoService<Case, CaseJoins> {
 
 		Predicate userFilter = createUserFilter(caseQueryContext);
 		Predicate criteriaFilter = criteria != null ? createCriteriaFilter(criteria, caseQueryContext) : null;
+		Predicate criteriaFilter2 = criteria != null ? createCriteriaFilter(criteria, caseQueryContext2) : null;
 		Expression<String> nameSimilarityExpr = cb.concat(person.get(Person.FIRST_NAME), " ");
 		nameSimilarityExpr = cb.concat(nameSimilarityExpr, person.get(Person.LAST_NAME));
 		Expression<String> nameSimilarityExpr2 = cb.concat(person2.get(Person.FIRST_NAME), " ");
@@ -1993,8 +1994,7 @@ public class CaseService extends AbstractCoreAdoService<Case, CaseJoins> {
 				cb.lessThanOrEqualTo(root2.get(Case.CREATION_DATE), DateHelper.getStartOfDay(criteria.getCreationDateFrom())),
 				cb.greaterThanOrEqualTo(root2.get(Case.CREATION_DATE), DateHelper.getEndOfDay(criteria.getCreationDateTo()))));
 
-		Predicate filter = CriteriaBuilderHelper
-			.and(cb, createDefaultFilter(cb, root), createDefaultFilter(cb, root2), userFilter, criteriaFilter, nameSimilarityFilter, diseaseFilter);
+		Predicate filter = CriteriaBuilderHelper.and(cb, userFilter, criteriaFilter, criteriaFilter2, nameSimilarityFilter, diseaseFilter);
 
 		if (!showDuplicatesWithDifferentRegion) {
 			Predicate regionFilter = cb.and(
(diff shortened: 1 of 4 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11652

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11652

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
```diff
@@ -1993,8 +1994,7 @@ public List<CaseMergeIndexDto[]> getCasesForDuplicateMerging(
 				cb.lessThanOrEqualTo(root2.get(Case.CREATION_DATE), DateHelper.getStartOfDay(criteria.getCreationDateFrom())),
 				cb.greaterThanOrEqualTo(root2.get(Case.CREATION_DATE), DateHelper.getEndOfDay(criteria.getCreationDateTo()))));
 
-		Predicate filter = CriteriaBuilderHelper
-			.and(cb, createDefaultFilter(cb, root), createDefaultFilter(cb, root2), userFilter, criteriaFilter, nameSimilarityFilter, diseaseFilter);
+		Predicate filter = CriteriaBuilderHelper.and(cb, userFilter, criteriaFilter, criteriaFilter2, nameSimilarityFilter, diseaseFilter);
```
leventegal-she, line 1997: Not sure this is needed. It was not asked by the issue you worked on.

sergiupacurariu: fixed

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11652

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11652

PR #11669, opened 2023-03-16
#11502 - Fix backend event location validation

### LINKED ISSUE #11502: [Events] Events without jurisdiction can not be saved
### Problem Description

Due to a validation that was added in https://github.com/hzi-braunschweig/SORMAS-Project/issues/10110 events can not be saved (created or edited) without a jurisdiction set.

It used to be possible to create events without a jurisdiction, or edit them without having a jurisdiction, as long as the associated event participants had a jurisdiction set.

Additionally, it seems like the region and district fields are not marked as mandatory when no country is selected in the country dropdown (maybe this is only applied after doing an initial change, or it never happens), even when a server country is specified in the properties file. The fields are only supposed to be optional when no server country is specified in the properties.

#### Steps to Reproduce

1. Open the 'Create new event' form - the fields for region and district are mandatory (if the country is not set or if the country is set to Germany);
2. For an event already created, set a region and a district for the associated event participant; then remove the region and district that were previously on the event page;

### Actual Behavior

The events that do not have a region and district filled can not be saved.

### Expected Behavior

API level/backend validation should match the UI validation here, i.e., region and district are mandatory except for the following scenarios:

1) A different country than the server country is selected (if NO country is selected, the fields still have to be mandatory)
2) No server country has been specified in the properties file and any country has been selected (makes sense to review whether this is a good idea, but right now that's how we do it in the UI, so we should reproduce this behavior)

Additionally, make sure that the region and district fields are always marked as mandatory if a country is specified in the properties file and no country is selected in the country dropdown.

#### Screenshots

image
image

#### System Details

* Device: Windows 10 
* SORMAS version: 1.80.0; 1.78.0
* Server URL: local machine
* User Role: Admin+NatUser

MateStrysewske:
Align backend event location validation with UI validation

Fixes #11502

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventFacadeEjb.java
@@ -86,6 +86,7 @@ import de.symeda.sormas.api.i18n.Captions;
 import de.symeda.sormas.api.i18n.I18nProperties;
 import de.symeda.sormas.api.i18n.Strings;
 import de.symeda.sormas.api.i18n.Validations;
+import de.symeda.sormas.api.infrastructure.country.CountryReferenceDto;
 import de.symeda.sormas.api.infrastructure.facility.FacilityDto;
 import de.symeda.sormas.api.infrastructure.region.RegionReferenceDto;
 import de.symeda.sormas.api.location.LocationDto;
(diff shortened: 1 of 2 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11669

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11669

sergiupacurariu (approved):

PR #11673, opened 2023-03-17
#11627 - Duplicate case query is triggered twice on every search

### LINKED ISSUE #11627: Duplicate case query is triggered twice on every search
#### Steps to Reproduce

1.
2.

#### System Details

* Device:
* SORMAS version:
* Android version/Browser:
* Server URL:
* User Role:

sergiupacurariu:
Fixes #11627

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/MergeCasesView.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/MergeCasesView.java
@@ -96,10 +96,6 @@ public class MergeCasesView extends AbstractView {
 			ValoTheme.BUTTON_PRIMARY);
 
 		addHeaderComponent(btnBack);
-
-		if (!viewConfiguration.isFiltersApplied()) {
-			reloadAndUpdateDuplicateCount();
-		}
 	}
 
 	private void reloadAndUpdateDuplicateCount() {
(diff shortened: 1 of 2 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11673

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/MergeCasesView.java
```diff
@@ -97,7 +97,7 @@ public MergeCasesView() {
 
 		addHeaderComponent(btnBack);
 
-		if (viewConfiguration.isFiltersApplied()) {
+		if (!viewConfiguration.isFiltersApplied()) {
```
leventegal-she, line 100: the is the same condition as in the `enter` method but it does the opposite si it is confusing. Sometimes we reload when the filters are not applied and sometimes we show a message to the user, it should be consistent.

Also by default this flag is false so it loads and shows the user a message the he should click on the Apply filters button to see the duplicates

leventegal-she (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/caze/MergeCasesView.java
sergiupacurariu: update the change

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11673

leventegal-she (approved):

PR #11680, opened 2023-03-17
#10858 - Sync person problem

### LINKED ISSUE #10858: Mobile users can't push persons if they don't have the user right to manage symptom journal
### Bug Description

If a mobile user tries to push changes to a person (either directly to the person or a connected entity, e.g. an immunization) to the web app without having the user right `MANAGE_EXTERNAL_SYMPTOM_JOURNAL`, the synchronization fails if the person in question has any symptom journal status in the web database (e.g. "Unregistered"). The code responsible for throwing an exception here is PersonFacadeEjb:validateUserRights.

This happens because the `symptomJournalStatus` field present in the Person and PersonDto classes in the web app/api does not exist in the mobile app, which means that whenever a person is pushed to the web app, it's pushed with the `symptomJournalStatus` being set to null.

This ALSO means that any user of the mobile app who DOES have this user right will overwrite the symptom journal status of all persons they push with null.

### Steps to Reproduce

1. Have a person in the database with any symptom journal status
2. Make sure that this person is associated with e.g. an immunization that a mobile user without the `MANAGE_EXTERNAL_SYMPTOM_JOURNAL` has access to, e.g. a hospital informant
3. Make any changes to the person in the mobile app and try to synchronize

### Expected Behavior

Ensure that mobile app users are always allowed to push persons irrespective of the symptom journal status, and that the symptom journal status is NOT overwritten when a push is done.

Do this by not updating the symptom journal status when pushing persons via the mobile app; since we know when a push is done via the mobile app or not, this should be doable.

### System Details

* Device:
* SORMAS version: 1.77.0-SNAPSHOT, certainly older released versions.
* Android version/Browser:
* Server URL:
* User Role: Hospital Informant

### Additional Information

- The mentioned field `Person.symptomJournalStatus` was introduced in #1970.
- The mentioned validation in `PersonFacadeEjb.validateUserRights` was introduced by #8508.

MateStrysewske:
Ignore symptom journal status when pushing persons through the mobile app

Fixes #10858

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/person/PersonFacadeEjb.java
@@ -68,6 +68,7 @@ import com.google.i18n.phonenumbers.PhoneNumberUtil;
 import com.google.i18n.phonenumbers.Phonenumber;
 
 import de.symeda.sormas.api.Disease;
+import de.symeda.sormas.api.RequestContextHolder;
 import de.symeda.sormas.api.caze.AgeAndBirthDateDto;
 import de.symeda.sormas.api.caze.BirthDateDto;
 import de.symeda.sormas.api.caze.CaseClassification;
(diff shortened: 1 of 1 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11680

sergiupacurariu (approved):

PR #11684, opened 2023-03-17
[11661] use stable release instead of main

### LINKED ISSUE #11661: Use MobSF Action to security scan the App
### Problem Description
We already make good use of GitHub's security features and security scanners for the SORMAS application, however, we do not cover the Android App as of now. This should be no longer the case.

### Proposed Solution
Use the [mobsfscan](https://github.com/marketplace/actions/mobsfscan) action. I have used Mobsf in the past and its quite a nice tool for mobile security.

### Possible Alternatives
None, we should really take care of this.

JonasCir:
Fixes #11661 

Just adjust the version to avoid breaking changes from main.


StefanKock (approved):

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11684

PR #11751, opened 2023-03-28
#11690 S2S - Handing over the case should select place of stay only w…

### LINKED ISSUE #11690: S2S - Handing over the case should select place of stay only when the jurisdiction between 2 systems is different
### Problem Description

At the moment when sending a case with ownership to another system the responsible jurisdiction is automatically changed with the one from the receiving system/department even of the sending system has already completed the corrects jurisdiction of the receiving system.

Example: 
Green system has as jurisdiction Green-Region and Green-District
The Red system creates a case with the above jurisdiction and completes only the mandatory fields
The Red system sends the case to Green system with handover ownership
Green system accepts the case.

At this point in the case in the receiving system it can be seen that responsible jurisdiction remains the same but there is checked that the place of stay differes from the responsible jurisdiction but has the same Region and District as the responsible Region and District

### Proposed Change

When sending a case to another health department with handover ownership and Responsible Region and District are already the ones used by the receiving system, then the checkbox with Patient having a different place of stay should not be selected

### Added Value/Benefit
 This ticket is an extension that was not covered when was added this functionality https://github.com/hzi-braunschweig/SORMAS-Project/issues/8113

### Acceptance Criteria

- [ ]
- [ ]
- [ ]

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

leventegal-she:
…hen the jurisdiction between 2 systems is different

Fixes #11690

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/common/ConfigFacadeEjb.java
@@ -159,7 +159,7 @@ public class ConfigFacadeEjb implements ConfigFacade {
 
 	private static final String SORMAS2SORMAS_ETCD_KEY_PREFIX = "sormas2sormas.etcd.keyPrefix";
 
-	private static final String SORMAS2SORMAS_DISTRICT_EXTERNAL_ID = "sormas2sormas.districtExternalId";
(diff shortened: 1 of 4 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11751

carina29 (approved):

carina29:
This PR will be merged after the release.

PR #11771, opened 2023-03-31
#8213 - Enter home address for existing persons

### LINKED ISSUE #8213: Allow entering of home address for existing persons with empty home address [1]
### Problem Description

When creating a case with a chosen existing person the checkbox "Enter home address of the case person now" is always disabled, even if the home address of the person is empty.

### Proposed Change

Enable the checkbox if the address of the person is empty because in that case, users might want to add it at this point rather than having to enter the person page and enter it there afterwards.

### Acceptance Criteria

- [x] When creating a new case for an existing person, the checkbox is disabled if the home address of the person has at least one field that is not empty
- [x] Otherwise, the checkbox is enabled and the user can create a home address for the existing person
- [x] `LocationDto.checkIsEmptyLocation` is used to check this, and its implementation has been extended according to the specifications below

### Implementation Details
Extend `LocationDto.checkIsEmptyLocation`:
- Use `StringUtils.isNotBlank` for String fields instead of a simple null check
- Add all missing fields that are currently not checked in this method

MateStrysewske:
Fixes #8213

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonCreateForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonCreateForm.java
@@ -53,6 +53,7 @@ import de.symeda.sormas.api.person.PresentCondition;
 import de.symeda.sormas.api.person.SimilarPersonDto;
 import de.symeda.sormas.api.symptoms.SymptomsDto;
 import de.symeda.sormas.api.utils.DateHelper;
+import de.symeda.sormas.api.utils.LocationHelper;
 import de.symeda.sormas.api.utils.fieldaccess.UiFieldAccessCheckers;
 import de.symeda.sormas.api.utils.fieldvisibility.FieldVisibilityCheckers;
 import de.symeda.sormas.ui.ControllerProvider;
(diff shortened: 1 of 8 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11771

sormas-ui/src/main/java/de/symeda/sormas/ui/person/PersonCreateForm.java
```diff
@@ -286,9 +292,9 @@ protected Button createPersonSearchButton(String personSearchLoc) {
 					SimilarPersonDto pickedPerson = personSearchField.getValue();
 					if (pickedPerson != null) {
 						// add consumer
-						searchedPerson = FacadeProvider.getPersonFacade().getByUuid(pickedPerson.getUuid());
-						setPerson(searchedPerson);
-						enablePersonFields(false);
+						person = FacadeProvider.getPersonFacade().getByUuid(pickedPerson.getUuid());
+						setPerson(person);
+						enablePersonFields(false, true);
```
sergiupacurariu, line 297: when a person is picked then all fields are disabled no matter if they are empty or not. This does not satisfy the comment added by @adinaflorea9 in the ticket.

MateStrysewske: I think Adina's comment is out of scope for this issue because it would involve a lot more changes (e.g. this would need to be done for all create forms that include person details, not only the case create form). I've created a follow-up issue for it: https://github.com/hzi-braunschweig/SORMAS-Project/issues/11781

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11771

sergiupacurariu (approved):

PR #11775, opened 2023-04-03
fix Mergen without hand over the ownership - fill only mandatory fiel…

pk-sgent:
…ds while creating case

Fixes mergen without hand over the ownership - fill only mandatory fields while creating case
image


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11775

PR #11789, opened 2023-04-04
#11596 Add merge confirmation pop-up to contacts view and rephrase co…

### LINKED ISSUE #11596: Add merge confirmation pop-up to contacts view and rephrase content [1]
### Problem Description
At the moment when selecting Merge Case a confirmation pop-up appears informing user that the action can take a long time. 
This ticket https://github.com/hzi-braunschweig/SORMAS-Project/issues/9054 introduce a limit that user can use to retrieve the list of merging. Having the pop-up at this moment is also not consistent with Merge Contacts where the pop-up is not in place
Screenshot 2023-03-07 at 13 49 44

### Proposed Change
Since the performance issues should largely have been sorted out now, we should still keep the hint (in order to avoid users running into time-outs in situations where the performance is still bad), but change its text. Also, the hint should be added to the contact merge view for consistency reasons.

**Updated popup content:**

```
Calculating and displaying potential duplicates may take a lot of time under specific circumstances. It is recommended to use the filters on top of this view to reduce the number of cases that have to be compared at the same time, e.g. by choosing a narrow creation date interval and a low result count limit. If the potential duplicates are loading fast, extending these values should be safe.

In order to avoid performance issues when entering this view, potential duplicates are only loaded once you click on "Confirm Filters".
```

### Acceptance Criteria

- [x] The content of the pop-up has been changed to the text specified above
- [x] The same pop-up is also shown when entering the contact merge view.
  - [x] As in the case merge view, the pop-up is only shown when the user enters the view for the first time during their current session; afterwards it's not displayed anymore, but the user is still required to click on "Confirm filters" before the duplicates are loaded

leventegal-she:
…ntent

Fixes #11596


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11789

sergiupacurariu (approved):

PR #11794, opened 2023-04-05
#8213 - Reset continent on country removal

### LINKED ISSUE #8213: Allow entering of home address for existing persons with empty home address [1]

MateStrysewske:
Fixes #8213

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/utils/LocationHelper.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/utils/LocationHelper.java
@@ -95,4 +95,11 @@ public final class LocationHelper {
 
 		return true;
 	}
+
+	public static void resetContinentFieldsIfCountryRemoved(LocationDto newLocation, LocationDto oldLocation) {
+		if (newLocation.getCountry() == null && oldLocation.getCountry() != null) {
+			newLocation.setContinent(null);
+			newLocation.setSubcontinent(null);
+		}
+	}
 }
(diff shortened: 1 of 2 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11794

sormas-api/src/main/java/de/symeda/sormas/api/utils/LocationHelper.java
```diff
@@ -95,4 +95,11 @@ public static boolean checkIsEmptyLocation(LocationDto location) {
 
 		return true;
 	}
+
+	public static void resetContinentFieldsIfCountryRemoved(LocationDto newLocation, LocationDto oldLocation) {
```
sergiupacurariu, line 99: I believe the Locations that are already saved in database in a state that now is prevented should be cleaned up as well.

MateStrysewske: Since a state in which the continent is present, but the country is not, can be valid for locations (exposures and activities as case), I don't think we can do that without potentially messing up existing data.

sergiupacurariu (approved):

PR #11827, opened 2023-04-11
#11710 fix Delete buttons are enabled for entities outside the user's…

### LINKED ISSUE #11710: Delete buttons are enabled for entities outside the user's jurisdiction
### Bug Description

#11509 has introduced logic that enables the delete button even if the user does not have edit rights for the respective entity. However, we now face the situation that the delete button is enabled even if the entity is outside the user's jurisdiction; in this case, a backend check would prevent the deletion and an error message is displayed (except for campaign forms and event actions).

### Reproduction Steps

1. Create user A that is assigned to a jurisdiction and has e.g. the delete cases right, but NOT the edit cases right
2. Create a case outside the user's jurisdiction with user B (e.g. an admin/national user)
3. Open the case with user A; the case should be read-only, but the delete button is enabled
4. Try to delete the case; an error message appears informing the user about the case not being in their jurisdiction

### Expected Behavior

The delete button should not be enabled if the user would not be able to delete the entity in the first place, i.e. we need to check whether it is in the user's jurisdiction or owned by the user on UI level and disable the button if it's not.

### Affected Area(s)

Web app

### SORMAS Version

1.82.0

### Android version/Browser

Any browser

### User Role(s)

e.g. Surveillance Supervisor

### Server URL

localhost

### Additional Information

_No response_

leventegal-she:
… jurisdiction

Fixes #11710

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/action/ActionController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/action/ActionController.java
@@ -39,9 +39,12 @@ public class ActionController {
 	/**
 	 * Show a create form for an action.
 	 *
-	 * @param context of the action
-	 * @param entityRef of the action
-	 * @param callback on save
+	 * @param context
+	 *            of the action
+	 * @param entityRef
+	 *            of the action
+	 * @param callback
+	 *            on save
 	 */
 	public void create(ActionContext context, ReferenceDto entityRef, Runnable callback) {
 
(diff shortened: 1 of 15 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11827

MateStrysewske (changes requested):
Bug is not fixed for me. I can still reproduce it at least with cases (haven't tried it with other entities).

- Create district user with case delete, but without case edit rights in jurisdiction A
- Create case with a different user in jurisdiction B
- Log in with the district user and open the case (e.g. by using the URL), delete button is still clickable while the rest of the form is read only

MateStrysewske (approved):

PR #11836, opened 2023-04-13
#11819 remove backend logic for setting the user's jurisdiction on ev…

### LINKED ISSUE #11819: District-level users can't change only the responsible region of an event participant
### Bug Description

When changing the region of an event participant as a user with an assigned district, and leaving the responsible district empty, a warning message appears that tells the user they will lose access to the event participant when performing the action. However, after confirming the change, the fielids are simply reset. Changing both the region and district works as expected.

### Reproduction Steps

1. Create an event with a national user, and create an event participant with a responsible district
2. Log in with a district-level user with the same assigned district as the event participant
3. Open the event participant and change the region, leave the district empty
4. Confirm that you want to change the jurisdiction even though you would lose access to it
5. The jurisdiction fields are reset

### Expected Behavior

The region and district should be changed according to the user input (only if they are not required of course, because otherwise the district in fact has to be filled in), and the event participant should no longer be editable for the user.

### Affected Area(s)

Web app

### SORMAS Version

SORMAS 1.83.0-SNAPSHOT

### Android version/Browser

Google Chrome

### User Role(s)

Surveillance Officer

### Server URL

localhost

### Additional Information

_No response_

leventegal-she:
ent participant because it should have been on the UI only but it's no longer needed

Fixes #11819

remove backend logic for setting the user's jurisdiction on event participant because it should have been on the UI only but it's no longer needed

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
@@ -85,9 +85,7 @@ import de.symeda.sormas.api.i18n.Strings;
 import de.symeda.sormas.api.i18n.Validations;
 import de.symeda.sormas.api.immunization.MeansOfImmunization;
 import de.symeda.sormas.api.importexport.ExportConfigurationDto;
-import de.symeda.sormas.api.infrastructure.district.DistrictReferenceDto;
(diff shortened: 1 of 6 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11836

MateStrysewske (approved):

PR #11837, opened 2023-04-13
#4959 - Enforce DTO fields marked as @Required to be present in the r…

### LINKED ISSUE #4959: Enforce DTO fields marked as @Required to be present in the request

sergiupacurariu:
…equest - fix error case creation with Point Of Entry origin

Fixes #4959

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseDataDto.java
@@ -334,7 +334,6 @@ public class CaseDataDto extends SormasToSormasShareableDto {
 	@SensitiveData(mandatoryField = true)
 	private FacilityType facilityType;
 	@Outbreaks
-	@NotNull(message = Validations.validFacility)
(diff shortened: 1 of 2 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11837

MateStrysewske (approved):

PR #11843, opened 2023-04-13
#11838 S2S - Cannot accept case in target system if Case has a sample

### LINKED ISSUE #11838: S2S - Cannot accept case in target system if Case has a sample
### Problem Description

S2S - Cannot accept case in target system if Case has a sample. An error message appears with 500 code error

### Reproduction Steps

1. Login as Admin+NatUser
2. Create a Case with a Sample and a pathogen test positive
3. Share the case with handover
4. Accept the case in the target system

### Expected Behavior

_No response_

### Affected Area(s)

Web app

### SORMAS Version

1.83.0-snapshot

### Android version/Browser

-

### User Role(s)

Admin+NatUser

### Server URL

_No response_

### Additional Information

[s2s500.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/11219883/s2s500.txt)

leventegal-she:
Fixes #11838

Removed the @NotNull annotation of reportingUser on the SampleDto because there was the only place this annotation was added and it is ensured by business logic that the field is always set. 

In s2s context all reporting users are set tu null before sending data and theyare reset on the target system when accepting the share request

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleDto.java
@@ -14,13 +14,12 @@
  */
 package de.symeda.sormas.api.sample;
 
-import de.symeda.sormas.api.feature.FeatureType;
-import de.symeda.sormas.api.utils.DependingOnFeatureType;
 import java.util.Date;
 import java.util.Set;
 
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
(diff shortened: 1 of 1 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11843

MateStrysewske:
@leventegal-she Please open another PR linked to #4959 (with the commit also linked to #4959) instead because that's the original cause of the problem.

PR #11880, opened 2023-04-20
#11327 - Show warning hint before restore popup

### LINKED ISSUE #11327: Introduce "Restore" bulk action [3]
### Problem Description

With #8465  the new grid for soft deleted entities was introduced for cases, contacts, events, event participants, immunizations, samples and travel entries. 

In this grid, bulk edit mode and bulk actions are possible. However, among the bulk actions there is currently 'Delete' but no 'Undo Deletion' bulk action, similar to how there is archive and de-archive.

### Proposed Change

Introduce the bulk action 'Undo deletion' in the grids that display soft deleted entities.

### Acceptance Criteria

- [x] An "Undo deletion" bulk action is shown in all directories specified above when filtering by deleted entities
- [x] Clicking the bulk action brings up a pop-up:
  - Caption: "Confirm undo deletion"
  - Content: "Are you sure you want to undo the deletion for all {selected_entry_count} selected {entity_name}s?"
  - Options: "No" and "Yes"
  - Should look similar to the pop-up that appears when e.g. bulk de-archiving cases
- [x] Grid is reloaded after confirming the action; entities for which the deletion has been undone no longer appear in the grid (since it should still be filtered by deleted entities)
- [x] Clicking the bulk action without selecting any entity brings up the usual warning message
- [x] Hide the "Delete" bulk action when filtering cases by deleted cases

### Additional Information
Make sure that the user is always allowed to undelete all the cases in the list (i.e. check whether there's any restriction in this regard for the archive/de-archive bulk action; if yes, we need to have something similar here, if no, we're probably fine).

MateStrysewske:
Fixes #11327

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/DeleteRestoreController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/utils/DeleteRestoreController.java
@@ -17,6 +17,16 @@ import de.symeda.sormas.api.utils.HtmlHelper;
 public class DeleteRestoreController<F extends DeletableFacade> {
 
 	public void restoreSelectedItems(List<String> entityUuids, F entityFacade, CoreEntityRestoreMessages messages, Runnable callback) {
+
+		if (entityUuids.isEmpty()) {
+			new Notification(
+				I18nProperties.getString(messages.getHeadingNoSelection()),
+				I18nProperties.getString(messages.getMessageNoSelection()),
+				Notification.Type.WARNING_MESSAGE,
+				false).show(Page.getCurrent());
+			return;
+		}
+
 		Label restoreConfirmationMessage = new Label();
 		restoreConfirmationMessage.setValue(
 			String.format(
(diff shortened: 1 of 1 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11880

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/DeleteRestoreController.java
```diff
@@ -37,47 +47,40 @@ public void restoreSelectedItems(List<String> entityUuids, F entityFacade, CoreE
...
+			} catch (Exception e) {
+				countNotRestoredEntities++;
+				notRestoredEntities.append(selectedRow, 0, 6).append(", ");
+			}
+		}
+		if (notRestoredEntities.length() > 0) {
+			notRestoredEntities = new StringBuilder(" " + notRestoredEntities.substring(0, notRestoredEntities.length() - 2) + ". ");
+		}
+		callback.run();
+		if (countNotRestoredEntities == 0) {
```
carina29, line 66: I think that would be easier to read this method with some refactorings.
- by extracting countNotRestoredEntities == 0 check into a method with a proper naming for ex: allTheEntitiesAreRestored(countNotRestoredEntities) which would return a boolean

- the code from the else branch could go into a separate method for ex: displayNotRestoredEntities(countNotRestoredEntities, notRestoredEntities)  or some similar naming

carina29 (changes requested):

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11880

PR #11904, opened 2023-04-27
#11903 - Display expected columns in read-only mode

### LINKED ISSUE #11903: Case/contact follow-up visit grid displays additional columns when in read-only mode
### Bug Description

When entering the follow-up view of a case or contact in read-only mode, instead of the usual columns, all columns that are part of the DTO are instead added to the grid, resulting in a much higher number of columns and some that should never be displayed in the frontend (e.g. whether or not the visit is pseudonymized or its database ID).

### Reproduction Steps

1. Copy the URL of a case that a user e.g. on district level does not have access to
2. Log in as that district user and enter the case via the URL
3. Open the follow-up view and check the columns in the grid

### Expected Behavior

Only the columns defined in the code should be displayed.

### Affected Area(s)

Web app

### SORMAS Version

1.83.0-SNAPSHOT

### Android version/Browser

Chrome

### User Role(s)

e.g. Surveillance Officer

### Server URL

localhost

### Additional Information

_No response_

MateStrysewske:
Fixes #11903

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/visit/VisitGrid.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/visit/VisitGrid.java
@@ -75,17 +75,15 @@ public class VisitGrid extends FilteredGrid<VisitIndexDto, VisitCriteria> {
 			return new Label(displayText);
 		}).setId(VisitIndexDto.ORIGIN);
 
-		if (isEditAllowed) {
-			setColumns(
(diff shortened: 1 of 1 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11904

sergiupacurariu (approved):

PR #11911, opened 2023-04-28
#11597 - Added a bulk progress layout

### LINKED ISSUE #11597: Add a progress bar to "Edit" bulk actions [5]
### Feature Description

#11442 has introduced batches for bulk editing in order to avoid running into timeouts. Right now, the user gets a popup every 200 processed entities or 3 minutes that asks them whether they want to continue the bulk operation. This is suboptimal UX because it requires them to interact with the app frequently, and while it tells the user how many entities are left to process when the popup appears, there's no further feedback about the process in between those popups.

### Proposed Change

Introduce a progress bar popup to bulk editing, similar to the progress bar we have for imports, that shows the user how many entities out of the amount of entities selected have been edited so far and allows them to cancel the procedure.

### Acceptance Criteria
- [x] In all directories that support the "Edit" bulk action, when using it with AT LEAST 40 entries selected, a progress screen is shown (see mockup 1 below) that includes the number of selected entries, processed entries so far, a progress bar that updates according to the ratio, and a button that allows the user to cancel the bulk action
- [x] When at least 40 entries are selected and the progress screen is shown, the respective backend save method is called in batches of 20 entries instead of passing it all selected entries at once
- [x] When a batch has been successfully processed, the progress bar and process entry count on the progress screen are updated accordingly and the save method is called with the next batch
- [x] Once all entries have been processed, the progress indicator icon is replaced with a checkmark, the text is updated and the "Cancel" button is replaced with a "Close" button (see mockup 2 below); clicking on "Close" closes the window; no success notification is needed afterwards, but the directory has to be reloaded
- [x] Clicking on the "Cancel" button at any time results in a popup being shown (see mockup 3 below) that allows the user to confirm the cancellation; this only happens after the current batch of entries have been saved (and the progress screen updated accordingly), so there might be a delay until the popup window is shown to the user
- [x] Aborting the cancellation closes the popup window and continues the bulk processing (i.e. the save method is called with the next batch)
- [x] Confirming the cancellation closes the popup window and the progress screen, reloads the directory and displays a warning notification: "The bulk action has been cancelled. <processed entry count> out of <selected entry count> have been processed. All unprocessed entries are still selected."
- [x] When the bulk action is cancelled, all entries that have not yet been processed still need to be selected (this is already implemented, i.e. verify that it's still working after the other changes in this issue have been done)
- [x] When using the "Edit" bulk action with LESS THAN 40 entries selected, the progress screen is not shown and instead all selected entries are passed to the backend for processing; a success notification is shown after the bulk action has been completed as usual
- [x] Any now unused functionality introduced in #10461 has been removed from the code (e.g. the popup that asks the user whether they want to continue the bulk processing, the entryand time limit, etc.); make sure to also remove no longer used strings from strings.properties

#### Bulk processing in progress
bulk_in_progress
#### Bulk processing completed
bulk_completed
#### Cancel confirmation
bulk_cancel

### Implementation Details

- Use the import progress layout as the basis for implementation; ideally, add a general ProgressLayout superclass that is reused by the import progress layouts and the new bulk progress layout
- The thresholds specified above should be stored in constants so that we can easily change them, i.e. `BULK_ACTION_BATCH_SIZE` = 20 and `BULK_ACTION_PROGRESS_THRESHOLD` = 40; add comments to these fields and explain what they do

MateStrysewske:
Fixes #11597


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11911

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/BulkOperationHandler.java
```diff
@@ -0,0 +1,168 @@
...
+			} else {
+				NotificationHelper.showNotification(I18nProperties.getString(Strings.messageEntriesEdited), Notification.Type.HUMANIZED_MESSAGE, -1);
+			}
+			bulkOperationDoneCallback.accept(Collections.emptyList());
+		} else {
+			UI currentUI = UI.getCurrent();
+
+			BulkProgressLayout bulkProgressLayout = new BulkProgressLayout(currentUI, selectedEntries.size(), this::handleCancelButtonClicked);
+			Window window = VaadinUiUtil.createPopupWindow();
+			window.setCaption(I18nProperties.getString(Strings.headingDataImport));
```
sergiupacurariu, line 63: the data import header should be replaced with something more suitable for Bulk Edit

MateStrysewske: Aah, great, I've already created that string, but forgot to use it :D

sergiupacurariu (approved):

PR #11923, opened 2023-05-03
Change 11922 remove unnecessarry validations

### LINKED ISSUE #11922: Remove unnecessary validation constraints introduced by 4959
### Problem Description

In #4959 we have replaced the (not properly working) Required annotation with a (properly working) NotNull annotation. Some external systems (e.g. survnet converter) now run into problems, because fields that were not required before now have to be filled.

Example: A case now needs to have a case classification and an investigation status. From an API point of view this seems very inconvenient, because both fields have clear default values ("not classified" and "pending").

### Proposed Change

Fields that should not have a NotNull annotation:

- [x] CaseDataDto.caseClassification (NOT_CLASSIFIED)
- [x] CaseDataDto.investigationStatus (PENDING)
- [x] ContactDto.multiDayContact (is boolean anyway)
- [x] ContactDto.contactClassification (UNCONFIRMED)
- [x] ContactDto.epiData (same as for case - when it does not exist it will be created)
- [x] ContactDto.healthConditions (same)
- [x] ImmunizationDto.immunizationStatus (PENDING)
- [x] ImmunizationDto.immunizationManagementStatus (SCHEDULED)
- [x] PathogenTestDto.labUser (just not needed)
- [x] Make sure all default values mentioned above are automatically set, when missing.
- [x] Same for CaseDataDto outcome and caseOrigin

### Added Value/Benefit

No unnecessary problems with external systems.

### Acceptance Criteria

* Case classification of an already classified case is not reset, when sending the case with a null value for the classification
* Same for the other enum fields listed in the proposed change
* When sending data without the fields listed above, no validation error is perceived and default values are applied where needed (e.g. no exceptions as we had reported in #4937 )

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

MartinWahnschaffe:
Fixes #11922


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11923

MartinWahnschaffe:
@MateStrysewske You can ignore the trivy warnings - they are not related to the changes in this PR.

MateStrysewske (approved):

PR #11939, opened 2023-05-08
Change 6699 java 11

### LINKED ISSUE #6699: Update to Java 11 (?)
### Feature Description
Once #6686 is done we finally have Java 8 support for the SORMAS API project. 

### Problem Description
With Java 11 there is a newer LTS version available that brings some benefits (see https://medium.com/testvagrant/migrate-java-8-to-java-11-12-9ab082f335e8 - the Oracle part is not relevant for us, since we are relying on OpenJDK anyway).

With the upgrade to Android Gradle Plugin 7+, we now also have support for Java 11 in Android:
https://developer.android.com/studio/releases/gradle-plugin#java-11

This roadmap provided by azul (zulu) is a bit confusing, though: https://www.azul.com/products/azul-support-roadmap/
It says the support for Java 8 will be longer for Java 11.

### Proposed Change

- [ ] clarify whether the azul roadmap is a reason to stay with Java 8
- [ ] update the used java version of all projects to 11
- [ ] test Java 11 for Android. I had a try with it and was able to use Java 11 API (String.isBlank) in an SORMAS API method that was then used by the Android app and ran on a version 24 emulator. I was not able to directly use String.isBlank in Android, though. It should be clarified why this is the case. There is currently no documentation available for this - maybe the release of Android Studio 2021 bumblebee will clarify this.

MartinWahnschaffe:
Fixes #6699


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11939

StefanKock (approved):

PR #11960, opened 2023-05-15
#11818 - add handling of the inexistent uuid response

### LINKED ISSUE #11818: Add ReST endpoint to provide effective user rights of a user [1]
### Feature Description

Add a ReST endpoint to provide the user rights of a user based on the user's uuid.

### Added Value/Benefit

A client can easily find out which user rights a user (especially the current one) has without having to check and retrieve all user roles.

### Acceptance Criteria

- [x] Calling the endpoint returns the effective user rights of a single user based on its UUID
- [x] When no UUID is passed, the user rights of the current user are returned
- [x] The endpoint returns a `List<UserRight>`
- [x] User rights are checked (ideally use existing backend methods).
- [x] A unit test covers the backend method logic

### Implementation Details

Suggestion: Add UserResource.getUserRights with url users/rights/{uuid}. 

### Mockups

_No response_

### Additional Information

I've also added the full UUID of the user to the user edit form, as well as the UUID of the currently logged in user to the user settings popup.

carina29:
Fixes #11818


carina29:
@Martin I did not created a separate Exception in sormas -api for EntityNotFoundException ( as we have for example for AccessDeniedException in de.symeda.sormas.api.utils) I've used the one from javax.persistence but if you want I can create one in sormas-api utils to be consistent.

Another problem I've found is not related to this ticket but is affecting the HTTP response:
image

Right now only the response for 200 is working correctly which retrieves the list of rights.
For all the other cases ( AccessDenied and EntityNotFoundException) instead of returning the correct HTTP response according to the AccessDeniedExceptionMapper and EntityNotFoundExceptionMapper, the exceptions are caught by 
CatchAllExceptionMapper all the time and 500 Internal server will be returned due to the CatchAllExceptionMapper.

The exceptions (AccessDenied, EntityNotFound) thrown from UserFacadeEjb - > getUserRights will be wrapped in a TransactionRolledBack Exception and will not be anymore an instance of AccessDeniedException, EntityNotFoundException but in the returned log we can identify the correct, initial cause of the exception with the forbidden/not found message. 

image

image

I've investigated a bit this problem and one possibility would have had been to use a PersistenceExceptionMapper and check the instance of the getCause(): 
https://www.tabnine.com/code/java/methods/javax.persistence.PersistenceException/getCause
https://stackoverflow.com/questions/3293599/jax-rs-using-exception-mappers

In the way we have now our implementation I've tried and the getCause will not work because we the TransactionRolledBack exception is mapped earlier and the initial cause is kind of list.
This problem is present for all the REST endpoints we have when an exception is thrown.

This would mean that we should catch earlier the initial exception in order to know the instance of the causing exception in the Exception Mappers or change the way we throw the exceptions. 

But probably this part related to the HTTP responses and the ExceptionMappers should be treated in a bug ticket created for 10251 since the ExceptionMappers were introduced here: https://github.com/hzi-braunschweig/SORMAS-Project/pull/11885/files

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11960

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11960

PR #11994, opened 2023-05-24
#11980 Vaccination delete without edit rights prevents the Vaccinatio…

### LINKED ISSUE #11980: Vaccination delete without edit rights prevents the Vaccination status update
### Bug Description

Given the following situation:
 - a user having the following rights:
    - cases: Delete right - Yes, Edit right - No
    - immunization and vaccination - Delete right - Yes, Edit rights - no
- feature configuration - Immunization module = 'reduced'

In case a user wants to delete a vaccination that could impact the vaccination status of the case then if the user wants to update the status it will be restricted. This happens because in order to update the vaccination status user must have Case Edit Rights.

### Reproduction Steps
1. Activate "IMMUNIZATION_MANAGEMENT" feature and set Reduced on true;
2. Create a user with the following rights: CASE: View and Delete rights; IMMUNIZATION: View and Delete rights;
3.  With and Admin+NatUser - create a case (or contact/event participant) who has a relevant vaccination;
4. With the user from step 2 open the vaccination and click to delete it;
5. In the pop-up window "Vaccination status update" select "Yes".
Actual behavior:
The vaccination is deleted in the process but the Case vaccination status cannot be updated when the user selects "Yes".

image

Logs from the local machine:
[vaccination logs.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/11544085/vaccination.logs.txt)

### Expected Behavior
The vaccination status should get updated in this particular flow even though the user does not have edit rights over the case/contact/event participant - (the solution would probably again be to have some kind of a dedicated method that only updates the vaccination status as discussed with @MateStrysewske).

### Affected Area(s)

Web app

### SORMAS Version

SORMAS 1.84.0

### Android version/Browser

Google Chrome

### User Role(s)

Admin+NatUser; Configured user that has the rights described in the steps to reproduce

### Server URL

_No response_

### Additional Information
In the server logs, regardless of whether the user tries to delete a vaccination associated with a case, or contact or event participant, it is always displayed that the missing rights are CASE_CREATE and CASE_EDIT.

leventegal-she:
…n status update

Fixes #11980

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseFacade.java
@@ -89,6 +89,8 @@ public interface CaseFacade extends CoreFacade<CaseDataDto, CaseIndexDto, CaseRe
 
 	CaseDataDto updateFollowUpComment(@Valid @NotNull CaseDataDto dto) throws ValidationRuntimeException;
 
+	void updateVaccinationStatus(CaseReferenceDto caseRef, VaccinationStatus status);
+
 	CaseDataDto save(@Valid @NotNull CaseDataDto dto, boolean systemSave) throws ValidationRuntimeException;
 
 	CoreAndPersonDto<CaseDataDto> save(@Valid @NotNull CoreAndPersonDto<CaseDataDto> dto) throws ValidationRuntimeException;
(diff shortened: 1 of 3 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=11994

MateStrysewske (approved):

PR #12038, opened 2023-06-05
#11518 - Enable the cards on the person page based on the rights

### LINKED ISSUE #11518: Enable the cards on the person page for users that have view rights/edit rights [2]
### Problem Description

The cards on the person page are not enabled for users that have view rights, and in the case of the 'Samples' card, it is not enabled when the user has even edit rights over samples. Two scenarios can therefore be observed:

I: A user that has viewing rights over all entities in the app, and all of the rights associated with samples, the 'Samples' card on the person page is read only, as opposed to the sample cards on the other entities where the card is editable.

- on the person pages (from the 'Persons' directory): 
image

- on the other entities (cases, contacts and event participants): 
image

II: A user that has only view rights over all entities in SORMAS, on the person pages, the buttons to 'See ... for this person' are not clickable.

### Proposed Change

I: The samples card should behave on the person page the same way it behaves on cases/contacts/event participants - it should be editable as long as the user has edit rights on the samples.

II: For a user that has view rights over the respective entities the buttons to 'See ... for this person' should be enabled.

### Added Value/Benefit
- easy navigation;
- users can action the rights they have from every flow that they have access to.

### Acceptance Criteria

I:
- [x] the user that has edit rights over an entity (ex: case, contact, event participant, samples, vaccination/immunization, travel entry) should see the card of the entity on the person page as editable ( pencil icon should also appear);
- [x] the button 'See samples for this person' should be clickable;

II:
- [x] if the user has view rights over the respective entity (such as 'View existing cases' + 'View existing persons') the buttons 'See ... for this person' should be enabled on the respective entity card;
- [x] when the respective entities are not within the user's jurisdiction the logic should remain as it currently is: the user can click the buttons, they are redirected to the respective grid filtered based on the person and based on whether the entities are owned or in the user's jurisdiction;
- [x] the buttons should also be enabled for persons whose associated entities are only 'To view' after sharing via S2S as long as the user has the view rights;
- [x] buttons to be covered are: 
- See cases for this person
- See contacts for this person
- See samples for this person
- See events for this person

carina29:
Fixes #11518

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CasePersonView.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CasePersonView.java
@@ -60,7 +60,7 @@ public class CasePersonView extends AbstractCaseView implements PersonSideCompon
 		DetailSubComponentWrapper componentWrapper = addComponentWrapper(editComponent);
 		CustomLayout layout = addPageLayout(componentWrapper, editComponent);
 		setSubComponent(componentWrapper);
-		addSideComponents(layout, CoreEntityType.CASE, caseData.getUuid(), person.toReference(), this::showUnsavedChangesPopup);
+		addSideComponents(layout, CoreEntityType.CASE, caseData.getUuid(), person.toReference(), this::showUnsavedChangesPopup, isEditAllowed());
 		setEditPermission(
 			editComponent,
 			UserProvider.getCurrent().hasUserRight(UserRight.PERSON_EDIT),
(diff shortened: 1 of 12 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12038

carina29:
PersonDataView and PersonSideComponentsElement classes were refactored to have similar behaviour with CaseDataView, EventDataView and other similar entities with SideComponents.

isEditAllowed was added to the list components of each entity.

MateStrysewske (approved):

PR #12044, opened 2023-06-06
#12043 - add missing check for TRAVEL_ENTRY_EDIT

### LINKED ISSUE #12043: Pencil icon appears for not editable Travel entry
### Bug Description

Pencil icon appears for not editable Travel entry instead of eye icon

### Reproduction Steps

1. Log in with a user without having "Edit existing travel entries" right
2. Go to Persons page and check the Travel entries card: the "pencil" icon is appearing instead of the "eye" icon

Additional note: the travel entry is readonly after clicking the pencil icon ( as it should be)

### Expected Behavior

The pencil icon should not appear if the user does not have "Edit travel entries" right.

### Affected Area(s)

Web app

### SORMAS Version

1.84

### Android version/Browser

Chrome

### User Role(s)

Configured user

### Server URL

_No response_

### Additional Information

The problem is that the pencil icon is displayed based on the isEditAllowed flag and should be displayed only if isEditAllowed is true and the user also have "Edit travel entries" right.

carina29:
Fixes #12043


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12044

sergiupacurariu (approved):

PR #12064, opened 2023-06-09
#12008 - fix link event group problems

### LINKED ISSUE #12008: [Event groups] Clicking 'Link event group' leads to an error when the user does not have the 'Create new event group' right
### Bug Description

When a user does not have the right 'Create new event groups' and they try to link an event to a group, an error appears.

### Reproduction Steps

1. Configure a user that has all the rights except the 'Create new event groups';
2. On an existing event, click on the 'Link event group' button.
Actual behavior: An error appears.
image

Logs from the local machine:
[logs create new event groups.txt](https://github.com/SORMAS-Foundation/SORMAS-Project/files/11591769/logs.create.new.event.groups.txt)

### Expected Behavior

The user should be able to successfully link the existing event groups to the event. The missing right should only prevent them from creating new event groups. 

### Affected Area(s)

Web app

### SORMAS Version

1.85.0; 1.84.0; 1.83.0

### Android version/Browser

Chrome

### User Role(s)

Configured user that does not have the 'Create new event groups' right

### Server URL

local machine; https://release-international.sormas.netzlink.com/

### Additional Information

_No response_

carina29:
Fixes #12008


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12064

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12064

MateStrysewske (changes requested):
Even though it's currently technically correct (because there's no way to create an event group in the web app without also linking an event to it), I don't really like the dependency that users need to have the event group link right for the event group create right; as soon as we e.g. invent a "Create event group" button that creates an empty event group, this part of the logic will break. Semantically, the dependency doesn't make much sense to me.

MateStrysewske:
@carina29 We can go with the new implementation, however, that means we also need to make some migration because the changes would only ever be applied to new user roles.

That means we have to:
- Give all default user roles that have the user right to create event groups, but not the one to link them, the user right to link event groups (if any exist)
- Write an SQL migration script that applies this change to ALL user roles on the given system; i.e. give the USER_RIGHT_LINK user right to all user roles that have the USER_RIGHT_CREATE user right. We have examples of how to do this in sormas_schema.sql
- Add a hint to the release notes Confluence page that this has been done because we need to add it to the release notes

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12064

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12064

MateStrysewske (approved):

PR #12091, opened 2023-06-16
#11931 - extract deleteAllSelected items from CaseController to Delet…

### LINKED ISSUE #11931: Add bulk action progress window to "Delete" and "Restore" bulk actions [2]
### Feature Description

https://github.com/hzi-braunschweig/SORMAS-Project/issues/11597 has introduced a progress window for the "Edit" bulk actions. This window needs to be applied to other bulk actions as well.

Add the progress window to the "Delete" and "Restore" bulk actions of all directories. In preparation, move all "deleteSelected" methods from the individual controllers to the DeleteRestoreController introduced in #11327.

- [x] Extract deleteAllSelectedItems method from CaseController to DeleteRestoreController and do the necessary in order to be similar with the restoreSelectedCases method from DeleteRestoreController

- [x] Reuse the deleteAllSelectedItems from DeleteRestoreController for ContactController
- [x] Reuse the deleteAllSelectedItems from DeleteRestoreController for EventController and do the necessary adjustments
- [x] Reuse the deleteAllSelectedItems from DeleteRestoreController for SampleController
- [x] Reuse the deleteAllSelectedItems from DeleteRestoreController for EventParticipantController
- [x] Reuse the deleteAllSelectedItems from DeleteRestoreController for ImmunizationController
- [x] Create a new Controller for permanent deletion (PermanentDeleteController) and a new interface (PermanentlyDeletableFacade)
- [x] Reuse the deleteAllSelectedItems from PermanentDeleteController for VisitController
- [x] Reuse the deleteAllSelectedItems from DeleteRestoreController for TravelEntryController
- [x] Reuse the deleteAllSelectedItems from DeleteRestoreController for TaskController
- [x] Reuse the deleteAllSelectedItems from DeleteRestoreController for ExternalMessageController

### Added Value/Benefit

Consistency with the other bulk actions and increased usability because users will know how many entries have already been edited and still have to be edited, and can cancel the process at any time.

### Acceptance Criteria

- [x] All bulk methods used to delete multiple items at the same time have been moved from individual controllers to the DeleteRestoreController
- [x] All "Delete" and "Restore" bulk actions in all directories use the bulk action progress window introduced in #11597 when at least 40 entries are selected
- [x] The respective backend calls are done in batches of 20 entities when at least 40 entries are selected
- [x] The bulk action also works in the case detailed and contact detailed directories, which use different grids
- [x] Cancelling the action before it finishes leaves the selected entries that have not yet been processed still selected (as implemented in #11597)

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

carina29:
…eRestoreController, add missing messageEntitiesNotRestored

Fixes #11931


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

carina29:
@MateStrysewske the  doBulkOperation method will be better structured after implementing https://github.com/SORMAS-Foundation/SORMAS-Project/issues/12257.

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-api/src/main/resources/strings.properties
```diff
@@ -751,7 +760,7 @@ headingEventGroupUnlinkEventIssue = Issue while unlinking events to event group
 headingExportUserRightsFailed = Export user rights failed
 headingExternalMessageDownload = Download message
 headingNoExternalMessagesSelected = No messages selected
-headingExternalMessagesDeleteProcessed = Delete processed message(s)
+headingExternalMessagesEligibleForDeletion = Delete processed message(s)
```
MateStrysewske, line 763: What's the reason for this renaming?

sormas-api/src/main/resources/strings.properties
```diff
@@ -991,7 +1000,10 @@ infoNoAccessToPersonEntities = You don't have the necessary user rights to view
 infoDashboardFinalLaboratoryResult = When a case has multiple samples, only the final laboratory result of the sample with the latest sample collection date is considered.
 infoBulkProcess = %d selected entries are currently being processed. This may take a while. The progress will be updated for each 20 entries that have been processed.
 infoBulkProcessFinished = <b>All selected entries have been processed!</b><br/>You can now close this window.
-infoBulkProcessFinishedWithSkips = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, most likely because they were outside your jurisdiction or already archived.
+infoBulkProcessFinishedWithIneligibleItems = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, because they were not eligible.
```
MateStrysewske, line 1003: I find the word "eligible" to be very vague in this context. Can't we be more precise and tell the user potential reasons why some of the entries could not be processed (as we do in e.g. the ...FinishedWithSkips string)? What exactly does "not eligible" mean in this context?

sormas-api/src/main/resources/strings.properties
```diff
@@ -991,7 +1000,10 @@ infoNoAccessToPersonEntities = You don't have the necessary user rights to view
...
+infoBulkDeleteProcessFinishedWithSkips = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, most likely because they were outside your jurisdiction.
```
MateStrysewske, line 1004: What's the purpose of this string? It's called "...DeleteProcess...", but there's no reference to deletion in the string; instead, it's basically the next string minus the eligible keyword.

sormas-api/src/main/resources/strings.properties
```diff
@@ -991,7 +1000,10 @@ infoNoAccessToPersonEntities = You don't have the necessary user rights to view
...
+infoBulkProcessFinishedWithSkips = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, most likely because they were outside your jurisdiction, already archived or not eligible.
+infoBulkProcessNoEligibleEntries = <b>Bulk process has been cancelled because there are no eligible entries for this operation!
```
MateStrysewske, line 1006: See above

sormas-api/src/main/resources/strings.properties
```diff
@@ -1097,27 +1119,35 @@ messageEventParticipantDearchived = Event participant has been de-archived
...
-messageCountEventsNotDeleted = %s events not deleted. Event Ids not deleted: %s
-messageEventsNotDeletedReason = No events that contain event participants linked will be deleted. Please remove the events participants from the event in order to be able to delete it.
-messageCountEventsNotDeletedExternalSurveillanceTool=%s events not deleted due to the reporting tool communication. Event Ids not deleted: %s
-messageEventsNotDeletedReasonExternalSurveillanceTool=Some events were not deleted because the communication with the reporting tool failed.
+messageCountVisitsNotDeleted = %s visits not deleted. UUIDs of visits not deleted: %s
+messageCountEventsNotDeleted = %s events not deleted. UUIDs of events not deleted: %s
+messageEventsNotDeletedExternalReason = The events mentioned in the above list of UUIDs could not be deleted because the communication with the reporting tool failed.
+messageEventsNotDeletedLinkedEntitiesReason = No events that contain event participants linked will be deleted. Please remove the events participants from the event in order to be able to delete it.
```
MateStrysewske, line 1150:
```suggestion
messageEventsNotDeletedLinkedEntitiesReason = No events that contain event participants linked will be deleted. Please remove the event participants from the event in order to be able to delete it.
```

sormas-backend/src/main/java/de/symeda/sormas/backend/contact/ContactFacadeEjb.java
```diff
@@ -555,12 +555,44 @@ public void delete(String contactUuid, DeletionDetails deletionDetails) {
...
 
+	@Override
+	@RightsAllowed(UserRight._CONTACT_DELETE)
+	public void restore(List<String> uuids) {
+		restoreContacts(uuids);
+	}
+
+	@Override
+	@RightsAllowed(UserRight._CONTACT_DELETE)
+	public List<String> restoreContacts(List<String> uuids) {
```
MateStrysewske, line 578: Since this implementation seems to be the same for all classes, please move this into a default implementation of the restore method in the `DeletableFacade`, and make the logged error more generic (i.e. replace the concrete entity name with "entity"; the entity in question will be pretty clear from the stack trace anyway)

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
```diff
@@ -513,12 +509,66 @@ public void delete(String uuid, DeletionDetails deletionDetails) throws External
...
 
+	@Override
+	@RightsAllowed(UserRight._EVENTPARTICIPANT_DELETE)
+	public void delete(List<String> uuids, DeletionDetails deletionDetails) {
+		deleteEventParticipants(uuids, deletionDetails);
+	}
+
+	@Override
+	@RightsAllowed(UserRight._EVENTPARTICIPANT_DELETE)
+	public List<String> deleteEventParticipants(List<String> uuids, DeletionDetails deletionDetails) {
```
MateStrysewske, line 520: Probably makes sense to also add a default implementation for the delete method because we have the same implementation at least for some entities here as well, i.e. this would reduce code duplication

sormas-backend/src/main/java/de/symeda/sormas/backend/task/TaskFacadeEjb.java
```diff
@@ -914,19 +914,34 @@ public TaskDto getByUuid(String uuid) {
 
 	@Override
 	@RightsAllowed(UserRight._TASK_DELETE)
-	public void deleteTask(TaskDto taskDto) {
-		Task task = taskService.getByUuid(taskDto.getUuid());
+	public void delete(String uuid) {
+		if (!userService.hasRight(UserRight.TASK_DELETE)) {
```
MateStrysewske, line 918: Why is this needed? Shouldn't the `@RightsAllowed` annotation already handle this?

sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventController.java
```diff
@@ -1006,95 +1006,41 @@ public EventDto createNewEvent(Disease disease) {
...
+		ControllerProvider.getDeleteRestoreController()
+			.deleteAllSelectedItems(
+				selectedRows,
+				getEligibleEvents(selectedRows),
+				getIneligibleEvents(selectedRows),
+				DeleteRestoreHandlers.forEvent(),
+				bulkOperationCallback(eventGrid, null));
+	}
+
+	public Collection<EventIndexDto> getEligibleEvents(Collection<EventIndexDto> selectedRows) {
```
MateStrysewske, line 1020: Please remove this and the `getIneligibleEvents` method and instead build those lists directly in `deleteAllSelectedItems`, because this produces performance overhead (e.g. the same method gets called twice unnecessarily). I.e. build the list of ineligible events there (using a stream instead of a for loop), store it in a variable and use the stored list to build the list of eligible events.

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/BulkOperationHandler.java
```diff
@@ -38,78 +41,158 @@ public class BulkOperationHandler<T extends HasUuid> {
...
+		initialEligibleEntryCount = getInitialEligibleEntryCount(selectedEntries, selectedIneligibleEntries, selectedEligibleEntries);
+		selectedEligibleEntries = !areIneligibleEntriesSelected(selectedIneligibleEntries) ? selectedEntries : selectedEligibleEntries;
+
 		if (selectedEntries.size() < BULK_ACTION_PROGRESS_THRESHOLD) {
-			successfulEntryCount = bulkOperationFunction.apply(selectedEntries);
-			if (initialEntryCount > successfulEntryCount) {
+			if (initialEligibleEntryCount > 0) {
+				if (areIneligibleEntriesSelected(selectedIneligibleEntries)) {
+					successfulEntryCount =
+						areIneligibleEntriesSelected(selectedIneligibleEntries) ? bulkOperationFunction.apply(selectedEligibleEntries) : 0;
```
MateStrysewske, line 99: Calling `areIneligibleEntriesSelected` seems to be redundant here because it's already done in the if condition

sormas-ui/src/main/java/de/symeda/sormas/ui/utils/BulkOperationHandler.java
```diff
@@ -38,78 +41,158 @@ public class BulkOperationHandler<T extends HasUuid> {
...
-				NotificationHelper
-					.showNotification(I18nProperties.getString(allEntriesProcessedMessageProperty), Notification.Type.HUMANIZED_MESSAGE, -1);
+				if (areIneligibleEntriesSelected(selectedIneligibleEntries)) {
+					String description = getErrorDescription(
+						selectedIneligibleEntries.stream().map(HasUuid::getUuid).collect(Collectors.toList()),
+						I18nProperties.getString(countEntriesNotDeletedMessageProperty),
+						I18nProperties.getString(ineligibleEntriesNotProcessedMessageProperty));
+
+					Window response =
+						VaadinUiUtil.showSimplePopupWindow(I18nProperties.getString(headingSomeEntitiesNotDeleted), description, ContentMode.HTML);
```
MateStrysewske, line 125: See above; don't use specific strings

MateStrysewske (changes requested):
What's the point of the explicit "deleteCases", "restoreCases", "deleteContacts", "restoreContacts", etc. methods when we have the general "delete" and "restore" methods inherited by the DeletableFacade interface? I feel like we should get rid of all these explicit methods and instead move the code into the inherited methods.

sormas-api/src/main/resources/strings.properties
carina29: The exact reason  ( for < 40 entities) in bulk progress bar in bulk progress bar why some entities will not be processed we've agreed to be handled in https://github.com/SORMAS-Foundation/SORMAS-Project/issues/12257.

@adinaflorea9 agreed that for this ticket for the bulk progress bar's content would be ok a general message if the deletion is stopped for these items - without the specific reasons of the problem since this will be handled in https://github.com/SORMAS-Foundation/SORMAS-Project/issues/12257 but if you want to change this, please let me know.  For less than 40 selected items also appears that the events with event participants can not be deleted.
I'll explain below also what scenarios we have till now.

/**********************************************************************************************************************/
Right now this message with the ineligible items looks like this:
image. These 'ineligible' item will be prefiltered since because of some specific reason for entity - bulk action should not be processed at all. 

I've used eligible in order to be able to use this message for all the entities and bulk processes where will be necessary ( I do not know for archive/dearchive/edit and other bulk processes what will be considered ineligible - I still have to check them here https://github.com/SORMAS-Foundation/SORMAS-Project/issues/12257 when I'll work on the ticket but for Delete ( Soft and Permanent) and Restore the following ineligible scenarios we'll have:

1)If the selected events contains events with event participants the items with event participants should not be deleted and a message should appear to the user ( as it was also in the past before the bulk progress bar).
image   -> This part is implemented in this ticket for < 40 selected entities the reason of ineligible is also shown, > 40 selected entities only the fact there are ineligible items will be shown and those will not be deleted

2) In case of external messages the ineligible items would be the messages with "PROCESSED" status.
In the past we had a functionality which I've preserved in this ticket too:
If we select a number of external messages and at least 1 is PROCESSED ( ineligible for deletion), a message will be displayed to the user and the whole deletion ( for all the entities selected) will be rolledback.
image
image

image


In https://github.com/SORMAS-Foundation/SORMAS-Project/issues/12257 I also thought to prefilter the cases/events shared with survnet if there is no communication with Survnet in order to show the entities which due to Survnet problems were not processed.

Other ineligible scenarios can appear from other bulk actions ( archive/dearchive/edit/ other bulk actions), I still have to check all the possibilites.
For each controller and bulk process we would know what would mean ineligible  for instance:
-  events with event partcipants can not be deleted
-   cases without archive rights can not be archived ( for this we have a future ticket to not show the archive at all but is not solved yet and is also blocking the bulk progress) and so on.

Please let me know how would you want to proceed after you read this comment :)

sormas-api/src/main/resources/strings.properties
carina29: I've thought that it would be more consistent if the heading, the message and in the code appears the same naming with eligible.

Now (after my changes) we have like this in the notification:
headingExternalMessagesEligibleForDeletion = Delete processed message(s)
messageExternalMessagesEligibleForDeletion = Only unprocessed messages can be deleted
image


And we display the notification from here:
image
image

If you want, I can change back the namings to headingExternalMessagesDeleteProcessed and 
messageExternalMessagesDeleteProcessed.

sormas-api/src/main/resources/strings.properties
carina29: infoBulkProcessFinishedWithIneligibleItems will appear for all the bulk processes which will be ineligible ( as I've mentioned in one comment from above ex: for now the archive without archive rights (till this will not be solved by hiding the button, the delete of events with event participants and any other scenario which I still have to check for other bulk processes). It will be a general message for all bulk processes. 
image

infoBulkDeleteProcessFinishedWithSkips = will be used only for delete because in a call @adinaflorea9 observed that the message contains also the archive and for delete is not correct to displayed that the reason for unprocessed entities is "outside your jurisdiction or already archived" because archived entities can also be deleted so for the delete I've used a message without the "archiving" part and we also know that is not the "not eligible" problem because I've handled in other condition that one.
This specific message ( only for delete) will be passed to the DeleteRestoreController ( for soft delete) and PermanentDeleteController ( for permanent delete) 

image

For all the other bulk processes, we will use the general message:
infoBulkProcessFinishedWithSkips = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, most likely because they were outside your jurisdiction, already archived or not eligible.

For Edit:
image

For Archive/Dearchive:
image

Should I change it? If yes, how would you want to be changed?

sormas-api/src/main/resources/strings.properties
carina29: A part of the messages, I've explained above, related to the infoBulkProcessNoEligibleEntries string:
Is used here (when > 40 entitiesare selected but none of the selected entities are eligible):
image

It looks like this now (the reason why is not eligible can be handled here https://github.com/SORMAS-Foundation/SORMAS-Project/issues/12257):
image

sormas-api/src/main/resources/strings.properties
carina29: Fixed

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-backend/src/main/java/de/symeda/sormas/backend/task/TaskFacadeEjb.java
carina29: It was like this in the past ( on development): 
image

And when the user does not have the TASK_DELETE right the delete will still appear in the bulk dropdown.
The annotation helps the interceptor to catch the problem and will be visible in the server.log but the user will not be notified, the task will remain selected and nothing will happen after clicking the delete.  
image
image

With the added exception in the same scenario the user will be notified about what happened:
image


I saw that we solved like this the problem on short term in case of other methods too:
image
image

The long term solution would be to implement this ticket and do not display at all the button/ menu item when the user does not have the proper right:
https://github.com/SORMAS-Foundation/SORMAS-Project/issues/12015

carina29:
> What's the point of the explicit "deleteCases", "restoreCases", "deleteContacts", "restoreContacts", etc. methods when we have the general "delete" and "restore" methods inherited by the DeletableFacade interface? I feel like we should get rid of all these explicit methods and instead move the code into the inherited methods.

@MateStrysewske I also had the same idea and I wanted at first to remove all the deleteEntities((List<String> contactUuids, DeletionDetails deletionDetails) kind of methods but I've found out that we can propagate the exceptions with the help of these:

For ContactsFacade for example:
When we want to delete multiple contacts ( bulk mode) we'll call this method:
image

Which calls the deleteContacts which will not return a void type but the actual List of the deleted entities ( without exceptions):
image

In this way, after the calling of the apply method from BulkHandler , the successFulEntryCount will be the correct number: the 
actual deletedContactUuids.
image

I've debugged a lot this problem and without this actual list of the successfully deleted entities, the successfulEntryCount were all the time the number of the selected entities even if not all the entities were successfully processed.
For < 40 entities is working fine for > 40 entities I still have to catch some exceptions and handle them in order to not remain the Progress bar blocked. This second part should be done here : https://github.com/SORMAS-Foundation/SORMAS-Project/issues/12257

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-api/src/main/resources/strings.properties
MateStrysewske: I think I would just remove the heading from that popup and simply display the error message; the heading doesn't really provide anything meaningful anyway

sormas-api/src/main/resources/strings.properties
MateStrysewske: I'm fine with leaving it as it is :)

sormas-api/src/main/resources/strings.properties
MateStrysewske: I think I would just call the property infoBulkProcessFinishedWithSkips, simply because there's no deletion reference in its text, even if we (right now) exclusively use it in the deletion process

sormas-api/src/main/resources/strings.properties
MateStrysewske: Fine by me :)

MateStrysewske:
@carina29 Wouldn't it then be an option to change the return type of the interface methods to List<String> instead?

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-api/src/main/resources/strings.properties
carina29: It's not visible from the above PR section, but we already have the infoBulkProcessFinishedWithSkips.
This will be used for other bulk processes ( not delete) because contains the "archive part" which would be not true for deletion.

image

So right now we have the following:

- **infoBulkProcessFinishedWithSkips** = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, most likely because they were outside your jurisdiction, already archived or not eligible.
- **infoBulkDeleteProcessFinishedWithSkips** = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, most likely because they were outside your jurisdiction.

How would you want to rename the infoBulkDeleteProcessFinishedWithSkips? (since infoBulkProcessFinishedWithSkips is already taken :D)

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-api/src/main/resources/strings.properties
MateStrysewske: I'd go with `infoBulkProcessFinishedWithSkipsOutsideJurisdiction`

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventParticipantFacadeEjb.java
carina29: Same as above

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

sormas-api/src/main/resources/strings.properties
```diff
@@ -991,7 +999,10 @@ infoNoAccessToPersonEntities = You don't have the necessary user rights to view
...
 infoBulkProcessFinished = <b>All selected entries have been processed!</b><br/>You can now close this window.
-infoBulkProcessFinishedWithSkips = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, most likely because they were outside your jurisdiction or already archived.
+infoBulkProcessFinishedWithIneligibleItems = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, because they were not eligible.
+infoBulkProcessFinishedWithSkipsOutsideJurisdiction = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, most likely because they were outside your jurisdiction.
+infoBulkProcessFinishedWithSkips = <b>Bulk process has been successfully completed!</b><br/>However, some of the selected entries could not be processed, most likely because they were outside your jurisdiction, already archived or not eligible.
+infoBulkProcessNoEligibleEntries = <b>Bulk process has been cancelled because there are no eligible entries for this operation!
```
MateStrysewske, line 1005: Bold tag is not closed

MateStrysewske (changes requested):
Just one minor thing. You can merge the pull request after fixing it. :)

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12091

PR #12100, opened 2023-06-20
#11946: Allow users to link/unlink event groups to events they own

### LINKED ISSUE #11946: Users can not link/unlink event groups to events owned by them
### Bug Description

For a user restricted by responsible jurisdiction, if they own an event outside their jurisdiction they can not link/un-link any event group to it.

### Reproduction Steps

I:

1. With a user limited by a responsible jurisdiction, create a new event that has the responsible jurisdiction in another region and district;
2. From the 'Event groups' card on the event page, click on 'Link event group'.

Actual behavior: A red warning message appears informing the user that they can not edit the event as it is outside their jurisdiction.
image

II:
1. With a user limited by a responsible jurisdiction create an event group;
2. With another users (Admin+NatUser) link multiple events to that event group (from different jurisdictions);
3. Log back with the user used at step 1, open the event group and click on the 'Link event' button;
4. From the list of events pick an event that is owned by the user, but that is outside their responsible jurisdiction.

Actual behavior:
An error appears and the user can not link the event.
image

Logs from the local machine that appear when executing scenario II:
[Event group error.txt](https://github.com/hzi-braunschweig/SORMAS-Project/files/11432227/Event.group.error.txt)

### Expected Behavior

The user should be able to link event groups to events owned by them.

### Affected Area(s)

Web app

### SORMAS Version

1.84.0; 1.82.0

### Android version/Browser

Chrome

### User Role(s)

Admin+NatUser; Configured user that is restricted by responsible jurisdiction

### Server URL

local machine; release-international

### Additional Information

If the actual behavior noted in the ticket is actually the expected one and was implemented this way on purpose, then at least the warning message that appears should be updated and changed from the red message that generally appears when an error is thrown or when a mandatory field is not filled, to a regular yellow informative note message.

ChristopherRiedel:
Fixes #11946


carina29:
I've tested the feature branch and on my side and the first problem related to the linking of the eventgroups when the owner has a different jurisdiction is still reproducible on this feature branch:
image

The logs:
image

The user I'm logged in has the following region and district:
image


The event created by this user has the following region and district:
image
image

The user has the following rights:
image


The second problem described by @adinaflorea9 I could not reproduce anymore but I've found a very similar scenario related to the unlinking:
1) I've added the eventgroup as it was described by Adina (with the same user described above)
2) I've logged in with another user (National user) and linked multiple events to that event group (from different jurisdictions)
3) Logged back with the initial user and opened the event to which the eventgroup earilier was added and tried to unlink it, a message is displayed that the unlink is not possible because of the jurisdiction:
image

sormas-api/src/main/java/de/symeda/sormas/api/event/EventFacade.java
```diff
@@ -86,4 +86,6 @@ Integer saveBulkEvents(
 		boolean eventStatusChange,
 		boolean eventInvestigationStatusChange,
 		boolean eventManagementStatusChange);
+
+	boolean inJurisdictionOrOwned(String uuid);
```
carina29, line 90: I've added the findings as a comment to the PR

carina29 (changes requested):

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12100

carina29 (approved):

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12100

PR #12221, opened 2023-06-28
#12219 Added guava where it is directly used

### LINKED ISSUE #12219: List guava as dependency in apps
### Problem Description

Found when developing https://github.com/SORMAS-Foundation/sormas-esante-adapter/issues/11 that `sormas-api/DateHelper` uses `guava` classes (search for example `import com.google.common.collect.Sets`), but it is not listed as dependency and therefore not found in the classpath when `sormas-api` is used as lib.

### Proposed Solution

List `guava` in all sormas modules that directly use classes from it (search `import com.google.common.`).

### Additional Information

This change does fix the problem when using `sormas-api` as dependency. Currently one has to make sure that such transitive dependency of `sormas-api` is then listed as dependency as well (relevant for unit tests). That's not pretty and a consequence of #11963.

_No response_

StefanKock:
Fixes #12219

### DIFF
--- a/sormas-api/pom.xml
+++ b/sormas-api/pom.xml
@@ -55,6 +55,11 @@
 		</dependency>
 
 		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
+		</dependency>
+
+		<dependency>
 			<groupId>commons-codec</groupId>
 			<artifactId>commons-codec</artifactId>
 		</dependency>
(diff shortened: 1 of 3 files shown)


MateStrysewske (approved):

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12221

PR #12594, opened 2023-09-29
#12548 - Skip selection when processing messages

### LINKED ISSUE #12548: Skip the "Pick or create entry" window when processing external messages if only one of the features is applicable [2]
### Problem Description

When only one of the features case surveillance, contact tracing, and event surveillance is enabled, or the user only has the user right to create entities for one of those, the external message processing workflow still prompts the user to select which kind of entity they want to create.

### Proposed Change

Hide all options that are not eligible to be used based on the features that are enabled on the server and the create rights of the user. Furthermore, prevent the processing of external messages if the user is not allowed to create either cases, contacts, or event participants.

### Added Value/Benefit

The workflow will be simplified for systems that only use one of the feature types (e.g. Luxembourg), and it will no longer be possible to create entities that aren't supported in the first place.

### Acceptance Criteria

- [x] The "Pick or create entry" window only shows radio buttons for the entities that are enabled on the server AND that the user has the create right for
- [x] If the user can only create entries for one of the entities (case, contact, event participant) based on the criteria above, skip this window completely UNLESS there is a potential duplicate found
- [x] If there is a potential duplicate found, the window is still shown, but only the "Create new" radio button for the entity that the user has access to is shown besides the "Select" radio button for the same entity
- [x] Additionally, disable the "Process" button if the user does not have the "create and edit" user rights for either cases, contacts, or event participants, or if none of the features is enabled on the server

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

MateStrysewske:
Fixes #12548


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12594

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12594

sergiupacurariu (approved):

PR #12618, opened 2023-10-09
#12550 - Hide jurisdiction fields when configured

### LINKED ISSUE #12550: Add an option to hide the jurisdiction fields from cases and use default values [5]
### Feature Description

For epidemiological surveillance right now, Luxembourg does not divide the country into several jurisdictions like e.g. Nigeria or Germany do. Because of that, the case jurisdiction fields are redundant and actually would be more of a hassle to be used by the users there. Therefore, a feature configuration is to be added to the system that hides the jurisdiction fields from cases (and other entities that are directly connected to cases and that aren't a separate feature) and automatically fills them with a default option.

Please note that it's not necessary to implement most of this in the mobile app, except for adding the new field to the infrastructure entities and the database for consistency reasons.

### Added Value/Benefit

Users in countries that don't divide the country/area in which SORMAS is used into several jurisdictions won't have to bother with the jurisdiction fields, easing and speeding up their workflow.

### Acceptance Criteria

- [x] A feature type property `HIDE_JURISDICTION_FIELDS` has been added to the `CASE_SURVEILANCE` feature, disabled by default
- [x] The `InfrastructureAdo` and `InfrastructureDto` classes, and all infrastructure tables in the database have been extended with a boolean `default` field, `false` by default; this also needs to be applied to the mobile app
- [x] All infrastructure create/edit forms in the UI have been extended with a checkbox for the new field (caption: "Default jurisdiction"); only show this field when the new feature type property is enabled
- [x] Upon saving an infrastructure with the checkbox checked, there's a backend validation that ensures that there is not already another default jurisdiction of the same type in the database; if there is, a ValidationRuntimeException with the text "There already is another default {entity name} in the database. Only one default {entity name} is allowed." is thrown
- [x] Add a `private DTO defaultInfrastructure` field to `AbstractInfrastructureFacadeEjb`, and a method called `getDefaultInfrastructure` that checks whether the field is already filled; if yes, return its value, otherwise query the database for the first entry of the ADO type with `default == true`; make sure that this method doesn't fail if there are no or multiple default entries; basically, this field should work like a cache so that we only have to query for default jurisdictions once and not every time a user requests those
- [x] If the new feature property is enabled:
  - [x] In the case creation and edit forms, hide all jurisdiction fields (and related fields, e.g. the "Place of stay differs from responsible jursidiction" checkbox) up to facility and automatically fill the region, district, and community fields with the respective default options by using the new `getDefaultInfrastructure` method; health facility fields will remain visible
  - [x] In the location form, hide the region, district, and community fields and automatically fill them with the respective default options

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

MateStrysewske:
Fixes #12550


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12618

sergiupacurariu:
the server cannot be started because in the StartupShutdownService the "createConstantFacilities()" method runs before "upgrade()" method this leads to an error as the "defaultInfrastructure" field does not exist yet.

sergiupacurariu:
- In the Configuration menu is not possible to check whether there is or not a default value set for an infrastructure type. It is very unpleasant to change it since the only way to do it is to 
  find it and deselect it then go and select default a new one.
  - Environment forms created and edited have to be modified the same way as for cases. Currently, the facilities available in the form are connected to the user's jurisdiction and not related to the default infrastructure set in Configuration.
  - The same changes are needed on the Environment Samples form as on the Environment forms.
  - it is possible to set default infrastructure that is not related, meaning you can choose default Region1 and default District from different region. On case save all the Infrastructure fields on the case (region_id, district_id, community_id) will be null. After saving, the case cannot be seen in the case grid. Even when the default infrastructure is selected consistently (e.g. default District is from the default Region) the case is saved with empty infrastructure fields.
  - Person grid - no persons are seen in the grid despite the creation of a few cases that also have a person who should be accessible

MateStrysewske:
@sergiupacurariu 

- Only regions, districts and communities can be default infrastructures now
- Added a column to those grids so users can see which is the default
- Added validation that prevents e.g. a district being set as default when it's not in the default region

Apart from that:
- Android doesn't need to be covered
- Environments don't need to be covered for now; I will add a respective warning to the Wiki and release notes
- It's also fine for now that users won't see data entered by them if they're not in the default jurisdiction, because this is not how this feature is supposed to be used

This is still pretty messy and I'd like to have it more consistent, but that would mean A LOT more effort that won't really be necessary at this point.

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/common/DatabaseHelper.java
```diff
@@ -3141,6 +3141,12 @@ public void onUpgrade(SQLiteDatabase db, ConnectionSource connectionSource, int
 				getDao(PathogenTest.class)
 					.executeRaw("ALTER TABLE pathogentest ADD COLUMN testedPathogen varchar(255);");
 
+			case 353:
```
sergiupacurariu, line 3144: this version needs to be updated to 354

sergiupacurariu:
- on all three Region, District and Community grids - the defaultInfrastructure field is not added in the sorting properties so filtering by Default crashes the grid
  - the defaultInftrastructureCache is not populated after the default Region is saved and also when the default region is changed conducting to an error while trying to save a district from the default Region. Because of this I even managed to save a district that was from the old region which was still stored in the cache leading to a situation of having a default district from outside the default region.
  - if there is a default district and community set it is possible to remove the default district while we still have a default community (without a default district). After this, it is also possible to save a new default district that is different from the default community district.

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12618

MateStrysewske:
@sergiupacurariu Updated DatabaseHelper. Yes, that needs to be mentioned, but since these changes are only on dev until the next release, it doesn't make sense to add these hints right now. I'll add them to the release notes and the Wiki page during the release.

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12618

PR #12674, opened 2023-10-19
#8337 - Show icon to create a case from contact as we have in the webapp

### LINKED ISSUE #8337: Show icon to create a case on contact  edit forms  on mobile app always [1]
### Proposed Change
Always show the button to convert contacts to cases in the mobile app, even if the contact classification is not yet confirmed. Right now, the button is hidden until the contact has been confirmed, which is inconsistent with the web app.

### Acceptance Criteria

- [x] Always show the icon  "create a case for this contact person"
- [x] If the contact classification is not confirmed and the user click on the button, show the same error message that is already shown in the web app for the same flow ("You can only convert contacts to cases that have been confirmed. Please confirm this contact before creating a case for its contact person.")

carina29:
Fixes #8337

### DIFF
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseNewActivity.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseNewActivity.java
@@ -27,13 +27,13 @@ import android.content.Context;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.view.Menu;
-
 import androidx.annotation.NonNull;
-
 import de.symeda.sormas.api.Disease;
 import de.symeda.sormas.api.caze.CaseClassification;
 import de.symeda.sormas.api.contact.ContactClassification;
 import de.symeda.sormas.api.contact.ContactStatus;
+import de.symeda.sormas.api.i18n.I18nProperties;
+import de.symeda.sormas.api.i18n.Strings;
 import de.symeda.sormas.api.utils.DataHelper;
 import de.symeda.sormas.api.utils.ValidationException;
 import de.symeda.sormas.api.utils.YesNoUnknown;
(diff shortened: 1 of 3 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12674

sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseNewActivity.java
```diff
@@ -80,8 +81,16 @@ public static void startActivityWithEmptyReportDate(Context fromActivity) {
 		BaseEditActivity.startActivity(fromActivity, CaseNewActivity.class, buildBundleWithEmptyReportDate());
 	}
 
-	public static void startActivityFromContact(Context fromActivity, String contactUuid) {
-		BaseEditActivity.startActivity(fromActivity, CaseNewActivity.class, buildBundleWithContact(contactUuid));
+	public static void startActivityFromContact(Context fromActivity, Contact contact) {
+		if (contact.getContactClassification() == ContactClassification.UNCONFIRMED) {
```
MateStrysewske, line 85:
```suggestion
		if (contact.getContactClassification() != ContactClassification.CONFIRMED) {
```

The other classifications also don't allow users to convert the contact

sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseNewActivity.java
```diff
@@ -80,8 +81,16 @@ public static void startActivityWithEmptyReportDate(Context fromActivity) {
...
+			NotificationHelper.showNotification(
+				getActiveActivity(),
+				NotificationType.WARNING,
+				getActiveActivity().getResources().getString(R.string.message_contact_to_case_confirmation_required));
```
MateStrysewske, line 89: Please use the same message that we use in the web app with I18nProperties; no need to duplicate it in the mobile app

MateStrysewske (changes requested):

sormas-app/app/src/main/java/de/symeda/sormas/app/caze/edit/CaseNewActivity.java
carina29: Yes, but the NO_CONTACT classification can not happen at all because the button will not be displayed on this case.
That's why I thought that it's enough to check the UNCONFIRMED , but I can change as you've suggested. :)

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12674

PR #12677, opened 2023-10-20
#12643 - Users can create cases through contacts and event participan…

### LINKED ISSUE #12643: Users can create cases through contacts and event participants if the CASE_SURVEILANCE feature is disabled
### Bug Description

Have the Feature config **CASE_SURVEILANCE** set on **False**. When creating a Contact/Event Participant with a positive Sample, user will be asked if he wants to create a case based on the positive sample. From here user will create a case even if the feature config is not available for the system.
Also in the User Role config the rights for Case are available to select.

### Reproduction Steps
I: Contacts:
1. Login with an Admin+NatUser that have a system with CASE_SURVEILANCE set on false
2. Create a Contact with a Sample having Positive result
3. A confirmation pop-up appears asking if a case is wanted to be created
4. Select Yes and continue creating the new Case

II: Event Participants:
1. Log in with a default Admin+NatUser;
2. Disable the 'CASE SURVEILLANCE' feature from the featureconfiguration in the DB;
3. Create an event with an event participant;
4. Add a positive sample with positive pathogen test to the event participant;
Actual behavior: After saving the sample with positive pathogen test, the user is asked whether they want to create a resulting case. Confirming opens the 'Create new case' form and the user can successfully create a case even though the feature for cases is not available from the feature configuration and the 'Cases' directory is not visible.

### Expected Behavior

User should not be able to create a case if the feature is not available for his system.

### Affected Area(s)

Web app

### SORMAS Version

1.89.0

### Android version/Browser

-

### User Role(s)

NatUser + Admin

### Server URL

_No response_

### Additional Information

_No response_

sergiupacurariu:
…ts if the CASE_SURVEILANCE feature is disabled

Fixes #12643

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/samples/PathogenTestController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/samples/PathogenTestController.java
@@ -51,6 +51,7 @@ import de.symeda.sormas.api.environment.environmentsample.EnvironmentSampleRefer
 import de.symeda.sormas.api.event.EventDto;
 import de.symeda.sormas.api.event.EventParticipantDto;
 import de.symeda.sormas.api.event.EventParticipantReferenceDto;
+import de.symeda.sormas.api.feature.FeatureType;
 import de.symeda.sormas.api.i18n.Captions;
 import de.symeda.sormas.api.i18n.I18nProperties;
 import de.symeda.sormas.api.i18n.Strings;
@@ -326,7 +327,7 @@ public class PathogenTestController {
 			.filter(t -> t.getTestResult() == PathogenTestResultType.NEGATIVE && t.getTestResultVerified())
 			.findFirst();
 
-		if (positiveWithSameDisease.isPresent()) {
+		if (positiveWithSameDisease.isPresent() && FacadeProvider.getFeatureConfigurationFacade().isFeatureEnabled(FeatureType.CASE_SURVEILANCE)) {
 			showChangeAssociatedSampleResultDialog(positiveWithSameDisease.get(), (accepted) -> {
 				if (accepted) {
 					checkForDiseaseVariantUpdate(positiveWithSameDisease.get(), caze, suppressNavigateToCase, this::showConfirmCaseDialog);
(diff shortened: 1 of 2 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12677

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12677

MateStrysewske (changes requested):
Wouldn't it make more sense to just add the check for the enabled feature configuration to the first if clause in the handleAssociatedCase method?

Also, please revert the changes to the user role edit form; we currently don't hide sections there if the respective feature configuration is disabled; a change like this doesn't belong in this issue. I've changed the description accordingly.

sergiupacurariu:
> Wouldn't it make more sense to just add the check for the enabled feature configuration to the first if clause in the handleAssociatedCase method?
> 
> Also, please revert the changes to the user role edit form; we currently don't hide sections there if the respective feature configuration is disabled; a change like this doesn't belong in this issue. I've changed the description accordingly.

Indeed it would be more readable but the handleAssociated.... methods also contain a confirmation logic when you already have a positive pathogen test and you want to add a negative pathogentTest. This logic will be removed as well. Do we want this logic to be removed?
The changes from the userRoleEditForm were removed.

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12677

MateStrysewske (commented):
I would say if there's logic that should be executed even if the case surveillance feature is off, it also should be executed when the user does not have case edit rights; in that case, this logic should be executed before the early return. However, we also need to make sure that there's no logic involving cases executed in this circumstance.

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12677

MateStrysewske (approved):

PR #12678, opened 2023-10-20
#9905- make sync work for users without role

### LINKED ISSUE #9905: Make user sync more resilient to missing user roles
### Problem Description

#9903 showed that users without user roles make data sync breaking.

`UserDtoHelper` dto to entity is to complicated: 
1. Fiddles around with userroles if not empty. This helps to avoid currupting existing data (makes data different to server then!), but the initial sync 
2. Calculates `JurisdictionLevel` despite being sent in the DTO, that fails if `userRoles == null`.

### Proposed Change

Make data sync work even if users don't have user roles.

carina29:
Fixes #9905


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12678

sergiupacurariu (approved):

PR #12764, opened 2023-11-20
Feature 6340 customizable enum interface

### LINKED ISSUE #6340: Add a user interface for customizable enum values [13]
### Situation Description

Thus far, the customizable enum value table only exists in the database and there is no user access via the UI. Since this table allows customization of a users Sormas instance, admins should be able to manage entries via an interface in the configurations menu

### Feature Description

Add a new tab to the configuration view that allows users with the respective rights to edit and remove existing and add new enum values to customizable enums.

### Added Value/Benefit
Users don't have to rely on technical admins to manipulate the database. Also, going through the interface instead of directly doing changes in the database means that the data can actually be validated.

### Acceptance Criteria
- [x] Add a new `CUSTOMIZABLE_ENUM_MANAGE` user right granted to user roles deriving from the default admin role
- [x] Add a new tab called 'Customizable Enum Values' to the configurations screen; make sure to update the logic that grants users access to the configurations with the new user right
- [x] Add a table which includes the following columns
    - edit icon (no caption)
    - `datatype`
    - `value`
    - `caption`
    - `diseases`
    - `properties`
- [x] Add general filters above the list
    - free text search field for: value, caption
    - datatype - drop-down only for data types included in the list
    - diseases - drop-down only for diseases included in the list
- [x] Add a creation button at the top right
    - When clicked, open a customizable enum value entry form
        - datatype (drop-down including the available data types) - compulsory
        - value (free text) - compulsory, validate for uniqueness in datatype and proper format
        - caption (free text) - compulsory
- [x] When clicking the edit icon, open an edit window which is similar to the creation form plus additional fields:
    - translations (should include a multi selection option with the available languages - `languageCode` - plus a subsequent free text `value` translation field
      - the same `languageCode` should not be able to be selected more than once
    - properties
    - diseases (multi selection option to specify any available diseases)
    - Also show the uuid of the entry (non-editable) at the top
    - Add buttons to discard and save
    - Disable datatype and value fields
- [x] Add an info button to the left of the creation button
    - Upon clicking, open a pop-up with explanations on how to use the customizable enum value configuration section

### Additional Information

Configurations customizable enum values
New value
Edit value

MateStrysewske:
Fixes #6340


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12764

sergiupacurariu:
UI findings
- when a new customizable enum value is added the value is not available in a combo box unless the server is restarted do we want this behavior (should instruct admins about this)
- once added there is no possibility of removing a customizable enum even if it hasn't been used at all
- the "Occupation type" for example is present on a combo related to a person entity. Does setting a disease restriction change anything in the behavior? Should we have this for any data type?
	Same with the Pathogens on Environment sample(There is no disease on the environment entity)
	
- Pathogen test form
	- if a disease variant had additional details set(has details) and then removed, the pathogens that had "details" still show the details text box without showing "Tested disease variant" combo
	
image

Android app:
- CustomizableEnumValueDtoHelper.fillInnerFromDto - setDiseases fail because the type inconsistency (Set vs List)
	- diseases field type must be changed to Set (after that the synchronization works fine :))

MateStrysewske:
@sergiupacurariu 

- Fixed the Android issue
- Fixed the issues with server restart being required after changing enums, and the details field still being shown after removing it from a value (same problem = the cache has to be reset after saving an enum value)
- The fact that removing an existing value isn't possible is true; this is because checking whether or not the value is in use would be a VERY fragile method that would require a lot of maintenance (i.e. whenever a customizable enum is added to another entity, the respective code would need to be manually extended) and could easily lead to pretty significant errors, which is why I decided against implementing a delete functionality; there will be a follow-up issue for adding an archive functionality instead
- Diseases not being relevant for some enum types is true; however, this a) might change if the enum type is added to other entities, and b) is not really relevant for this issue here because it would be a general change to the customizable enum system; I'll think about creating a follow-up issue for this as well

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12764

sergiupacurariu:
- if a Customizable enum has some diseases set then that Enum will not be shown in an environment where no disease is involved (e.g. the Pathogen data type used along with Environment sample that does not have any link to a particular disease) - is this a wanted behavior?
- Customizable enum grid - for the Pathogen data type the grid does not show the Properties column correctly (all Pathogen rows show "Has details" while most of them have this property set to false)
- Create / Edit pathogen test result - The "Tested pathogen details" text field is not shown when a Pathogen "Has Details". It seems there is no support on the form for this situation. Not sure whether this part should be fixed here because it is linked to a different feature. However, this directly impacts the Customizable enum feature.

MateStrysewske:
@sergiupacurariu 
- Yes, that's wanted behaviour
- Fixed
- Created a new issue for this: https://github.com/SORMAS-Foundation/SORMAS-Project/issues/12785

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12764

PR #12782, opened 2023-11-24
#12729 Detect duplicate persons by national health ID

### LINKED ISSUE #12729: Detect duplicate persons by national health ID [1]
### Feature Description

For Luxembourg (and potentially other systems in the future), persons can be unambigiously identified by their national health ID. This e.g. has impact on the automatic processing of external messages, i.e. automatic processing is not possible if there's already a person with the same national health ID in the system. However, when manually processing a message, this currently does not lead to a duplicate person detection if the name is not similar (which should be an edge case, but is possible), making it impossible to properly process the message under these circumstances.

The purpose of this issue is to alter the person duplicate detection to add a configuration that uses the national health ID as a "single source of truth" when it's present.

### Added Value/Benefit

Persons can be detected as duplicates based on their unique national health ID.

### Acceptance Criteria

- [x] A new property has been added to sormas.properties: `duplicatechecks.nationalhealthidoverridescriteria`; default false
- [x] If the property value is true, when checking for duplicate persons (PersonService.buildSimilarityCriteriaFilter), IF a national health ID is specified in the criteria, persons are automatically detected as duplicates if they have a national health ID specified and it matches the one specified in the criteria; persons that don't have a national health ID specified or one that doesn't match the one in the criteria, the other existing logic is considered
- [x] This logic is properly integrated in the places where duplicate persons are checked; especially during e.g. case creation, i.e. a new case for a new person with a national health ID that already exists in the system, but with a different name, has to be flagged as a potential duplicate

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

leventegal-she:
Fixes #12729

### DIFF
--- a/sormas-base/setup/sormas.properties
+++ b/sormas-base/setup/sormas.properties
@@ -150,7 +150,7 @@ app.url=
 # Default: false
 # Possible Values: true, false
 #duplicatechecks.excludepersonsonlylinkedtoarchivedentries=false
-# Determines whether the national health ID is used to identify duplicate persons.
+# Determines whether the national health ID is used to unambiguously identify duplicate persons irrespective of whether other fields match or not.
 # Default: false
 # Possible Values: true, false
 #duplicatechecks.nationalhealthidoverridescriteria=false
--- a/sormas-base/setup/sormas.properties
+++ b/sormas-base/setup/sormas.properties
@@ -150,6 +150,10 @@ app.url=
 # Default: false
 # Possible Values: true, false
 #duplicatechecks.excludepersonsonlylinkedtoarchivedentries=false
+# Determines whether the national health ID is used to identify duplicate persons.
+# Default: false
+# Possible Values: true, false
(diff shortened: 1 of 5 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12782

sormas-base/setup/sormas.properties
```diff
@@ -150,6 +150,10 @@ app.url=
 # Default: false
 # Possible Values: true, false
 #duplicatechecks.excludepersonsonlylinkedtoarchivedentries=false
+# Determines whether the national health ID is used to identify duplicate persons.
```
MateStrysewske, line 153:
```suggestion
# Determines whether the national health ID is used to unambiguously identify duplicate persons irrespective of whether other fields match or not.
```

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12782

PR #12797, opened 2023-11-29
#12604 - Add a line-listing feature to event participants - fix "Line…

### LINKED ISSUE #12604: Add a line-listing feature to event participants [5]
### Feature Description

-  Users are required to return to the event participant directory each time a new ep is recorded, thus leading to much time wasted.
-  As a user, I would like to list all the ep as fast as possible in one view (like the line listing feature in the web app) or sormas should present me with the option to "save and create another" after filling in the creation form for an ap.
- Add the line listing to ep directory in the web app
- Add the option  "save and create another" to the mobile app just like already implemented for cases (when line listing is enabled for the respective disease from the Configuration -> Line Listing view of the web app)

### Added Value/Benefit

- users would be able to create event participants faster and then complement the non-requited information at a later.

### Acceptance Criteria
Web app:
- [x] A "Line Listing" button has been added to the event participants view
- [x] When clicking on the new button, a popup is opened that allows the user to create as many event participants for the event as they wish, similar to the case line listing window
  - Responsible region and district are displayed under a heading "Shared information"; whether or not they're required depends on whether there's a region and district specified for the event (same logic as when creating individual event participants)
  - First name, last name, sex (all required) and involvement description are displayed in one line below a "List of new event participants" heading
  - An "Add line" button allows users to add new event participants to the list
  - Delete buttons (displayed as a trash can icon) are added to all lines and delete the respective line; when only one line is present, the button is disabled
  - Saving saves all the individual event participants ONLY if there are no validation errors (i.e. only when all required fields have been filled in)
  - Person AND event participant duplicate detection is done live for each event participant; i.e. when two identical event participants are saved, the second one should be detected as a duplicate event participant (similar to how we already do it for cases and contacts in line listing)

Mobile app:
- [x] Add a checkbox to the top of the event participant creation fragment: "Use rapid event participant entry"
- [x] If the checkbox is checked, after saving, don't redirect the user to the event participant edit fragment, but instead allow them to create another event participant; take over the responsible region and district from the previously entered event participant

### Implementation Details

See rapid case entry in the mobile app and implement the rapid event participant entry similarly

### Mockups

_No response_

sergiupacurariu:
… listing" button readOnly behavior

Fixes #12604

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventParticipantsView.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/events/EventParticipantsView.java
@@ -305,6 +305,8 @@ public class EventParticipantsView extends AbstractEventView implements HasName 
 			final ExpandableButton lineListingButton = new ExpandableButton(Captions.lineListing)
 				.expand(e -> ControllerProvider.getEventParticipantController().openLineListingWindow(getEventRef()));
 			addHeaderComponent(lineListingButton);
+			lineListingButton
+				.setEnabled(!isEventDeleted() && isEditAllowed() && UserProvider.getCurrent().hasUserRight(UserRight.EVENTPARTICIPANT_EDIT));
 		}
 
 		topLayout.addStyleName(CssStyles.VSPACE_3);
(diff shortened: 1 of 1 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12797

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12797

MateStrysewske (approved):

PR #12814, opened 2023-12-06
#6340 - Update displayed diseases

### LINKED ISSUE #6340: Add a user interface for customizable enum values [13]

MateStrysewske:
Fixes #6340

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/i18n/Strings.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/i18n/Strings.java
@@ -1578,6 +1578,7 @@ public interface Strings {
 	String promptContactEpiWeekTo = "promptContactEpiWeekTo";
 	String promptContactsSearchField = "promptContactsSearchField";
 	String promptCreationDateFrom = "promptCreationDateFrom";
+	String promptCustomizableEnumSearchField = "promptCustomizableEnumSearchField";
 	String promptCustomizableEnumTranslationCaption = "promptCustomizableEnumTranslationCaption";
 	String promptCustomizableEnumTranslationLanguage = "promptCustomizableEnumTranslationLanguage";
 	String promptDateTo = "promptDateTo";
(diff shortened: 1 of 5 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12814

sergiupacurariu (approved):

PR #12879, opened 2024-01-11
#6024 Alphabetical sorting should not consider capitalization

### LINKED ISSUE #6024: Alphabetical sorting should not consider capitalization [3]
### Bug Description

When columns are sorted alphabetically, the column displays capitalized entries alphabetically first and then starts again with lower case text which in the end disrupts the alphabetical order.
### Steps to Reproduce

1. Enter e.g. cases with first names differing in capitalization and first letter
2. Sort the first name column and check the sorted list

### Expected Behavior

All alphabetically sorted columns should disregard capitalization for alphabetical sorting.
### Screenshots

sortingAlphabetically

### System Details

* Device:
* SORMAS version: 1.61.0

### Developer Specifications

**Prerequisite:** #10249

Make sure IndexLists use case-insensitive ordering where necessary:

- instead of using e.g. `ORDER BY name`, use `ORDER BY lower(name)`. As mentioned below, it might be necessary to use `ORDER BY lower(name), name` to prevent certain problems regarding `DISTINCT`
- Do this for every `indexList` which involves freetext-fields which are later displayed in sortable tables/grids
  - This sorting is only necessary for the freetext fields; Predefined fields (e.g. enums) should not be affected, as each value inside of an enum should use the same capitalization
- Investigate potential performance impact. Make sure no tables are initially sorted by any of the freetext fields
- Write some unit tests for sorting (?)

leventegal-she:
Fixes #6024

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseFollowUpDto.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/caze/CaseFollowUpDto.java
@@ -36,7 +36,7 @@ public class CaseFollowUpDto extends FollowUpDto {
 	public CaseFollowUpDto(String uuid, Date changeDate, String personFirstName, String personLastName,
 						   Date reportDate, Date symptomsOnsetDate, Date followUpUntil, SymptomJournalStatus symptomJournalStatus,
 						   Disease disease,
-						   boolean isInJurisdiction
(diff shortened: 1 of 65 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12879

MateStrysewske (approved):

PR #12882, opened 2024-01-12
Feature 12804 custom enum values active field

### LINKED ISSUE #12804: Add an "active" field to customizable enum values [5]
### Feature Description

Customizable enum values currently can't be deleted, archived or anything like that. However, deletion would imply a lot of potential issues (e.g. the need to manually query all tables where the customizable enum in question is in use in order to determine whether it is used somewhere, which is code that needs to be maintained).

Because of that, the purpose of this issue is to add an "active" field to customizable enum values, similar to the one we have for jurisdictions.

### Added Value/Benefit

Customizable enum values that have been created accidentally or that are no longer in use can be removed from selection fields in order to prevent users from selecting them/seeing too many options.

### Acceptance Criteria

- [ ] A boolean `active` field has been added to customizable enum values - defaults to true
- [ ] The edit form for customizable enum values has been extended with a checkbox for the new field (caption: "Active")
- [ ] The customizable enum values grid has been extended with an "Active" column (would be great to have a renderer that renders true as a checkmark and false as an empty string, otherwise just use our boolean renderer)
- [ ] An "Active/Inactive" dropdown filter has been added (right-aligned) that allows to toggle between active and inactive customizable enum values
- [ ] The customizable enum values grid has been extended with a filter for the new field (dropdown with the selection options "Active" and "Inactive")
- [ ] All code that populates fields with customizable enum values (both in the web and mobile app) has been extended to only show active values of the respective data type; if the currently selected value is an inactive value, it still needs to be shown and added to the field (like we e.g. do it for diseases)

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

MateStrysewske:
Fixes #12804

### DIFF
--- a/sormas-api/src/main/java/de/symeda/sormas/api/customizableenum/CustomizableEnumFacade.java
+++ b/sormas-api/src/main/java/de/symeda/sormas/api/customizableenum/CustomizableEnumFacade.java
@@ -79,19 +79,26 @@ public interface CustomizableEnumFacade
 	boolean existsEnumValue(CustomizableEnumType type, String value, Disease disease);
 
 	/**
+	 * See getEnumValues(CustomizableEnumType, String, Disease);
+	 */
+	<T extends CustomizableEnum> List<T> getEnumValues(CustomizableEnumType type, Disease disease);
+
+	/**
 	 * Retrieves the cached contents of all enum value instances of the specified type. The results are already
 	 * internationalized based on the user's language, or the server language as a fallback. If the enum values for the
 	 * specified type and disease have not been requested yet, the cache is extended with them on demand.
-	 * 
+	 *
 	 * @param type
 	 *            The type for which to retrieve the enum values
 	 * @param disease
-	 *            The disease for which to retrieve the enum values. If null, all enum values that are disease-independent are retrieved
+	 *            The disease for which to retrieve the enum values; if null, all enum values that are disease-independent are retrieved
+	 * @param selectedValue
+	 *            An already selected value that needs to be included even if inactive
 	 * @param <T>
 	 *            The specific extension of {@link CustomizableEnum} for type safety
 	 * @return A list of all enum instances containing their values, internationalized captions, and optional properties
 	 */
-	<T extends CustomizableEnum> List<T> getEnumValues(CustomizableEnumType type, Disease disease);
+	<T extends CustomizableEnum> List<T> getEnumValues(CustomizableEnumType type, String selectedValue, Disease disease);
 
 	/**
 	 * Indicates if the specified type has any entry in the database for a specific disease.
--- a/sormas-app/app/src/main/java/de/symeda/sormas/app/backend/customizableenum/CustomizableEnumValueDao.java
+++ b/sormas-app/app/src/main/java/de/symeda/sormas/app/backend/customizableenum/CustomizableEnumValueDao.java
@@ -15,17 +15,17 @@
 
 package de.symeda.sormas.app.backend.customizableenum;
 
-import com.j256.ormlite.dao.Dao;
-
-import org.apache.commons.collections4.CollectionUtils;
-import org.apache.commons.lang3.StringUtils;
-
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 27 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12882

sormas-backend/src/main/java/de/symeda/sormas/backend/customizableenum/CustomizableEnumFacadeEjb.java
```diff
@@ -91,6 +91,10 @@ public class CustomizableEnumFacadeEjb
 	 * enum values as its values.
 	 */
 	private final EnumMap<CustomizableEnumType, Map<String, Map<String, Object>>> enumProperties = new EnumMap<>(CustomizableEnumType.class);
+	/**
+	 * Maps a customizable enum type to a map with all enum values of this type as its keys and their active status as its values.
+	 */
+	private final EnumMap<CustomizableEnumType, Map<String, Boolean>> enumActiveStatuses = new EnumMap<>(CustomizableEnumType.class);
```
leventegal-she, line 97: Would be cleaner to change the above field to include the active flag too, I mean use a custom class instead of `Map<String, Map<String, Object>>`

Or event better refactor the `enumValues` field

Keeping this structure the next developer will introduce yet another map for another field on the customizable enum, and in my opinion it just makes the code more complicated and harder to read and understand

sormas-app/app/src/main/java/de/symeda/sormas/app/backend/customizableenum/CustomizableEnumValueDao.java
```diff
@@ -57,6 +57,10 @@ public class CustomizableEnumValueDao extends AbstractAdoDao<CustomizableEnumVal
 	 * enum values as its values.
 	 */
 	private final Map<CustomizableEnumType, Map<String, Map<String, Object>>> enumProperties = new HashMap<>();
+	/**
+	 * Maps a customizable enum type to a map with all enum values of this type as its keys and their active status as its values.
+	 */
+	private final Map<CustomizableEnumType, Map<String, Boolean>> enumActiveStatuses = new HashMap<>();
```
leventegal-she, line 63: same as in the Facade

sormas-api/src/main/java/de/symeda/sormas/api/customizableenum/CustomizableEnumFacade.java
```diff
@@ -78,20 +78,27 @@ public interface CustomizableEnumFacade
...
-	 *            The disease for which to retrieve the enum values. If null, all enum values that are disease-independent are retrieved
+	 *            The disease for which to retrieve the enum values; if null, all enum values that are disease-independent are retrieved
+	 * @param selectedValue
+	 *            An already selected value that needs to be included even if inactive
 	 * @param <T>
 	 *            The specific extension of {@link CustomizableEnum} for type safety
 	 * @return A list of all enum instances containing their values, internationalized captions, and optional properties
 	 */
-	<T extends CustomizableEnum> List<T> getEnumValues(CustomizableEnumType type, Disease disease);
+	<T extends CustomizableEnum> List<T> getEnumValues(CustomizableEnumType type, String selectedValue, Disease disease);
```
leventegal-she, line 101: Is this really needed, we already have a working solution for inactive infrastructure values:
show it as value in the input suffixed with "(inactive)", and missing in the list of selectable values

This would be a second way to handle the same problem and it doesn't event mark that the value in not active anymore.

Post probably it's the same in the mobile app too

leventegal-she (changes requested):

MateStrysewske:
@leventegal-she United the properties and activeStatuses caches with a dedicated class, and changed the logic in the web app to behave similarly to the infrastructure fields. In the mobile app, the refactoring effort would be significantly higher, which is why I kept the current logic there.

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12882

leventegal-she (approved):

PR #12915, opened 2024-01-25
#12914 - Fix config screen access

### LINKED ISSUE #12914: Error when accessing the configuration menu without outbreak or region access
### Bug Description

When trying to access the configuration screen without access to the Outbreaks and Regions view, e.g. with a user that can only configure line listings, errors are displayed because the system tries to navigate to the Regions view by default, which the user is unable to access.

### Reproduction Steps

1. Configure a user role to only have access to the line listings configuration among all the configuration screens
2. Log in with that user and open the configurations screen
3. Observe the error messages; you can click on the "Line listing" view to actually properly access it after dismissing the errors

### Expected Behavior

The user should be redirected to a screen they can actually access.

### Affected Area(s)

Web app

### SORMAS Version

SORMAS 1.94.0-SNAPSHOT

### Android version/Browser

Chrome

### User Role(s)

Any user role

### Server URL

localhost

### Additional Information

_No response_

MateStrysewske:
Fixes #12914

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/MainScreen.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/MainScreen.java
@@ -343,10 +343,9 @@ public class MainScreen extends HorizontalLayout {
 		}
 
 		if (UserProvider.getCurrent().hasConfigurationAccess()) {
-			AbstractConfigurationView.registerViews(navigator);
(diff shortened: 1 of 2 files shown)


carina29 (approved):

carina29:
@MateStrysewske I also tested the bug before and after your changes

PR #12939, opened 2024-02-06
#12930 - Only disable config when enabled

### LINKED ISSUE #12930: Don't disable feature configurations that are already disabled
### Bug Description

During server startup, SORMAS is checking whether there's any feature configuration in the database that has a disabled dependent feature configuration. If such a configuration is found, it is disabled.

However, this will be repeated on every startup because the code is not checking whether the depending feature configuration is already disabled. This a) slows startup time and b) potentially interferes with read-only instances.

### Reproduction Steps

1. Disable a feature configuration that has depending feature configurations, e.g. case surveillance
2. Disable one of the depending feature configurations, e.g. contact tracing
3. Put a breakpoint in FeatureConfigurationService line 190
4. Observe that the contact tracing feature configuration is saved even though it has not changed (since it was already disabled before)

### Expected Behavior

Feature configurations that are already disabled should be excluded from this logic.

### Affected Area(s)

Web app

### SORMAS Version

1.95.0-SNAPSHOT

### Android version/Browser

Any browser

### User Role(s)

Doesn't matter

### Server URL

localhost

### Additional Information

_No response_

MateStrysewske:
Fixes #12930

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/feature/FeatureConfigurationService.java
@@ -162,7 +162,7 @@ public class FeatureConfigurationService extends AdoServiceWithUserFilterAndJuri
 			FeatureConfiguration configuration = configs.get(featureType);
 			boolean persistingNeeded = false;
 
-			if (featureType.isServerFeature() && featureType.isDependent()) {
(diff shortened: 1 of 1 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12939

sergiupacurariu (approved):

PR #12944, opened 2024-02-07
#12662 - Don't show case creation prompt

### LINKED ISSUE #12662: Do not trigger the 'Create resulting' case for contacts and event participants when creating a positive sample [1]
### Problem Description

Currently, when the rights to create new cases are missing for a user, if they create a positive sample for a contact or event participant, the user is asked whether the want to create a resulting case based on that sample. The form for 'Create new case' that gets triggered afterwards is completely read only, since the user does not have create rights.

### Proposed Change

No longer trigger the 'create resulting case' pop-up when saving the positive sample if the user does not have the 'Create new cases' right.

### Added Value/Benefit

- not triggering a form that the user can not fill and does not have access to in any other flow.

### Acceptance Criteria

- [ ] do not trigger the 'Create resulting case' pop-up saving positive samples for contacts or event participants if the user does not have 'Create new cases' right;

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

MateStrysewske:
Fixes #12662

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/samples/PathogenTestController.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/samples/PathogenTestController.java
@@ -405,12 +405,10 @@ public class PathogenTestController {
 			.filter(t -> t.getTestResult() == PathogenTestResultType.NEGATIVE && t.getTestResultVerified())
 			.findFirst();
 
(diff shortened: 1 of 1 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=12944

leventegal-she (approved):

PR #13005, opened 2024-02-23
 #11995 Deleting a pathogen test without edit right over the associat…

### LINKED ISSUE #11995: Deleting a pathogen test without edit right over the associated case does not update the case classification
### Bug Description

When a user that does not have edit rights over a case deletes a pathogen test that was used to automatically classify the case, the case's classification is not updated back.

### Reproduction Steps

1. With a regular Admin+NatUser create a case and add to it a sample with positive pathogen test result (Covid-19);
2. Once the pathogen test and the sample are saved, the case's classification is automatically updated to 'Confirmed case with unknown symptoms' or 'Confirmed case' (if the symptoms are also present);
3. Log in with a user that has the following rights over cases and samples:
image
image
4. Open the sample and delete the pathogen test; or delete the whole sample (if the right to 'Delete existing samples' is also given to the user).

### Expected Behavior

Once the pathogen test is deleted, the Case classification should be automatically updated to the value that can be met by the current case classification criteria. (solution should be similar to #11980).

### Affected Area(s)

Web app

### SORMAS Version

1.84.0

### Android version/Browser

Chrome

### User Role(s)

Admin+NatUser; Configured user with the rights from the steps to reproduce

### Server URL

local machine

### Additional Information

_No response_

leventegal-she:
…ed case does not update the case classification

Fixes #11995

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -2193,43 +2193,7 @@ public class CaseFacadeEjb extends AbstractCoreFacadeEjb<Case, CaseDataDto, Case
 
 		updateTasksOnCaseChanged(newCase, existingCase);
 
-		// Update case classification if the feature is enabled
-		CaseClassification classification = null;
-		boolean setClassificationInfo = true;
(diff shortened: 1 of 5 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=13005

MateStrysewske (approved):

PR #13022, opened 2024-02-29
#11552 Consider source case region and district for contacts in detai…

### LINKED ISSUE #11552: Consider source case region and district for contacts in detailed sample export
### Problem Description

Currently, if a sample is associated with a contact that takes its jurisdiction from a source case, in the detailed export file of samples, that sample does not have a jurisdiction. In the main grid of the 'Samples' directory it is visible that the sample takes the district of the source case as that one is displayed in the grid, and based on that jurisdiction the sample can be searched/filtered for.

### Proposed Change

Fill the contact region, contact district and contact community columns in the detailed sample export file with the information from the responsible jurisdiction of the source case if the contact itself does not have a responsible region, district and community (if the contact has a region and district, but not a community, DON'T take over the case community).

leventegal-she:
…led sample export

Fixes #11552


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=13022

carina29 (approved):

PR #13032, opened 2024-03-05
#11406 - Show only active entities in Person Info Cards

### LINKED ISSUE #11406: Show only active entities in Person Info Cards
### Problem Description

Currently, on each entity grid, a user with enough rights is able to see entities in 3 statuses: Active, Archived and Deleted. The entities displayed on the info cards in the Person View could have Active or Archived statuses. When "See cases for this person" is used the default filter of the grid that opens is "Active". This conducts to inconsistency. Let's say the person open has 2 cases one is Active and one Archived. On the info card from the person view 2 cases are shown but when the user checks "See cases for this person" only one case is shown (the Active one). This situation happens on all entities that have info cards on Person View.
In the case of shared cases with ownership, the situation is the same. The info card on the Person view shows both local and shared cases while the "See cases for this person" filters out the shared with ownership ones. This logic difference can conduct as well to have the users seeing different number of cases in info cards vs filtered case grid. 

### Proposed Change

Show only the active entities on the info card. 

### Possible Alternative

An alternative could be that we keep the current way that the entities are displayed on the info cards on each person page (both the active and archived/shared entities) but to have a marker to distinguish between them (just how we already have a marker that helps the user distinguish the case that they entered the person page from, or contact and so on; i.e how that entity is highlighted).

In this way, the user sees all the entities associated with the person (unless the entities are already soft deleted or permanently deleted) and the marks will indicate that some are active, some are archived, some are shared with ownership to another system. In this way, when the user navigates to 'See cases for this person' they will already know that not all cases from the info cards are active, and therefore, they will know to filter the cases grid based on the 'Active/Archived/All cases' or 'With ownership/To view/All Cases' filter options. 

If the archived or shared cases are not displayed at all, then it would be unlikely that the user would take the extra step to go check the archived cases grid/ to view grid just on the chance that the person might also have other entities associated.

sergiupacurariu:
Fixes #11406

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseService.java
@@ -1473,15 +1473,15 @@ public class CaseService extends AbstractCoreAdoService<Case, CaseJoins> {
 				default:
 				}
 
-				filter = CriteriaBuilderHelper.or(cb, filter, specialCaseAccessService.createSpecialCaseAccessFilter(cb, joins.getSpecialCaseAccesses()));
+				filter =
+					CriteriaBuilderHelper.or(cb, filter, specialCaseAccessService.createSpecialCaseAccessFilter(cb, joins.getSpecialCaseAccesses()));
 			}
 
 			// get all cases based on the user's contact association
 			if (userFilterCriteria == null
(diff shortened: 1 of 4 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=13032

leventegal-she (approved):

PR #13052, opened 2024-03-21
#11928 Limited disease users should not be able to edit event groups …

### LINKED ISSUE #11928: Limited disease users should not be able to edit event groups that have at least one event of a different disease
### Problem Description

#11715 introduced changes so that users limited to a certain disease could no longer edit entities that they do not own of a different disease than the one they are limited by.

Event groups were not included in the respective ticket as more investigation needed to be done in order to specify the expected behavior. 

Currently. for a user limited by a certain disease, as long as one event of their disease is linked to an event group, the user can edit every aspect of that event group (archive it, de-archive it, delete and restore it, update the group name, link and unlink events). This is prohibited however if the user is restricted by a responsible jurisdiction instead of a disease one, so the behavior needs to be updated to be consistent.

### Proposed Change

The users that are limited by a disease should have similar restriction in what concerns event groups as users limited by responsible jurisdiction when they access an event group that has entities outside their jurisdiction.

### Added Value/Benefit

- consistency in what concerns behaviors related to user roles and rights in the event groups 

### Acceptance Criteria

- [x] in the 'Groups' grid the user should see the groups that have at least one event associated that has the same disease as the user's disease ~~(or at least one event owned by the user);~~
- [ ] ~~in the 'Groups' grid the user should see the groups owned by them (that they created);~~
- [x] on the event group page, if the event group does not contain any event of the disease of the user, ~~or any event owned by the user,~~ they should not be able to edit any aspect of the event group (archive, delete, update the name);
- [x] on the event group page, if the event group has at least one event of the user's disease and other events of different diseases the user should not be able to edit the event group information; they should be able to unlink only the event of their disease;
- [x] on the event group page. if the event group has only events of the same disease as the user's, then all action should be enabled (as long as the user has the respective user rights and the events are all within the user's responsible jurisdiction);
- [x] the user should still see all events related to the event group regardless of their responsible or disease jurisdiction (to be consistent with how the user sees for example entities associated with a person - on the person page, all associated entities are displayed regardless of their responsible or disease jurisdiction);
- [x] ~~unless owned by~~, the user should be forbidden from unlinking events that are outside their disease jurisdiction;
- [x] linking events to the event group should follow the same logic as is currently implemented (if the user has rights to link the event and the event is within their jurisdiction or owned by them then they should still be able to do so). 

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

leventegal-she:
…that have at least one event of a different disease

Fixes #11928

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventGroupFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventGroupFacadeEjb.java
@@ -55,6 +55,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
+import de.symeda.sormas.api.EditPermissionType;
 import de.symeda.sormas.api.common.Page;
 import de.symeda.sormas.api.common.progress.ProcessedEntity;
 import de.symeda.sormas.api.common.progress.ProcessedEntityStatus;
@@ -332,6 +333,10 @@ public class EventGroupFacadeEjb implements EventGroupFacade {
 		EventGroup existingEventGroup = eventGroupService.getByUuid(dto.getUuid());
 		FacadeHelper.checkCreateAndEditRights(existingEventGroup, userService, UserRight.EVENTGROUP_CREATE, UserRight.EVENTGROUP_EDIT);
 
+		if (!isEditAllowed(dto.getUuid())) {
+			throw new AccessDeniedException(I18nProperties.getString(Strings.errorAccessDenied));
+		}
+
 		EventGroup eventGroup = fillOrBuildEntity(dto, existingEventGroup, checkChangeDate);
 
 		final JurisdictionLevel jurisdictionLevel = currentUser.getJurisdictionLevel();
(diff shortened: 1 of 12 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=13052

sormas-backend/src/main/java/de/symeda/sormas/backend/event/EventGroupFacadeEjb.java
```diff
@@ -432,15 +440,14 @@ public void linkEventToGroup(Event event, User currentUser, List<EventGroup> eve
 				"User " + currentUser.getUuid() + " is not allowed to link events from another region to an event group.");
 		}
 
-		// Check that the event group is not already related to this event
-		List<EventGroup> filteredEventGroups = getFilteredEventGroups(event, eventGroups);
+		List<EventGroup> eventGroupsToLinkTo = getEventGroupsNotLinkedTo(event, eventGroups);
```
carina29, line 443: Related to the last acceptance criteria, I'm not sure if the linking should not be possible if the user has rights to link the event but the jurisdiction is different but the user is the owner of the event ( since is crossed the "owned by them" part).
I've checked the code and tested it, and right now is possible, as it was in the past.
image

I've changed the jurisdiction of the above used Test User from Nation to Region level (Hamburg):
image
image
I was able to link an event with the same disease as the user has (the owner is the logged in test user) from another jurisdiction ( if the owner part is still requested, everything is fine).
image

image

Also tried to link another event ( with a different owner and different region but same disease).
The save is prevented with an Access Denied exception as it is checked in the code:
image

What is not clear, why this event appears at all in the Event section ( different jurisdiction ,different owner, same disease) but the display of this event in Event Directory is not related to this ticket:
image

image
image

carina29 (changes requested):

sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=13052

carina29 (approved):

PR #13061, opened 2024-03-27
#13042 - Add feature configuration for Keycloak synchronization plus …

### LINKED ISSUE #13042: Add feature configuration for Keycloak synchronization plus SORMAS adjustments [3]
### Feature Description

Ticket: #13033 
[In Luxembourg we will have the situation, that users are created and maintained by their own IAM/TAM System.
This service will be linked to keycloak, which will be used as proxy.](https://github.com/SORMAS-Foundation/SORMAS-Project/issues/13033)

The implementation of this ticket and going live with it would lead to the situation, that other SORMAS users cannot synch their users from SORMAS to keycloak anymore.

Therefore a feature configuration is needed.

But as this is the expected behavior, and the new implementation was done especially for Luxembourg, we have to add a feature configuration that keeps the sync-direction for all other instances as it was before (DEFAULT) and offers the option to change the SYNC direction from Keycloak to SORMAS with all the consequences.

### Added Value/Benefit

The implementation of this ticket and going live with it would lead to the situation, that other SORMAS users cannot synch their users from SORMAS to keycloak anymore.

Therefore a feature configuration is needed.

But as this is the expected behavior, and the new implementation was done especially for Luxembourg, we have to add a feature configuration that keeps the sync-direction for all other instances as it was before (DEFAULT) and offers the option to change the SYNC direction from Keycloak to SORMAS with all the consequences.

### Acceptance Criteria

- [ ] Feature configuration to enable/disable SORMAS => KEYCLOAK // KEYCLOAK => SORMAS (It's either or)
- [ ] When KEYCLOAK => SORMAS is enabled, you cannot edit users anymore (except roles and rights)

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

sergiupacurariu:
…SORMAS adjustments - enable jurisdiction fields

Fixes #13042

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserEditForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserEditForm.java
@@ -24,6 +24,7 @@ import static de.symeda.sormas.ui.utils.LayoutUtil.fluidRowLocsCss;
 import static de.symeda.sormas.ui.utils.LayoutUtil.loc;
 import static java.util.function.Predicate.not;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


sormas-ui/src/main/java/de/symeda/sormas/ui/user/UserEditForm.java
```diff
@@ -245,12 +255,11 @@ protected void addFields() {
 
         if (FacadeProvider.getFeatureConfigurationFacade().isFeatureEnabled(FeatureType.AUTH_PROVIDER_TO_SORMAS_USER_SYNC)) {
             this.getFieldGroup().getFields().forEach(userField ->{
-                List<String> excludedFields = Arrays.asList(UserDto.USER_ROLES, UserDto.REGION, UserDto.DISTRICT, UserDto.COMMUNITY);
                 if (!excludedFields.contains(userField.getId())) {
                     userField.setEnabled(false);
                 }
             });
-            this.getField(UserEditForm.RESTRICT_DISEASES_CHECKBOX_LOC).setEnabled(false);
+            this.getField(UserEditForm.RESTRICT_DISEASES_CHECKBOX_LOC).setEnabled(true);
```
leventegal-she, line 262: this line is not needed, fields are enabled by default

PR #13164, opened 2024-10-14
#13156 Hide Case Classification for RSV and Automatically Set Default…

### LINKED ISSUE #13156: Hide Case Classification for RSV and Automatically Set Default Values [1]
### Problem Description

For RSV cases, the case classification process should be hidden, as every entered case is automatically considered a confirmed case. Additionally, certain default values should be automatically set in the background to streamline case entry.

### Proposed Change

- [x] Hide Case Classification: The "Case Classification" field should not be displayed for RSV cases.

**Automatically set the following values in the background for each RSV case:**
- [x] Classification: Set as "Confirmed case" in the database for every entered case.
- [x] Date of Classification: Set equal to the "Report Date."
- [x] Classifying User: Set as the "Reporting User."

### Added Value/Benefit

**Streamlined Case Management:**
By automatically classifying every RSV case as "Confirmed" and hiding the case classification field, the data entry process becomes much simpler and quicker for users. This reduces the administrative burden on health authorities and speeds up the workflow.
 **Reduced Human Error:**
Automatically setting the classification date to the report date and assigning the classifying user to the reporting user ensures consistency in the data. This minimizes the risk of human error, as users won’t need to manually enter or select these values.

### Acceptance Criteria

- [x] The "Case Classification" field should not appear in the UI for RSV cases.

In the background, every RSV case should be:
- [x] Automatically classified as "Confirmed" and saved in the database as such.
- [x] The "Date of Classification" should be equal to the "Report Date."
- [x] The "Classifying User" should be the same as the "Reporting User."

- [x] These changes should apply only to RSV cases without affecting other diseases.

### Implementation Details

_No response_

### Mockups

_No response_

### Additional Information

_No response_

leventegal-she:
… Values

Fixes #13156

### DIFF
--- a/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
+++ b/sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
@@ -2066,7 +2066,8 @@ public class CaseFacadeEjb extends AbstractCoreFacadeEjb<Case, CaseDataDto, Case
 	@PermitAll
 	public void onCaseSampleChanged(Case associatedCase) {
 		// Update case classification if the feature is enabled
-		if (configFacade.getCaseClassificationCalculationMode(associatedCase.getDisease()).isAutomaticEnabled()) {
+		if (configFacade.getCaseClassificationCalculationMode(associatedCase.getDisease()).isAutomaticEnabled()
+			& associatedCase.getDisease() != Disease.RESPIRATORY_SYNCYTIAL_VIRUS) {
 			if (associatedCase.getCaseClassification() != CaseClassification.NO_CASE) {
 				Long pathogenTestsCount = pathogenTestService.countByCase(associatedCase);
 				if (pathogenTestsCount == 0) {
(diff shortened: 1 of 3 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=13164

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
```diff
@@ -2066,7 +2066,8 @@ private void updateTasksOnCaseChanged(Case newCase, CaseDataDto existingCase) {
 	@PermitAll
 	public void onCaseSampleChanged(Case associatedCase) {
 		// Update case classification if the feature is enabled
-		if (configFacade.getCaseClassificationCalculationMode(associatedCase.getDisease()).isAutomaticEnabled()) {
+		if (configFacade.getCaseClassificationCalculationMode(associatedCase.getDisease()).isAutomaticEnabled()
+			& associatedCase.getDisease() != Disease.RESPIRATORY_SYNCYTIAL_VIRUS) {
```
sergiupacurariu, line 2070: the above condition will be hard to follow if the number of diseases with this attributes start to grow. Shouldn't we have a single place where we test the disease?

sergiupacurariu:
will this disease features be applied only on Luxembourg servers?

sormas-backend/src/main/java/de/symeda/sormas/backend/caze/CaseFacadeEjb.java
leventegal-she: Can be changed when it gets hard to follow, at the moment we don't know how it will be

leventegal-she:
> will this disease features be applied only on Luxembourg servers?

For all servers

sergiupacurariu (approved):

PR #13232, opened 2025-01-15
#13231 - Birthdate filter does not hold the selected values

### LINKED ISSUE #13231: Birthdate filter does not hold the selected values
### Bug Description

An user filters the Case grid and uses the birthdate fields in the filter. After applying the filter the filtered grid is listed. One of the results (eg. a Case) is accessed. After finishing the work for that Case the user might want to go back to the filtered grid and pick a new case from the filtered grid. When the user goes back to the grid it will find the filtered Cases but the filter will not have the birthdate fields filled like all other fields in the filter. In order to have the same query the birthdate filters must be added again.
The above described behavior must be fixed for: Persons, Cases and Contacts directories.

### Reproduction Steps

1. Go to cases
2. Add dates to birthdate filter
3.Apply the filters
4.Go into one of the filtered grid results
5. From the accessed case go back to filtered grid
6. The values from the birthdate fields are not displayed anymore
7. If any other adjustment is made to the filter it will also need to enter once more the birthdate fields
...

### Expected Behavior

The birthdate fields should keep the dates inserted in the first place. No need to be inserted again once returning from a record listed in the grid.

### Affected Area(s)

Web app

### SORMAS Version

Sormas 1.100.0

### Android version/Browser

Chrome

### User Role(s)

National admin user

### Server URL

_No response_

### Additional Information

_No response_

sergiupacurariu:
Fixes #13231

### DIFF
--- a/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseFilterForm.java
+++ b/sormas-ui/src/main/java/de/symeda/sormas/ui/caze/CaseFilterForm.java
@@ -783,6 +783,14 @@ public class CaseFilterForm extends AbstractFilterForm<CaseCriteria> {
 				weekAndDateFilter.getDateToFilter().setValue(criteria.getNewCaseDateTo());
 			}
 		}
+
+		//Birthdate Filter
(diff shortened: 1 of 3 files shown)


sormas-vitagroup:
SonarCloud analysis: https://sonarcloud.io/dashboard?id=SORMAS-Project&pullRequest=13232

leventegal-she (approved):

Issue #14036, opened 2026-06-24
Antibiotic Susceptibility Testing (AST) for Tuberculosis shows unnecessary columns


Christophe-SF:
### Bug Description

The AST must provide the following for each antibiotic: a value, a method, and an interpretation/conclusion, arranged in three columns as it was the case some time ago. Currently, five columns are displayed.

<img width="1079" height="658" alt="Image" src="https://github.com/user-attachments/assets/a97ffec8-efe0-4e08-97f3-70d0447c99aa" />

### Reproduction Steps

1.
2.
...

### Expected Behavior

Antibiotic susceptibility testing (AST) evaluates the resistance or susceptibility of the disease-causing pathogen to various antibiotics. In the context of surveillance, these antibiotics are disease-specific and are hard-coded in SORMAS.
Various methods exist, yielding specific values ​​that must be interpreted by the laboratory.
In this case, the test must provide the result, the method, and the interpretation for each antibiotic. 

<img width="873" height="481" alt="Image" src="https://github.com/user-attachments/assets/54480ddd-7294-40b6-b08b-05b6356d812c" />

### Affected Area(s)

Web app

### SORMAS Version

1.104

### Android version/Browser

chrome

### User Role(s)

Nat user

### Server URL

qa-lux

Christophe-SF:
Works fine

Issue #13951, opened 2026-05-29
Laboratory - Introduce test categories and complete the method list


roldy:
### Problem Description

The list of pathogen test types today mixes broad categories (for example "Antigen detection") with
specific methods (for example "Sequencing") and a few temporary "Other …" placeholder entries. The
specification asks for the tests to be organised into **categories**, each containing a clear set of
**methods**, and for several missing methods to be added.

### Proposed Solution

* When adding a pathogen test, the user first chooses a **category**, then a **method** within that
  category.
* The seven categories are available: **Molecular Assays**, **Serological Tests**, **Antigen
  Detection**, **Culture & Isolation**, **Microscopy & Staining**, **Antimicrobial Susceptibility
  Testing**, **Functional Immune Assays**.
* The methods from the approved test list are selectable, including those currently missing — for
  example: RT-PCR, PCR (qPCR & conventional), Multiplex PCR, Digital PCR, TMA, LAMP, Line Probe
  Assay, Sanger Sequencing, Whole-Genome / Next-Generation Sequencing, DNA Microarray; ELISA
  (IgM/IgG/IgA), Western Blot, Neutralization, Immunofluorescence (IFA/DFA), Rapid Antibody Test;
  Lateral Flow, Latex/Slide agglutination, Quellung, Hemagglutination Inhibition, RDT; Bacterial
  Culture, Viral Isolation, Fungal Culture, MALDI-TOF; Gram Stain, Acid-Fast (smear grade),
  Dark-Field, Giemsa, Histopathology, FISH, Immunohistochemistry, Electron Microscopy, blood smears;
  Disk Diffusion / MIC, Genotypic Resistance Test; IGRA, Tuberculin Skin Test, Flow Cytometry.
* The temporary "Other …" placeholder methods are removed once the categories are in place.
* Methods that were merged or retired (for example the old separate "IgM/IgG/IgA serum antibody"
  entries now under ELISA, and removed entries such as "Incubation time" and "CQ value detection")
  no longer appear when adding a new test, **but tests already recorded with them still display
  correctly**.
* The method list stays sorted alphabetically (English) within its grouping.

### Additional Information

The authoritative list is the "Test to add" sheet of `4 Tests & categories X results.xlsx`
(update 28/07/2025), and the seven categories are on its "Categories" sheet. Each method's expected
result values and value type are defined there.

Christophe-SF:
Works fine
The reference table has been updated with 
[...]

PR #14316, opened 2026-08-31
Gonococcal infection - Adapt samples and pathogen tests

### LINKED ISSUE #14312: Gonococcal infection - Adapt samples and pathogen tests
# Gonococcal infection - Adapt samples and pathogen tests 

### Samples

- [x] Make the existing `SampleMaterial.GENITAL_SWAB` selectable for Gonococcal infection.
- [x] Make the existing `SampleMaterial.CONJUNCTIVAL_SWAB` selectable for Gonococcal infection on Luxembourg systems.

### Test types

- [x] Add `NON_AMPLIFIED_NUCLEIC_ACID_PROBE_TEST` to `PathogenTestType` as a molecular-assay test with a qualitative result, and make it selectable for Gonococcal infection.
- [x] Make the existing `PathogenTestType.NAAT` selectable for Gonococcal infection as a qualitative test with a single `CQ/CT value` input and without the disease-specific PCR target fields.
- [x] Make the existing `PathogenTestType.GENOTYPING` selectable for Gonococcal infection. It must not have a positive/negative result.
- [x] Make the existing `PathogenTestType.ANTIBIOTIC_SUSCEPTIBILITY` selectable for Gonococcal infection. AST must continue to use its dedicated drug-susceptibility result form instead of a pathogen-test-level result.
- [x] Add `PENICILLINASE_ACTIVITY` to `PathogenTestType` as a qualitative test and make it selectable for Gonococcal infection.

### NAAT

- [x] Extend the Cq/Ct applicability and form visibility logic so Gonococcal `NAAT` displays one `CQ/CT value` field.
- [ ] Do not display the existing pathogen-target-specific Ct fields (`ctValueE`, `ctValueN`, `ctValueRdrp`, `ctValueS`, `ctValueOrf1`, or `ctValueRdrpS`) for Gonococcal `NAAT`.

### Genotyping

- [x] Add four Gonococcal-genotyping text fields to the pathogen-test persistence model, DTO, mapper, form, and relevant exports/imports:
  - `porBAllele`
  - `tbpBAllele`
  - `sequenceType`
  - `genogroup`
- [x] Display these fields only for Gonococcal `GENOTYPING` and store the pathogen-test result as `NOT_APPLICABLE`/null rather than implicitly positive.

### Antimicrobial susceptibility testing

- [x] Extend the existing per-drug `DrugSusceptibility` model and form so a Gonococcal AST contains entries for exactly:
  - Existing drugs: `CIPROFLOXACIN`, `CEFTRIAXONE`, `AZITHROMYCIN`
  - New drugs: `CEFIXIME`, `TETRACYCLINE`, `GENTAMICIN`, `SPECTINOMYCIN`
- [ ] For every listed antibiotic, persist and display:
  - `AST method`: `BREAKPOINT`, `ETEST`, or `MIC`
  - `MIC value (mg/L)`: a decimal numeric value
  - `AST result`: `SUSCEPTIBLE`, `INTERMEDIATE`, `RESISTANT`, or `UNKNOWN`
- [x] Add the missing `BREAKPOINT` and `MIC` values to `SusceptibilityMethod`; retain `ETEST`.
- [x] Add the four new drugs and their fixed method/MIC/susceptibility field triplets consistently across `Drug`, `DrugSusceptibilityDto`, `DrugSusceptibility`, mapping, forms, external-message mapping, and lab-result display/export.
- [ ] Validate MIC as a non-negative decimal in mg/L. Preserve meaningful decimal precision and do not impose an unsupported clinical maximum; keep serialization compatible with the existing free-text MIC representation if changing the stored type would break existing data.
- [ ] Keep laboratory code/question mappings as metadata around the existing per-drug structure so later mapping changes do not require another persistence redesign.

Pa-Touche:
Fixes #14312

## Summary by CodeRabbit

* **New Features**
  * Added gonococcal infection support for pathogen testing, including PorB allele, TbpB allele, sequence type, and genogroup fields.
  * Added non-amplified nucleic acid probe testing, NAAT, genotyping, and penicillinase activity options.
  * Expanded antibiotic susceptibility testing with cefixime, tetracycline, gentamicin, and spectinomycin, including MIC, method, and susceptibility details.
  * Added relevant sample materials, a dedicated gonococcal infection section, and treatment options.
  * Expanded rubella and congenital rubella sample and test availability.
  * Added captions and labels for new fields and options.

* **Tests**
  * Added coverage for gonococcal test types and sample material visibility.


sormas-api/src/main/java/de/symeda/sormas/api/therapy/SusceptibilityMethod.java
```diff
@@ -26,7 +26,9 @@ public enum SusceptibilityMethod {
 	BROTH_MICRODILUTION,
 	AGAR_DILUTION,
 	DISK_DIFFUSION,
+	BREAKPOINT,
```
roldy, line 29: BREAKPOINT looks like it reintroduces the concept that #14036 retired as BREAKPOINT_CATEGORICAL. That change also migrated every stored value to OTHER across all 20 drugs in sormas_schema.sql, so older breakpoint records can no longer be told apart from genuine Other entries.

Could we resurrect BREAKPOINT_CATEGORICAL instead, the way the test category work did in #13951? That keeps the historical values recoverable.

Separately, MIC as a method sits a little awkwardly next to AUTOMATED_MIC and next to the per drug MIC value column. Might be worth confirming with Christophe what the lab means by it.

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleMaterial.java
```diff
@@ -701,8 +720,6 @@ public enum SampleMaterial {
 		Disease.INVASIVE_PNEUMOCOCCAL_INFECTION,
 		Disease.INVASIVE_MENINGOCOCCAL_INFECTION,
 		Disease.SHIGELLOSIS }, hide = true)
-	@HideForCountries(countries = {
-		CountryHelper.COUNTRY_CODE_LUXEMBOURG })
 	CONJUNCTIVAL_SWAB,
```
roldy, line 723: Dropping the Luxembourg country hide here opens conjunctival swab for every disease not already in the hide list above, which is more than 50 diseases in the Luxembourg instance rather than just Gonococcal infection. Rubella, congenital rubella, HIV, influenza A and B, adenovirus and trachoma are among them.

The Rubella work on development took the other route and extended the hide list instead. Could we lift the country hide and also add the diseases that should not offer this material, so the widening stays intentional?

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/diseasesection/GonococcalInfectionSectionComponent.java
```diff
@@ -0,0 +1,111 @@
...
+				eventBus.fire(new SetTestResultEvent(null));
+			}
+		}));
+	}
+
+	@Override
+	public void setDto(PathogenTestDto dto) {
+		super.setDto(dto);
+		PathogenTestType type = dto == null ? null : dto.getTestType();
+		setGenotypingVisible(type == PathogenTestType.GENOTYPING);
```
roldy, line 69: setDto calls setGenotypingVisible(false) for any test type other than Genotyping, and that method clears the four bound fields. Since the binder is bean bound, the clear writes null straight back into the DTO that was just loaded.

ExternalMessageMapper maps porBAllele, tbpBAllele, sequenceType and genogroup with no test type condition, so a lab message carrying an allele on a NAAT test would save correctly and then be wiped the first time someone opens that test and saves.

Clearing on load is not really needed, the fields are hidden anyway and cleanup already handles the section swap. Could the clear be kept on the test type change path only?

sormas-api/src/main/java/de/symeda/sormas/api/sample/PathogenTestType.java
```diff
@@ -788,7 +820,9 @@ public enum PathogenTestType {
 		Disease.INVASIVE_MENINGOCOCCAL_INFECTION,
 		Disease.GIARDIASIS,
 		Disease.CRYPTOSPORIDIOSIS,
-		Disease.SALMONELLOSIS }, hide = true)
+		Disease.SALMONELLOSIS,
+		Disease.GONOCOCCAL_INFECTION,
+		Disease.GONOCOCCAL_INFECTION }, hide = true)
```
roldy, line 825: Small one, Disease.GONOCOCCAL_INFECTION appears twice in this hide list. It compiles fine, but these lists get merged constant by constant when branches meet, so the duplicate makes the next merge harder to read.

sormas-api/src/test/java/de/symeda/sormas/api/sample/PathogenTestTypeGonococcalInfectionTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import de.symeda.sormas.api.Disease;
+import de.symeda.sormas.api.utils.fieldvisibility.checkers.DiseaseFieldVisibilityChecker;
+
+public class PathogenTestTypeGonococcalInfectionTest {
+
+	private final DiseaseFieldVisibilityChecker checker = new DiseaseFieldVisibilityChecker(Disease.GONOCOCCAL_INFECTION);
+
+	@Test
+	public void requiredTestTypesAreSelectable() {
```
roldy, line 19: These read well. One suggestion though, both tests assert only that the wanted values are visible. The bulk of the change is the roughly 50 hide list edits, and none of those is covered, which is how the duplicate entry on MALDI_TOF got through.

An assertion on the exact visible set for Gonococcal infection, matching the nine sample materials and eight test types in the requirements document, would pin the specification down and protect it when other diseases touch the same lists.

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/PathogenTestController.java
```diff
@@ -78,7 +78,11 @@ public class PathogenTestController {
 
 	// Antibiotic susceptibility test is applicable for TB(Lux), IMI, IPI and Shigellosis. For others it should be null.
```
roldy, line 79: The comment just above still says the AST test applies to TB Lux, IMI, IPI and Shigellosis, but Gonococcal infection is now in the list.

Two other comments drifted the same way in this change. The javadoc on cqInputApplies still lists only three disease and method pairs, and the comment above the coercion in TestResultComponent.setDto says only AST is coerced, while Genotyping for Gonococcal infection is now coerced too. Could we refresh all three?

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/components/TestResultComponent.java
```diff
@@ -348,7 +348,9 @@ public void setDto(PathogenTestDto dto) {
 		// QUALITATIVE in their @ResultValueTypeRel (WESTERN_BLOT, BACTERIAL_CULTURE, typing methods, ...)
```
roldy, line 348: The Gonococcal section fires SetTestResultEvent with NOT_APPLICABLE for Genotyping and for AST. In the listener further up, the branch handling a non null result sets the value without calling ensureNotApplicableSelectable first, and NOT_APPLICABLE is filtered out of the item list for non Luxembourg instances.

Gonococcal infection is not country scoped in the Disease enum, so a non Luxembourg server that enables it would end up unable to save the form when a result is required. This is pre existing and shared with Shigellosis, but the change makes it easier to reach.

sormas-ui/src/main/java/de/symeda/sormas/ui/therapy/DrugSusceptibilityForm.java
```diff
@@ -277,9 +285,41 @@ protected void addFields() {
...
 		FieldHelper.hideFieldsNotInList(getFieldGroup(), List.of(), true);
 	}
 
+	private void addDrugRow(String labelLoc, Drug drug, String methodId, String micId, String susceptibilityId) {
+		addDrugLabel(labelLoc, drug, micId);
+		addMethodField(methodId);
+		addMicField(micId);
+		addResistanceResultField(susceptibilityId).setInputPrompt(I18nProperties.getString(Strings.promptResistanceResult));
+	}
+
```
roldy, line 322: Nice helper. Would you consider applying it to the 22 existing drug blocks as well? At the moment the file has two ways of building the same row, so any future change to a row has to be made in 23 places. Converting the rest would also take out around 70 lines.

sormas-backend/src/main/java/de/symeda/sormas/backend/therapy/DrugSusceptibilityMapper.java
```diff
@@ -252,6 +276,19 @@ public static boolean hasData(DrugSusceptibilityDto dto) {
 			|| dto.getAmpicillinMethod() != null
```
roldy, line 276: Adding a drug now means remembering it in four places, the entity, both copy blocks in this file and this boolean chain. Missing the last one is silent data loss, since hasData returns false and the whole grid is dropped on save. The chain already checks getErythromycinSusceptibility twice, which shows how easy that is to do.

AnnotationFieldHelper is already used against this DTO in DrugSusceptibilityForm, so driving hasData from the declared fields would make it addition proof.

sormas-api/src/main/java/de/symeda/sormas/api/externalmessage/labmessage/TestReportDto.java
```diff
@@ -213,6 +214,22 @@ public class TestReportDto extends EntityDto {
...
+	private String tetracyclineMic;
+	private DrugSusceptibilityType tetracyclineSusceptibility;
+	private SusceptibilityMethod tetracyclineMethod;
+	private String gentamicinMic;
+	private DrugSusceptibilityType gentamicinSusceptibility;
+	private SusceptibilityMethod gentamicinMethod;
+	private String spectinomycinMic;
+	private DrugSusceptibilityType spectinomycinSusceptibility;
+	private SusceptibilityMethod spectinomycinMethod;
+	private String porBAllele;
```
roldy, line 229: These four fields have no @Size, while their twins on PathogenTestDto use CHARACTER_LIMIT_DEFAULT and the TestReport columns are bounded to the same length. An over long value arriving from a lab message would pass validation and then fail at the database layer rather than giving a readable error naming the field. Worth adding @Size for consistency with the other text fields in this class.

roldy (commented):
Thanks for this. I went through it against the Luxembourg requirements document, and the scoping work holds up well. The sample list, the test list and the AST drug set all resolve to exactly what the document specifies, once @NotSelectableForNewTests and the deprecated materials are taken into account. 

Two things to sort out before merge, plus some smaller notes inline.

**Database migration is missing**

The branch adds 32 new persisted columns, 4 on pathogentest, 12 on drugsusceptibility and 16 on testreport, but sormas_schema.sql is untouched and the last version is still 665. Against a real Postgres database, opening or saving any pathogen test would fail with a missing column error. The history tables need the same columns too, the azithromycinmic entries around lines 16223 and 16248 show the expected shape.

**Rebase first, then write the migration**

Development has moved some commits ahead since this branch last took it, which scoped the same two enums. Some conflicts to sort out

1. CONJUNCTIVAL_SWAB. Development kept the Luxembourg country hide and added Rubella and CRS to the hide list, this branch removed the country hide. Taking either side wholesale loses something.
2. GENOTYPING. This one is an allow list rather than a hide list, so a union means the opposite of what it means in the surrounding 44 hunks.
3. MALDI_TOF. Contains the duplicate entry noted inline, so this is a good moment to drop it.

**Worth confirming with Christophe**

Four places where the code and the document disagree. In each case the code looks right to me and the document probably needs updating:

1. The MIC value is free text rather than a float, following the decision taken in #14036.
2. The AST method dropdown offers all nine susceptibility methods rather than the three listed in the document.
3. Genotyping sets the result to Not applicable rather than always positive.
4. Row GONO-14 says the sequence type goes into the genogroup field, which contradicts section 3.2.2 where sequence type has its own field. The code follows 3.2.2.

roldy (changes requested):
Switching this to request changes so the migration does not get missed. Everything is already covered in my review above, there is nothing new here.

The blocker is the database migration. The 32 new persisted columns have no entry in sormas_schema.sql, so the branch would fail against a real database and no test in CI would flag it beforehand.

The rebase onto current development is the other thing worth doing before merge, ideally before the migration is written so that it does not get renumbered underneath a database that has already run the old number.

The remaining inline notes are suggestions and are fine to take or leave. Happy to take another look once the migration is in.

sormas-api/src/main/java/de/symeda/sormas/api/externalmessage/labmessage/TestReportDto.java
Pa-Touche: Agreed: unified across types

sormas-api/src/main/java/de/symeda/sormas/api/therapy/SusceptibilityMethod.java
Pa-Touche: Discussed with @SORMAS-JanBoehme and also with Chris. 

Categorical suffix does not add semantic to the enum member name. 

No PROD data can exist for BREAKPOINT_CATEGORICAL as this was not added before the latest release. 
So for now will be kept as BREAKPOINT, in case we missed something please let us now @roldy and we can adjust before the next release. 

Many thanks in advance

sormas-ui/src/main/java/de/symeda/sormas/ui/therapy/DrugSusceptibilityForm.java
Pa-Touche: Okay, done

sormas-api/src/main/java/de/symeda/sormas/api/sample/PathogenTestType.java
Pa-Touche: correct

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleMaterial.java
Pa-Touche: I'll check this with @XavCol

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/diseasesection/GonococcalInfectionSectionComponent.java
Pa-Touche: Agree, I removed the clear

sormas-api/src/test/java/de/symeda/sormas/api/sample/PathogenTestTypeGonococcalInfectionTest.java
Pa-Touche: Nice catch! Added abstract class to make it simpler to test this

sormas-backend/src/main/java/de/symeda/sormas/backend/therapy/DrugSusceptibilityMapper.java
Pa-Touche: Can you please make those changes yourself ? I'm not confident making those changes. You now this better than I do, but I think it's a good idea!

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/PathogenTestController.java
Pa-Touche: agree

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleMaterial.java
Pa-Touche: Checked with @XavCol: as there still is the restriction on the diseases, this is fine for him

raulbob: I just unresolved this as a tripple check.
Hope you're sure on this because :
- the restriction list holds 21 diseases; the `Disease` enum has 72 constants. 
- removing `@HideForCountries(LUXEMBOURG)` therefore opens conjunctival swab to all other disease not listed in the annotation for Luxembourg.

sormas-api/src/test/java/de/symeda/sormas/api/sample/SampleMaterialGonococcalInfectionTest.java
```diff
@@ -0,0 +1,17 @@
+package de.symeda.sormas.api.sample;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+import de.symeda.sormas.api.Disease;
+import de.symeda.sormas.api.utils.fieldvisibility.checkers.DiseaseFieldVisibilityChecker;
+
+public class SampleMaterialGonococcalInfectionTest {
```
raulbob: This has to be revisited because the sample material list contain 'UNKOWN'.
In the spec document there are 9 sample materials, the implementation enables the additional 'UNKOWN'.
Please clarify if 10 is intended vs 9 in spec.

From the spec:
- Genital swab
- Clinical sample
- Urine
- Rectal swab
- Oropharyngeal swab
- Conjunctival swab
- Blood
- Cerebrospinal fluid
- Synovial fluid

sormas-backend/src/main/resources/sql/sormas_schema.sql
```diff
@@ -17318,4 +17318,82 @@ ALTER TABLE testreport_history ADD COLUMN IF NOT EXISTS virulencegenesdetected b
...
+ALTER TABLE testreport_history ADD COLUMN IF NOT EXISTS spectinomycinsusceptibility varchar(255);
+ALTER TABLE testreport_history ADD COLUMN IF NOT EXISTS spectinomycinmethod varchar(255);
+ALTER TABLE testreport_history ADD COLUMN IF NOT EXISTS porballele varchar(512);
+ALTER TABLE testreport_history ADD COLUMN IF NOT EXISTS tbpballele varchar(512);
+ALTER TABLE testreport_history ADD COLUMN IF NOT EXISTS sequencetype varchar(512);
+ALTER TABLE testreport_history ADD COLUMN IF NOT EXISTS genogroup varchar(512);
+
+-- Match the updated TestReport text length.
+ALTER TABLE testreport ALTER COLUMN serogroupspecificationtext TYPE varchar(512);
+ALTER TABLE testreport_history ALTER COLUMN serogroupspecificationtext TYPE varchar(512);
```
raulbob: `testreport` `serogroupspecificationtext`  512 vs. `pathogentest` `serogroupspecificationtext` at 255

sormas-api/src/main/java/de/symeda/sormas/api/therapy/SusceptibilityMethod.java
raulbob: I'm continuing where Harold left off, the second part of the comment.
`SusceptibilityMethod.MIC` is with `AUTOMATED_MIC` and the per-drug MIC **value**.
So the two new options now also show up in the TB (Lux), IMI, IPI and Shigellosis susceptibility grids. 

I checked with @Christophe-SF, he is fine with both keys present, although from his point of view `AUTOMATE_MIC` could be enough for the users.
As there is no apparent problem with having both at the moment we will keep both values.

sormas-backend/src/main/java/de/symeda/sormas/backend/externalmessage/labmessage/TestReport.java
```diff
@@ -905,6 +1066,7 @@ public void setSeroGroupSpecification(SeroGroupSpecification seroGroupSpecificat
 		this.seroGroupSpecification = seroGroupSpecification;
 	}
 
+	@Column(length = CHARACTER_LIMIT_DEFAULT)
```
raulbob: Wider then corresponding field in PathogenTest.
Seems that PathogenTest.seroGroupSpecificationText is missing the `@Column` annotation completely.
Maybe we should make sure that all these correlated fields have the same length and annotations.

sormas-api/src/main/java/de/symeda/sormas/api/sample/PathogenTestType.java
```diff
@@ -492,12 +513,18 @@ public enum PathogenTestType {
...
 		Disease.RUBELLA,
 		Disease.CONGENITAL_RUBELLA })
 	@PathogenTestCategoryRel(PathogenTestCategory.MOLECULAR_ASSAYS)
 	@ResultValueTypeRel(ResultValueType.TEXT)
 	GENOTYPING,
 
+	@Diseases(Disease.GONOCOCCAL_INFECTION)
+	@PathogenTestCategoryRel(PathogenTestCategory.ANTIMICROBIAL_SUSCEPTIBILITY_TESTING)
+	@ResultValueTypeRel(ResultValueType.QUALITATIVE)
+	PENICILLINASE_ACTIVITY,
```
raulbob: Wrong placement, this should be under the `Molecular Assays` comment header line 1194.

sormas-api/src/main/java/de/symeda/sormas/api/therapy/DrugSusceptibilityDto.java
```diff
@@ -89,6 +89,18 @@ public class DrugSusceptibilityDto extends PseudonymizableDto {
 	public static final String AZITHROMYCIN_MIC = "azithromycinMic";
 	public static final String AZITHROMYCIN_SUSCEPTIBILITY = "azithromycinSusceptibility";
 	public static final String AZITHROMYCIN_METHOD = "azithromycinMethod";
+	public static final String CEFIXIME_MIC = "cefiximeMic";
+	public static final String CEFIXIME_SUSCEPTIBILITY = "cefiximeSusceptibility";
+	public static final String CEFIXIME_METHOD = "cefiximeMethod";
+	public static final String TETRACYCLINE_MIC = "tetracyclineMic";
```
raulbob, line 95: There are some fields that are not added to the sorting list in `DrugSusceptibilityResultPanel:74`.

Maybe we need to add the following fields to the list:

- `AZITHROMYCIN_SUSCEPTIBILITY`
- `CEFIXIME_SUSCEPTIBILITY`
- `TETRACYCLINE_SUSCEPTIBILITY`
- `GENTAMICIN_SUSCEPTIBILITY`
- `SPECTINOMYCIN_SUSCEPTIBILITY`
- `CEFTAZIDIME_SUSCEPTIBILITY`
- `CEFOTAXIME_SUSCEPTIBILITY`
- `AMPICILLIN_SUSCEPTIBILITY`
- `TRIMETHOPRIM_SULFAMETHOXAZOLE_SUSCEPTIBILITY`

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/diseasesection/GonococcalInfectionSectionComponent.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * SORMAS® - Surveillance Outbreak Response Management & Analysis System
+ * Copyright © 2016-2026 SORMAS Foundation gGmbH
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License.
```
raulbob: Cut off header.

raulbob (changes requested):

raulbob (commented):
In principle ok, there are some issues that could be addressed see comments.

sormas-backend/src/main/java/de/symeda/sormas/backend/externalmessage/labmessage/TestReport.java
Pa-Touche: Indeed, nice catch

sormas-backend/src/main/resources/sql/sormas_schema.sql
Pa-Touche: correct fixed

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/diseasesection/GonococcalInfectionSectionComponent.java
raulbob, line 1: It seems that the following fields are not cleared at save :
- porBAllele
- tbpBAllele
- sequenceType
- genogroup

create a test select `GENOTYPING`, enter values in those fields, switch the test type to `CULTURE`, save, hidden values are now written to db.

raulbob (changes requested):

sormas-ui/src/main/java/de/symeda/sormas/ui/samples/diseasesection/GonococcalInfectionSectionComponent.java
raulbob: Now I do not think that it is needed to override the setDto method.
Unless there is a specific reason that might elude me.
Visibility should be managed through other means as is the case for other disease section components. This is the only section component implementation that overrides setDto.

sormas-api/src/main/java/de/symeda/sormas/api/sample/SampleMaterial.java
XavCol: @raulbob @Pa-Touche The sampling technique is not specific to Luxembourg but disease specifique, so I don't see the point to have specifique rules for luxembourg. Also as this data is generally fill automatically with the lab xml it is not a big deal to have more sample in the list.
So we can keep it like this

